Issue #64, opened 2022-04-02
Cornerstone3D and Vite


ty-ler:
I have been attempting to utilize Cornerstone3D in a Svelte web app which uses Vite. Unfortunately, when I try to import @cornerstonejs/core into the app, the build fails. 

![image](https://user-images.githubusercontent.com/18401956/161401704-dc1ccab2-fd29-4ab3-9e08-f1c6cb1139a9.png)

![image](https://user-images.githubusercontent.com/18401956/161401622-fc17f059-5410-4106-93d7-a6dcffa92e72.png)

**Note:** If I import things from `@cornerstonejs/core/dist/umd/index` everything compiles and works fine, but I cannot utilize the TS types.

**Reproduction steps:** 
1. Generate new Svelte app using Vite (`npm create vite@latest`)
2. `npm install`
3. `npm install @cornerstonejs/core` (etc...)
4. `npm run dev`
5. Above error is thrown
  
I've created a repo with nothing but a fresh Svelte app and Cornerstone3d to demonstrate the problem:
https://github.com/ty-ler/cornerstone3d-svelte-vite-repro

swederik:
Thanks for the reproduction repo!

I was actually also wondering what we should be using for the main/module etc fields in package.json for Cornerstone. We currently output three things: UMD, CommonJS, and ESM.

My understanding is that we need UMD for browser-based script-tag usage, so I put that as the default `main`. 

I was under the impression that ESM was not widely supported yet, but was the intended future format, so I put that as `module` and also pointed the `types` to the index.d.ts in the ESM folder. 

I left CommonJS in there in case someone needed it for something.

We could potentially switch `main` to the ESM version and then add the `browser` property to point to UMD instead (https://docs.npmjs.com/cli/v8/configuring-npm/package-json#browser). You could try doing this, rebuilding, and then using yarn link to see if this resolves your issue. If it does, and there's no other downsides I'd be happy to merge it.

My only hesitation with switching to ESM as `main` is that we are currently using the ESM output generated by TypeScript, and it does not generate valid ESM modules, since they are missing the `.js` in their imports (see this thread https://github.com/microsoft/TypeScript/issues/40878). It doesn't look like the TypeScript team plans to fix this issue, so we would probably need to find a workaround for that. One super-ugly workaround that people describe is to use `.js` while importing from a `.ts` file - the .js won't get transpiled out and so the result will be valid. I hate the idea of doing that though but haven't had a chance to look for a better alternative.

[...]

PR #65, opened 2022-04-03
fix: change package.json 'module' field to use '.js' extension from '.ts'

ty-ler:
The 'module' field in package.json should probably use a '.js' extension instead of '.ts'. The generated output files located in `packages/core/dist/esm` don't actually include any .ts files. Changing the extension resolves build issues experienced in Vite when importing the ESM build of @cornerstonejs/core.

Discussion here: #64

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -4,7 +4,7 @@
   "description": "",
   "main": "dist/umd/index.js",
   "types": "dist/esm/index.d.ts",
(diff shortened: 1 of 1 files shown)


swederik:
Oh wow I'm a moron yeah that should definitely be pointing to the .js index. Probably a typo from when I was experimenting with it. In our tests and examples we consume the src via aliases, not the actual built version so that's probably why we hadn't noticed yet.

PR #73, opened 2022-04-12
fix: Remove resemblejs from dependencies, add detect-gpu

swederik:


.webpack/webpack.base.js
```diff
@@ -68,15 +66,7 @@ module.exports = (env, argv, { DIST_DIR }) => {
...
-      new CleanWebpackPlugin(),
-      new CopyPlugin({
-        patterns: [
-          {
-            from: '../../node_modules/cornerstone-wado-image-loader/dist/dynamic-import/',
-            to: DIST_DIR,
-          },
-        ],
-      }),
+      new CleanWebpackPlugin()
```
sedghi, line 69: So we are not using the dynamic import by default? Since we don't have the `from 'cornerstone-wardo-image-loader/dist/dynamic-import` anywhere in the code

sedghi (approved):
Looks good, one comment for WADO

.webpack/webpack.base.js
swederik: good point, I think we need the alias instead of just the copyPlugin. I'll fix that before we merge this.

PR #74, opened 2022-04-12
fix: renderToCanvas to use CPU rendering

sedghi:
- Fix the renderToCanvas to use the CPU rendering pipeline
- Add stackPrefetch utility function

### DIFF
--- a/packages/core/src/utilities/loadImageToCanvas.ts
+++ b/packages/core/src/utilities/loadImageToCanvas.ts
@@ -1,15 +1,13 @@
-import { IImage, CPUFallbackEnabledElement } from '../../types';
+import { IImage } from '../types';
 
-import getDefaultViewport from './cpuFallback/rendering/getDefaultViewport';
-import calculateTransform from './cpuFallback/rendering/calculateTransform';
-import drawImageSync from './cpuFallback/drawImageSync';
-import { loadAndCacheImage } from '../../imageLoader';
-import * as metaData from '../../metaData';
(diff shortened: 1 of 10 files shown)


packages/core/src/utilities/loadImageToCanvas.ts
```diff
@@ -0,0 +1,86 @@
...
+ * @param requestType - The type of request (default to interaction), can be 'interaction' or 'prefetch' or 'thumbnail'
+ * the order of loading for the pool manager is interaction, thumbnail, prefetch
+ * @param priority - The priority of the request within the request type (lower is higher priority)
+ * @returns - A promise that resolves when the image has been rendered with the imageId
+ */
+export default function loadImageToCanvas(
+  canvas: HTMLCanvasElement,
+  imageId: string,
+  requestType = RequestType.Interaction,
+  priority = -5
```
JamesAPetts, line 31: Shall we just hardcode this as something super high priority? It feels like no one, not even power users, would ever specify anything other than _now_ for this? Setting a priority so it loads in some queue you have no knowledge of when making this call feels realllly weird.

JamesAPetts (approved):
This looks good to me, minor comment/question, but otherwise fine.

packages/core/src/utilities/loadImageToCanvas.ts
sedghi: I think it has some value to keep the parameters and give default options.

PR #76, opened 2022-04-14
feat: add scrollToSlice for element

sedghi:


packages/tools/src/utilities/viewport/scrollToSlice.ts
```diff
@@ -0,0 +1,60 @@
...
+/**
+ * It uses the metadata to scroll to the slice that is intended. Currently,
+ * only supports imageIdIndex in the metadata, but could be extended to
+ * support other types of metadata such as camera position, focal point, etc.
+ * for volume viewports.
+ * @param element - the HTML Div element scrolling inside
+ * @param metadata - the metadata used for scrolling
+ * @returns Promise that resolves to ImageIdIndex
+ */
+function scrollToSlice(
```
JamesAPetts, line 36: jumpToSlice?
jumpToPosition?

packages/tools/src/utilities/viewport/scrollToSlice.ts
```diff
@@ -0,0 +1,60 @@
...
+   * The FrameOfReferenceUID
+   */
+  FrameOfReferenceUID: string;
+};
+
+/**
+ * It uses the metadata to scroll to the slice that is intended. Currently,
+ * only supports imageIdIndex in the metadata, but could be extended to
+ * support other types of metadata such as camera position, focal point, etc.
+ * for volume viewports.
```
JamesAPetts, line 31: I really don't like docs that talk about the future, as they often never get updated.

Instead can you describe what it does _now_, i.e. jumps to imageIdIndex for an element enabled as a stack viewport?

You can write the intention of the future in @privateRemarks.

packages/tools/src/utilities/viewport/scrollToSlice.ts
```diff
@@ -0,0 +1,60 @@
...
+ * @param element - the HTML Div element scrolling inside
+ * @param metadata - the metadata used for scrolling
+ * @returns Promise that resolves to ImageIdIndex
+ */
+function scrollToSlice(
+  element: HTMLDivElement,
+  metadata: Metadata
```
JamesAPetts, line 38: Metadata is hugely overloaded term, can we call this something like options, with a type of `ScrollToSliceOptions`/`JumpToSliceOptions` or something?

packages/tools/src/utilities/viewport/scrollToSlice.ts
```diff
@@ -0,0 +1,60 @@
+import { Types, getEnabledElement, StackViewport } from '@cornerstonejs/core';
+
+type Metadata = {
```
JamesAPetts, line 3: The type needs to be exposed.

JamesAPetts: Can we only have the actually used properties on here for now? We can add more when they are actually used.

JamesAPetts (changes requested):
Minor changes requested for clarity/easy of API use.

sedghi:
jumpToAnnotationUID

JamesAPetts (approved):
LGTM

PR #78, opened 2022-04-18
fix: jumpToSlice and scaling of images in renderToCanvas

sedghi:
- Separate resetZoom and resetPan for camera reset as the consumer need to decide which reset is desired (by default reset zoom was applied)
- Remove the need for passing element to `removeAnnotation` for the annotation to get removed 
- Fixed the scaling of the images in the `renderToCanvas` function
- cleanup the stackScrollTool to use the `scrollThroughStack` utility and fix the problem with deltaFrames = 0 by caching previous direction
- Fixed the jumpToSlice bug for imagerIdIndex out of bound

### DIFF
--- a/packages/core/src/RenderingEngine/RenderingEngine.ts
+++ b/packages/core/src/RenderingEngine/RenderingEngine.ts
@@ -298,12 +298,12 @@ class RenderingEngine implements IRenderingEngine {
    * This is left as an app level concern as one might want to debounce the changes, or the like.
    *
    * @param immediate - Whether all of the viewports should be rendered immediately.
-   * @param resetPanZoomForViewPlane - Whether each viewport gets centered (reset pan) and
+   * @param resetPan - Whether the viewport should reset its pan.
+   * @param resetZoom - Whether the viewport should reset its zoom.
    * its zoom gets reset upon resize.
    */
-  public resize(immediate = true, resetPanZoomForViewPlane = true): void {
+  public resize(immediate = true, resetPan = true, resetZoom = true): void {
     this._throwIfDestroyed();
-
     // 1. Get the viewports' canvases
     const viewports = this._getViewportsAsArray();
 
@@ -320,14 +320,16 @@ class RenderingEngine implements IRenderingEngine {
 
     this._resizeVTKViewports(
       vtkDrivenViewports,
-      immediate,
-      resetPanZoomForViewPlane
+      resetPan,
+      resetZoom,
+      immediate
     );
 
     this._resizeUsingCustomResizeHandler(
       customRenderingViewports,
-      immediate,
-      resetPanZoomForViewPlane
+      resetPan,
+      resetZoom,
+      immediate
     );
   }
 
@@ -526,8 +528,9 @@ class RenderingEngine implements IRenderingEngine {
 
   private _resizeUsingCustomResizeHandler(
     customRenderingViewports: StackViewport[],
-    immediate = true,
-    resetPanZoomForViewPlane = true
+    resetPan = true,
+    resetZoom = true,
+    immediate = true
   ) {
     // 1. If viewport has a custom resize method, call it here.
     customRenderingViewports.forEach((vp) => {
@@ -536,7 +539,7 @@ class RenderingEngine implements IRenderingEngine {
 
     // 3. Reset viewport cameras
     customRenderingViewports.forEach((vp) => {
-      vp.resetCamera(resetPanZoomForViewPlane);
+      vp.resetCamera(resetPan, resetZoom);
     });
 
     // 2. If render is immediate: Render all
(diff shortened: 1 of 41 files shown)


packages/tools/src/stateManagement/annotation/annotationState.ts
```diff
@@ -97,25 +97,22 @@ function addAnnotation(element: HTMLDivElement, annotation: Annotation): void {
...
  */
 function removeAnnotation(
-  element: HTMLDivElement,
-  annotationUID: string
+  annotationUID: string,
+  element?: HTMLDivElement
 ): void {
-  const annotationManager = getViewportSpecificAnnotationManager(element);
+  let annotationManager = getDefaultAnnotationManager();
+  if (element) {
```
JamesAPetts, line 104: I think this is a great solution that makes sense, allows us to have the annotation-only-api whilst also allowing for element specific managers if we ever need them. I know @swederik  is away, but want to tag him for visibility about this decision (sure we can merge before he sees it, can always make another change).

packages/tools/src/tools/StackScrollTool.ts
```diff
@@ -34,6 +36,21 @@ export default class StackScrollTool extends BaseTool {
...
+    let volumeId;
+    if (viewport instanceof VolumeViewport) {
+      volumeId = targetId.split('volumeId:')[1];
+    }
+
+    // We need this check since the deltaFrames can be 0 when the user is
+    // scrolling very slowly so in that case we use the previous direction
+    let direction;
+    if (deltaFrames === 0) {
+      direction = this.previousDirection;
```
JamesAPetts, line 48: Hmm, do we actually want to do this? Want it cause overshooting when you change direction?

packages/tools/src/types/EventTypes.ts
```diff
@@ -49,10 +49,6 @@ type AnnotationModifiedEventDetail = {
  * The data that is passed to the event handler when an annotation is completed drawing.
  */
 type AnnotationRemovedEventDetail = {
-  /** unique id of the viewport */
-  viewportId: string;
-  /** unique id of the rendering engine */
-  renderingEngineId: string;
   /** The annotation that is being added to the annotations manager. */
```
JamesAPetts, line 52: Thinking about this, I think we need the `FrameOfReferenceSpecificAnnotationManager`'s UID? For most cases this will be the default UID, but in the case of still supporting element specific state management it is required to uniquely identify the annotation?

It still keeps it decoupled from the view, as the only information in the event is "This is the annotation, and this is the annotation manager it belongs to."

packages/tools/src/utilities/stackScrollTool/scrollThroughStack.ts
```diff
@@ -8,28 +8,29 @@ import clip from '../clip';
 import getSliceRange from './getSliceRange';
 import snapFocalPointToSlice from './snapFocalPointToSlice';
 
+type ScrollOptions = {
```
JamesAPetts, line 11: Doesn't this type need to be exposed to the consumer? Its public API.

JamesAPetts (changes requested):
Some minor comments to address, but otherwise good.

How are we doing publishing of cornerstone3D beta btw? We are still currently changing public API a lot during this phase, but we shouldn't increment major versions until its out of beta. Do we make each breaking change a feature change. I.e. 0.X.0.?

packages/core/src/RenderingEngine/RenderingEngine.ts
```diff
@@ -320,14 +320,16 @@ class RenderingEngine implements IRenderingEngine {
...
+      resetPan,
+      resetZoom,
+      immediate
     );
 
     this._resizeUsingCustomResizeHandler(
       customRenderingViewports,
-      immediate,
-      resetPanZoomForViewPlane
+      resetPan,
```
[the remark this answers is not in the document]
sedghi: True

packages/tools/src/types/EventTypes.ts
```diff
@@ -51,6 +51,8 @@ type AnnotationModifiedEventDetail = {
 type AnnotationRemovedEventDetail = {
   /** The annotation that is being added to the annotations manager. */
   annotation: Annotation;
+  /** FrameOfReferenceUID */
+  FrameOfReferenceUID: string;
```
JamesAPetts, line 55: What's needed is the _managers_ UID, not the FrameOfReferenceUID, right?

JamesAPetts: Oh you are using the correct value, but the name is wrong, its not a FrameOfReferenceUID, its an `FrameOfReferenceSpecificAnnotationManagerUID`..., which I understand is very long. Maybe `annotationManagerUID`?

sedghi: oh right, my bad

PR #82, opened 2022-04-21
fix: selection API, requestPoolManager and VOI and Scaling

sedghi:
- change the selection api to not rely on the annotation object (UID is enough)
- add PRE_NEW_STACK_IMAGE event for consumers
- add a check for if the loaded image is prescaled (if it was loaded with the scaling parameters), if so it handles the PT non scaled images
- Fix the CPU rendering for PT images 
- Add id to the pool managers
- Fix a bug in the requestPoolManager when a new image is added (interaction priority), it was waiting for a request to be finished before sending the high priority image request
- Fix a bug in requestPool to set the maximum number of requests based on the requestType, previously it was summing all up and comparing them with the current number of requests, it should be per requestType
- Fix prefetching imageIds requests to properly set the prefetch requestType
- Port over the orientation helpers from cornerstone-tools

### DIFF
--- a/packages/core/src/RenderingEngine/StackViewport.ts
+++ b/packages/core/src/RenderingEngine/StackViewport.ts
@@ -126,6 +126,7 @@ class StackViewport extends Viewport implements IStackViewport {
   // TODO: These should not be here and will be nuked
   public modality: string; // this is needed for tools
   public scaling: Scaling;
+  private scalingCache: { [key: string]: ScalingParameters } = {};
 
   /**
    * Constructor for the StackViewport class
@@ -367,11 +368,6 @@ class StackViewport extends Viewport implements IStackViewport {
 
     // todo: some tools rely on the modality
     this.modality = modality;
-    // Compute the image size and spacing given the meta data we already have available.
-    // const metaDataMap = new Map()
-    // imageIds.forEach((imageId) => {
-    //   metaDataMap.set(imageId, metaData.get('imagePlaneModule', imageId))
-    // })
 
     let imagePlaneModule = metaData.get('imagePlaneModule', imageId);
 
@@ -1269,6 +1265,7 @@ class StackViewport extends Viewport implements IStackViewport {
         triggerEvent(this.element, Events.STACK_NEW_IMAGE, eventDetail);
 
         const metadata = this._getImageDataMetadata(image) as ImageDataMetaData;
+        image.isPreScaled = this._isImagePreScaled(imageId);
 
         const viewport = getDefaultViewport(
           this.canvas,
@@ -1361,6 +1358,8 @@ class StackViewport extends Viewport implements IStackViewport {
         suvbw: suvFactor.suvbw,
       };
 
+      this.scalingCache[imageId] = scalingParameters;
+
       // Todo: Note that eventually all viewport data is converted into Float32Array,
       // we use it here for the purpose of scaling for now.
       const type = 'Float32Array';
@@ -1410,7 +1409,6 @@ class StackViewport extends Viewport implements IStackViewport {
         };
 
         triggerEvent(this.element, Events.STACK_NEW_IMAGE, eventDetail);
-
         this._updateActorToDisplayImageId(image);
 
         // Trigger the image to be drawn on the next animation frame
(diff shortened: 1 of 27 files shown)


packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -126,6 +126,7 @@ class StackViewport extends Viewport implements IStackViewport {
   // TODO: These should not be here and will be nuked
   public modality: string; // this is needed for tools
   public scaling: Scaling;
+  private scalingCache: { [key: string]: ScalingParameters } = {};
```
JamesAPetts, line 129: It looks like this will get bigger and bigger as the stack is reused. We should nuke its contents when setstack changes the imageIds hosted on the stack.


Also instead of `{ [key: string]: ScalingParameters } = {};`, its more natural to use Record here in TypeScript:

`private scaleCache: Record<string, ScalingParameters> = {}`

packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -1608,11 +1645,17 @@ class StackViewport extends Viewport implements IStackViewport {
 
     // set voi for the first time
     const { windowCenter, windowWidth } = imagePixelModule;
-    const voiRange =
+    let voiRange =
```
JamesAPetts, line 1648: So what happens is this ends up `undefined`? Shouldn't we then default to the range of the image? (min and max pixel values)?

packages/core/src/requestPool/imageRetrievalPoolManager.ts
```diff
@@ -8,7 +8,7 @@ import { RequestPoolManager } from './requestPoolManager';
  *
  * Retrieval (usually) === XHR requests
  */
-const imageRetrievalPoolManager = new RequestPoolManager();
+const imageRetrievalPoolManager = new RequestPoolManager('imageRetrievalPool');
```
JamesAPetts, line 11: I must of missed this, why do we have an `imageLoadPool` and an `imageRetrievalPool` manager?

packages/core/src/requestPool/requestPoolManager.ts
```diff
@@ -91,14 +91,17 @@ class RequestPoolManager {
    * of the request types, is created. Maximum number of requests of each type
    * is set to 6.
    */
-  constructor() {
+  constructor(id?: string) {
```
JamesAPetts, line 94: Having a default id is pretty dangerous. If one isn't provided we should just autogenerate one?

packages/tools/src/utilities/orientation/getOrientationString.ts
```diff
@@ -0,0 +1,52 @@
...
+ *
+ * @param vector - Input array
+ * @returns The orientation in the patient coordinate system.
+ */
+export default function getOrientationString(vector: Types.Point3): string {
+  // Thanks to David Clunie
+  // https://sites.google.com/site/dicomnotes/
+
+  let orientation = '';
+  const orientationX = vector[0] < 0 ? 'R' : 'L';
```
JamesAPetts, line 15: I understand this is just pulled from CSTLegacy, but we have to be really careful here!

This method is actually get DICOM Orientation string, it assumes we are in LPS. This is not generic and will show the incorrect values if the volume is RAS, for example. We need to be very explicit what this helper is for.

Also its weird to me that have F and H (foot and head?) for Inferior (I) and Superior (S). Was there a good reason for this originally @swederik ?

JamesAPetts (changes requested):
Some minor comments.

packages/core/src/RenderingEngine/StackViewport.ts
sedghi: yeah I thought about this after I logged off for the day, I will take care of it

packages/core/src/RenderingEngine/StackViewport.ts
sedghi: if undefined, it will get the range of pixels in the setVOIRange

packages/core/src/requestPool/imageRetrievalPoolManager.ts
sedghi: It has changed for a long time, loading and retrieval has been separated to two different pools since we want to send request as soon as possible, but decoding is another story

JamesAPetts (approved):

PR #83, opened 2022-04-22
Fix/camera events

sedghi:
- change the camera interface to include flipHorizontal and flipVertical
- move the rotation to the camera modified event detail
- remove the flipHorizontal and flipVertical properties from the setProperties and add it to the setCamera
- fix the CPU rendering for the non-scaled PT images 
- trigger camera modified for the setRotation and setFlip methods
- fix the tools to use the constructor toolName to avoid problem with inherited tools
- fix the windowLevel tool to work for non-scaled PT images


swederik (approved):

JamesAPetts (approved):
LGTM

PR #84, opened 2022-04-26
feat: Add annotation completed event 

sedghi:
Add annotation completed event


swederik (approved):

JamesAPetts (changes requested):
The stuff we got here is great for normal completion, but we've missed cancel.

Annotations persist on cancel, and we then still need to trigger the measurement completed event.

JamesAPetts (approved):
LGTM :)

xiaoxiao5:
The ANNOTATION tool does not trigger events such as ANNOTATION COMPLETED
image
cornerstone3D-beta-1.1.5
@sedghi

PR #85, opened 2022-04-27
feat: Add worldToImage and imageToWorld utilities

sedghi:
add worldToImage and imageToWorld utilities

### DIFF
--- a/packages/core/src/RenderingEngine/StackViewport.ts
+++ b/packages/core/src/RenderingEngine/StackViewport.ts
@@ -1004,8 +1004,8 @@ class StackViewport extends Viewport implements IStackViewport {
     const xVoxels = image.columns;
     const yVoxels = image.rows;
 
-    const zSpacing = imagePlaneModule.sliceThickness || EPSILON;
-
+    // Note: For rendering purposes, we use the EPSILON as the z spacing.
+    const zSpacing = EPSILON;
     const zVoxels = 1;
 
     const numComps =
--- a/packages/core/src/RenderingEngine/StackViewport.ts
+++ b/packages/core/src/RenderingEngine/StackViewport.ts
@@ -1004,7 +1004,8 @@ class StackViewport extends Viewport implements IStackViewport {
     const xVoxels = image.columns;
     const yVoxels = image.rows;
 
-    const zSpacing = image.sliceThickness || EPSILON;
+    const zSpacing = imagePlaneModule.sliceThickness || EPSILON;
+
     const zVoxels = 1;
 
     const numComps =
--- a/packages/core/src/RenderingEngine/StackViewport.ts
+++ b/packages/core/src/RenderingEngine/StackViewport.ts
@@ -1005,6 +1005,9 @@ class StackViewport extends Viewport implements IStackViewport {
     const yVoxels = image.rows;
 
     // Note: For rendering purposes, we use the EPSILON as the z spacing.
+    // This is purely for internal implementation logic since we are still
+    // technically rendering 3D objects with vtk.js, but the abstracted intention
+    //  of the stack viewport is to render 2D images
     const zSpacing = EPSILON;
     const zVoxels = 1;
 
--- a/packages/core/src/utilities/imageToWorldCoords.ts
+++ b/packages/core/src/utilities/imageToWorldCoords.ts
@@ -2,29 +2,19 @@ import { vec3 } from 'gl-matrix';
 import { metaData } from '..';
 import { Point2, Point3 } from '../types';
 
-type Options = {
-  imageId?: string;
-};
-
 /**
- * Given the 2d coordinates on the image space with [0,0] being the top left corner
+ * Given the imageId and a 2d coordinates on the image space with [0,0] being the top left corner
  * of the top left pixel, and options which includes the imageId, it returns the
  * 3d coordinates on the world space.
+ * @param imageId - The image id
  * @param imageCoords - The 2d coordinates on the image
- * @param options - options which includes the imageId
  * @returns The 3d coordinates on the world.
  *
  */
 export default function imageToWorldCoords(
-  imageCoords: Point2,
-  options: Options
+  imageId: string,
+  imageCoords: Point2
 ): Point3 | undefined {
-  const { imageId } = options;
-
-  if (!imageId) {
-    throw new Error('imageId is required for the imageToWorldCoords function');
-  }
-
   const imagePlaneModule = metaData.get('imagePlaneModule', imageId);
 
   if (!imagePlaneModule) {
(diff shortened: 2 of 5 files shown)


packages/core/src/utilities/imageToWorldCoords.ts
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * Given the 2d coordinates on the image space with [0,0] being the top left corner
+ * of the top left pixel, and options which includes the imageId, it returns the
+ * 3d coordinates on the world space.
+ * @param imageCoords - The 2d coordinates on the image
+ * @param options - options which includes the imageId
+ * @returns The 3d coordinates on the world.
+ *
+ */
+export default function imageToWorldCoords(
```
JamesAPetts, line 18: Tripple checked and logic looks correct to me.

packages/core/src/utilities/imageToWorldCoords.ts
```diff
@@ -0,0 +1,64 @@
...
+  imageCoords: Point2,
+  options: Options
+): Point3 | undefined {
+  const { imageId } = options;
```
JamesAPetts, line 22: It feels weird to me that this is in options, things in options I expect to have a default or suchlike.

You will always need an imageId for this method, so I think it should be (imageId, imageCoords)?

packages/core/src/utilities/worldToImageCoords.ts
```diff
@@ -0,0 +1,133 @@
...
+import { Point2, Point3 } from '../types';
+
+type Options = {
+  imageId?: string;
+};
+
+/**
+ * Given the 3d coordinates on the world space, and options which includes the imageId
+ * it returns the image coordinates (IJ) on the image space. The image space is
+ * defined with [0,0] being on the top left corner of the first top left pixel,
```
JamesAPetts, line 12: "first top left" is kind of degenerate with just "top left".

packages/core/src/utilities/worldToImageCoords.ts
```diff
@@ -0,0 +1,133 @@
...
+ * @param worldCoords - The 3d coordinates on the world.
+ * @param options - options which includes the imageId
+ * @returns The 2d coordinates on the image.
+ *
+ */
+function worldToImageCoords(
+  worldCoords: Point3,
+  options: Options
+): Point2 | undefined {
+  const { imageId } = options;
```
JamesAPetts, line 23: Again, think imageId should be an explicit parameter of the API.

packages/core/src/utilities/worldToImageCoords.ts
```diff
@@ -0,0 +1,133 @@
...
+    imageCoords[0] >= columns ||
+    imageCoords[1] < 0 ||
+    imageCoords[1] >= rows
+  ) {
+    throw new Error(
+      `The image coordinates are outside of the image, imageCoords: ${imageCoords}`
+    );
+  }
+
+  return Array.from(imageCoords) as Point2;
```
JamesAPetts, line 130: So this returns a float image coordinate rather than a floored pixel. I think this is what we want, but we should make clear to document this.

packages/core/src/utilities/worldToImageCoords.ts
```diff
@@ -0,0 +1,133 @@
...
+    0,
+    0,
+    1
+  );
+
+  // The scale matrix for the world to image coordinates
+  const scale = mat4.create();
+  mat4.fromScaling(
+    scale,
+    vec3.fromValues(columnPixelSpacing, rowPixelSpacing, sliceThickness)
```
JamesAPetts, line 92: Why do we need the slice thickness? We are going to be only mapping X and Y points. And as stated before the slice thickness does not help us here.

packages/core/src/utilities/worldToImageCoords.ts
```diff
@@ -0,0 +1,133 @@
...
+ * it returns the image coordinates (IJ) on the image space. The image space is
+ * defined with [0,0] being on the top left corner of the first top left pixel,
+ * the [1,1] being on the bottom right corner of the first top left pixel.
+ * @param worldCoords - The 3d coordinates on the world.
+ * @param options - options which includes the imageId
+ * @returns The 2d coordinates on the image.
+ *
+ */
+function worldToImageCoords(
+  worldCoords: Point3,
```
JamesAPetts, line 20: My only concern with this method is unlike the very simiple image to world, this calculates an entire matrix first, which is then re-usable.


If we call this for 1000 points, the matrix will be calculated again and again for each point.

Maybe both methods should instead take an array of points and map them all?

imageToWorldCoords would just be a loop, but worldToImageCoords could then re-use the matrix?

packages/core/src/utilities/worldToImageCoords.ts
```diff
@@ -0,0 +1,133 @@
...
+  // The origin is the image position patient, but since image coordinates start
+  // from [0,0] for the top left hand of the first pixel, and the origin is at the
+  // center of the first pixel, we need to account for this.
+  const newOrigin = vec3.create();
+
+  vec3.scaleAndAdd(newOrigin, origin, columnCosines, -columnPixelSpacing / 2);
+  vec3.scaleAndAdd(newOrigin, newOrigin, rowCosines, -rowPixelSpacing / 2);
+
+  // Translation matrix for the world to image coordinates
+  const translationMatrix = mat4.create();
```
JamesAPetts, line 58: I wonder if the following is easier to the reader and semantically the same, since we are only doing a very simple transformation, using a general matrix:

For x:
- calculate the distance in the x (rowCosines) direction between the point and the `newOrigin`, Divide this by the rowPixelSpacing.
for y:
- calculate the distance in y (columnCosines) direction between the point and the `newOrigin`, Divide this by the columnPixelSpacing.

This will result in a much shorter, simpler calculation, in a similar vein to the imageToWorldCoords calculation, which might void my other point about generalising these methods for multiple points.

JamesAPetts (changes requested):
A few points we should discuss before continuing here :).

packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -1004,7 +1004,8 @@ class StackViewport extends Viewport implements IStackViewport {
     const xVoxels = image.columns;
     const yVoxels = image.rows;
 
-    const zSpacing = image.sliceThickness || EPSILON;
+    // Note: For rendering purposes, we use the EPSILON as the z spacing.
```
JamesAPetts, line 1007: Could we add a tiny bit more info here?

Maybe something like "Note: For rendering purposes, we use the EPSILON as the z spacing. This is purely for internal implementation logic since we are still technically rendering 3D objects with vtk.js, but the asbtracted intention of the stack viewport is to render 2D images"

Or better wording if you can come up with it ^_^.

JamesAPetts (approved):
Perfect, the imageToWorldCoords is much simpler now.
One minor comment about... comments. But otherwise LGTM.

PR #86, opened 2022-04-29
feat: Add toolGroup specific configuration for tools

sedghi:


packages/tools/src/tools/base/AnnotationTool.ts
```diff
@@ -402,8 +402,27 @@ abstract class AnnotationTool extends BaseTool {
...
+    // if the toolGroup setting exists in the custom settings.
+    // We check here if the requested property directly exists on the
+    // Object settings (passed in as argument) if not AND toolGroup settings has that
+    // property we use that. Otherwise, since toolGroup settings is a custom settings
+    // with no fallback to runtime and default settings, if the toolGroupSettings don't
+    // have that property we fallback to the Object Setting since it has the prototype chain
+    // which includes the Runtime and then Default and there is ALWAYS a value in
+    // default for the property.
+    // The order for taking the styles is Annotation Object Setting > ToolGroup Setting > Runtime Setting > Default Settings
+    const toolGroupSettings = Settings.getCustomSettings(this.toolGroupId);
```
JamesAPetts, line 416: Since we are just using a custom settings map and hoping someone put a toolgroup tag, this is extremely cryptic without diving into the code. I think we really need some docs explaining how this fits in the hierarchy of settings state/fallbacks for styling?

packages/tools/src/tools/base/AnnotationTool.ts
```diff
@@ -402,8 +402,27 @@ abstract class AnnotationTool extends BaseTool {
     property: string,
     annotation?: Annotation
   ): unknown {
+    // Since we would like to have a toolGroup settings and cornerstone core Settings
```
JamesAPetts, line 405: in the api contract:

```
settings: Settings,
property: string,
annotation?: Annotation
```
It feels a bit weird to me that we pass in a settings that we may then not use. I think it may be easier for the consumer above to understand the following:

```
property: string,
annotation?: Annotation
```

Then, if annotation is present, we can get the annotation scoped settings via: 


```
const settings = Settings.getObjectSettings(
    annotation,
    this.getToolName()
);
```


And the toolGroupSettings can still be got via `this.toolGroupId`.

packages/tools/src/tools/base/AnnotationTool.ts
```diff
@@ -402,8 +402,27 @@ abstract class AnnotationTool extends BaseTool {
...
+    // property we use that. Otherwise, since toolGroup settings is a custom settings
+    // with no fallback to runtime and default settings, if the toolGroupSettings don't
+    // have that property we fallback to the Object Setting since it has the prototype chain
+    // which includes the Runtime and then Default and there is ALWAYS a value in
+    // default for the property.
+    // The order for taking the styles is Annotation Object Setting > ToolGroup Setting > Runtime Setting > Default Settings
+    const toolGroupSettings = Settings.getCustomSettings(this.toolGroupId);
+    const toolStyleProperty = `tool.style.${property}`;
+    const settingsToUse =
+      toolGroupSettings.hasOwn(toolStyleProperty) &&
```
JamesAPetts, line 419: Maybe I'm going in circles, but didn't we agree that exact annotation scoped toolstyle is more granular and takes precidence over toolGroup styling?

priority left to right:
i.e annotation > toolGroup > tool > global?

JamesAPetts (changes requested):
Few comments, can discuss later.

sedghi:
Closing in favor of re-doing the `Settings` re-architecture later as its limitations is becoming more evident

PR #88, opened 2022-04-29
feat: Add ArrowTool and remove toolName from drawing API

sedghi:


packages/tools/src/tools/annotation/ArrowTool.ts
```diff
@@ -0,0 +1,695 @@
...
+          viewportIdsToRender
+        );
+        this.editData = null;
+        this.isDrawing = false;
+        return;
+      }
+      annotation.data.text = text;
+    }
+
+    triggerAnnotationRenderForViewportIds(renderingEngine, viewportIdsToRender);
```
JamesAPetts, line 340: Can't you move these down below the next if block, so you can combine both the `if (newAnnotation)` blocks into one if block?

JamesAPetts (approved):
LGTM, one minor comment.

packages/tools/src/tools/annotation/ArrowTool.ts
sedghi: done

PR #89, opened 2022-04-29
Planar freehand roi tool

JamesAPetts:
WIP Planar Freehand ROI Tool.

- Still need to fix some issues so the api docs will build.
- One crossover issue left/interaction issues to fix.

More features to come in a different PR.

**Live example:** https://626f9a9da152cf7e6d378757--cs3d-docs.netlify.app/live-examples/planarfreehandroitool

Deploy preview: https://626f9a9da152cf7e6d378757--cs3d-docs.netlify.app/


packages/tools/src/tools/annotation/PlanarFreehandROITool.ts
```diff
@@ -0,0 +1,446 @@
...
+  InteractionTypes,
+} from '../../types';
+import { PlanarFreehandROIAnnotation } from '../../types/ToolSpecificAnnotationTypes';
+import { PlanarFreehandROICommonData } from '../../utilities/math/polyline/planarFreehandROIInternalTypes';
+
+const { pointCanProjectOnLine } = polyline;
+
+/**
+ * PlanarFreehandROITool lets you draw annotations that define an arbitrarily drawn region.
+ * You can use the PlanarFreehandROITool in all perpendicular views (axial, sagittal, coronal),
```
sedghi, line 41: Is the comment correct? Does it support all perpendicular views?

packages/tools/src/tools/annotation/PlanarFreehandROITool.ts
```diff
@@ -0,0 +1,446 @@
...
+    toolProps: PublicToolProps = {},
+    defaultToolProps: ToolProps = {
+      supportedInteractionTypes: ['Mouse', 'Touch'],
+      configuration: {
+        shadow: true,
+        preventHandleOutsideImage: false,
+        allowOpenContours: true,
+        // Radius in canvas coordinates used to join contours.
+        closeContourProximity: 10,
+        subPixelResolution: 4,
```
sedghi, line 132: can we add comment on subPixelResolution?

packages/tools/src/utilities/math/polyline/getSpacingAndXYDirections.ts
```diff
@@ -0,0 +1,86 @@
...
+const EPSILON = 1e-3;
+
+/**
+ * Gets the desired spacing for points in the polyline for the
+ * `PlanarFreehandROITool` in the x and y canvas directions, as well as
+ * returning these canvas directions in world space.
+ */
+const getSpacingAndXYDirections = (
+  viewport: Types.IStackViewport | Types.IVolumeViewport,
+  subPixelResolution: number
```
sedghi, line 14: Needs documentation for the function parameters

packages/tools/src/tools/annotation/planarFreehandROITool/openContourEditLoop.ts
```diff
@@ -0,0 +1,548 @@
...
+  const { viewport } = enabledElement;
+
+  const prevCanvasPoints = annotation.data.polyline.map(viewport.worldToCanvas);
+
+  const { spacing, xDir, yDir } = getSpacingAndXYDirections(
+    viewport,
+    this.configuration.subPixelResolution
+  );
+
+  this.commonEditData = {
```
sedghi, line 39: we use this.editData elsewhere in the annotation tools, is this `commonEditData` different than `editData`?

packages/tools/src/tools/annotation/planarFreehandROITool/openContourEditLoop.ts
```diff
@@ -0,0 +1,548 @@
...
+
+  const eventDetail = evt.detail;
+  const { currentPoints, element } = eventDetail;
+  const canvasPos = currentPoints.canvas;
+  const enabledElement = getEnabledElement(element);
+  const { viewport } = enabledElement;
+
+  const prevCanvasPoints = annotation.data.polyline.map(viewport.worldToCanvas);
+
+  const { spacing, xDir, yDir } = getSpacingAndXYDirections(
```
sedghi, line 34: this is not spacing as I understand, it is spacing/subPixelResolution, can we name it differently so that it does not get confused with image spacings?

packages/tools/src/utilities/math/polyline/planarFreehandROIInternalTypes.ts
```diff
@@ -0,0 +1,30 @@
+import type { Types } from '@cornerstonejs/core';
+import { PlanarFreehandROIAnnotation } from '../../../types/ToolSpecificAnnotationTypes';
+
+type PlanarFreehandROIDrawData = {
```
sedghi, line 4: are these files exported to the app? if yes, we need documentations so that it appears correctly at https://www.cornerstonejs.org/api/tools/namespace/Types

packages/tools/src/tools/annotation/planarFreehandROITool/openContourEditLoop.ts
```diff
@@ -0,0 +1,548 @@
...
+  }
+
+  const numPointsAdded = addCanvasPointsToArray(
+    element,
+    editCanvasPoints,
+    canvasPos,
+    this.commonData
+  );
+
+  const currentEditIndex = editIndex + numPointsAdded;
```
sedghi, line 138: I first thought `editIndex` is the index of the contour being edited, but apparently it is not ...? and it is related the the number of points recently added somehow.

packages/tools/src/tools/annotation/planarFreehandROITool/editLoopCommon.ts
```diff
@@ -0,0 +1,278 @@
...
+/**
+ * Check if the `editCanvasPoints` have crossed the `prevCanvasPoints` during
+ * an edit.
+ *
+ * @privateRemarks The following tricks are required to make the UX smooth and
+ * the editing not very picky on exactly where you click:
+ * - If we don't cross after 2 points, but projecting the line backwards the
+ * proximity distance means we cross, extend the line back.
+ * - If we travel the full proximity in canvas points but don't cross a line, we
+ * are likely drawing along the line, which is intutive to the user. At this point
```
sedghi, line 17: typo

packages/tools/src/tools/annotation/planarFreehandROITool/openContourEditLoop.ts
sedghi: So is it like you associate an index to each point being added and you use it for crossing etc? I think I see what you have done

packages/tools/src/utilities/math/polyline/getIntersectionWithPolyline.ts
```diff
@@ -0,0 +1,182 @@
...
+
+  if (orientationValue === 0) {
+    return 0; // Colinear
+  }
+
+  return orientationValue > 0 ? 1 : 2;
+}
+
+/**
+ * Checks if point `q` lines on the segment (`p`,`r`).
```
sedghi, line 167: lies?

packages/tools/src/tools/annotation/planarFreehandROITool/editLoopCommon.ts
```diff
@@ -0,0 +1,278 @@
...
+    canvasPos,
+    lastCanvasPoint,
+    isClosedContour
+  );
+
+  if (!crossedLineSegment) {
+    return false;
+  }
+
+  return true;
```
sedghi, line 176: maybe I'm wrong, but the function logic is, if it doesn't not have first crossing it has second crossing? what about none?

packages/tools/src/utilities/math/polyline/calculateAreaOfPoints.ts
```diff
@@ -0,0 +1,19 @@
+import type { Types } from '@cornerstonejs/core';
+
+/**
+ * cCalculates the area of an array of `Point2` points using the shoelace algorithm.
+ */
+export default function calculateAreaOfPoints(points: Types.Point2[]): number {
```
sedghi, line 6: what is the unit for this return? we should document its unit

packages/tools/src/utilities/math/polyline/pointCanProjectOnLine.ts
```diff
@@ -0,0 +1,52 @@
+import type { Types } from '@cornerstonejs/core';
+import { vec2 } from 'gl-matrix';
+
+/**
+ * Returns `true` if the point `p` can project onto point (`p1`, `p2`), and if
+ * this projected point is less than `proximity` units away.
+ */
+const pointCanProjectOnLine = (
```
sedghi, line 8: The function name does not match its output, I expected a boolean output, but it is number | false

packages/tools/src/tools/annotation/PlanarFreehandROITool.ts
```diff
@@ -0,0 +1,446 @@
...
+
+    // NOTE: It is implemented this way so that we do not double calculate
+    // points when number crunching adjacent line segments.
+    let previousPoint = viewport.worldToCanvas(points[0]);
+
+    for (let i = 1; i < points.length; i++) {
+      const p1 = previousPoint;
+      const p2 = viewport.worldToCanvas(points[i]);
+
+      const distance = pointCanProjectOnLine(canvasCoords, p1, p2, proximity);
```
sedghi, line 299: `distance = pointCanProjectOnLine(` seems weird, since it is used later as a boolean check maybe do `canProject` or just `if (pointCanProject...)`

packages/tools/src/tools/annotation/PlanarFreehandROITool.ts
```diff
@@ -0,0 +1,446 @@
...
+    if (annotation.data.isOpenContour) {
+      // Contour is open, don't check last point to first point.
+      return false;
+    }
+
+    // check last point to first point
+    const pStart = viewport.worldToCanvas(points[0]);
+    const pEnd = viewport.worldToCanvas(points[points.length - 1]);
+
+    const distance = pointCanProjectOnLine(
```
sedghi, line 317: same here

packages/tools/src/tools/annotation/planarFreehandROITool/renderMethods.ts
```diff
@@ -0,0 +1,300 @@
...
+): void {
+  const { viewport } = enabledElement;
+
+  const settings = Settings.getObjectSettings(annotation, this.getToolName());
+
+  // Todo -> Its unfortunate that we have to do this for each annotation,
+  // Even if its unchanged. Perhaps we should cache canvas points per element
+  // on the tool? That feels very weird also as we'd need to manage it/clean
+  // them up.
+  const canvasPoints = annotation.data.polyline.map((worldPos) =>
```
sedghi, line 43: Yes I agree, this was not problem in other annotation tools ....

packages/tools/src/tools/annotation/planarFreehandROITool/renderMethods.ts
```diff
@@ -0,0 +1,300 @@
...
+
+  const options = {
+    color: color === undefined ? undefined : <string>color,
+    width: lineWidth === undefined ? undefined : <number>lineWidth,
+    connectLastToFirst: false,
+  };
+
+  const polylineUID = '1';
+
+  drawPolylineSvg(
```
sedghi, line 100: The logic for drawPolyLineSVG is almost the same for open/closed contour, can you extract them both in a common logic which accepts `connectLastToFirst`?

packages/tools/src/tools/annotation/PlanarFreehandROITool.ts
```diff
@@ -0,0 +1,446 @@
...
+
+    if (!annotations?.length) {
+      return;
+    }
+
+    const isDrawing = this.isDrawing;
+    const isEditingOpen = this.isEditingOpen;
+    const isEditingClosed = this.isEditingClosed;
+
+    if (!(isDrawing || isEditingOpen || isEditingClosed)) {
```
sedghi, line 404: Which case it is? so it is not drawing, and not editing? does it means it is on hover? can we add that comment?

packages/tools/src/tools/annotation/planarFreehandROITool/renderMethods.ts
```diff
@@ -0,0 +1,300 @@
...
+  const lineWidth = this.getStyle(settings, 'lineWidth', annotation);
+  // const lineDash = this.getStyle(settings, 'lineDash', annotation);
+  const color = this.getStyle(settings, 'color', annotation);
+
+  const options = {
+    color: color === undefined ? undefined : <string>color,
+    width: lineWidth === undefined ? undefined : <number>lineWidth,
+  };
+
+  drawPolylineSvg(
```
sedghi, line 156: again this shares the same logic up to here

packages/tools/src/tools/annotation/planarFreehandROITool/renderMethods.ts
```diff
@@ -0,0 +1,300 @@
...
+
+  const options = {
+    color: color === undefined ? undefined : <string>color,
+    width: lineWidth === undefined ? undefined : <number>lineWidth,
+    connectLastToFirst: true,
+  };
+
+  const polylineUIDToRender = 'preview-1';
+
+  drawPolylineSvg(
```
sedghi, line 233: same

packages/tools/src/tools/annotation/planarFreehandROITool/renderMethods.ts
```diff
@@ -0,0 +1,300 @@
...
+
+  const options = {
+    color: color === undefined ? undefined : <string>color,
+    width: lineWidth === undefined ? undefined : <number>lineWidth,
+    connectLastToFirst: false,
+  };
+
+  const polylineUIDToRender = 'preview-1';
+
+  drawPolylineSvg(
```
sedghi, line 273: same

sedghi:
Looks like an awesome tool. Had couple of comments, please see them.

More general comment:
- Color green on mouse hover gives an impression of edit mode, and that is correct. But for a newly drawn contour that is not the case, even if the mouse is so far it is still green (since it is in the selected mode). How about we set the contour as yellow (unselected) when it is finished drawing? 


https://user-images.githubusercontent.com/7490180/166289489-d41e92f0-e204-4d2f-9b70-aa3d1c802784.mp4

JamesAPetts:
Hmmmm. For your last comment/video, I think this is more of an issue of our default color selection for Cornerstone3D. All Cornerstone3D annotations currently go to selected state after being drawn, which I think makes sense?

We are using the same color for selected and highlighted, when they mean semantically different things, for _all_ annotations, you can only grab and interact in the selected state.

Maybe our default on CS3D should have a different color/shared for selected, we don't demonstrate this well at all.

packages/tools/src/drawingSvg/drawPolyline.ts
```diff
@@ -0,0 +1,85 @@
...
+
+/**
+ * Draws an SVG polyline with the given points.
+ *
+ * The `connectLastToFirst` option, if true, draws a closed polyline, with the
+ * last point connected to the first.
+ */
+export default function drawPolyline(
+  svgDrawingHelper: any,
+  toolName: string,
```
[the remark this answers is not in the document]
JamesAPetts: I'll make this change once your PR is merged.

packages/tools/src/types/ToolSpecificAnnotationTypes.ts
```diff
@@ -173,3 +173,35 @@ export interface RectangleROIStartEndThresholdAnnotation extends Annotation {
...
+      activeHandleIndex: number | null;
+      textBox: {
+        hasMoved: boolean;
+        worldPosition: Types.Point3;
+        worldBoundingBox: {
+          topLeft: Types.Point3;
+          topRight: Types.Point3;
+          bottomLeft: Types.Point3;
+          bottomRight: Types.Point3;
+        };
```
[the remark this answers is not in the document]
JamesAPetts: It doesn't yet as that isn't in scope for my client's piece of work at this moment. I believe. However its here because it inherits from Annotation, and it would be more painful to make it not inherit and be different from everything else for now. I'll add a comment, I think it will come in a future PR at some point.

packages/tools/src/tools/annotation/PlanarFreehandROITool.ts
JamesAPetts: Yeah, the only reason it doesn't support obliques actually is just because of the implementation in `getSpacingAndXYDirections`, which could late be improved.

When I say all perpendicular views, I mean the same as all other planar annotations. These are just planar contours for now, so no rendering of planar annotations drawn in one view, in another view, same as e.g. Length.

JamesAPetts: I'll clarify the comment.

packages/tools/src/tools/annotation/planarFreehandROITool/openContourEditLoop.ts
JamesAPetts: Yeah, its derived from image spacing, but is a different concept. Renamed.

packages/tools/src/utilities/math/polyline/getSpacingAndXYDirections.ts
JamesAPetts: Done.

packages/tools/src/tools/annotation/PlanarFreehandROITool.ts
JamesAPetts: Done.

packages/tools/src/utilities/math/polyline/planarFreehandROIInternalTypes.ts
JamesAPetts: They are internal types only used by the drawing loop. I was unsure where to put them. Any ideas? I'll write a comment that they are internal for now.

packages/tools/src/tools/annotation/planarFreehandROITool/openContourEditLoop.ts
JamesAPetts: This is because I had commonEdit data and some edit data specific to the type of edit. I've actually changed this, so it can be edit data now. Will change.

packages/tools/src/tools/annotation/planarFreehandROITool/openContourEditLoop.ts
JamesAPetts: Its the index of the (invisible) edit line being used to generate the edit preview. Analogous to `handleIndex` of BidirectionalTool's `editData` pointing to its own handle, but for a different line used for calculation. I'll add some comments.

packages/tools/src/tools/annotation/planarFreehandROITool/editLoopCommon.ts
```diff
@@ -0,0 +1,278 @@
+import { polyline } from '../../../utilities/math';
+import { vec2 } from 'gl-matrix';
+import { EventTypes } from '../../../types';
+import { Types } from '@cornerstonejs/core';
```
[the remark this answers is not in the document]
JamesAPetts: Sure, I'll do this for all files in the PR.

packages/tools/src/tools/annotation/planarFreehandROITool/editLoopCommon.ts
JamesAPetts: check for second crossing only checks for the second crossing (i.e. end of the edit line). This method is not even called if there is no first crossing yet.

Maybe this is just due to some confusing of the `getFirstIntersectionWithPolyline`, which is a generic helper actually. The helper returns the _first crossing of the lines given_. The parameters given to this helper inside the method are to try to find the second crossing.

I'll add a comment to clarify this. I see how it can be confusing.

packages/tools/src/utilities/math/polyline/calculateAreaOfPoints.ts
JamesAPetts: In whatever units the points are in. Right now this is used for a calculation on canvas (on the fuse contour logic), but I can see it being used later for area calculation in other units. I'll doc this.

packages/tools/src/tools/annotation/planarFreehandROITool/renderMethods.ts
JamesAPetts: This is comment for the future. This will happen for any significantly large annotation. Its an optimisation we don't have time to tackle here yet.

packages/tools/src/tools/annotation/planarFreehandROITool/editLoopCommon.ts
```diff
@@ -0,0 +1,278 @@
...
+    lastCanvasPoint,
+    isClosedContour
+  );
+
+  if (crossedLineSegment) {
+    this.commonEditData.startCrossingIndex = crossedLineSegment[0];
+
+    // On the first crossing, remove the first lines prior to the crossing
+    this.removePointsUpUntilFirstCrossing(isClosedContour);
+  } else if (editCanvasPoints.length > 6) {
```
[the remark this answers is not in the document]
JamesAPetts: Its actually a magic number, happens to be the proximity as it felt good. I'll move it to the configuration.

packages/tools/src/tools/annotation/planarFreehandROITool/renderMethods.ts
JamesAPetts: I just tried doing this and its awkward as the methods at the end of renderOpenContour actually requires all the logic you could refactor out (also needs the canvasPoints, options, styles, etc). I could do it but it would require recalcation of world points etc.

To all your points, I've refactored out the settings + options bit, which is common, combining these two is not so useful though.

JamesAPetts:
Worked on some comment, still need to resolve more.

packages/tools/src/tools/annotation/PlanarFreehandROITool.ts
JamesAPetts: The renderAnnotation callback renders all tools, this just says render them all as tooldata, and there will be no previews, which need different render paths, will add a comment.

sedghi (approved):
LGTM

PR #90, opened 2022-05-02
feat hidden annotations

ladeirarodolfo:
It includes
- isVisible property on annotation (Implementation based on locking feature)


packages/tools/src/tools/annotation/RectangleROITool.ts
```diff
@@ -631,6 +631,11 @@ export default class RectangleROITool extends AnnotationTool {
 
     for (let i = 0; i < annotations.length; i++) {
       const annotation = annotations[i] as RectangleROIAnnotation;
+
+      if (isAnnotationHidden(annotation)) {
```
sedghi, line 635: hmmm, I think this skipping will make a problem for the cachedStats calculations and they become non-synced with the other targets. We should only skip the rendering portion of the `renderAnnotation` logic

packages/tools/src/stateManagement/annotation/annotationHide.ts
```diff
@@ -0,0 +1,178 @@
+import { eventTarget, triggerEvent } from '@cornerstonejs/core';
+import { Events } from '../../enums';
+import { Annotation } from '../../types';
+import { AnnotationHideChangeEventDetail } from '../../types/EventTypes';
+
+/*
+ * Constants
+ */
+const globalHiddenAnnotationsSet: Set<Annotation> = new Set();
```
sedghi, line 9: I know you did the same logic as locking, but we don't need the full `annotation` for hiding it - It will cause a lot of headache in the consumer as it needs the full annotation - `AnnotationUID` is enough in fact. I have revised the annotaiton selection, and fixing locking was on my todo list. Follow the `annotationSelection` I would recommend

sedghi (changes requested):
Added some comments, mainly

1) We shouldn't skip the whole cachedStats calculation for the annotation, only the rendering 
2) Hidying API can work with the AnnotationUID too, we shouldn't require the full annotation as arguments
3) `RectangleROIThresholdTool` & `RectangleROIStartEndThresholdTool` should also implement the hiding as they are AnnotationTools (they extend RectangleROITool)

packages/tools/src/stateManagement/annotation/annotationHide.ts
JamesAPetts: Agreed, this should use UIDs. In practice we found you always have the UID, and in this old style used to always call a get of the annotation from the UID, before passing it in to these methods.

JamesAPetts (changes requested):
Shouldn't we use `isVisible` like we did in Cornerstone Legacy, rather than the inverse "hidden"/"isHidden"? As such the module would then be the visibility module. Annotations should have a `visible` option that is added as true by default?

As for the CS3D perspective:
- On the one hand, we use visible for labelmaps, so should use the same nomencleture for annotations?
- On the other hand, hidden behaves more like locked, so its a good flag to have?



I haven't done a full review for this reason, as I would be suggesting e.g. ANNOTATION_HIDE_CHANGE => ANNOTATION_VISIBILITY_CHANGE etc. I can do a full review once we have agreement on the direction.


Otherwise I agree with Alireza's points.

packages/tools/src/stateManagement/annotation/annotationHide.ts
ladeirarodolfo: Ok I felts the same :D, I will change thanks

packages/tools/src/tools/annotation/RectangleROITool.ts
ladeirarodolfo: hm, but once the annotation is hidden it means anytime method enters here the annotation would have exactly the same value (cachedStats) as prior being hidden. The only moment I see it being out of sync is if when user is moving handle he press a short key, then it turns to hidden prior to updating the cachedStats: is that possible?

ladeirarodolfo: I can follow your input, Im just checking if we are doing unnecessary maths

sedghi: For any usecase that changes the metadata for the images (e.g., SUV re-calculations based on new tags) we need to update the stats, It is safe to update stats

ladeirarodolfo:
> Shouldn't we use `isVisible` like we did in Cornerstone Legacy, rather than the inverse "hidden"/"isHidden"? As such the module would then be the visibility module. Annotations should have a `visible` option that is added as true by default?
> 
> As for the CS3D perspective:
> 
> * On the one hand, we use visible for labelmaps, so should use the same nomencleture for annotations?
> * On the other hand, hidden behaves more like locked, so its a good flag to have?
> 
> I haven't done a full review for this reason, as I would be suggesting e.g. ANNOTATION_HIDE_CHANGE => ANNOTATION_VISIBILITY_CHANGE etc. I can do a full review once we have agreement on the direction.
> 
> Otherwise I agree with Alireza's points.

@sedghi  what do you think.
It totally agree with James and his perspectives capture exactly my questions. I followed approach 2, hidden act as locking, as it allows incrementally adding this feature with none (or last say few) regression possibility.

sedghi:
> > Shouldn't we use `isVisible` like we did in Cornerstone Legacy, rather than the inverse "hidden"/"isHidden"? As such the module would then be the visibility module. Annotations should have a `visible` option that is added as true by default?
> > As for the CS3D perspective:
> > 
> > * On the one hand, we use visible for labelmaps, so should use the same nomencleture for annotations?
> > * On the other hand, hidden behaves more like locked, so its a good flag to have?
> > 
> > I haven't done a full review for this reason, as I would be suggesting e.g. ANNOTATION_HIDE_CHANGE => ANNOTATION_VISIBILITY_CHANGE etc. I can do a full review once we have agreement on the direction.
> > Otherwise I agree with Alireza's points.
> 
> @sedghi what do you think? It totally agree with James and his perspectives capture exactly my questions. I followed approach 2, hidden act as locking, as it allows incrementally adding this feature with none (or lets say few) regression possibilities.

I think `visible` default to true makes a lot of sense to me (as per visibility controller for segmentation too)

packages/tools/src/tools/annotation/RectangleROITool.ts
ladeirarodolfo: ok

ladeirarodolfo:
@sedghi @JamesAPetts  I did the changes proposed, can you take a look again?

sedghi (changes requested):
I thought we decided to use the `visible` approach, I see a lot of `Hidden` state and hide controllers. Can you please double check each file you have changed?

common/reviews/api/tools.api.md
```diff
@@ -177,6 +179,8 @@ enum AnnotationStyleStates {
     // (undocumented)
     Default = "",
     // (undocumented)
+    Hidden = "Hidden",
```
wayfarer3130, line 182: Still a hidden here.

common/reviews/api/tools.api.md
```diff
@@ -216,6 +220,17 @@ export abstract class AnnotationTool extends BaseTool {
     abstract toolSelectedCallback(evt: EventTypes_2.MouseDownEventType, annotation: Annotation, interactionType: InteractionTypes): void;
 }
 
+// @public (undocumented)
+type AnnotationVisibilityChangeEventDetail = {
+    added: Array<string>;
+    removed: Array<string>;
+    hidden: Array<string>;
```
wayfarer3130, line 227: Why both hidden and visible"

ladeirarodolfo: we agreed on using a var to represent the visible uids. But having both, for hidden and visible we can:
- Differentiate between being hidden or not existing
- We can do a show all, hide all
- The implementation is highly decoupled from the rest of code, which means we can easily tell how many are visible/hidden

wayfarer3130: I still don't understand what the difference is - it seems like hidden and visible are the same type of thing.

JamesAPetts: > We can do a show all, hide all.

Sure, but this is just toggling _one_ variable, as Bill says. I also don't see any benefit to having properties for these.

sedghi: @ladeirarodolfo I don't think I agree on having both hidden and visible, if that is what I communicated, I meant change everything to visible, visibility, isVisible

ladeirarodolfo: ok to move forward lets consider the api for it. If is that what we want I can proceed, having just visible/isVisible.
Just to bring to our attention: an not visible annotation data (based on utility isAnnotationVisible) does not mean its hidden (it could be inexistent) 

export {
  getAnnotationUIDsVisible,
  getAnnotationUIDsVisibleCount,
  setAnnotationVisibility,
  isAnnotationVisible, 
  checkAndDefineIsVisibleProperty,
};

sedghi: > an not visible annotation data (based on utility isAnnotationVisible) does not mean its hidden (it could be inexistent) . I say that because on getState method, things might be very trick and inconsistent as far as annotation state combination is concerned

I need more context for this sentence, can you please elaborate? Why does a non visible annotation not mean it is hidden?

ladeirarodolfo: Lets say by the context of this helper, it will only store visible uids.

If, for instance we invoke isAnnotationVisible(ABCUid) and this returns false.
It means:
1- We created ABCUid (it will be set as visible true) then we setVisibility false. Here isAnnotationVisible returning false is accurate.
2- There is no ABCUid annotation at all, isAnnotationVisible returning false is not consistent, because false might cause a wrong assumption the tool is hidden, but it actually does not exist at all (and we can not know that, if we only consider the proposed api).

Differently from locking, visible is actually the default state, but what we really want is to know if that is hidden or not, so with option 2 we might be lead to mistakes or false positives

JamesAPetts: There doesn't need to be a "default" state, its just that we should change all the tools `addAnnotation` callback so that when an annotation is created the visible tag is created, and ad `visible: boolean` as a mandatory field on the `Annotation` type. Then we can just avoid all of the complex logic, it is `visible` (`true`) or not (`false`).

ladeirarodolfo: An example for a possible wrong assumption would be the usage of getState method.

Lets say we change [this](packages/tools/src/stateManagement/annotation/config/getState.ts) to Visible instead.
Also for consistency lets say we only do the conditional `if(isVisible(ABCUID)` at end of clauses (prior to `AnnotationStyleStates.Default;`). So, the if there is no ABCUID, `isVisible(ABCUID)` is false, should we return here, should we go to other return (`AnnotationStyleStates.Default`)

ladeirarodolfo:
Branch name is not accurate. closing it

PR #93, opened 2022-05-05
feat: Add toolStyles and DragProbe

sedghi:
We will start by looking at the style hierarchy. The style hierarchy is as follow.

- Annotation-level settings (with UID) **set/getAnnotationToolStyle**
  - Viewport-level tool settings **set/getViewportToolStyle**
    - Per-tool this layer: Length on this viewport
    - Global this layer: All tools in this viewport
      - toolGroup settings (for any tool specified in this toolGroup in all viewports of the toolGroup) **set/getToolGroupToolStyle**
        - Per-tool layer: Angle on this toolGroup in all viewports
        - Global this layer: All tools in this toolGroup in all viewports
          - Default level: **set/getDefaultToolStyle**
            - Per-tool layer (Length) settings
            - Global (app-level) settings (we provide a default).

In annotation rendering loop, upon getting a style for a certain property (`color`, `lineDash`, `lineThickness`)
we check whether the style is set at the annotation level (highest priority).
If not, we check whether any viewport-level setting is set (for the viewport annotation is drawing on); however,
in the viewportLevel, we first check whether the tool-level setting is set. If not, we check in the "global" (all tools in the viewport) level.
If not found, we move to the next level for toolGroup level. If not found, we move to the next level for global level which is last
level to check.

Other stuff:
- Add Drag probe 
- Fix a bug where textlines changing would not appear


packages/tools/src/stateManagement/annotation/config/helpers.ts
```diff
@@ -0,0 +1,55 @@
...
+ */
+function getStyleProperty(
+  property: string,
+  StyleSpecifications: StyleSpecifications,
+  state?: AnnotationStyleStates,
+  mode?: ToolModes
+): string {
+  // `alternatives` is a list of property names with priority in ascending
+  // order like: ['color', 'colorSelected', 'colorSelectedActive']
+  // Thus, we attempt resolving property names in reverse order
```
JamesAPetts, line 42: We can avoid this reversing, and a comment why, if we just return them in the correct order from the helper, which I think would be so much simpler?

packages/tools/src/stateManagement/annotation/config/helpers.ts
```diff
@@ -0,0 +1,55 @@
...
+  const alternatives = getStyleAlternatives(property, state, mode);
```
JamesAPetts, line 43: Hmm, alternatives is a weird name. How about something like `getHierarchalPropertyStyles` or something? 🤔

JamesAPetts (approved):
Minor comments, LGTM, other than type renamings we talked about in meeting.

packages/tools/src/stateManagement/annotation/config/helpers.ts
sedghi: sure

packages/tools/src/stateManagement/annotation/config/helpers.ts
sedghi: it will include a lot of if/else statements based on the availability of the state and/or mode. I'll expand more on the comment and since it is internal I would say it is not that bad

JamesAPetts (changes requested):
Looks good other than the Drag Probe Tool which needs a rethink.

packages/tools/src/tools/annotation/DragProbeTool.ts
```diff
@@ -39,34 +44,138 @@ export default class DragProbeTool extends ProbeTool {
     super(toolProps, defaultToolProps);
   }
 
-  _mouseUpCallback = (
-    evt: EventTypes.MouseUpEventType | EventTypes.MouseClickEventType
-  ) => {
+  addNewAnnotation = (
```
JamesAPetts, line 47: This is almost perfect, my only qualm with this now is that since we don't add annotations with this tool, we should swap this callback to `postMouseDownCallback`? Using `addNewAnnotation` implies something will be added to state.

JamesAPetts (changes requested):
One change, lets discuss on zoom later today.

PR #94, opened 2022-05-05
Feat/annotation visibility

ladeirarodolfo:
It includes
- Annotation Visibility feat

IMPORTANT
- Opened a new PR because the old one has a totally wrong branch name.
- I have done compared to other PR:
- Covered the changes we discussed on chat
- Added documentation to some methods and also md files
- Added changes to Freehand tool

### DIFF
--- a/packages/docs/docs/concepts/cornerstone-tools/annotation/visibility.md
+++ b/packages/docs/docs/concepts/cornerstone-tools/annotation/visibility.md
@@ -1,0 +1,42 @@
+---
+id: visibility
+title: Visibility
+---
+
+# Visibility
+
+Annotations can have their visibility changed. You can use
+the visibility API to show/hide annotations.
+
+## API
+
+There are various APIs for showing and hiding annotations along with get/set methods
+
+```js
+import { annotations } from '@cornerstonejs/tools';
+
+// changing an annotation visibility to be visible (implicit visible param).
+annotations.visibility.setAnnotationVisibility(annotationUID);
+
+// changing an annotation visibility to NOT be visible.
+annotations.visibility.setAnnotationVisibility(annotationUID, false);
+
+// get all the hidden annotations uids.
+annotations.visibility.getAnnotationUIDsHidden();
+
+// get the number of hidden annotations.
+annotations.visibility.getAnnotationUIDsHiddenCount();
+
+// show all annotation(hidden)
+annotations.visibility.showAllAnnotations();
+
+// get if an annotation is visible or not.
+// Possible results are: undefined if there is no annotation for given UID, true if visible and false if not.
+annotations.visibility.isAnnotationVisible(annotationUID);
+```
+
+## Read more
+
+:::note TIP
+Read more about the visibility API [here](/api/tools/namespace/annotation#visibility)
+:::
--- a/packages/docs/docs/concepts/cornerstone-tools/annotation/visibility.md
+++ b/packages/docs/docs/concepts/cornerstone-tools/annotation/visibility.md
@@ -21,12 +21,6 @@ annotations.visibility.setAnnotationVisibility(annotationUID);
 // changing an annotation visibility to NOT be visible.
 annotations.visibility.setAnnotationVisibility(annotationUID, false);
 
-// get all the hidden annotations uids.
-annotations.visibility.getAnnotationUIDsHidden();
-
-// get the number of hidden annotations.
-annotations.visibility.getAnnotationUIDsHiddenCount();
-
 // show all annotation(hidden)
 annotations.visibility.showAllAnnotations();
 
--- a/packages/tools/src/stateManagement/annotation/FrameOfReferenceSpecificAnnotationManager.ts
+++ b/packages/tools/src/stateManagement/annotation/FrameOfReferenceSpecificAnnotationManager.ts
@@ -9,6 +9,7 @@ import cloneDeep from 'lodash.clonedeep';
 import { Enums, eventTarget, Types, utilities } from '@cornerstonejs/core';
 
 import { checkAndDefineIsLockedProperty } from './annotationLocking';
+import { checkAndDefineIsVisibleProperty } from './annotationVisibility';
 
 interface FilterInterface {
   FrameOfReferenceUID?: string;
@@ -179,6 +180,7 @@ export default class FrameOfReferenceSpecificAnnotationManager {
 
     toolSpecificAnnotations.push(annotation);
     checkAndDefineIsLockedProperty(annotation);
+    checkAndDefineIsVisibleProperty(annotation);
   };
 
   /**
--- a/packages/tools/src/stateManagement/annotation/annotationState.ts
+++ b/packages/tools/src/stateManagement/annotation/annotationState.ts
@@ -123,7 +123,7 @@ function removeAnnotation(
  * Get the Annotation object by its UID
  * @param annotationUID - The unique identifier of the annotation.
  * @param element - The element that the tool is being used on.
- * @returns A Annotation object.
+ * @returns A Annotation object
  */
 function getAnnotation(
   annotationUID: string,
--- a/packages/tools/src/stateManagement/annotation/annotationVisibility.ts
+++ b/packages/tools/src/stateManagement/annotation/annotationVisibility.ts
@@ -49,15 +49,6 @@ function showAllAnnotations(): void {
 }
 
 /**
- * Returns an array of all the annotation uids that is currently hidden.
- * @returns An array of tool specific annotation UIDs.
- *
- */
-function getAnnotationUIDsHidden(): Array<string> {
-  return Array.from(globalHiddenAnnotationUIDsSet);
-}
-
-/**
  * Given an annotation UID, return true if it is visible, false if hidden and undefined if does not exist.
  * @param annotationUID - The annotation uid to tell if is visible or not.
  * @returns A boolean value or value if does not exist.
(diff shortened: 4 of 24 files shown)


packages/tools/src/utilities/planar/filterAnnotationsForDisplay.ts
```diff
@@ -34,6 +34,10 @@ export default function filterAnnotationsForDisplay(
     // and we don't have a proper way to check distance either since a stack can be
     // composed of multiple unrelated images
     return annotations.filter((annotation) => {
+      if (!annotation.isVisible) {
```
ladeirarodolfo, line 37: @JamesAPetts  is this accurate? as far as planarFreehand is concerned!

packages/tools/src/utilities/planar/filterAnnotationsWithinSlice.ts
```diff
@@ -41,6 +41,9 @@ export default function filterAnnotationsWithinSlice(
     const data = annotation.data;
     const point = data.handles.points[0];
 
+    if (!annotation.isVisible) {
```
ladeirarodolfo, line 44: @JamesAPetts  is this accurate? as far as planarFreehand is concerned!

packages/tools/src/stateManagement/annotation/FrameOfReferenceSpecificAnnotationManager.ts
```diff
@@ -179,6 +180,7 @@ export default class FrameOfReferenceSpecificAnnotationManager {
 
     toolSpecificAnnotations.push(annotation);
     checkAndDefineIsLockedProperty(annotation);
+    checkAndDefineIsVisibleProperty(annotation);
```
ladeirarodolfo, line 183: This (and also locked property line above) causes a code smell. I.e there is an inappropriate intimacy between AnnotationManger and Each Tool, but they should be actually totally decoupled.

My recommendation is to:
base/AnnotationTool be responsible decorate the annotation (on addnew annotation method)

Each Tool to invoke its super method to allow this decoration.

packages/tools/examples/annotationSelectionAndLocking/index.ts
```diff
@@ -147,7 +147,7 @@ async function run() {
       },
     ],
   });
-  toolGroup.setToolPassive(BidirectionalTool.toolName);
+  toolGroup.setToolPassive(RectangleROITool.toolName);
```
sedghi, line 150: Curious of why you changed bidirectional to rectangleROI?

packages/tools/src/stateManagement/annotation/annotationState.ts
```diff
@@ -123,7 +123,7 @@ function removeAnnotation(
  * Get the Annotation object by its UID
  * @param annotationUID - The unique identifier of the annotation.
  * @param element - The element that the tool is being used on.
- * @returns A Annotation object.
+ * @returns A Annotation object or undefined if that does not exist.
```
sedghi, line 126: I think the return type is also incorrect (has happend before this PR)

packages/tools/src/tools/annotation/LengthTool.ts
```diff
@@ -568,6 +570,10 @@ class LengthTool extends AnnotationTool {
 
       let activeHandleCanvasCoords;
 
+      if (!isAnnotationVisible(annotationUID)) {
```
sedghi, line 573: put it after cachedStats

packages/tools/src/types/AnnotationTypes.ts
```diff
@@ -7,6 +7,8 @@ type Annotation = {
   highlighted?: boolean;
   /** If the annotation is locked for manipulation */
   isLocked?: boolean;
+  /** If the annotation is visible for manipulation */
+  isVisible?: boolean;
```
sedghi, line 11: Why this is optional? I thought it is always there and default to true

packages/docs/docs/concepts/cornerstone-tools/annotation/visibility.md
```diff
@@ -0,0 +1,42 @@
...
+import { annotations } from '@cornerstonejs/tools';
+
+// changing an annotation visibility to be visible (implicit visible param).
+annotations.visibility.setAnnotationVisibility(annotationUID);
+
+// changing an annotation visibility to NOT be visible.
+annotations.visibility.setAnnotationVisibility(annotationUID, false);
+
+// get all the hidden annotations uids.
+annotations.visibility.getAnnotationUIDsHidden();
```
sedghi, line 25: hmm, not sure if it is me only, but It feels weird that the API has both `visible` and `hidden` methods. Can't we say `getAnnotationUIDsInvisible`? or `getAnnotationUIDsInvisibleCount`? This might be only me since I'm not well known for naming things :D @JamesAPetts

packages/tools/src/stateManagement/annotation/FrameOfReferenceSpecificAnnotationManager.ts
ladeirarodolfo: ok
https://github.com/cornerstonejs/cornerstone3D-beta/issues/95

packages/tools/examples/annotationSelectionAndLocking/index.ts
ladeirarodolfo: I shouldnt commit that at all.. it was just because of testing purpose

packages/tools/src/stateManagement/annotation/annotationState.ts
ladeirarodolfo: Glad that I forgot that.
I will remove this, but I would like to do an observation that the signature of it is not consistent to
annotationManager.getAnnotation(annotationUID);

packages/tools/src/types/AnnotationTypes.ts
ladeirarodolfo: This is indeed a very trick one. It tells us something about our architecture: 
As far as OOO is concerned, this type is not what it meant to be. 
When I see this type Annotation, I understand this represents an instance of an AnnotationTool. If that is true, each annotation instance has at least to 2 phases/shapes types. I.e at construction time and after adding the handles.
So, with all that said that is why many of flags here are optional, because they are completed (if that happens) at run time. 
In summary, AnnotationType is being used as a representation of instance (at construction time) and considering this flag is added by decoration at run time (not at construction time), I believe would be fine to let it optional. This should be fixed whenever we move the decoration to BaseAnnotationLayer instead of AnnotationManager layer

packages/tools/src/stateManagement/annotation/annotationVisibility.ts
```diff
@@ -73,7 +73,7 @@ function isAnnotationVisible(annotationUID: string): boolean | undefined {
 /**
  * Get the number of hidden annotation objects in the global set of hidden annotation
  * objects uids.
- * @returns {number} The number of hidden annotation objects.
+ * @returns The number of hidden annotation objects.
  *
  */
 function getAnnotationUIDsHiddenCount(): number {
```
sedghi, line 79: Also this doesn't make sense it has UID in the name, can we have `getNumberOfHiddenAnnotations` and any one call .length on it

packages/docs/docs/concepts/cornerstone-tools/annotation/visibility.md
ladeirarodolfo: This is not a requirement on my impl. So I will remove it. 
If there is a need for something like that in future we can add it back.

packages/tools/src/stateManagement/annotation/annotationVisibility.ts
```diff
@@ -0,0 +1,171 @@
...
+import { AnnotationVisibilityChangeEventDetail } from '../../types/EventTypes';
+
+/*
+ * It stores all hidden annotation uids.
+ */
+const globalHiddenAnnotationUIDsSet: Set<string> = new Set();
+
+/*
+ * Interface (Public API)
+ */
```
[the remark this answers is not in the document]
JamesAPetts: Its certainly some pretty strange code on both fronts (not criticising this PR, it was obviously following the same pattern):

`checkAndDefineIsVisibleProperty` defines the getter and setter of the property but it does seem like they aren't used. There doesn't seem to be anywhere we are setting isVisible directly anywhere, but it also looks like if you traverse the `setAnnotationVisibility` tree you don't hit the defined properties. But we _are_ checking the properties of the annotations, so they must be being set somehow. We should clean up this implementation so its less esoteric at some point, it really needn't be this complicated.

packages/tools/src/types/AnnotationTypes.ts
JamesAPetts: >When I see this type Annotation, I understand this represents an instance of an AnnotationTool

This is incorrect. The`Annotation` is _data_ created by an `AnnotationTool`, not an instance of the `AnnotationTool`. `Annotations` are like `toolData` in Cornerstone Legacy.

`isVisible` should be mandatory, and set true by default when the annotation is defined in an `AnnotationTool`'s `addNewAnnotation` method (or as you said before,as a super called to `BaseAnnotationTool`).

There should be no scenario where `isVisible` is undefined, and we should really do the same for `isLocked` too (set it `false` on creation).

cc @sedghi .

packages/tools/src/utilities/planar/filterAnnotationsWithinSlice.ts
JamesAPetts: Same as above.

JamesAPetts (changes requested):
Some minor changes to example required, the rest are discussion points, and don't necessarily need to be dealt with in this PR.

JamesAPetts (approved):
Looks good for now, we can make the proposed improvements to base annotation tool etc in the future, this PR is good to merge.

PR #99, opened 2022-05-12
feat: Add CINE tool via playClip

sedghi:


packages/tools/src/utilities/cine/playClip.ts
```diff
@@ -0,0 +1,246 @@
...
+      'playClip: element must be a stack viewport, volume viewport playClip not yet implemented'
+    );
+  }
+
+  const stackData = {
+    currentImageIdIndex: viewport.getCurrentImageIdIndex(),
+    imageIds: viewport.getImageIds(),
+  };
+
+  const playClipToolData = getToolState(element);
```
JamesAPetts, line 47: Can't you just do `let playClipData = getToolState(element)` and then if its undefined, define it and add it to the state? Then you can avoid having both the `playClipToolData` and `playClipData` variables.

packages/tools/src/utilities/cine/playClip.ts
```diff
@@ -0,0 +1,246 @@
...
+const { triggerEvent } = utilities;
+
+/**
+ * Starts playing a clip or adjusts the frame rate of an already playing clip.  framesPerSecond is
+ * optional and defaults to 30 if not specified.  A negative framesPerSecond will play the clip in reverse.
+ * The element must be a stack of images
+ * @param element - HTML Element
+ * @param framesPerSecond - Number of frames per second
+ */
+function playClip(element: HTMLDivElement, framesPerSecond: number): void {
```
JamesAPetts, line 18: It feels weird to me that `framesPerSecond` is some special option which can be added here, yet all the others need to be added to toolstate first if you want them to be set on the first play.

Can't we add all the options which aren't internal ids etc as optional params here, rather than just the framesPerSecond, i.e.:
```
playClip(element: HTMLDivElement, options: PlayClipOptions) {...}

Type PlayClipOptions =  {
  framesPerSecond?: number;
  frameTimeVector?: number[];
  reverse?: boolean;
  loop:? boolean;
  speed?: number // Also we should rename this, as its actually just for the frame time vectors. Maybe something like `frameTimeVectorSpeedMultiplier`?
```

packages/tools/src/utilities/cine/playClip.ts
```diff
@@ -0,0 +1,246 @@
...
+  };
+
+  const playClipToolData = getToolState(element);
+
+  if (!playClipToolData) {
+    playClipData = {
+      intervalId: undefined,
+      framesPerSecond: 30,
+      lastFrameTimeStamp: undefined,
+      frameRate: 0,
```
JamesAPetts, line 54: `frameRate` doesn't appear to be used? Should we remove it? It seems like a duplicate of `framesPerSecond`?

packages/tools/src/utilities/cine/state.ts
```diff
@@ -0,0 +1,31 @@
...
+  framesPerSecond: number;
+  lastFrameTimeStamp: number | undefined;
+  frameRate: number;
+  frameTimeVector: number[] | undefined;
+  ignoreFrameTimeVector: boolean;
+  usingFrameTimeVector: boolean;
+  speed: number;
+  reverse: boolean;
+  loop: boolean;
+  data?: unknown[];
```
JamesAPetts, line 14: Why do we need this field? It doesn't appear to be used. Unless I'm mistaken.

packages/tools/src/utilities/cine/playClip.ts
```diff
@@ -0,0 +1,246 @@
...
+function _getPlayClipTimeouts(vector: number[], speed: number) {
+  let i;
+  let sample;
+  let delay;
+  let sum = 0;
+  const limit = vector.length;
+  const timeouts = [];
+
+  // Initialize time varying to false
+  // @ts-ignore
```
JamesAPetts, line 190: Can we fix types instead of using `@ts-ignore`?

packages/tools/src/utilities/cine/playClip.ts
```diff
@@ -0,0 +1,246 @@
...
+  // First element of a frame time vector must be discarded
+  for (i = 1; i < limit; i++) {
+    // eslint-disable-next-line no-bitwise
+    delay = (Number(vector[i]) / speed) | 0; // Integral part only
+    timeouts.push(delay);
+    if (i === 1) {
+      // Use first item as a sample for comparison
+      sample = delay;
+    } else if (delay !== sample) {
+      // @ts-ignore
```
JamesAPetts, line 206: Can we fix types instead of using `@ts-ignore`?

packages/tools/src/utilities/cine/playClip.ts
```diff
@@ -0,0 +1,246 @@
...
+      timeouts.isTimeVarying = true;
+    }
+
+    sum += delay;
+  }
+
+  if (timeouts.length > 0) {
+    // @ts-ignore
```
JamesAPetts, line 214: Can we fix types instead of using `@ts-ignore`?

packages/tools/src/utilities/cine/events.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * CINE Tool Events
+ */
+enum Events {
+  CLIP_STOPPED = 'CORNERSTONE_CINE_TOOL_STOPPED',
```
JamesAPetts, line 5: Is a clip started event useful? I can imagine some UI updating based on this? I.e. the user clicks a hotkey to start CINE, and then some UI updates to show its playing?

JamesAPetts (changes requested):
Some minor comments, but one major one:

 At the minimum: we need an example page with a single stack viewport that demonstrates `playClip` and `stopClip`, and has a slider for the framerate from like 1-90 fps.

packages/tools/src/utilities/cine/playClip.ts
```diff
@@ -44,14 +48,13 @@ function playClip(element: HTMLDivElement, framesPerSecond: number): void {
...
+  let playClipData = getToolState(element);
 
-  if (!playClipToolData) {
+  if (!playClipData) {
     playClipData = {
       intervalId: undefined,
       framesPerSecond: 30,
       lastFrameTimeStamp: undefined,
-      frameRate: 0,
       frameTimeVector: undefined,
```
JamesAPetts, line 58: Can you set the other `playClipOptions` if they are defined?

JamesAPetts (approved):
One tiny task then feel free to merge.

PR #100, opened 2022-05-12
Feat/freehand improvements

JamesAPetts:
- Fix a bunch of known issues with the planar freehand tool interaction.
- Add the (slightly hidden) option for cardiac annotations within the planar tool, but add this example to the demo. In the future this should likely be pulled out to a seperate tool for cleanliness.


common/reviews/api/tools.api.md
```diff
@@ -2728,6 +2728,8 @@ interface PlanarFreehandROIAnnotation extends Annotation {
         polyline: Types_2.Point3[];
         label?: string;
         isOpenContour?: boolean;
+        isOpenCardiacAnnotation?: boolean;
+        openCardiacAnnotationVectorToPeak?: Types_2.Point3[];
```
sedghi, line 2732: Are you claiming that the horseshoe shape is only for cardiac and not being used anywhere else? I think it makes sense to say something like more general, maybe is horseShoe shaped? that is aweful too but still not claiming that we know it all and this is being used only in cardiac

packages/tools/src/tools/annotation/planarFreehandROITool/findOpenCardiacAnnotationVectorToPeak.ts
```diff
@@ -0,0 +1,71 @@
+import type { Types } from '@cornerstonejs/core';
+import { PlanarFreehandROIAnnotation } from '../../../types/ToolSpecificAnnotationTypes';
+import { vec2 } from 'gl-matrix';
+
+/**
+ * Finds the length of the longest line from the midpoint of the line
+ * that joins the start and end of the open contour, to the surface of the
+ * open contour.
+ */
+export default function findOpenCardiacAnnotationVectorToPeak(
```
sedghi, line 10: same in the function name here

packages/tools/src/tools/annotation/planarFreehandROITool/renderMethods.ts
```diff
@@ -51,7 +52,25 @@ function renderContour(
   annotation: PlanarFreehandROIAnnotation
 ): void {
   if (annotation.data.isOpenContour) {
-    this.renderOpenContour(enabledElement, svgDrawingHelper, annotation);
+    if (annotation.data.isOpenCardiacAnnotation) {
+      if (!annotation.data.openCardiacAnnotationVectorToPeak) {
```
sedghi, line 56: Not important, but this can be better re-written to be more readable with gaurd clauses

packages/tools/src/utilities/planar/filterAnnotationsForDisplay.ts
```diff
@@ -35,6 +35,13 @@ export default function filterAnnotationsForDisplay(
     // composed of multiple unrelated images
     return annotations.filter((annotation) => {
       const imageId = annotation.metadata.referencedImageId;
+
+      if (imageId === undefined) {
```
sedghi, line 39: Is there a situation that a StackViewport annotation doesn't have imageId? How is it rendering it if it is undefined?

sedghi (commented):
Please see my comments

packages/tools/src/tools/annotation/planarFreehandROITool/renderMethods.ts
JamesAPetts: Simplified a little and added comments.

packages/tools/src/utilities/planar/filterAnnotationsForDisplay.ts
sedghi: oh I see great

sedghi (approved):

PR #101, opened 2022-05-16
Freehand: Prevent very small annotations from being drawn

JamesAPetts:
- This PR prevents very small annotations that may even be too small to render from being created.
- These annotations are removed on mouse up, and no annotation completed event is fired.


sedghi (approved):

sedghi (approved):

JamesAPetts:
Pulled in changes from #94 .

JamesAPetts:
Will merge after checks.

PR #102, opened 2022-05-16
fix: Selection event to have the correct state of annotation, when cha…

ladeirarodolfo:
…nging its visibility

Includes
- Ensure selection evt (triggered by changing visibility state) has the correct visibility value.


sedghi (approved):
LGTM

PR #104, opened 2022-05-18
feat: Add VOLUME_NEW_IMAGE event and Add jumpToSlice and default VOI for volume viewport

sedghi:
**Fix:** 
- Disabling element should not reset camera 
- WindowLevel tool for volume bug in PT scaling 

**Refactor**
- Move various `@cornerstonejs/core` camera-related utility functions out of their incorrect location in the `@cornertsontejs/tools`

**Add**
- Add a new VOLUME_NEW_IMAGE event which publish the `imageIndex` of the image that has rendered in the volume viewport (any camera view)
- Add the capability in `jumpToSlice` to jump to a slice in the `volumeViewport`
- Add default VOI set for the volumes (from metadata or by calculating the min and max of the middle slice)

### DIFF
--- a/packages/core/src/RenderingEngine/helpers/createVolumeActor.ts
+++ b/packages/core/src/RenderingEngine/helpers/createVolumeActor.ts
@@ -60,6 +60,12 @@ async function createVolumeActor(
       }
       triggerVOIModified(element, viewportId, volumeActor);
     });
+  } else {
+    if (callback) {
+      callback({ volumeActor, volumeId });
+    }
+
+    triggerVOIModified(element, viewportId, volumeActor);
   }
 
   return volumeActor;
--- a/packages/core/src/RenderingEngine/Viewport.ts
+++ b/packages/core/src/RenderingEngine/Viewport.ts
@@ -398,7 +398,7 @@ class Viewport implements IViewport {
   /**
    * Add a list of actors (actor entries) to the viewport
    * @param resetCameraPanAndZoom - force reset pan and zoom of the camera,
-   *        default vaule is false.
+   *        default value is false.
    * @param actors - An array of ActorEntry objects.
    */
   public addActors(
--- a/packages/streaming-image-volume-loader/package.json
+++ b/packages/streaming-image-volume-loader/package.json
@@ -32,7 +32,7 @@
   },
   "devDependencies": {
     "@cornerstonejs/calculate-suv": "1.0.2",
-    "@cornerstonejs/core": "^0.8.0",
+    "@cornerstonejs/core": "^0.9.0",
     "cornerstone-wado-image-loader": "^4.1.2"
   },
   "contributors": [
--- a/packages/tools/package.json
+++ b/packages/tools/package.json
@@ -32,7 +32,7 @@
     "gl-matrix": "^3.4.3"
   },
   "devDependencies": {
-    "@cornerstonejs/core": "^0.8.0",
+    "@cornerstonejs/core": "^0.9.0",
     "@cornerstonejs/streaming-image-volume-loader": "^0.2.27",
     "@kitware/vtk.js": "24.0.0"
   },


packages/core/examples/volumeEvents/index.ts
```diff
@@ -72,7 +72,7 @@ function updateLastEvents(number, eventName, detail) {
 
 let eventNumber = 1;
 
-const { IMAGE_RENDERED, CAMERA_MODIFIED, STACK_NEW_IMAGE } = Enums.Events;
```
JamesAPetts, line 75: I see you removed STACK_NEW_IMAGE which makes sense, but could we then sub to VOLUME_NEW_IMAGE now and log that out in the last events?

packages/core/src/RenderingEngine/RenderingEngine.ts
```diff
@@ -221,7 +221,11 @@ class RenderingEngine implements IRenderingEngine {
     }
 
     // 8. Resize the offScreen canvas to accommodate for the new size (after removal)
-    this.resize();
+    // Note: Resize should not reset pan and zoom in disableElement
```
JamesAPetts, line 224: I know this seems obvious to you as to why, but can you be a bit more verbose here? Something like:
"Note: Resize should not reset pan and zoom when disabling an element. This is because we are only resizing the offscreen canvas to deal with the element which was removed, and do not wish to alter the current state of any other currently enabled element"?

Or something :)

packages/core/src/RenderingEngine/VolumeViewport.ts
```diff
@@ -79,6 +83,21 @@ class VolumeViewport extends Viewport implements IVolumeViewport {
     return false;
   }
 
+  private initializeVolumeNewImageEventDispatcher(): void {
+    this.element.addEventListener(
```
JamesAPetts, line 87: You need to also remove this event handler when the element is disabled. Otherwise if you re-enable it as, say, a stack viewport, we will still be reacting to CAMERA_MODIFIED events and this will likely break or just fire erroneous misleading events.

packages/core/src/RenderingEngine/helpers/volumeNewImageEventDispatcher.ts
```diff
@@ -0,0 +1,61 @@
...
+function volumeNewImageEventDispatcher(
+  cameraEvent: EventTypes.CameraModifiedEvent,
+  viewportImageData: IImageData
+): void {
+  const { renderingEngineId, viewportId } = cameraEvent.detail;
+  const renderingEngine = getRenderingEngine(renderingEngineId);
+  const viewport = renderingEngine.getViewport(viewportId);
+
+  if (!(viewport instanceof VolumeViewport)) {
+    return;
```
JamesAPetts, line 34: I would actually throw if this is hit, as this means that something very wrong has occured?

packages/core/src/RenderingEngine/helpers/volumeNewImageEventDispatcher.ts
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  if (!state[viewport.id]) {
+    state[viewport.id] = 0;
```
JamesAPetts, line 38: We probably need to clean up parts of this state when elements with that volume ID are disabled (i.e. listen to element disabled). Otherwise if the same viewportId is re-used you may hit weird edge cases where an event should have fired as the index from the old viewport was the same.

packages/core/src/types/EventTypes.ts
```diff
@@ -171,6 +173,21 @@ type StackNewImageEventDetail = {
...
+  viewportId: string;
+  /** unique id for the renderingEngine */
+  renderingEngineId: string;
+};
+
+/**
+ * VOLUME_NEW_IMAGE Event's data
+ */
+type VolumeNewImageEventDetail = {
+  imageData: vtkImageData;
```
JamesAPetts, line 188: Should we be leaking vtk here? Why is this required on this event?

packages/core/src/utilities/getImageSliceDataForVolumeViewport.ts
```diff
@@ -0,0 +1,50 @@
...
+  const { volumeActor } = actor;
+  const sliceRange = getSliceRange(volumeActor, viewPlaneNormal, focalPoint);
+
+  const { min, max, current } = sliceRange;
+
+  // calculate number of steps from min to max with current normal spacing in direction
+  const numberOfSlices = Math.floor((max - min) / spacingInNormalDirection + 1);
+
+  // calculate the imageIndex based on min, max, current
+  const imageIndex = Math.round(
```
JamesAPetts, line 40: Shouldn't this be floor rather than round? If you are within the spacing between where slice 0 and slice 1 is defined, you are viewing slice 0, right?

JamesAPetts: round also allows you to report a value past the last index.

e.g. slices go from 0-1, 1-2, 2-3 (3 slices, ultra simplified units), if you round a camera viewing depth 2.75, you'll get index 3, when its clearly index 2 being viewed.

JamesAPetts (changes requested):
Mostly looks good. Some changes required mostly about unsubscribing to events, and cleaning up state.

JamesAPetts (changes requested):
One consideration on the element disabled events.

packages/core/src/RenderingEngine/VolumeViewport.ts
```diff
@@ -84,16 +88,34 @@ class VolumeViewport extends Viewport implements IVolumeViewport {
...
-        if (!viewportImageData) {
-          return;
-        }
+      volumeNewImageHandler
+    );
 
-        volumeNewImageEventDispatcher(cameraEvent, viewportImageData);
+    eventTarget.addEventListener(
+      Events.ELEMENT_DISABLED,
+      (evt: EventTypes.ElementDisabledEvent) => {
```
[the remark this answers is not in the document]
sedghi: good catch you are right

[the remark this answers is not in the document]
JamesAPetts: > We also need to remove this element disabled event listener right haha?

This one hasn't been addressed. When this element is disabled, we need to remove THIS ELEMENT_DISABLED handler also, right?

packages/core/src/RenderingEngine/helpers/createVolumeActor.ts
```diff
@@ -43,11 +48,45 @@ async function createVolumeActor(
...
+  viewportId: string,
+  volumeActor: VolumeActor,
+  voiRange: [number, number]
+) => {
+  const newVoiRange = volumeActor
+    .getProperty()
+    .getRGBTransferFunction(0)
+    .getRange();
+
+  if (newVoiRange[0] === voiRange[0] && newVoiRange[1] === voiRange[1]) {
```
JamesAPetts, line 77: I'm confused, don't we just want to fire a voi modified change regardless here? This is just for the initial voi of the viewport, like the first stack new image.

When the volume actor is created, we wanted to emit a voi modified event so that things like overlays can display the initial values. Here if no callback is given, a voi is still set by default, but we aren't emiting a value, so UI etc can't update.

I think we don't need a check here, and we don't need to calculate the old voi. Just after the (optional) callback, find whatever the current voi is, and fire the voi_modified event.

packages/core/src/types/Metadata.ts
```diff
@@ -18,6 +18,8 @@ type Metadata = {
   PixelRepresentation: number;
   /** Image Modality */
   Modality: string;
+  /** SeriesInstanceUID */
+  SeriesInstanceUID: string;
```
JamesAPetts, line 22: The comment is rather superflous, its just the same as the property name haha.

packages/tools/src/enums/JumpPresets.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * Jump Presets - This enum defines the 3 jump states which are available
+ * to be used with the jumpToSlice utility function.
+ */
+enum JumpPresets {
```
JamesAPetts, line 5: Hmm, I was asumming this would be at the app level, this seems like a rather opinionated thing to have at the library level, when camera position/imageIndex covers all basis and is more generic?

JamesAPetts (changes requested):
Thing there is still an event issue, plus some other comments.

packages/core/src/RenderingEngine/VolumeViewport.ts
```diff
@@ -88,40 +88,47 @@ class VolumeViewport extends Viewport implements IVolumeViewport {
...
       if (!viewportImageData) {
         return;
       }
 
       volumeNewImageEventDispatcher(cameraEvent);
-    };
+    }
+
+    function volumeNewImageCleanUp(evt) {
+      eventTarget.removeEventListener(
```
JamesAPetts, line 102: Doesn't this have to come after the if block on line 109-111? Right now _another_ viewport being disabled will cause this handler to be removed from this viewport.

JamesAPetts (changes requested):
I still think there is a dangling event listener issue here, more in line.

packages/core/src/RenderingEngine/helpers/cpuFallback/rendering/generateLut.ts
```diff
@@ -36,9 +36,15 @@ export default function (
 
   const lut = image.cachedLut.lutArray;
 
-  const mlutfn = getModalityLut(image.slope, image.intercept, modalityLUT);
+  let mlutfn = getModalityLut(image.slope, image.intercept, modalityLUT);
   const vlutfn = getVOILUT(windowWidth, windowCenter, voiLUT);
 
+  if (image.isPreScaled) {
+    // if the image is already preScaled, it means the slop, intercept
```
JamesAPetts, line 43: > it means the slop

typo

JamesAPetts: Also shouldn't it be "the slope and the intercept", not "slope, intercept", there are only two items in the list.

packages/core/src/RenderingEngine/helpers/createVolumeActor.ts
```diff
@@ -48,45 +49,40 @@ async function createVolumeActor(
   const volumeActor = vtkVolume.newInstance();
   volumeActor.setMapper(volumeMapper);
 
-  const voiRange = volumeActor
-    .getProperty()
-    .getRGBTransferFunction(0)
-    .getRange()
-    .slice();
+  // If the volume is composed of imageIds, we can apply a default VOI based
```
JamesAPetts, line 52: Might be good to write an example of when there might not be imageIds, as this won't be obvious to new users.

packages/core/src/RenderingEngine/helpers/createVolumeActor.ts
```diff
@@ -48,45 +49,40 @@ async function createVolumeActor(
...
+  // on either the metadata or the min/max of the middle slice.
+  if (imageVolume.imageIds) {
+    await setDefaultVolumeVOI(volumeActor, imageVolume);
```
JamesAPetts, line 55: Hmm, i don't think we should await here, this is blocking the creation of the volumeActor on the frame request to set the VOI.

We should instead pass through and return the volumeActor, and only the callback + triggerVOIModified should await the frame request?

packages/core/src/RenderingEngine/helpers/setDefaultVolumeVOI.ts
```diff
@@ -0,0 +1,234 @@
...
+  const imageIdIndex = Math.floor(imageIds.length / 2);
+  const imageId = imageVolume.imageIds[imageIdIndex];
+  const generalSeriesModule =
+    metaData.get('generalSeriesModule', imageId) || {};
+  const { modality } = generalSeriesModule;
+  const modalityLutModule = metaData.get('modalityLutModule', imageId) || {};
+
+  const numImages = imageIds.length;
+  const bytesPerImage = scalarData.byteLength / numImages;
+  const length = scalarData.length / numImages;
```
JamesAPetts, line 130: Can we call this variable `voxelsPerImage` or something? Think its a bit cleaner than a generic `length`.

packages/core/src/RenderingEngine/helpers/setDefaultVolumeVOI.ts
```diff
@@ -0,0 +1,234 @@
...
+      scalingParametersToUse = {
+        ...scalingParameters,
+        suvbw: suvFactor.suvbw,
+      };
+    }
+  }
+
+  const options = {
+    targetBuffer: {
+      arrayBuffer: scalarData.buffer,
```
JamesAPetts, line 162: In `_getImageScalarDataFromImageVolume` you do a bunch of calculations again, but I think you more or less already have all the variables you need here to get the array view you need right?

packages/core/src/RenderingEngine/helpers/setDefaultVolumeVOI.ts
```diff
@@ -0,0 +1,234 @@
...
+  };
+}
+
+function _getImageScalarDataFromImageVolume(imageVolume, imageIdIndex) {
+  const { scalarData, imageIds, dimensions } = imageVolume;
+  const numImages = imageIds.length;
+  const bytesPerImage = scalarData.byteLength / numImages;
+
+  const numVoxels = dimensions[0] * dimensions[1] * dimensions[2];
+  const numComponents = scalarData.length / numVoxels;
```
JamesAPetts, line 209: We aren't really dealing with the number of components for anything are we (all the logic below probably only works for 1)? I wonder how we even calculate the window level for RGB in this case?

JamesAPetts (changes requested):
Mostly there, comments in line.

PR #109, opened 2022-05-20
Clipping planes

Punzo:
Change how the clipping in the rendering is done. 

We now use clipping planes which are associated for each actor.

### DIFF
--- a/packages/core/src/RenderingEngine/Viewport.ts
+++ b/packages/core/src/RenderingEngine/Viewport.ts
@@ -1,6 +1,7 @@
 import type { vtkCamera } from '@kitware/vtk.js/Rendering/Core/Camera';
 import vtkMatrixBuilder from '@kitware/vtk.js/Common/Core/MatrixBuilder';
 import vtkMath from '@kitware/vtk.js/Common/Core/Math';
+import vtkPlane from '@kitware/vtk.js/Common/DataModel/Plane';
 
 import { vec3, mat4 } from 'gl-matrix';
 import _cloneDeep from 'lodash.clonedeep';
@@ -501,7 +502,7 @@ class Viewport implements IViewport {
     const previousCamera = _cloneDeep(this.getCamera());
 
     const bounds = renderer.computeVisiblePropBounds();
-    const focalPoint = [0, 0, 0];
+    const focalPoint = <Point3>[0, 0, 0];
     const imageData = this.getDefaultImageData();
 
     // Todo: remove this, this is just for tests passing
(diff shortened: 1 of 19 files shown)


packages/core/src/RenderingEngine/vtkClasses/vtkStreamingOpenGLVolumeMapper.js
```diff
@@ -246,24 +246,12 @@ function vtkStreamingOpenGLVolumeMapper(publicAPI, model) {
...
-      crange[1] = cameraMidpoint + slabThickness;
-      cam.setSlabThicknessActive(true);
-    } else {
-      cam.setSlabThicknessActive(false);
-    }
+    // NOTE: the actual slab thickness clipping is done with clipping planes,
+    // but here we still need to have these values here, otherwise
+    // the rendering will be clipped before the clipping planes.
+    // The clipping range is set in the Viewports (in resetCamera) and
+    // have values: [0.01, model.distance * 2]
```
sedghi, line 254: I think this comment is invalid, "here we still need to have these values", which values?

packages/core/src/RenderingEngine/Viewport.ts
```diff
@@ -801,10 +763,21 @@ class Viewport implements IViewport {
...
 
+    this.triggerCameraModifiedEvent(previousCamera, updatedCamera);
+  }
+
+  /**
+   * Trigger camera modified event
+   * @param cameraInterface - ICamera
+   * @param cameraInterface - ICamera
+   */
+  public triggerCameraModifiedEvent(
```
sedghi, line 777: The name of this function implies that it always triggers, which is not the case in its logic

packages/core/src/RenderingEngine/VolumeViewport.ts
```diff
@@ -274,31 +261,82 @@ class VolumeViewport extends Viewport implements IVolumeViewport {
    * Reset the camera for the volume viewport
    */
   public resetCamera(resetPan = true, resetZoom = true): boolean {
-    return super.resetCamera(resetPan, resetZoom);
+    const success = super.resetCamera(resetPan, resetZoom);
```
sedghi, line 264: super.resetCamera() always return `true` why we are calling it success, is there any scenario that is doesn't succeed?

packages/core/src/RenderingEngine/VolumeViewport.ts
```diff
@@ -274,31 +261,82 @@ class VolumeViewport extends Viewport implements IVolumeViewport {
...
+      clipPlane1.setOrigin(newOrigin1);
+
+      clipPlane2.setNormal(vtkMath.multiplyScalar(viewPlaneNormal, -1));
+      const newOrigin2 = <Point3>[0, 0, 0];
+      vtkMath.add(focalPoint, scaledDistance, newOrigin2);
+      clipPlane2.setOrigin(newOrigin2);
+
+      const mapper = actor.volumeActor.getMapper();
+      mapper.addClippingPlane(clipPlane1);
+      mapper.addClippingPlane(clipPlane2);
```
sedghi, line 300: Can we factor out this logic to the viewport since it is reused in two places?

PR #110, opened 2022-05-20
feat: Add Clipping planes for rendering

Punzo:
Applied review from https://github.com/cornerstonejs/cornerstone3D-beta/pull/109 and rebased


sedghi:
I tried the examples and here are my comments 

1. The colormap for fusion doesn't loop the same as the one we have currently

new
image

old
image

You can try a simpler fusion case here https://www.cornerstonejs.org/live-examples/multivolumeapi and locally `cd packages/core` and then `yarn run example multiVolumeAPI`

2. Slab Thickness change difference

new
image

old
image

3. Stack Viewport World transformation seems to be wrong by 0.05 (the third element in LPS world), you can run `stackcanvastoworld` example


3d slicer
image

old
image

new
image


4. Volume viewport World transformation seems to be wrong by 0.1 (third element in LPS)

3D Slicer 

image

Old
image


New
image

sedghi (changes requested):
Saw my comment above, I had some minor comments in the previous opened PR too

packages/core/examples/multiVolumeCanvasToWorld/index.ts
```diff
@@ -105,7 +105,7 @@ async function run() {
     type: ViewportType.ORTHOGRAPHIC,
     element,
     defaultOptions: {
-      orientation: ORIENTATION.SAGITTAL,
+      orientation: ORIENTATION.AXIAL,
```
sedghi, line 108: This is intentional to show to be SAGITTAL as it shows it is a volume

packages/core/examples/multiVolumeAPI/index.ts
```diff
@@ -116,11 +116,13 @@ addButtonToToolbar({
     } else {
       // Add the PET volume to the viewport. It is in the same DICOM Frame Of Reference/worldspace
       // If it was in a different frame of reference, you would need to register it first.
+      const slabThickness = 0.1;
```
sedghi, line 119: Is this required? if always required can we bake it in the addVolumes? e.g., if not defined use 0.1? Also 1-2 lines of explanation for this 0.1 is good

packages/core/examples/multiVolumeCanvasToWorld/index.ts
Punzo: nice catch, I changed it for testing and forgot to put it back

packages/core/examples/multiVolumeAPI/index.ts
Punzo: no it is not required. I will remove it. If it will be undefined, when setting the clipping planes in resetCamera it will be used the default value (0.05)

packages/core/src/RenderingEngine/VolumeViewport.ts
```diff
@@ -164,12 +163,15 @@ class VolumeViewport extends Viewport implements IVolumeViewport {
...
       const uid = actorUID || volumeId;
-      volumeActors.push({ uid, volumeActor });
+      volumeActors.push({ uid, volumeActor, slabThickness });
     }
 
     this.addActors(volumeActors);
 
     if (immediate) {
+      // set the clipping planes
+      this.resetCamera(false, false);
```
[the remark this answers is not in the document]
Punzo: you are right, I moved the the reset of the camera into the addActors method. It makes more sense: i.e. everytime there is a new actor, it will add the clipping planes to the actors and it is independent from the request of rendering

packages/tools/src/tools/CrosshairsTool.ts
```diff
@@ -151,6 +158,9 @@ export default class CrosshairsTool extends AnnotationTool {
     this._getReferenceLineSlabThicknessControlsOn =
       toolProps.configuration?.getReferenceLineSlabThicknessControlsOn ||
       defaultReferenceLineSlabThicknessControlsOn;
+
+    this._applySlabThicknessToOnlyFirstActor =
```
JamesAPetts, line 162: Is there any case where we want like 2/3 actors mipped? I'm asking because then we should make the config item an array, and use that as a filter for what to MIP.

The tool configuration is public API, so i'd rather it not change again next time we realised we need this?

JamesAPetts (approved):
LGTM, some minor comments.

packages/core/src/RenderingEngine/VolumeViewport.ts
```diff
@@ -321,7 +321,25 @@ class VolumeViewport extends Viewport implements IVolumeViewport {
   }
 
   /**
-   * Gets the largest slab thickness from all actors.
+   * Sets the slab thickness to a specific actor in the viewport.
+   *
+   * @param actorUID - The unique ID of the actor.
+   * @param slabThickness - The slab thickness to set.
+   */
+  public setSlabThicknessForActor(
```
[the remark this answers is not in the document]
Punzo: yes, you are right. I have started looking into non volume actor and yes I would apply the full camera clipping range (adn yes just use "normal" clipping planes)

packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -1595,7 +1596,8 @@ class StackViewport extends Viewport implements IStackViewport {
 
       // This is necessary to initialize the clipping range and it is not related
       // to our custom slabThickness.
-      // activeCamera.setThicknessFromFocalPoint(0.1)
+      //activeCamera.setThicknessFromFocalPoint(0.1);
```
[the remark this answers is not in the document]
Punzo: yes it is not needed, I will remove the line

packages/core/src/types/IViewport.ts
```diff
@@ -47,6 +47,10 @@ interface IViewport {
   getActors(): Array<ActorEntry>;
   /** returns specific actor by its uid */
   getActor(actorUID: string): ActorEntry;
+  /** returns specific actor uid by array index */
+  getActorUIDbyIndex(index: number): string;
+  /** returns specific actor by array index */
+  getActorbyIndex(index: number): ActorEntry;
```
swederik, line 53: Fix the camelCase to getActorByIndex

Punzo: done https://github.com/cornerstonejs/cornerstone3D-beta/pull/110/commits/051517ca9dc6b1b312e0aba6a466a801a9e3d1d5

PR #113, opened 2022-05-26
fix: remove the need for slabThickness in volumeAPI for tools

sedghi:
- Removes the need for slabThicknessEnabled 
- Simplifies the API for setSlabThickness
- Fix bugs for volumeNewImageEventDispatcher
- Fix a bug for Synchronizer that triggered source viewport event when it didn't belong to the synchronizer anymore
- Fix RectangleROIStartEndThreshold tool various bugs with the new API
- Fix arrow tool to accept external api for change/set text


Punzo:
Nice fix. One comment you set the infrastructure in crosshair to set the slabthickness by volumeIDs, but then iin the petCt example you don't use it. You should exclude in the thirs row of viewport the pet volume.

i.e. now the example renders like:

image
 
but it should be  (i.e., no slab thickness for the pet):

image

packages/core/src/RenderingEngine/Viewport.ts
```diff
@@ -798,17 +801,16 @@ class Viewport implements IViewport {
...
+    this.TriggerCameraModifiedEventIfNecessary(previousCamera, updatedCamera);
   }
 
   /**
    * Trigger camera modified event
    * @param cameraInterface - ICamera
    * @param cameraInterface - ICamera
    */
-  public checkAndTriggerCameraModifiedEvent(
+  public TriggerCameraModifiedEventIfNecessary(
```
JamesAPetts, line 813: Should be camelCase rather than PascalCase.

packages/core/src/RenderingEngine/Viewport.ts
```diff
@@ -830,16 +832,14 @@ class Viewport implements IViewport {
    * Updates the actors clipping planes orientation from the camera properties
    * @param updatedCamera - ICamera
    */
-  public updateActorsClippingPlanesOnCameraModified(
-    updatedCamera: ICamera
-  ): void {
+  public updateActorsClippingPlanes(updatedCamera: ICamera): void {
```
JamesAPetts, line 835: - Any reason for this to be public? It seems like a private function.
- The method name is very weird to me due to the possessive plural in the middle. Maybe `updateClippingPlanesForActors`?

packages/core/src/RenderingEngine/VolumeViewport.ts
```diff
@@ -89,6 +89,9 @@ class VolumeViewport extends Viewport implements IVolumeViewport {
   }
 
   private initializeVolumeNewImageEventDispatcher(): void {
+    const volumeNewImageHandlerBound = volumeNewImageHandler.bind(this);
```
JamesAPetts, line 92: If you just replace all the following methods to arrow functions, you won't have to use any of these bindings and can just use the function definitions directly:

- `initializeVolumeNewImageEventDispatcher`
- `volumeNewImageHandler`
- `volumeNewImageCleanUp`

packages/core/src/RenderingEngine/VolumeViewport.ts
```diff
@@ -383,52 +379,31 @@ class VolumeViewport extends Viewport implements IVolumeViewport {
...
 
   /**
-   * Sets the slab thickness to all the volume actors in the viewport.
+   * It sets the slabThickness of the actors of the viewport. If actorUIDs are
+   * provided, it will only set the slabThickness of the actors with the given
+   * uid, but if actorUIDs are not provided, it will set the slabThickness of
+   * all the actors in the viewport.
    *
-   * @param slabThickness - The slab thickness to set.
+   * @param slabThickness - number -The slab thickness to set.
```
JamesAPetts, line 387: We don't need `number` here right, A) the docs should use the types defined in the funciton definition when parsing typescript, B) its not formatted in jsdoc style anyway.

packages/core/src/RenderingEngine/VolumeViewport.ts
```diff
@@ -383,52 +379,31 @@ class VolumeViewport extends Viewport implements IVolumeViewport {
...
+   * @param actorUIDs - An array of actor UIDs. If this is not
```
JamesAPetts, line 388: Hmmm, since this is actually a filter, shall we name it as such, at first glance its a bit weird that if you specify it only does it that one, otherwise it does it to _all_. In most other cases throughout the project, if you omit a uid, it applies it to the first item found.

Maybe `filterActorUIDs` or something?

JamesAPetts: Propogate this comment to all other API of this format here.

packages/tools/src/tools/CrosshairsTool.ts
```diff
@@ -76,7 +76,7 @@ interface CrosshairsAnnotation extends Annotation {
 }
 
 function defaultReferenceLineColor() {
-  return 'rgb(200, 200, 200)';
+  return 'rgb(0, 200, 0)';
```
JamesAPetts, line 79: Do we really need this funciton at all? It's weird that it defines its own color seperate from everything else. Shouldn't the default just be one of the colors from the default tool style? I.e. `colorHighlighted`? Then we could just remove this function entirely and nuke this isolated state.

packages/tools/src/tools/CrosshairsTool.ts
```diff
@@ -88,7 +88,7 @@ function defaultReferenceLineDraggableRotatable() {
 }
 
 function defaultReferenceLineSlabThicknessControlsOn() {
-  return false;
+  return true;
```
JamesAPetts, line 91: What is the motivation for changing the default?

packages/tools/src/tools/CrosshairsTool.ts
```diff
@@ -137,6 +140,9 @@ export default class CrosshairsTool extends AnnotationTool {
           enabled: false,
           panSize: 10,
         },
+        // actorUIDs for slabThickness application, if not defined, the slab thickness
+        // will be applied to all actors of the viewport
+        actorUIDsForSlabThickness: [],
```
JamesAPetts, line 145: I guess this is also a filter right?

Something like: `filterActorUIDsToSetSlabThickness`... or something.

packages/tools/src/tools/segmentation/RectangleROIStartEndThresholdTool.ts
```diff
@@ -198,7 +200,7 @@ export default class RectangleROIStartEndThresholdTool extends RectangleROITool
     return annotation;
   };
 
-  // Todo: make it work for other acquisition planes
+  // Todo: make it work for other than acquisition planes
```
JamesAPetts, line 203: `// Todo: make it work for planes other than acquisition planes`

JamesAPetts (changes requested):
Minor comments, I think the all-by-default or just use these isn't very clear in the API, and minor changes could fix that.

Otherwise I agree with Davide, the petCt example needs to be updated to reflect these changes.

packages/tools/src/tools/CrosshairsTool.ts
sedghi: That is more work to do I will do it in another PR

packages/tools/examples/crossHairs/index.ts
```diff
@@ -136,7 +135,50 @@ function getReferenceLineSlabThicknessControlsOn(viewportId) {
...
+addDropdownToToolbar({
+  options: {
+    values: [
+      'Maximum Intensity Projection',
+      'Minimum Intensity Projection',
+      'Average Intensity Projection',
+    ],
+    defaultValue: 'Maximum Intensity Projection',
+  },
+  onSelectedValueChange: (selectedValue) => {
```
JamesAPetts, line 153: Sent you a video @sedghi . Because the blend mode only changes when you manipulate the lines that corresponds to a viewport, you can end up with a weird combination of different blendmodes. I think you just need to set all actors to that blendmode and re-render on your swap button.

JamesAPetts (approved):
Minor comment on demo, otherwise LGTM.

PR #114, opened 2022-05-26
chore: Increase compatibility with Windows development environment

bajiu:
Modifying the webpack configuration file helps to debug all examples on windows


swederik:
Thanks!

PR #115, opened 2022-05-27
Add the option to always render handles on open contours

JamesAPetts:
Add optional configuration for the planar freehand roi tool:

- allowsRenderOpenContourHandles: boolean (default false)
    - When true, always render end points when you have an open contour, rather than just rendering a line.
- alwaysRenderOpenContourHandlesRadius: number (default 2)
    - When `allowsRenderOpenContourHandles` is true, use this radius to draw the endpoints when not hovering.

### DIFF
--- a/packages/tools/src/tools/annotation/PlanarFreehandROITool.ts
+++ b/packages/tools/src/tools/annotation/PlanarFreehandROITool.ts
@@ -136,12 +136,13 @@ class PlanarFreehandROITool extends AnnotationTool {
       configuration: {
         shadow: true,
         preventHandleOutsideImage: false,
-        // When true, always render end points when you have an open contour, rather
-        // than just rendering a line.
-        allowsRenderOpenContourHandles: true,
-        // When `allowsRenderOpenContourHandles` is true, use this radius to draw
-        // the endpoints when not hovering.
-        alwaysRenderOpenContourHandlesRadius: 2,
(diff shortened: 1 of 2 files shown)


packages/tools/src/tools/annotation/PlanarFreehandROITool.ts
```diff
@@ -136,6 +136,12 @@ class PlanarFreehandROITool extends AnnotationTool {
       configuration: {
         shadow: true,
         preventHandleOutsideImage: false,
+        // When true, always render end points when you have an open contour, rather
+        // than just rendering a line.
+        allowsRenderOpenContourHandles: true,
+        // When `allowsRenderOpenContourHandles` is true, use this radius to draw
+        // the endpoints when not hovering.
+        alwaysRenderOpenContourHandlesRadius: 2,
```
sedghi, line 144: If these are working together, shouldn't they become an object 

```
renderOpenContourHandles = {
  enable: true,
  radius: 2 
}
```

sedghi (commented):
See my comment

sedghi (approved):

packages/tools/src/tools/annotation/PlanarFreehandROITool.ts
JamesAPetts: done :)

PR #123, opened 2022-06-07
feat: remove unnecessary event firing for annotations

sedghi:
Fix:
- ANNOTATION_RENDER events were firing even if no annotation were rendered 
- Camera API to be consistent between CPU and GPU rendering routes

Feat:
- Add new `suppressEvents` to StackViewport `setProperties`

### DIFF
--- a/packages/core/src/RenderingEngine/StackViewport.ts
+++ b/packages/core/src/RenderingEngine/StackViewport.ts
@@ -62,6 +62,7 @@ import {
 } from '../types/EventTypes';
 import getScalingParameters from '../utilities/getScalingParameters';
 import cache from '../cache';
+import correctShift from './helpers/cpuFallback/rendering/correctShift';
 
 const EPSILON = 1; // Slice Thickness
 
@@ -686,6 +687,8 @@ class StackViewport extends Viewport implements IStackViewport {
         viewPlaneNormal[2],
       ],
       viewUp: [viewUp[0], viewUp[1], viewUp[2]],
+      flipHorizontal: this.flipHorizontal,
+      flipVertical: this.flipVertical,
     };
   }
 
@@ -726,8 +729,13 @@ class StackViewport extends Viewport implements IStackViewport {
         vec2.fromValues(prevFocalPointPixel[0], prevFocalPointPixel[1])
       );
 
-      viewport.translation.x -= deltaPixel[0];
-      viewport.translation.y -= deltaPixel[1];
+      const shift = correctShift(
+        { x: deltaPixel[0], y: deltaPixel[1] },
+        viewport
+      );
+
+      viewport.translation.x -= shift.x;
+      viewport.translation.y -= shift.y;
     }
 
     if (parallelScale) {
@@ -749,7 +757,7 @@ class StackViewport extends Viewport implements IStackViewport {
       viewport.parallelScale = (clientHeight * rowPixelSpacing * 0.5) / scale;
     }
 
-    if (flipHorizontal || flipVertical) {
+    if (flipHorizontal !== undefined || flipVertical !== undefined) {
       this.setFlipCPU({ flipHorizontal, flipVertical });
     }
 
--- a/packages/core/src/RenderingEngine/StackViewport.ts
+++ b/packages/core/src/RenderingEngine/StackViewport.ts
@@ -637,7 +637,7 @@ class StackViewport extends Viewport implements IStackViewport {
   }
 
   private getCameraCPU(): Partial<ICamera> {
-    const { metadata, viewport, image } = this._cpuFallbackEnabledElement;
+    const { metadata, viewport } = this._cpuFallbackEnabledElement;
     const { direction } = metadata;
 
     // focalPoint and position of CPU camera is just a placeholder since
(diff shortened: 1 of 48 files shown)


packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -524,16 +524,19 @@ class StackViewport extends Viewport implements IStackViewport {
...
-    interpolationType,
-    rotation,
-  }: StackViewportProperties = {}): void {
+  public setProperties(
+    {
+      voiRange,
+      invert,
+      interpolationType,
+      rotation,
+      suppressEvents,
```
swederik, line 533: I think suppressEvents should be another argument, not part of the StackViewportProperties object. This makes it look like a property, which it is not.

JamesAPetts: Agreed, can keep the API softfork-ey by putting it as the second argument and defaulting it to false.

packages/tools/src/tools/CrosshairsTool.ts
```diff
@@ -1288,17 +1289,18 @@ export default class CrosshairsTool extends AnnotationTool {
...
     // Save new handles points in annotation
     data.handles.rotationPoints = newRtpoints;
     data.handles.slabThicknessPoints = newStpoints;
 
-    // render a circle to pin point the viewport color
-    // TODO: This should not be part of the tool, and definitely not part of the renderAnnotation loop
-
     if (!this.configuration.viewportIndicators) {
-      return;
+      return renderStatus;
```
JamesAPetts, line 1299: Wouldn't it be cleaner to just stick the reference color rendering inside:

`if (this.configuration.viewportIndicators) { ... }`

And then just have one `return renderStatus;` line at the end of the method?

packages/tools/src/tools/annotation/AngleTool.ts
```diff
@@ -523,15 +523,17 @@ class AngleTool extends AnnotationTool {
...
-  ): void => {
+  ): boolean => {
+    let renderStatus = false;
+
     const { viewport } = enabledElement;
     const { element } = viewport;
 
     let annotations = getAnnotations(element, this.getToolName());
 
     // Todo: We don't need this anymore, filtering happens in triggerAnnotationRender
```
JamesAPetts, line 534: Shall we just nuke this everywhere then?

JamesAPetts (approved):
Minor comments to change but mostly LGTM.

packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -656,11 +659,27 @@ class StackViewport extends Viewport implements IStackViewport {
       ) as Float32Array;
     }
 
+    const canvasCenter: Point2 = [
+      this.element.clientWidth / 2,
+      this.element.clientHeight / 2,
+    ];
+
+    // Focal point is the center of the canvas in world coordinate by design
```
JamesAPetts, line 667: Nitpicky I would change "by design" to "by construction"?

packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -656,11 +659,27 @@ class StackViewport extends Viewport implements IStackViewport {
...
+    const canvasCenterWorld = this.canvasToWorld(canvasCenter);
+
+    // parallel scale is half of the viewport height in the world units (mm)
+
+    const topLeftCanvas = this.canvasToWorld([0, 0]);
```
JamesAPetts, line 672: This is topLeftWorld, no? same for bottomLeft?

packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -671,57 +690,82 @@ class StackViewport extends Viewport implements IStackViewport {
...
-      // Note: as parallel scale is defined differently to the GPU version,
-      // We instead need to find the difference and move the camera in
-      // the other direction in this adapter.
+    if (parallelScale) {
+      // We need to convert he parallelScale which has a physical meaning to
+      // camera scale factor (since CPU works with scale). Since parallelScale represents
+      // half of the height of the viewport in the world unit (mm), we can use that
+      // to compute the scale factor which is the ratio of the viewport height in pixels
+      // to the current rendered image height.
+      const { rowPixelSpacing } = image;
```
JamesAPetts, line 739: Ooooh, what if the viewport is rotated? I don't think it correct.

packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -671,57 +690,82 @@ class StackViewport extends Viewport implements IStackViewport {
...
+      const scale = (clientHeight * rowPixelSpacing * 0.5) / parallelScale;
 
-      const diff = previousCamera.parallelScale - parallelScale;
+      viewport.scale = scale;
+      viewport.parallelScale = parallelScale;
+    }
 
-      viewport.scale += diff; // parallelScale; //viewport.scale < 0.1 ? 0.1 : viewport.scale;
+    if (scale) {
+      const { rowPixelSpacing } = image;
```
JamesAPetts, line 747: And here.

packages/tools/src/drawingSvg/drawCircle.ts
```diff
@@ -1,12 +1,13 @@
...
+import { SVGDrawingHelper } from '../types';
 
 import _getHash from './_getHash';
 
 import _setAttributesIfNecessary from './_setAttributesIfNecessary';
 import _setNewAttributesIfValid from './_setNewAttributesIfValid';
 
 function drawCircle(
-  svgDrawingHelper: any,
+  svgDrawingHelper: SVGDrawingHelper,
```
JamesAPetts, line 10: I thought we settled on `svgDraingHelpers`?

JamesAPetts (changes requested):
Looks like the camera stuff is almost there, but looks like it will break on rotation.

packages/core/src/RenderingEngine/StackViewport.ts
sedghi: it seems to work fine

JamesAPetts (approved):
LGTM

PR #127, opened 2022-06-14
fix: large image rendering, missing metadata for StackViewport, high DPI devices

sedghi:
Fix: 
- Switching `StackViewport` to use `imageMapper` to enable rendering of large images
- `StackViewport` rendering of images when metadata is missing

Still has a bug for `canvasToWorld` after the switch to imageMapper


packages/core/src/types/IImageVolume.ts
```diff
@@ -45,6 +45,8 @@ interface IImageVolume {
   imageIds?: Array<string>;
   /** volume referencedVolumeId (if it is derived from another volume) */
   referencedVolumeId?: string; // if volume is derived from another volume
+  /** whether the metadata for the pixel spacing is not undefined  */
+  hasPixelSpacing?: boolean;
```
JamesAPetts, line 49: How can a volume even be built if it doesn't have pixelspacing?

JamesAPetts (approved):
LGTM one minor question.

packages/core/src/RenderingEngine/RenderingEngine.ts
```diff
@@ -767,6 +773,7 @@ class RenderingEngine implements IRenderingEngine {
     const { clientWidth, clientHeight } = canvas;
 
     // Set the canvas to be same resolution as the client.
+    // TODO: Not necessary - breaks devicePixelRatio... why are we doing this?
```
JamesAPetts, line 776: Why do we have a TODO breaks, and then left as is?

packages/core/src/RenderingEngine/RenderingEngine.ts
```diff
@@ -832,11 +839,19 @@ class RenderingEngine implements IRenderingEngine {
     // Deal with vtkjs driven viewports
     if (viewportInputEntries.length) {
       // 1. Getting all the canvases from viewports calculation of the new offScreen size
-
       const vtkDrivenCanvases = viewportInputEntries.map((vp) =>
         getOrCreateCanvas(vp.element)
       );
 
+      // Ensure the canvas size includes any scaling due to device pixel ratio
+      vtkDrivenCanvases.forEach((canvas) => {
```
JamesAPetts, line 847: So it looks like you are doing the scaling here instead, can we nuke the scaling in my other comment above if its not needed?

packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -1785,6 +1797,23 @@ class StackViewport extends Viewport implements IStackViewport {
     this.initialVOIRange = voiRange;
     this.setProperties({ voiRange });
 
+    // At the moment it appears that vtkImageSlice actors do not automatically
+    // have an RGB Transfer Function created, so we need to create one
+    const cfun = vtkColorTransferFunction.newInstance();
+    let lower = 0;
+    let upper = 1024;
```
JamesAPetts, line 1804: Can we put a comment where this magic number comes from, even though the answer is "this is the default vtk.js would have made"?

JamesAPetts (commented):
Minor comments

JamesAPetts (approved):
LGTM

PR #130, opened 2022-06-15
Fix/freehand general fixes

ladeirarodolfo:
It includes:
- Prevent JS exception if when editing tool and drawing along the line( past the proximity for grabbing), but there are not the minimum of prevCanvasPoints.
- Ensure all paths to complete contour check if annotation should be completed or just be halted (logic used already exists, effort here is just to encapsulate in a method and also calls for all cases we want to complete the contour)

Those changes ensure cs3d freehand tool is more robust and consistent. I.e it will handle possible js exception and also it will not fire completion contour if that contour will not be displayed at all.

### DIFF
--- a/packages/tools/src/tools/annotation/planarFreehandROITool/editLoopCommon.ts
+++ b/packages/tools/src/tools/annotation/planarFreehandROITool/editLoopCommon.ts
@@ -40,84 +40,83 @@ function checkForFirstCrossing(
 
     // On the first crossing, remove the first lines prior to the crossing
     this.removePointsUpUntilFirstCrossing(isClosedContour);
-  } else if (
-    editCanvasPoints.length >
-    this.configuration.checkCanvasEditFallbackProximity
-  ) {
-    // At this point, likely we are drawing along the line, we are past the proximity for grabbing.
-    // Search for nearest line segment to the start of the edit.
-    // Set the crossing index to the lower index of the segment.
-
-    const firstEditCanvasPoint = editCanvasPoints[0];
-
-    const distanceIndexPairs = [];
-
     // prevent continue if there are not the minimum of points for this op.
-    if (prevCanvasPoints.length < 2) {
-      return;
-    }
+  } else if (prevCanvasPoints.length >= 2) {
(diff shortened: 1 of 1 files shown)


packages/tools/src/tools/annotation/planarFreehandROITool/drawLoop.ts
```diff
@@ -388,6 +373,56 @@ function cancelDrawing(element: HTMLElement) {
...
+     */
+    3
+  );
+  return canvasPoints.length < minPoints;
+}
+
+/**
+ * Check and halt a drawing for a given event. It returns true in case drawing is halted, otherswise false.
+ */
+function haltDrawing(element: HTMLDivElement, canvasPoints: any): boolean {
```
JamesAPetts, line 401: Shouldn't _this_ method be called `shouldHaltDrawing`. Its a conditional, but it reads like a command.

packages/tools/src/tools/annotation/planarFreehandROITool/editLoopCommon.ts
```diff
@@ -52,6 +52,11 @@ function checkForFirstCrossing(
 
     const distanceIndexPairs = [];
 
+    // prevent continue if there are not the minimum of points for this op.
+    if (prevCanvasPoints.length < 2) {
```
JamesAPetts, line 56: Good catch, but can you move this up to the top level conditional, aroun lines 43-46? I.e. we shouldn't even enter this block if this is true.

JamesAPetts (approved):
LGTM, address minor comments and its good to merge.

packages/tools/src/tools/annotation/planarFreehandROITool/drawLoop.ts
ladeirarodolfo: I dont think so, because its indeed a command it will halt if possible. The result of this command is used to cut operations, such as complete contour.
In other hand the name 'shouldHaltDrawing'( which already exists) should only check (which is not the case of haltDrawing).

PR #142, opened 2022-07-01
feat added bspline interpolation and interpolation config for planarf…

ladeirarodolfo:
…reehand

### Includes
- Interpolation flow using d3 bspline
- PlanarFreehand configuration to use interpolation
- PlanarFreehand impl to be interpolated after drawing done

### DIFF
--- a/packages/docs/docs/concepts/cornerstone-tools/tools.md
+++ b/packages/docs/docs/concepts/cornerstone-tools/tools.md
@@ -161,17 +161,17 @@ interactions.
 
 ## Tool interpolation
 
-Some of tools can be configured to use interpolation process. PlanarFreehandROITool is an example for that.
+Contour tools like the PlanarFreehandROITool can be configured can be configured to use interpolation process.
 The interpolation process uses b-spline algorithm and consider 3 configurations properties:
 
 - interpolation.enabled: to tell whether it should be interpolated or not (false by default)
-- interpolation.knotSampleSize: the distance between each knot on interpolation process (for editing it is considered the property editKnotSampleSize) ( default value 10)
-- interpolation.editKnotSampleSize: same as knotSampleSize but applicable only when editing the tool (default value 20)
+- interpolation.minKnotDistance: the minimum distance between each knot on interpolation process (for editing it is considered the property editMinKnotDistance) ( default value 20)
+- interpolation.editMinKnotDistance: same as minKnotDistance but applicable only when editing the tool (default value 10)
 
 So, with that said the interpolation might occur when:
 
-- drawing is done (i.e mouse is released) and interpolation.enabled is true. Interpolation algorithm uses knotSampleSize
-- edit drawing is done (i.e mouse is released) and interpolation.enabled. Interpolation algorithm uses editKnotSampleSize and its only applied to changed segment
+- drawing is done (i.e mouse is released) and interpolation.enabled is true. Interpolation algorithm uses minKnotDistance
+- edit drawing is done (i.e mouse is released) and interpolation.enabled. Interpolation algorithm uses editMinKnotDistance and its only applied to changed segment
 
 Interpolation does not occur when:
 
@@ -179,7 +179,7 @@ Interpolation does not occur when:
 - drawing still happening (editing or not)
 
 The result of interpolation will be a smoother set of segments.
-Changing tool configuration (see below) you can fine-tune the interpolation process and per experience, a less aggressive interpolation works better for long segments to be interpolated (usually first drawing) whereas a more aggressive interpolation works better for short segments (usually editing).
+Changing tool configuration (see below) you can fine-tune the interpolation process and per experience, a less aggressive interpolation (small values of minKnotDistance) works better for long segments to be interpolated (usually first drawing) whereas a more aggressive interpolation works better for short segments (usually editing).
 
 ## How to interpolate
 
@@ -190,12 +190,12 @@ ctToolGroup.setToolConfiguration(PlanarFreehandROITool.toolName, {
 });
 
 // in case we want to tune interpolation process
-// more agressive interpolation
+// more agressive interpolation. I.e there will less knot points and more interpolated points.
 ctToolGroup.setToolConfiguration(PlanarFreehandROITool.toolName, {
-  interpolation: { knotSampleSize: 5 },
+  interpolation: { minKnotDistance: 30 },
 });
-// less agressive interpolation
+// less agressive interpolation. I.e there will more knot points and less interpolated points.
 ctToolGroup.setToolConfiguration(PlanarFreehandROITool.toolName, {
-  interpolation: { knotSampleSize: 30 },
+  interpolation: { minKnotDistance: 5 },
 });
 ```
--- a/packages/docs/docs/concepts/cornerstone-tools/tools.md
+++ b/packages/docs/docs/concepts/cornerstone-tools/tools.md
@@ -158,3 +158,44 @@ interactions.
     </td>
   </tr>
 </table>
+
+## Tool interpolation
+
+Some of tools can be configured to use interpolation process. PlanarFreehandROITool is an example for that.
+The interpolation process uses b-spline algorithm and consider 3 configurations properties:
+
+- interpolation.enabled: to tell whether it should be interpolated or not (false by default)
+- interpolation.knotSampleSize: the distance between each knot on interpolation process (for editing it is considered the property editKnotSampleSize) ( default value 10)
+- interpolation.editKnotSampleSize: same as knotSampleSize but applicable only when editing the tool (default value 20)
+
+So, with that said the interpolation might occur when:
+
+- drawing is done (i.e mouse is released) and interpolation.enabled is true. Interpolation algorithm uses knotSampleSize
+- edit drawing is done (i.e mouse is released) and interpolation.enabled. Interpolation algorithm uses editKnotSampleSize and its only applied to changed segment
+
+Interpolation does not occur when:
+
+- interpolation.enabled is false
+- drawing still happening (editing or not)
+
+The result of interpolation will be a smoother set of segments.
+Changing tool configuration (see below) you can fine-tune the interpolation process and per experience, a less aggressive interpolation works better for long segments to be interpolated (usually first drawing) whereas a more aggressive interpolation works better for short segments (usually editing).
+
+## How to interpolate
+
+```
+// in case we want to turn it on
+ctToolGroup.setToolConfiguration(PlanarFreehandROITool.toolName, {
+  interpolation: { enabled: true },
+});
+
+// in case we want to tune interpolation process
+// more agressive interpolation
+ctToolGroup.setToolConfiguration(PlanarFreehandROITool.toolName, {
+  interpolation: { knotSampleSize: 5 },
+});
+// less agressive interpolation
+ctToolGroup.setToolConfiguration(PlanarFreehandROITool.toolName, {
+  interpolation: { knotSampleSize: 30 },
+});
+```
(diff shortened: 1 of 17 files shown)


packages/docs/docs/concepts/cornerstone-tools/tools.md
```diff
@@ -158,3 +158,44 @@ interactions.
     </td>
   </tr>
 </table>
+
+## Tool interpolation
+
+Some of tools can be configured to use interpolation process. PlanarFreehandROITool is an example for that.
```
wayfarer3130, line 164: I don't think it is "some of tools" - interpolation is really specific to the freehand tool, or anything else creating a region like the freehand tool.  Maybe something like "Contour tools like the FreehandTool can be configured..."

packages/tools/src/tools/annotation/planarFreehandROITool/drawLoop.ts
```diff
@@ -185,10 +189,14 @@ function completeDrawClosedContour(element: HTMLDivElement): boolean {
   // Remove last point which will be a duplicate now.
   canvasPoints.pop();
 
+  const updatedPoints = assertInterpolation(this.configuration)
```
wayfarer3130, line 192: Can you re-use the function from the previous request to figure out which set of points to get?  Eg the getInterpolatedOr... function?

packages/tools/src/tools/annotation/planarFreehandROITool/interpolatePoints.ts
```diff
@@ -0,0 +1,176 @@
+import { Types } from '@cornerstonejs/core';
+import { point } from '../../../utilities/math';
+import interpolateSegmentPoints from './interpolation/interpolateSegmentPoints';
+
+export function assertInterpolation(configuration: Record<any, any>): boolean {
```
wayfarer3130, line 5: This is the one to rename.

packages/tools/src/tools/annotation/planarFreehandROITool/interpolatePoints.ts
```diff
@@ -0,0 +1,176 @@
...
+  direction: number
+): number {
+  return (index + size + direction) % size;
+}
+
+/**
+ * Circular finding that returns the next index for two list where the criteria is met.
+ */
+function circularFindNextIndexBy(
+  pointContent: [number, number, Types.Point2[]],
```
wayfarer3130, line 53: This is too complex a type to really understand what it means from just the types - can you define it separately and have comments as to what the fields are, so that someone could re-use this function without having to necessarily trace through it?

packages/tools/src/tools/annotation/planarFreehandROITool/interpolatePoints.ts
```diff
@@ -0,0 +1,176 @@
...
+  const [, indexDelimiter, points] = pointContent;
+  const [, otherIndexDelimiter, otherPoints] = otherPointContent;
+
+  const pointsLength = points.length;
+  const otherPointsLength = otherPoints.length;
+
+  let startIndex = pointContent[0];
+  let otherStartIndex = otherPointContent[0];
+
+  function validIndex(
```
wayfarer3130, line 67: This function is an example of where you should really be using an arrow function - the optimizer MAY choose to optimize this to an arrow function, but it is better to declare it directly as the type you actually want.

packages/tools/src/tools/annotation/planarFreehandROITool/interpolatePoints.ts
```diff
@@ -0,0 +1,176 @@
...
+    _index: number,
+    _indexDelimiter: number,
+    _points: Types.Point2[]
+  ): boolean {
+    return _points[_index] && _index !== _indexDelimiter;
+  }
+
+  while (
```
wayfarer3130, line 75: What are the return conditions when:
1. Neither initial point is valid initially
2. One of the indices becomes invalid during the iteration
3. The criteria is met?
Are these all the same expected values, or are there edge cases to be considered here?

packages/tools/src/tools/annotation/planarFreehandROITool/interpolation/algorithms/bspline.ts
```diff
@@ -0,0 +1,36 @@
+import {
+  interpolateBasis as d3InterpolateBasis,
+  quantize as d3Quantize,
+} from 'd3-interpolate';
+import { zip as d3Zip } from 'd3-array';
+import { Types } from '@cornerstonejs/core';
+
+/**
+ * Returns a interpolated segment (b-spline algorithm) based on knotsIndexes.
+ * KnotsIndexes tells the first/last index of segment and also which index will be present on interpolation.
```
wayfarer3130, line 10: Would like more explanation here, as later developers may well want to know the intent of the algorithm.  Something like:
Interpolates a set of points indexed by knotsIndexes.  That is, it DISCARDS all points except those in knotsIndexes.  Then, a new set of points is created by using a bspline on the remaining points, in order to re-create a new set of points.   Returns an array of the originalPoints length containing the interpolated data.

packages/tools/src/tools/annotation/planarFreehandROITool/interpolation/interpolateSegmentPoints.ts
```diff
@@ -0,0 +1,59 @@
...
+ * The result total of points is equal to original points.
+ */
+export default function interpolateSegmentPoints(
+  points: Types.Point2[],
+  iniIndex: number,
+  endIndex: number,
+  knotSampleSize: number
+): Types.Point2[] {
+  const segmentSize = endIndex - iniIndex + 1;
+  if (isNaN(segmentSize) || !segmentSize || !knotSampleSize) {
```
wayfarer3130, line 18: You need to check that segmentSize is at least twice the knotSampleSize as you need at least 3 points for interpolation.

packages/tools/src/tools/annotation/planarFreehandROITool/interpolation/interpolateSegmentPoints.ts
```diff
@@ -0,0 +1,59 @@
...
+  );
+
+  const knotsIndexes = [];
+  let intIt = 0;
+  let nextKnotIndex = intIt * knotsDistance + interpolationIniIndex;
+
+  while (nextKnotIndex <= interpolationEndIndex) {
+    knotsIndexes[intIt] = nextKnotIndex;
+    intIt++;
+    nextKnotIndex = intIt * knotsDistance + interpolationIniIndex;
```
wayfarer3130, line 44: This calculation ends up putting all the fractional values into the last element, which produces an irregularity in the bspline result.  You should use the floor value here with the fractional knotsDistance.
Let me give an example - suppose the knotsDistance is 3 and the total length is 10 for the distance - you get:
0,3,6,9,10
as your points - note how 9 and 10 only differ by 1.  If you do this fractionally, you get:
0,2,5,8,10  so the split is on the first and last element.      It also means you can run this as a for loop, since you can calculate exactly how many indices you are going to need, and the last and first ones will both be exact.  That code should be easier to follow.

yarn.lock
```diff
@@ -12798,7 +12798,7 @@ lodash.flow@^3.3.0:
   resolved "https://registry.yarnpkg.com/lodash.flow/-/lodash.flow-3.5.0.tgz#87bf40292b8cf83e4e8ce1a3ae4209e20071675a"
   integrity sha512-ff3BX/tSioo+XojX4MOsOMhJw0nZoUEF011LX8g8d3gvjVbxd89cCio4BCXronjxcTUIJUoqKEUA+n4CqvvRPw==
 
-lodash.get@^4.4.2:
+lodash.get@4.4.2, lodash.get@^4.4.2:
```
wayfarer3130, line 12801: This usually means you have two different package imports, one @4.4.2 and one @^4.4.2 - could you fix it to make them both @^4.4.2 please?

packages/tools/src/tools/annotation/planarFreehandROITool/interpolatePoints.ts
```diff
@@ -0,0 +1,176 @@
...
+import interpolateSegmentPoints from './interpolation/interpolateSegmentPoints';
+
+export function assertInterpolation(configuration: Record<any, any>): boolean {
+  return configuration?.interpolation?.enabled === true;
+}
+
+/**
+ * Tells whether two points are equal by proximity or not as far as interpolation goes.
+ */
+function isEqualByProximity(pointA, pointB) {
```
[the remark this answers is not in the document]
wayfarer3130: If that is what the rest of it uses, then just leave it.

packages/tools/src/tools/annotation/planarFreehandROITool/interpolatePoints.ts
ladeirarodolfo: I have updated the function and its returning type to consider:
If return type is valid it should be [number,number] otherwise should be [undefined, undefined]

packages/tools/src/tools/annotation/planarFreehandROITool/closedContourEditLoop.ts
```diff
@@ -398,13 +402,20 @@ function completeClosedContourEdit(element: HTMLDivElement) {
   const { viewport, renderingEngine } = enabledElement;
 
   const { annotation, viewportIdsToRender } = this.commonData;
-  const { fusedCanvasPoints } = this.editData;
+  const { fusedCanvasPoints, prevCanvasPoints } = this.editData;
 
   if (fusedCanvasPoints) {
-    const worldPoints = fusedCanvasPoints.map((canvasPoint) =>
+    const updatedPoints = assertInterpolation(this.configuration)
+      ? getInterpolatedPoints(
```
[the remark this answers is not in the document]
ladeirarodolfo: The explanation/doc of each function is already on function itself.
I dont see a point of have a function (with a potential overladed name) just because we want to have a better explanation. I would say this is an unnecessary encapsulation we should avoid.

packages/tools/src/tools/annotation/planarFreehandROITool/interpolation/interpolateSegmentPoints.ts
```diff
@@ -0,0 +1,84 @@
...
+ * Formula for reference
+ * For given {x ∈ R | x ≥ 0} and {minDis ∈ R | minDis ≥ 0}, ∃ D(x) where D(x) ≥ a and D(x) ≤ b  =>
+ *         |
+ * D(x)  = |                  (b - a)
+ *         |  round( ------------------------ * x  )   + a
+ *         |                (b - a + 1)
+ *         |        round( -----------  )
+ *         |                 minDis
+ */
+function getContinuosUniformDistributionValues(
```
wayfarer3130, line 18: Minor quibble - getContinuousUniformDistributionValues
There is a second u in Continuous

packages/tools/src/tools/annotation/planarFreehandROITool/interpolation/interpolateSegmentPoints.ts
```diff
@@ -0,0 +1,84 @@
...
+
+  let x = 0;
+  let continuosDistributionValue =
+    Math.round(((intervalSize - 1) / (intensity - 1)) * x) + intervalIni;
+
+  while (continuosDistributionValue <= intervalEnd) {
+    result.push(continuosDistributionValue);
+    x++;
+    continuosDistributionValue =
+      Math.round(((intervalSize - 1) / (intensity - 1)) * x) + intervalIni;
```
wayfarer3130, line 36: Thank you - this code is easier to read.

wayfarer3130:
LGTM - just the one misspelling.

PR #147, opened 2022-07-14
feat: Add camera sync relative to the canvas

wayfarer3130:
The existing camera sync was a camera copy, which broke display of
images in different frame of reference uids/with different initial
viewing conditions, in a way that completely prevented viewing them.
This version preserves pan and zoom relative
to the initial viewing conditions.  As well, it takes input parameters
to set the initial pan/zoom conditions.

### DIFF
--- a/packages/core/src/RenderingEngine/Viewport.ts
+++ b/packages/core/src/RenderingEngine/Viewport.ts
@@ -304,8 +304,6 @@ class Viewport implements IViewport {
       actor.setUserMatrix(mat);
     });
 
-    this.resetInitialOffsets();
-
     this.getRenderingEngine().render();
   }
 
@@ -669,8 +667,7 @@ class Viewport implements IViewport {
     if (this.flipHorizontal || this.flipVertical) {
       this.flip({ flipHorizontal: false, flipVertical: false });
     }
-
-    this.resetInitialOffsets(resetOffsets);
+    this.resetInitialOffsets(!!resetOffsets);
 
     const RESET_CAMERA_EVENT = {
       type: 'ResetCameraEvent',
@@ -697,7 +694,9 @@ class Viewport implements IViewport {
    * @param resetOffsets can be passed to skip resetting, ie a no-op on this call
    */
   public resetInitialOffsets(resetOffsets = true) {
-    if (!resetOffsets) return;
+    if (!resetOffsets) {
+      return;
+    }
     this.initialCamera = this.getCamera();
   }
 
@@ -920,7 +919,7 @@ class Viewport implements IViewport {
       renderer.resetCameraClippingRange();
     }
 
-    this.resetInitialOffsets(resetOffsets);
+    this.resetInitialOffsets(!!resetOffsets);
 
     this.triggerCameraModifiedEventIfNecessary(previousCamera, updatedCamera);
   }
--- a/packages/core/src/RenderingEngine/Viewport.ts
+++ b/packages/core/src/RenderingEngine/Viewport.ts
@@ -3,7 +3,7 @@ import vtkMatrixBuilder from '@kitware/vtk.js/Common/Core/MatrixBuilder';
 import vtkMath from '@kitware/vtk.js/Common/Core/Math';
 import vtkPlane from '@kitware/vtk.js/Common/DataModel/Plane';
 
-import { vec3, mat4 } from 'gl-matrix';
+import { vec2, vec3, mat4 } from 'gl-matrix';
 import _cloneDeep from 'lodash.clonedeep';
 
 import Events from '../enums/Events';
@@ -66,6 +66,10 @@ class Viewport implements IViewport {
   /** A flag representing if viewport methods should fire events or not */
   readonly suppressEvents: boolean;
   protected hasPixelSpacing = true;
+  /** The camera that is initially defined on the reset for
+   * the relative pan/zoom
+   */
+  protected initialCamera: ICamera;
 
   constructor(props: ViewportInput) {
     this.id = props.id;
@@ -298,9 +302,9 @@ class Viewport implements IViewport {
       }
 
       actor.setUserMatrix(mat);
-
-      this.getRenderingEngine().render();
     });
+
+    this.resetInitialOffsets();
 
     this.getRenderingEngine().render();
   }
@@ -536,9 +540,14 @@ class Viewport implements IViewport {
    * is reset for the current view.
    * @param resetPan - If true, the camera focal point is reset to the center of the volume (slice)
    * @param resetZoom - If true, the camera zoom is reset to the default zoom
+   * @param resetOffsets - If true, the pan/zoom initial offsets are reset.
    * @returns boolean
    */
-  protected resetCamera(resetPan = true, resetZoom = true): boolean {
+  protected resetCamera(
+    resetPan = true,
+    resetZoom = true,
+    resetOffsets = true
+  ): boolean {
     const renderer = this.getRenderer();
     const previousCamera = _cloneDeep(this.getCamera());
 
@@ -623,11 +632,7 @@ class Viewport implements IViewport {
       activeCamera.setViewUp(-viewUp[2], viewUp[0], viewUp[1]);
     }
 
-    let focalPointToSet = focalPoint;
-
-    if (!resetPan) {
-      focalPointToSet = previousCamera.focalPoint;
-    }
+    const focalPointToSet = resetPan ? focalPoint : previousCamera.focalPoint;
 
     activeCamera.setFocalPoint(
       focalPointToSet[0],
@@ -661,14 +666,16 @@ class Viewport implements IViewport {
       RENDERING_DEFAULTS.MAXIMUM_RAY_DISTANCE
     );
 
+    if (this.flipHorizontal || this.flipVertical) {
+      this.flip({ flipHorizontal: false, flipVertical: false });
+    }
+
+    this.resetInitialOffsets(resetOffsets);
+
     const RESET_CAMERA_EVENT = {
       type: 'ResetCameraEvent',
       renderer,
     };
-
-    if (this.flipHorizontal || this.flipVertical) {
-      this.flip({ flipHorizontal: false, flipVertical: false });
-    }
 
     // Here to let parallel/distributed compositing intercept
     // and do the right thing.
@@ -680,6 +687,97 @@ class Viewport implements IViewport {
     );
 
     return true;
+  }
+
+  /**
+   * Resets the stored initial offset values for
+   * zoom and pan.  These values are the offset
+   * values used to allow the getPan and getZoom
+   * to return fixed initial values.
+   * @param resetOffsets can be passed to skip resetting, ie a no-op on this call
+   */
+  public resetInitialOffsets(resetOffsets = true) {
+    if (!resetOffsets) return;
+    this.initialCamera = this.getCamera();
+  }
+
+  /**
+   * Helper function to return the current canvas pan value.
+   *
+   * @returns a Point2 containing the current pan values
+   * on the canvas,
+   * computed from the current camera, where the initial pan
+   * value is [0,0].
+   */
+  public getPan(): Point2 {
+    const activeCamera = this.getVtkActiveCamera();
+    const focalPoint = activeCamera.getFocalPoint() as Point3;
+
+    const zero3 = this.canvasToWorld([0, 0]);
+    const initialCanvasFocal = this.worldToCanvas(
+      <Point3>vec3.subtract(vec3.create(), this.initialCamera.focalPoint, zero3)
+    );
+    const currentCanvasFocal = this.worldToCanvas(
+      <Point3>vec3.subtract(vec3.create(), focalPoint, zero3)
+    );
+    const result = <Point2>(
+      vec2.subtract(vec2.create(), initialCanvasFocal, currentCanvasFocal)
+    );
+    return result;
+  }
+
+  /**
+   * Sets the canvas pan value relative to the initial view position of 0,0
+   * Modifies the camera to perform the pan.
+   */
+  public setPan(pan: Point2, resetOffsets = false) {
+    const previousCamera = this.getCamera();
+    const { focalPoint, position } = previousCamera;
+    const zero3 = this.canvasToWorld([0, 0]);
+    const delta2 = vec2.subtract(vec2.create(), pan, this.getPan());
+    if (Math.abs(delta2[0]) < 1 && Math.abs(delta2[1]) < 1 && !resetOffsets) {
+      return;
+    }
+    const delta = vec3.subtract(
+      vec3.create(),
+      this.canvasToWorld(<Point2>delta2),
+      zero3
+    );
+    const newFocal = vec3.subtract(vec3.create(), focalPoint, delta);
+    const newPosition = vec3.subtract(vec3.create(), position, delta);
+    this.setCamera({
+      ...previousCamera,
+      focalPoint: newFocal as Point3,
+      position: newPosition as Point3,
+      resetOffsets,
+    });
+  }
+
+  /**
+   * Returns a current zoom level relative to the initial parallel scale
+   * originally applied to the image.  That is, on initial display,
+   * the zoom level is 1.  Computed as a function of the camera.
+   */
+  public getZoom() {
+    const activeCamera = this.getVtkActiveCamera();
+    const { parallelScale: initialParallelScale } = this.initialCamera;
+    return initialParallelScale / activeCamera.getParallelScale();
+  }
+
+  /** Zooms the image using parallel scale by updating the camera value.
+   * @param value is the relative parallel scale to apply.  It is relative
+   * to the initial offsets value.
+   * @param resetOffsets can be set to true to reset the zoom to the specified
+   *    value as a "1" value.
+   */
+  public setZoom(value: number, resetOffsets = false) {
+    const camera = this.getCamera();
+    const { parallelScale: initialParallelScale } = this.initialCamera;
+    const parallelScale = initialParallelScale / value;
+    if (camera.parallelScale === parallelScale && !resetOffsets) {
+      return;
+    }
+    this.setCamera({ ...camera, parallelScale, resetOffsets });
   }
 
   /**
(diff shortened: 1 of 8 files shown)


packages/core/src/RenderingEngine/Viewport.ts
```diff
@@ -682,6 +707,114 @@ class Viewport implements IViewport {
     return true;
   }
 
+  public resetPanOffset(widthWorld?: number, heightWorld?: number) {
```
wayfarer3130, line 710: The basic idea is to set the pan and zoom to [0,0] and 1 after applying the initial resetCamera conditions, as well as
including the initial center and canvas center.

wayfarer3130:
@rodrigoea - can you take a look at this PR please?  It provides support for both the "remember last zoom/pan" as well as the ability to synchronize zoom/pan on images not in the same initial frame of reference.

packages/core/src/RenderingEngine/Viewport.ts
```diff
@@ -682,6 +707,114 @@ class Viewport implements IViewport {
...
+        focalPoint[1] + distanceVec[1],
+        focalPoint[2] + distanceVec[2]
+      );
+    }
+
+    this.panOffset = [0, 0];
+    this.panOffset = this.getCanvasPan();
+  }
+
+  // The initial canvas pan of 0,0 means
```
wayfarer3130, line 774: Explain that this is just a function of the current camera position that returns values in terms of the initial position of the canvas.

packages/core/src/RenderingEngine/Viewport.ts
```diff
@@ -66,6 +82,8 @@ class Viewport implements IViewport {
   /** A flag representing if viewport methods should fire events or not */
   readonly suppressEvents: boolean;
   protected hasPixelSpacing = true;
+  protected initialScale = 1;
```
wayfarer3130, line 85: initialPanOffset
initialParallelScale

packages/core/src/RenderingEngine/Viewport.ts
```diff
@@ -24,6 +24,22 @@ import type {
 import type { ViewportInput, IViewport } from '../types/IViewport';
 import type { vtkSlabCamera } from './vtkClasses/vtkSlabCamera';
 
+// Pure function versions of vector operations
+const subtractVec2 = (p1: Point2, p2: Point2): Point2 => [
```
wayfarer3130, line 28: Use Vec3/Vec2

packages/core/src/RenderingEngine/Viewport.ts
```diff
@@ -682,6 +707,114 @@ class Viewport implements IViewport {
...
+    const activeCamera = this.getVtkActiveCamera();
+    const focalPoint = <Point3>activeCamera.getFocalPoint();
+    const center = <Point2>[0, 0];
+    const center3 = this.canvasToWorld(center);
+    const delta = this.worldToCanvas(subtractVec3(focalPoint, center3));
+    const result = subtractVec2(delta, this.panOffset);
+    return result;
+  }
+
+  /** Sets the canvas pan value relative to the initial view position of 0,0 */
```
wayfarer3130, line 786: Set it to 0,0 to reset to the initial pan position

packages/core/src/RenderingEngine/Viewport.ts
```diff
@@ -682,6 +707,114 @@ class Viewport implements IViewport {
...
+    });
+  }
+
+  /** Returns a current zoom level relative to the base zoom */
+  public getZoom() {
+    const activeCamera = this.getVtkActiveCamera();
+    return activeCamera.getParallelScale() / this.initialScale;
+  }
+
+  public setZoom(value: number) {
```
wayfarer3130, line 811: Add documentation here - this is relative to the initial values

packages/core/src/RenderingEngine/Viewport.ts
```diff
@@ -682,6 +707,114 @@ class Viewport implements IViewport {
...
+    const newFocal = addVec3(delta, focalPoint);
+    const newPosition = addVec3(delta, position);
+    this.setCamera({
+      ...previousCamera,
+      focalPoint: newFocal,
+      position: newPosition,
+    });
+  }
+
+  /** Returns a current zoom level relative to the base zoom */
```
wayfarer3130, line 805: Returns the parallel scale value relative to the original parallel scale.  That is, returns 1 on initial display, and the to magnify the image, values larger than 1 can be set (eg 2 would be twice the original size)

packages/tools/src/synchronizers/callbacks/cameraSyncCallback.ts
```diff
@@ -1,4 +1,4 @@
-import { getRenderingEngine, Types } from '@cornerstonejs/core';
+import { getRenderingEngine, Types, VolumeViewport } from '@cornerstonejs/core';
```
ladeirarodolfo, line 1: it seems VolumeViewport import is not being used.

ladeirarodolfo (changes requested):

packages/core/src/types/ViewportInputOptions.ts
```diff
@@ -10,6 +11,9 @@ type ViewportInputOptions = {
   orientation?: Orientation;
   /** whether the events should be suppressed and not fired*/
   suppressEvents?: boolean;
+  initialRange?: Point2;
```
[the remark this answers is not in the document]
wayfarer3130: No, it isn't fully supported with all those changes, but the plan is to do additional work over time to support this.

packages/tools/src/synchronizers/callbacks/cameraSyncCallback.ts
```diff
@@ -34,9 +34,14 @@ export default function cameraSyncCallback(
   const tViewport = renderingEngine.getViewport(targetViewport.viewportId);
 
   // TODO: only sync in-plane movements if one viewport is a stack viewport
+  if (camera.parallelProjection) {
```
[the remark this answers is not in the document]
wayfarer3130: Yes, for now that is acceptable - Eric and I discussed it, and we want an options object to be passed into the synchronizer so that exactly what gets synced can be controlled.  
If you sync everything on any viewport that doesn't start with the same initial conditions, it puts the images offscreen typically, so it wasn't working previously except in a very narrow case.  This version is much more robust, although there are still bugs.

packages/core/examples/panZoom/index.ts
```diff
@@ -0,0 +1,127 @@
...
+  'Click on index.ts to open source code for this example --------->'
+);
+
+const { ViewportType } = Enums;
+const renderingEngineId = 'myRenderingEngine';
+const viewportId = 'CT_STACK';
+
+// ======== Set up page ======== //
+setTitleAndDescription(
+  'Pan and Zoom',
```
sedghi, line 26: This example should be called differently to remove the confusion that this is how pan/zoom works. Maybe 'progarmatic pan/zoom'?

packages/core/examples/panZoom/index.ts
```diff
@@ -0,0 +1,127 @@
...
+  const renderingEngine = new RenderingEngine(renderingEngineId);
+
+  // Create a stack viewport
+  const viewportInput = {
+    viewportId,
+    type: ViewportType.STACK,
+    element,
+    defaultOptions: {
+      background: <Types.Point3>[0.2, 0, 0.2],
+      initialRange: [0.6, 0.6],
```
sedghi, line 101: Needs change after API change

packages/core/src/types/ViewportInputOptions.ts
```diff
@@ -10,6 +11,9 @@ type ViewportInputOptions = {
...
+  canvasCenter?: Point2;
+  initialCenter?: Point2;
```
sedghi, line 16: All require proper documentation, we use these commments in the api doc https://www.cornerstonejs.org/api/core/namespace/Types#ViewportInputOptions

packages/docs/docs/concepts/cornerstone-core/viewports.md
```diff
@@ -20,6 +20,34 @@ viewed via 4 viewports in a “4-up” view: Axial MPR, Sagittal MPR, Coronal MP
...
+the center of the canvas (at 512,400). This allows full control over alignment
+of images such as mammography images, without needing to know the actual size
+of the canvas and image beforehand.
+
+## Initial State helper functions
+
+There area few functions that act on the viewport relative to the initial state.
+The zoom function change the size of hte image relative to the initial zoom of 1,
+while the pan functions change the pan relative to the initial pan of [0,0].
+
```
sedghi, line 50: This document is a top level explanation of the viewports. We don't need to add complexity by explaining these optional arguments (similar to how we don't explain background, orientation etc. here). These comments should be moved to the Viewport (class) on top of the property for it to be properly shown in the api doc (see here https://www.cornerstonejs.org/api/core/class/StackViewport).

packages/tools/src/synchronizers/callbacks/cameraSyncCallback.ts
```diff
@@ -34,9 +37,15 @@ export default function cameraSyncCallback(
...
+    const sViewport = renderingEngine.getViewport(sourceViewport.viewportId);
+    const srcDelta = sViewport.getPan();
 
-  // Todo: we shouldn't set camera, we should set the focalPoint
-  // to the nearest slice center world position
-  tViewport.setCamera(camera);
+    // Do the zoom first, as the pan is relative to the zoom level
+    tViewport.setZoom(sViewport.getZoom());
```
sedghi, line 45: I don't like the inconsistensy in the set/get here. Can we have `const srcZoom = ...` similar to the `const srcDelta ...` instead of the `.setZoom(sViewport.getZoom())` ?

packages/tools/src/tools/PanTool.ts
```diff
@@ -29,6 +29,12 @@ export default class PanTool extends BaseTool {
     const enabledElement = getEnabledElement(element);
 
     const deltaPointsWorld = deltaPoints.world;
+    if (
+      deltaPointsWorld[0] === 0 &&
+      deltaPointsWorld[1] === 0 &&
+      deltaPointsWorld[2] === 0
+    )
+      return;
```
sedghi, line 37: The linter should warn you here (if not needs to be fixed), throughout the repo we always use curly brackets for the if statements

packages/core/src/RenderingEngine/Viewport.ts
```diff
@@ -599,7 +603,14 @@ class Viewport implements IViewport {
     }
 
     //const angle = vtkMath.radiansFromDegrees(activeCamera.getViewAngle())
-    const parallelScale = 1.1 * radius;
+    const [scaleX, scaleY] =
```
sedghi, line 606: If `initialRange` is not scale as we discussed, it should not be named scale

packages/core/src/RenderingEngine/Viewport.ts
```diff
@@ -47,6 +47,8 @@ class Viewport implements IViewport {
   protected flipHorizontal = false;
   protected flipVertical = false;
   protected rotation = 0;
+  static readonly INITIAL_RANGE = [1.1, 1.1];
+  static readonly CENTER = [0.5, 0.5];
```
sedghi, line 51: I don't understand why these are static properties? Do we ever need them without viewport instantiation?

packages/docs/docs/concepts/cornerstone-core/viewports.md
```diff
@@ -20,6 +20,34 @@ viewed via 4 viewports in a “4-up” view: Axial MPR, Sagittal MPR, Coronal MP
...
+### initialCenter and canvasCenter
+
+The initialCenter and canvasCenter are percentage of image size and canvas size,
+such that the two points correspond initially. That is, suppose the
+initial center is [0.5, 1] and the canvas center is 0.5,0.5, with a 512x512 image
+on a 1024x800 canvas, then the image point 256,512 should be displayed overtop
+the center of the canvas (at 512,400). This allows full control over alignment
+of images such as mammography images, without needing to know the actual size
+of the canvas and image beforehand.
```
sedghi, line 43: I love the details in this comment

packages/core/src/RenderingEngine/Viewport.ts
```diff
@@ -661,15 +669,19 @@ class Viewport implements IViewport {
       RENDERING_DEFAULTS.MAXIMUM_RAY_DISTANCE
     );
 
+    if (this.flipHorizontal || this.flipVertical) {
+      this.flip({ flipHorizontal: false, flipVertical: false });
+    }
+
+    if (resetPan && resetZoom) {
```
sedghi, line 676: why it needs `resetZoom` in order to `resetPanOffset`?

packages/core/src/RenderingEngine/Viewport.ts
```diff
@@ -682,6 +694,151 @@ class Viewport implements IViewport {
     return true;
   }
 
+  /**
+   * Resets the pan offset.  To do this, the options values
```
sedghi, line 698: It says `resets the pan offset` but it should define what is actually the resetting the offset instead of how to do this. The how to do this should be part of the function (preferrably step by step guide of how we are doing it), and the function documentation should be what this function achieves. I expected this function to be a simple `this.panOffset = [0,0]` but it is not.

packages/core/src/RenderingEngine/Viewport.ts
```diff
@@ -682,6 +694,151 @@ class Viewport implements IViewport {
...
+   * are applied to the vtk active camera, then the current pan
+   * value is read and is set as the initial conditions.  That
+   * sets current pan to 0,0
+   * The initialCenter is the IMAGE percentage coordinates for the
+   * image synchronization point.  That point is displayed at the
+   * percentage canvasCenter point.
+   * For example, suppose the imageCenter is (1,0.5) and the canvasCenter
+   * is (0.5,1), then on an image 513x513 on a 1025x801 canvas, the
+   * image point (256,512) will be displayed at canvas position (512,800)
+   * The initial pan values will then be 0,0.
```
sedghi, line 709: Again these should not be part of the function documentation, as it makes the doc harder to read. These are the definitions of the initialCenter and canvasCenter (which are not even arguments of this function)

packages/tools/src/synchronizers/callbacks/cameraSyncCallback.ts
```diff
@@ -34,9 +37,15 @@ export default function cameraSyncCallback(
...
+    tViewport.setPan(srcDelta);
```
sedghi, line 46: What happens to the scroll sync? Isn't scrolling sync handled by the cameraSyncCallback? Previously it was handled by .setCamera but now it is not (since only we setZoom and setPan here)

packages/core/src/RenderingEngine/Viewport.ts
sedghi: I still think resetCamera should neglect all the provided options and reset based on the FOV and actor bounds. IMO resetCamera should not go back to the state which you have defined with (`initialCenter` etc.). If you have put it in the `resetCamera` in order for the initial render to be correct, it can, instead, be a public API like `viewport.setPresentationState()` that accept those. 

CC'ing @swederik

sedghi: I'm even thinking that `PublicViewportInput` can ommit these presentational states, and the deverloper can just do 

`viewport.setStack(imageId).then(()=>{
  viewport.setPresentationState(initialRange, initialCenter, etc.) // we discussed these names should change
  viewport.render()
})

sedghi (changes requested):
Here are my comments: 


1. I still think resetCamera should neglect all the provided options and reset based on the FOV and actor bounds. IMO resetCamera should not go back to the state which you have defined with (initialCenter etc.).

2. re property names, as discussed, the namings should be revisited. `range` is certainly out of questions, and there is not center to initialCenter. Here is what I think out loud (I'm open to suggestions)

```
displayArea: {
    imageToCanvasMapping: [1,0, 0.5,0.5] // sets the top-right of the image at the center of canvas; 
    scale: [1, 1] // internally calculate the range based on the width, maybe?
}

```

3. I'm thinking that PublicViewportInput can ommit these presentational states, and the API can just be



```
viewport.setStack(imageId).then(()=>{
  viewport.setCamera(displayArea: {imageToCanvasMapping, scale}) // we discussed these names should change
  viewport.render()
}) 
```

in fact this is exactly a camera manipulation (involving zoom & pan), which should be done through camera API, similar to flip.

PS: I'm open to change presentationState if you have a better name

CC'ing @swederik

packages/core/src/RenderingEngine/Viewport.ts
wayfarer3130: They are just the defaults for the values.  Will delete them.

packages/core/src/RenderingEngine/Viewport.ts
```diff
@@ -682,6 +694,151 @@ class Viewport implements IViewport {
...
+      const activeCamera = this.getVtkActiveCamera();
+      const { viewPlaneNormal, focalPoint, position, viewUp } = camera;
+      const distanceVec = vec3.subtract(vec3.create(), position, focalPoint);
+
+      const { initialCenter = Viewport.CENTER } = this.options;
+      const { canvasCenter = initialCenter } = this.options;
+
+      if (!focalPoint || focalPoint[0] === null) return;
+
+      const worldSizeError = 1.15;
```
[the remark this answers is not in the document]
wayfarer3130: Removed - fixes the demo app now.  Something else got fixed so this wasn't needed, and was causing problems.

packages/tools/src/synchronizers/callbacks/cameraSyncCallback.ts
wayfarer3130: We probably need a separate call for that - options are:
1. Can pass scroll sync as the 3rd parameter in pan (eg setPan(Point3) instead of Point2)
2. Have separate setScroll/getScroll methods

It looks like almost all of this is defined in the scroll.ts file, plus scrollTool, but if this change is part of the overall change, then the scrolling should be defined as core functionality in Viewport, so that there aren't multiple methods dealing with scrolling.  Do you want this included right now?

wayfarer3130:
The camera fix sync doesn't need to be in the core cornerstoneJS code - the sync can be done externally entirely.

PR #151, opened 2022-07-20
fix: annotation unit hydration bug and more color image support

sedghi:
- Add default units for annotations
- Add example for local upload/drag and drop of dicom images


packages/tools/src/tools/annotation/EllipticalROITool.ts
```diff
@@ -917,7 +917,9 @@ export default class EllipticalROITool extends AnnotationTool {
     if (area) {
       areaLine = isEmptyArea
         ? `Area: Oblique not supported`
-        : `Area: ${area.toFixed(2)} ${areaUnit}${String.fromCharCode(178)}`;
+        : `Area: ${area.toFixed(2)} ${areaUnit ?? 'mm'}${String.fromCharCode(
```
wayfarer3130, line 920: Can you use:
 : `Area: ${area.toFixed(2)} ${areaUnit ?? 'mm'}$\xb2`
so that it is kept to one line?

packages/tools/src/tools/annotation/BidirectionalTool.ts
```diff
@@ -1172,8 +1172,8 @@ export default class BidirectionalTool extends AnnotationTool {
     // spaceBetweenSlices & pixelSpacing &
     // magnitude in each direction? Otherwise, this is "px"?
     const textLines = [
-      `L: ${length.toFixed(2)} ${unit}`,
-      `W: ${width.toFixed(2)} ${unit}`,
+      `L: ${length.toFixed(2)} ${unit ?? 'mm'}`,
```
wayfarer3130, line 1175: This should probably be 'px', as the default assumption is that units are in pixel values, not that they are in mm.  That is a patient safety issue, because an assumption of mm when it is actually px means the user could assume it has been calibrated, and use it, eg for surgery.  At a minimum, I would import the default unit from a common place.

packages/tools/src/tools/annotation/LengthTool.ts
```diff
@@ -690,7 +690,7 @@ class LengthTool extends AnnotationTool {
       return;
     }
 
-    const textLines = [`${length.toFixed(2)} ${unit}`];
+    const textLines = [`${length.toFixed(2)} ${unit ?? 'mm'}`];
```
wayfarer3130, line 693: Down in _calculateCachedStats, the unit is read - could you add a parent class function to AnnotationTool that gets the unit instead, and has the default unit for the image type?  Something like:
protected getUnit(image,position): string
This should be called for one of the points in the annotation - don't assume the units are consistent across the image, because they aren't for ultrasound (units can be Hertz in one part, mm in another, and mm/s in a third part).  Then, that call can be used in cached stats, and for the default cached stats, you can just assign the default image units, or reference it via an attribute, eg:
this.defaultUnit

That also allows fixing the units to show projected versus imaging units when relevant for CR/DX images.

sedghi: Hmmm, I hear what you are saying here, but for ultrasound, are you saying the pixelSpacing can be anything other than mm or px? Isn't the unit for pixelSpacing mm in dicom?

sedghi:
@wayfarer3130 I have gone to a different route (other than default unit), I'm checking if the cachedStats has any unit, if not re-calculate (which goes through the pixel spacing check)

packages/tools/src/tools/annotation/LengthTool.ts
wayfarer3130: Probably don't need to worry about different units right now, but having a function called in the parent would be useful, as there are other bits that are sometimes important to check.

sedghi: I ended up don't changing/assigning default units for this PR, can you review the latest?

sedghi:
Need https://github.com/cornerstonejs/cornerstoneWADOImageLoader/pull/457 to be merged first, and version bumped

packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -1067,7 +1067,11 @@ class StackViewport extends Viewport implements IStackViewport {
     // TODO: this function will need to have more logic later
     // see http://dicom.nema.org/medical/Dicom/current/output/chtml/part03/sect_C.7.6.3.html#sect_C.7.6.3.1.2
     let numberOfComponents = 1;
-    if (photometricInterpretation === 'RGB') {
+    if (
+      photometricInterpretation === 'RGB' ||
+      photometricInterpretation === 'YBR_FULL' ||
```
wayfarer3130, line 1072: photometricInterpretation.indexOf('YBR')!=-1 would be better - there are 5-6 YBR versions

packages/tools/src/tools/annotation/LengthTool.ts
wayfarer3130: Yes, the latest looks good.

wayfarer3130:
LGTM other than the YBR index of fix

packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -1067,7 +1067,11 @@ class StackViewport extends Viewport implements IStackViewport {
     // TODO: this function will need to have more logic later
     // see http://dicom.nema.org/medical/Dicom/current/output/chtml/part03/sect_C.7.6.3.html#sect_C.7.6.3.1.2
     let numberOfComponents = 1;
-    if (photometricInterpretation === 'RGB') {
+    if (
+      photometricInterpretation === 'RGB' ||
```
swederik, line 1071: Missing three of them:

YBR_ICT, YBR_RCT, PALETTE COLOR

PR #154, opened 2022-07-28
feat: volume viewport api with setProperties

sedghi:


packages/core/src/types/VolumeViewportProperties.ts
```diff
@@ -0,0 +1,11 @@
+import { VOIRange } from './voi';
+
+/**
+ * Stack Viewport Properties
+ */
+type StackViewportProperties = {
```
swederik, line 6: Wrong name

swederik (approved):
Fine to merge if you fix the name

PR #156, opened 2022-08-01
feat change interpolation approach to be based on a percentage of poi…

ladeirarodolfo:
## Includes
- Improved configuration for FreeHand (interpolation):
       - Ability to: enable interpolation for adding or editing, change interpolation level  for adding or editing
- Interpolation level configuration is based on percentage of points to be considered as knots
- Exposed interpolationAnnotation public api, so consumer can interpolate a Freehand anytime.


packages/docs/docs/concepts/cornerstone-tools/tools.md
```diff
@@ -164,22 +164,25 @@ interactions.
...
 
-- interpolation.enabled: to tell whether it should be interpolated or not (false by default)
-- interpolation.minKnotDistance: the minimum distance between each knot on interpolation process (for editing it is considered the property editMinKnotDistance) ( default value 20)
-- interpolation.editMinKnotDistance: same as minKnotDistance but applicable only when editing the tool (default value 10)
+- interpolation.enabled: to tell whether it should be interpolated or not (for editing it is considered the property editEnabled) (default: false)
+- interpolation.editEnabled: to tell whether it should be interpolated or not when editing (default: false)
+- interpolation.knotsRatioPct: percentage of points from Segment that are likely to be considered knots during interpolation (for editing it is considered the property editKnotsRatioPct) ( default: 40)
+- interpolation.editKnotsRatioPct: same as knotsRatioPct but applicable only when editing the tool (default: 40)
+- interpolation.forceAllSegment: tell whether interpolation occurs on edited segment only or all drawing (true means all drawing) (default: false)
```
sedghi, line 171: Hmmm, I don't think this is the right place to explain about all of this options. This page is a very high level concept of tools, and what is there to use rather than tool specific options. I even don't like the fact that we have `Tool interpolation` here (it was merged when I was out I guess). Can we keep this place as high level as possible, and you comment in the Contour tool class so that it appears in the API docs? Similar to here https://www.cornerstonejs.org/api/tools/class/WindowLevelTool for windowLevel tool.


Also please remove the `Tool Interpolation` and `how to interpolate` section from here. Where to put you might ask, I think the API doc is enough. We _can_ put a subsection (at the level of annotation tools) called contour tools, and just mention we have contour tools.

sedghi: Actually we have it here https://www.cornerstonejs.org/api/tools/class/planarfreehandroitool/ there is some doc, just move these two sections there

packages/tools/src/tools/annotation/PlanarFreehandROITool.ts
```diff
@@ -158,9 +158,11 @@ class PlanarFreehandROITool extends AnnotationTool {
         // be infinite as the lines become very computationally expensive to draw.
         subPixelResolution: 4,
         interpolation: {
+          editEnabled: false, // used for edit only
```
sedghi, line 161: I know you have put excellent docs on the .md file, can you copy paste here for each of these params in summary, and the full thing above the class so that it appears in the api doc?

sedghi: what is `editEnabled` vs `enabled`? maybe we choose another name? `editMode`?

packages/tools/src/tools/annotation/PlanarFreehandROITool.ts
```diff
@@ -158,9 +158,11 @@ class PlanarFreehandROITool extends AnnotationTool {
...
           enabled: false,
-          minKnotDistance: 20,
-          editMinKnotDistance: 10, // used for edit only
+          knotsRatioPct: 40,
```
sedghi, line 163: Do you mean `percentage` by `Pct`? I think `knowRatioPercentage` is more readable

packages/tools/src/tools/annotation/PlanarFreehandROITool.ts
sedghi: I think `editEnabled` means whether to interpolate on edit or not, hmmm, maybe we go with something else like `interpolateOnEdit`?

packages/tools/src/tools/annotation/PlanarFreehandROITool.ts
```diff
@@ -158,9 +158,11 @@ class PlanarFreehandROITool extends AnnotationTool {
...
+          editKnotsRatioPct: 40,
```
sedghi, line 164: Same here, instead of `editKnotsRatioPct`, maybe `knotsRatioPercentageOnEdit` ...?

packages/tools/src/tools/annotation/PlanarFreehandROITool.ts
```diff
@@ -158,9 +158,11 @@ class PlanarFreehandROITool extends AnnotationTool {
...
+          forceAllSegment: false,
```
sedghi, line 165: Not even sure what this is doing, force All segment to do what? IMO the variable name (since it is in config of the tool) should somehow tells what it does

sedghi (changes requested):
Looks like a great addition, I put some comments. I have another comment for the https://www.cornerstonejs.org/docs/examples page which is https://github.com/cornerstonejs/cornerstone3D-beta/blob/4f157dc5d7a8d0d80abb5b68c35ed17cb5f349ed/utils/ExampleRunner/example-info.json - can you include contours and interpolation in the text of the freehandtool example? since people might search for them as well.

packages/tools/src/tools/annotation/PlanarFreehandROITool.ts
ladeirarodolfo: removed it

packages/tools/src/utilities/planar/index.ts
```diff
@@ -2,17 +2,20 @@ import filterAnnotationsWithinSlice from './filterAnnotationsWithinSlice';
...
   getPointInLineOfSightWithCriteria,
+  interpolateAnnotation,
 };
 
 export {
   filterAnnotationsWithinSlice,
   getWorldWidthAndHeightFromCorners,
   filterAnnotationsForDisplay,
   getPointInLineOfSightWithCriteria,
+  interpolateAnnotation,
```
sedghi, line 20: I don't think it belongs here in the planar, since it is specific to the freehand as the annotation. Maybe we create the freehand name space and we can access this utility as `csTools.utilities.freeHandROITool.interpolatePoints/interpolateData`?

packages/tools/src/utilities/planar/interpolateAnnotation.ts
```diff
@@ -0,0 +1,80 @@
...
+    )
+  );
+
+  if (interpolatedPolyline === annotation.data.polyline) {
+    return false;
+  }
+
+  annotation.data.polyline = interpolatedPolyline;
+
+  enabledElement.viewport.render();
```
sedghi, line 77: make this an optional argument such as `immediate` we have such flags in the renderingEngine for the usecase that we don't need the render right away

sedghi (approved):
Getting there, just two small comments. The major one is that `csTools.utilities.planar.interpolateAnnotation` sounds like we can interpolate Length, Bidirectional, etc. while the interpolation is specific to the freehandROITool.

So approved given that change + addition of the immediate flag.

Nice work

packages/tools/src/utilities/planar/interpolateAnnotation.ts
ladeirarodolfo: ok I will change

packages/tools/src/utilities/planar/index.ts
ladeirarodolfo: ok I will change

PR #161, opened 2022-08-03
fix: unify handling of annotation units and remove 'MO'

pwespi:
This pull request addresses two points:
1) Unify handling of annotation units.

So far, each annotation (elliptical, rectangle, probe) was handling units on its own (and in a slightly different way). This is unified and the actual function is taken from https://github.com/OHIF/Viewers/blob/ae897c47e08611014ad1918b567df36444764329/extensions/cornerstone/src/utils/measurementServiceMappings/utils/getModalityUnit.js

2) By taking the function linked above, `MO` is not shown anymore. 

Before, for every modality that is not PT and not CT, `MO` was shown as unit for the annotations. I find this confusing, as `MO` doesn't have a physical meaning. I only recently learned that it stands for "modality unit" (as opposed to stored pixel value). See https://groups.google.com/g/cornerstone-platform/c/oCuC7_TIizY/m/fCyGBIEaBAAJ

### DIFF
--- a/packages/tools/src/utilities/getModalityUnit.ts
+++ b/packages/tools/src/utilities/getModalityUnit.ts
@@ -1,0 +1,11 @@
+function getModalityUnit(modality: string): string {
+  if (modality === 'CT') {
+    return 'HU';
+  } else if (modality === 'PT') {
+    return 'SUV';
+  } else {
+    return '';
+  }
+}
+
+export { getModalityUnit };
--- a/packages/tools/src/utilities/getModalityUnit.ts
+++ b/packages/tools/src/utilities/getModalityUnit.ts
@@ -1,7 +1,7 @@
-function getModalityUnit(modality: string): string {
+function getModalityUnit(modality: string, isPreScaled: boolean): string {
   if (modality === 'CT') {
     return 'HU';
-  } else if (modality === 'PT') {
+  } else if (modality === 'PT' && isPreScaled === true) {
     return 'SUV';
   } else {
     return '';
--- a/common/reviews/api/tools.api.md
+++ b/common/reviews/api/tools.api.md
@@ -1361,7 +1361,7 @@ export class EllipticalROITool extends AnnotationTool {
         hasMoved?: boolean;
     } | null;
     // (undocumented)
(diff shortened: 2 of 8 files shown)


salimkanoun:
sounds great but then we will need to differentiate different flavor of "suX" => Suv / sul jenma / sul james 
I see here PT == SUV which is not always true

pwespi:
> sounds great but then we will need to differentiate different flavor of "suX" => Suv / sul jenma / sul james I see here PT == SUV which is not always true

Yes, I agree. PT == SUV is also problematic, see #163.

packages/tools/src/utilities/getModalityUnit.ts
```diff
@@ -0,0 +1,11 @@
+function getModalityUnit(modality: string): string {
+  if (modality === 'CT') {
+    return 'HU';
+  } else if (modality === 'PT') {
```
sedghi, line 4: We can do both checks here for 'PT'. The check that sees if the data is preScaled. I guess this function can also get the `imaegId` or `volumeId` (can be extracted from `targetId`), and checks if the volume/image is prescaled. I have similar operation here (https://github.com/cornerstonejs/cornerstone3D-beta/blob/7c311f77f0532372ae82b6be2027bcd25925fa0d/packages/tools/src/tools/WindowLevelTool.ts#L71-L77) Let me know what you think

sedghi: So right now the StackViewport gives back the `preScale` flag, but we don't have that on the VolumeViewport since it can be either volumes (in case of fusion for example).

sedghi (approved):
Looks good to me, It can be extended to address the issue with https://github.com/cornerstonejs/cornerstone3D-beta/issues/163

packages/tools/src/utilities/getModalityUnit.ts
pwespi: Thanks for the link to your code. I added it to a helper function to check for pre-scaling and pass the flag to the modality unit function.

pwespi:
closes #163

sedghi (approved):
This is great contribution, thank you

PR #167, opened 2022-08-05
feat: camera sync canvas relative

wayfarer3130:
This feature adds a new  zoom/pan sync option, with either setting being able to be turned off.

### DIFF
--- a/packages/core/examples/programaticPanZoom/index.ts
+++ b/packages/core/examples/programaticPanZoom/index.ts
@@ -1,0 +1,160 @@
+import {
+  getRenderingEngine,
+  RenderingEngine,
+  Types,
+  Enums,
+} from '@cornerstonejs/core';
+import {
+  initDemo,
+  createImageIdsAndCacheMetaData,
+  setTitleAndDescription,
+  ctVoiRange,
+  addButtonToToolbar,
+} from '../../../../utils/demo/helpers';
+
+// This is for debugging purposes
+console.warn(
+  'Click on index.ts to open source code for this example --------->'
+);
+
+const { ViewportType } = Enums;
+const renderingEngineId = 'myRenderingEngine';
+const viewportId = 'CT_STACK';
+
+// ======== Set up page ======== //
+setTitleAndDescription(
+  'Programmatic Pan and Zoom with initial pan and zoom',
+  'Displays an image at the top of the viewport, half off the screen, and has pan/zoom buttons.'
+);
+
+const content = document.getElementById('content');
+const element = document.createElement('div');
+element.id = 'cornerstone-element';
+element.style.width = '500px';
+element.style.height = '500px';
+
+content.appendChild(element);
+// ============================= //
+addButtonToToolbar({
+  title: 'Set Pan (+5,0)',
+  onClick: () => {
+    // Get the rendering engine
+    const renderingEngine = getRenderingEngine(renderingEngineId);
+
+    // Get the stack viewport
+    const viewport = <Types.IVolumeViewport>(
+      renderingEngine.getViewport(viewportId)
+    );
+
+    const pan = viewport.getPan();
+    console.log('Current pan', JSON.stringify(pan));
+    viewport.setPan([pan[0] + 5, pan[1]]);
+    viewport.render();
+  },
+});
+
+addButtonToToolbar({
+  title: 'Set zoom * 1.05 ',
+  onClick: () => {
+    // Get the rendering engine
+    const renderingEngine = getRenderingEngine(renderingEngineId);
+
+    // Get the stack viewport
+    const viewport = <Types.IVolumeViewport>(
+      renderingEngine.getViewport(viewportId)
+    );
+
+    const zoom = viewport.getZoom();
+    console.log('Current zoom', zoom);
+    viewport.setZoom(zoom * 1.05);
+    viewport.render();
+  },
+});
+
+addButtonToToolbar({
+  title: 'Reset Original',
+  onClick: () => {
+    // Get the rendering engine
+    const renderingEngine = getRenderingEngine(renderingEngineId);
+
+    // Get the stack viewport
+    const viewport = <Types.IVolumeViewport>(
+      renderingEngine.getViewport(viewportId)
+    );
+    //viewport.resetCamera();
+    viewport.setZoom(0.8);
+    viewport.setPan([-128, 0]);
+    viewport.render();
+  },
+});
+
+addButtonToToolbar({
+  title: 'Reset current as initial',
+  onClick: () => {
+    // Get the rendering engine
+    const renderingEngine = getRenderingEngine(renderingEngineId);
+
+    // Get the stack viewport
+    const viewport = <Types.IVolumeViewport>(
+      renderingEngine.getViewport(viewportId)
+    );
+    viewport.setZoom(viewport.getZoom(), true);
+  },
+});
+
+/**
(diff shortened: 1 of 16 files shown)


packages/core/src/eventTarget.ts
```diff
@@ -16,6 +16,9 @@ class CornerstoneEventTarget implements EventTarget {
     if (!this.listeners[type]) {
       this.listeners[type] = [];
     }
+    if (this.listeners[type].indexOf(callback) !== -1) {
```
wayfarer3130, line 19: This is a standard event add check to prevent multiple callbacks - makes debugging easier, and for large viewport layouts, is faster and appears to reduce the occurance of a browser crash.

packages/core/examples/programaticPanZoom/index.ts
```diff
@@ -0,0 +1,160 @@
...
+);
+
+const { ViewportType } = Enums;
+const renderingEngineId = 'myRenderingEngine';
+const viewportId = 'CT_STACK';
+
+// ======== Set up page ======== //
+setTitleAndDescription(
+  'Programmatic Pan and Zoom with initial pan and zoom',
+  'Displays an image at the top of the viewport, half off the screen, and has pan/zoom buttons.'
```
sedghi, line 27: I tried this demo, it works fine. I'm just confused what the `reset current as initial` do actually? when I click on it it does nothing, but when I `reset original` it moves the camera. Can you add a sentence or two to say what is the expected behavior for the `reset current as initial`? is it `set current view as initial`? Can's you just trigger a render so that it actually set it? or call reset.

packages/core/src/RenderingEngine/Viewport.ts
```diff
@@ -682,6 +686,99 @@ class Viewport implements IViewport {
...
   }
 
+  /**
+   * Resets the stored initial offset values for
+   * zoom and pan.  These values are the offset
+   * values used to allow the getPan and getZoom
+   * to return [0,0] and 1 for the initial values.
+   * @param resetOffsets can be passed to skip resetting, ie a no-op on this call
+   */
+  protected resetInitialOffsets(resetOffsets = true) {
```
sedghi, line 696: The whole notion of `resetInitialOffset` does not make sense to me, this function just get the camera and set it as initialCamera. Shouldn't this function be like `setInitialCamera`? and then `resetOffsets` be like `storeInitialCamera` or something else? I don't understand what is being `reset` here?

packages/core/src/RenderingEngine/Viewport.ts
```diff
@@ -682,6 +686,99 @@ class Viewport implements IViewport {
...
+      vec2.subtract(vec2.create(), initialCanvasFocal, currentCanvasFocal)
+    );
+    return result;
+  }
+
+  /**
+   * Sets the canvas pan value relative to the initial view position of 0,0
+   * Modifies the camera to perform the pan.
+   */
+  public setPan(pan: Point2, resetOffsets = false) {
```
sedghi, line 732: again `resetOffsets` is confusing to me, I don't understand what it does. + function documentation miss arguments

packages/core/src/RenderingEngine/Viewport.ts
```diff
@@ -776,6 +873,7 @@ class Viewport implements IViewport {
       viewAngle,
       flipHorizontal,
       flipVertical,
+      resetOffsets,
```
sedghi, line 876: `resetOffsets` should not be on camera interface. It can be the second argument of `setCamera` though (`setCamera(cameraInterface, resetOffset)`) but again the `resetOffset` is vauge and not what it actually does (which is `set` initial camera)

packages/core/src/types/ICamera.ts
```diff
@@ -28,6 +28,11 @@ interface ICamera {
   flipHorizontal?: boolean;
   /** flip Vertical */
   flipVertical?: boolean;
+  /** resetOffsets set to true to reset the pan and zoom offsets
+   * such that the pan value becomes [0,0] and the zoom value 1.  Used to
+   * set the initial state.
+   */
+  resetOffsets?: boolean;
```
sedghi, line 35: This should not be here

packages/tools/src/synchronizers/callbacks/zoomPanSyncCallback.ts
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * targetViewport.options.syncZoom set to false to not sync the zoom
+ * targetViewport.options.syncPan set to false to not sync the pan
+
+ * @param synchronizerInstance - The Instance of the Synchronizer
+ * @param sourceViewport - The list of IDs defining the source viewport.
+ * @param targetViewport - The list of IDs defining the target viewport, different
+ *   from sourceViewport
+ */
+export default function cameraSyncCallback(
```
sedghi, line 15: Function name is wrong

packages/tools/src/synchronizers/synchronizers/createCameraPositionSynchronizer.ts
```diff
@@ -10,7 +10,7 @@ const { CAMERA_MODIFIED } = Enums.Events;
  * rendering event and calls the `cameraSyncCallback`.
  *
  * @param synchronizerName - The name of the synchronizer.
- *
+ * @param options - allows passing option values to the synchronizer on "this"
```
sedghi, line 13: No second argument is passed to the function, what is options doing here?

packages/tools/src/synchronizers/synchronizers/createZoomPanSynchronizer.ts
```diff
@@ -0,0 +1,26 @@
...
+import Synchronizer from '../../store/SynchronizerManager/Synchronizer';
+
+const { CAMERA_MODIFIED } = Enums.Events;
+
+/**
+ * A helper that creates a new `Synchronizer` which listens to the `CAMERA_MODIFIED`
+ * rendering event and calls the `cameraSyncCallback`.
+ *
+ * @param synchronizerName - The name of the synchronizer.
+ * @param options - allows passing option values to the synchronizer on "this"
```
sedghi, line 13: same here, what is options?

sedghi (changes requested):
See my comments, main one is that `resetOffsets` is vauge and should be renamed to represent what it does which is setting initial camera. initial camera is also something that should be documented more of what it is. If you fix these minor comments we can merge. Thanks

packages/core/examples/programaticPanZoom/index.ts
wayfarer3130: Added comments - basically it allows resetting the base/original conditions, so that a subsequent reset changes to display
the current setup.

packages/core/src/RenderingEngine/Viewport.ts
wayfarer3130: The idea is that it stores the initial camera, so I'm going to call it just that.  Will push an update momentarily.

packages/core/src/RenderingEngine/Viewport.ts
wayfarer3130: Renamed to storeInitialCamera

packages/core/src/RenderingEngine/Viewport.ts
wayfarer3130: OK, will make it the second argument

swederik (approved):

sedghi (approved):

PR #171, opened 2022-08-09
fix: default voi for volumes and webLoader

sedghi:


sedghi:
https://github.com/cornerstonejs/cornerstone3D-beta/issues/118

packages/core/examples/webLoader/hardcodedMetaDataProvider.ts
```diff
@@ -0,0 +1,54 @@
...
+      modality: 'SC',
+    };
+
+    return generalSeriesModule;
+  } else if (type === 'imagePlaneModule') {
+    const index = imageIds.indexOf(imageId);
+    // console.warn(index);
+    const imagePlaneModule = {
+      imageOrientationPatient: [1, 0, 0, 0, 1, 0],
+      imagePositionPatient: [0, 0, index * 5],
```
Georift, line 29: Does this positioning have any effect when we're loading it as a stack?

I just tried running the example without passing `imagePositionPatient` and it seems to work ok?

If it is required, what would be the correct way to have multiple stacks of images with this loader?

sedghi: no it does not, I was about to write a volume example as well, that is why I put it there. You can just ignore it

Georift: Ok thanks for the info :+1:

packages/core/examples/webLoader/hardcodedMetaDataProvider.ts
```diff
@@ -0,0 +1,54 @@
...
+      bitsAllocated: 24,
+      bitsStored: 24,
+      highBit: 24,
+      photometricInterpretation: 'RGB',
+      samplesPerPixel: 3,
+    };
+
+    return imagePixelModule;
+  } else if (type === 'generalSeriesModule') {
+    const generalSeriesModule = {
```
wayfarer3130, line 19: General Series Module has Series Instance UID, Series Number and Series Description, all of which are used in various places - even if you just add dummy values for them it would be good enough.  Series Data/Time might also be needed for sorting in display sets.

packages/core/examples/webLoader/hardcodedMetaDataProvider.ts
```diff
@@ -0,0 +1,54 @@
...
+      rows: 1216,
+      rowCosines: [1, 0, 0],
+      columnCosines: [0, 1, 0],
+    };
+
+    return imagePlaneModule;
+  } else if (type === 'voiLutModule') {
+    return {
+      windowWidth: [255],
+      windowCenter: [127],
```
wayfarer3130, line 44: 128

packages/core/examples/webLoader/index.ts
```diff
@@ -0,0 +1,110 @@
...
+element.id = 'cornerstone-element';
+element.style.width = '500px';
+element.style.height = '500px';
+
+content.appendChild(element);
+
+const renderingEngineId = 'myRenderingEngine';
+const viewportId = 'COLOR_STACK';
+
+// png images hosted on web (s3 with cors enabled)
```
wayfarer3130, line 41: Please indicate that these have no PHI in them and have permission to distribute.

wayfarer3130 (approved):
There are a couple of small changes to be included.

PR #176, opened 2022-08-11
fix: invalid keybindings Alt and Ctrl

burnpiro:
ref. https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode#non-printable_keys_function_keys

### DIFF
--- a/common/reviews/api/tools.api.md
+++ b/common/reviews/api/tools.api.md
@@ -2506,9 +2506,9 @@ type JumpToSliceOptions = {
 // @public (undocumented)
 enum KeyboardBindings {
     // (undocumented)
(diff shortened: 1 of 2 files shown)


burnpiro:
You might also want to consider changing `ToolBindingKeyboardType` type to extend `modifierKey` with other values. Right now if someone wants to use a different modifier than the 3 defined in `KeyboardBindings` it's not possible because of TS throwing an error (it has to be one of the enum values). You could use the broad TS definition 
https://github.com/nfriend/ts-keycode-enum/blob/master/Key.enum.ts

sedghi (approved):

PR #179, opened 2022-08-17
fix: add extra missing exports and no static code block at build

sedghi:


swederik:
I hate it but if it works I guess it's fine :-)

swederik (approved):

PR #182, opened 2022-08-19
Update type definition for vtkObject

jmannau:
Update import of vtkObject to match the new vtk.js types


sedghi (approved):

sedghi:
Our tests are not running, I will run it locally and push

PR #191, opened 2022-08-26
feat: draw center point of the ellipticalROI tool and make it configurable

md-prog:


sedghi:
You need to run the `yarn build:update-api` and include the generated markdown files. 
Also can you post here a screenshot of what it looks like?

md-prog:
> You need to run the `yarn build:update-api` and include the generated markdown files. Also can you post here a screenshot of what it looks like?

Thank you, it did the trick.

md-prog:
@sedghi  Please review

sedghi:
Can you put a screenshot of the annotation while the config is set to true?

sedghi:
Can you put a screenshot of the annotation while the config is set to true?

md-prog:
Sure. Here it is.
image

packages/tools/src/tools/annotation/EllipticalROITool.ts
```diff
@@ -125,6 +126,9 @@ class EllipticalROITool extends AnnotationTool {
       configuration: {
         shadow: true,
         preventHandleOutsideImage: false,
+        centerPointRadius: 0, // Radius of the circle to draw  at the center
```
sedghi, line 129: can you please update the class comment above to document this so that it appears in the docs here https://www.cornerstonejs.org/api/tools/class/ellipticalroitool/

packages/tools/src/tools/annotation/EllipticalROITool.ts
```diff
@@ -868,6 +874,27 @@ class EllipticalROITool extends AnnotationTool {
         }
       );
 
+      // draw center point, if "centerPointRadius" configuration is valid.
+      const minRadius = Math.min(
+        Math.abs(canvasCorners[0][0] - canvasCorners[1][0]) / 2, // horizontal radius
+        Math.abs(canvasCorners[0][1] - canvasCorners[1][1]) / 2 // vertical radius
+      );
```
sedghi, line 881: Can we put this calculations in the if statement below so that we don't add extra stuff when the radisu is 0?

md-prog:
> Please see comments

Fixed as per your review

sedghi (approved):

PR #199, opened 2022-09-02
fix: wrong ushape calculation when loading SR/freehand from server

### LINKED ISSUE #198: Wrong ushape calculation when loading SR/freehand from server
Steps to reproduce
1. Go to viewer that has Freehand integrated
2. Draw an opened contour and set its flag joinedOpenContour as true (it means the UShape must be drawn )
3. Save annotation
4. Reload the page
4.1 Ensure the request for loading SRs is faster than displaying the image

Expected
- When image is displayed contour should be displayed with appropriate UShape draw

Results
- Image is display, freehand is displayed on correct position (old one) but UShape is draw in totally wrong position

ladeirarodolfo:
- It includes
It just prevent doing calculation using camera when there is no camera/data full set yet

closes #198


sedghi (approved):
Makes sense, thanks

PR #202, opened 2022-09-06
[WIP]: Example/rtss

dxlin:
WIP - segmentToRTSS example has function for converting segmentation volume to ROIContour to be sent to dcmjs for export. Conversion to contour is performed via vtkImageMarchingSquares routine on each of the slices.

1. Currently missing Referenced Image UIDs, should this be populated instead in dcmjs?
2. Would the conversion function be best placed in a tool or as a general util?

### DIFF
--- a/packages/core/CHANGELOG.md
+++ b/packages/core/CHANGELOG.md
@@ -3,6 +3,17 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+## [0.15.3](https://github.com/cornerstonejs/cornerstone3D-beta/compare/@cornerstonejs/core@0.15.2...@cornerstonejs/core@0.15.3) (2022-09-02)
+
+
+### Bug Fixes
+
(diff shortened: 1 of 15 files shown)


sedghi:
Hey, glad to see that you have some progress on this. One small note, you need to add the name of the example folder [here](https://github.com/cornerstonejs/cornerstone3D-beta/blob/ed23f05b23063769942328f9e6797d792767ec49/utils/ExampleRunner/example-info.json) for it to show up in the demo page.

Looking forward to try it

dxlin:
Thanks for notice! I've updated with the changes.

The example at the moment only outputs the contour object in the console. The intention is for this object to be passed to dcmjs for export.

sedghi:
@dxlin Cornerstone3D does not have dependency on dcmjs (and we don't want to add that), hmmm, I'm wondering if then PR to Cornerstone is a bad idea. Let's try to implement the non-dcmjs part of the functionalities in the tools, and then the demo itself can have dependency on the dcmjs (right now some of them does).

sedghi:
Done in another PR https://github.com/cornerstonejs/cornerstone3D/pull/734

Issue #160, opened 2022-08-03
VolumeViewport Orientation when ImageOrientation metadata is unusual


LittleURF:
**TLDR:**
How to set proper orientation for a VolumeViewport when ImageOrientationPatient metadata doesn't match the provided constant orientations?
\---

Hi,
I'm having issues with finding out the proper Orientation values to set, when I'm rendering a volume with ImageOrientation metadata values different than the usual Axial/Sagital/Coronal values.

I believe that the ImageOrientation values map to orientations like this: 
```
    coronalImageOrientationPattern = [1, 0, 0, 0, 0, -1];
    // sliceNormal: <Point3>[0, 1, 0],
    // viewUp: <Point3>[0, 0, 1],

    sagittalImageOrientationPattern = [0, 1, 0, 0, 0, -1];
    // sliceNormal: <Point3>[1, 0, 0],
    // viewUp: <Point3>[0, 0, 1],

    axialImageOrientationPattern = [1, 0, 0, 0, 1, 0];
    // sliceNormal: <Point3>[0, 0, -1],
    // viewUp: <Point3>[0, -1, 0],
```
But when the ImageOrientation value is different(like below), the standard orientations don't result in the expected Axial/Coronal/Sagital views. 
`[0.85005025920695,-0.5267015823236,2.1648092e-8,0.00883004239504,0.01425087456269,-0.9998594615872]`.

I think cornerstone doesn't have any instructions about what to do in such case.  

So I attempted to create a custom orientation based on the ImageOrientation, I found out that the `sliceNormal` part is created by crossing the ImageOrientation like this: `sliceNormal = cross(imageOrientation.slice(0, 3), imageOrientation.slice(3, 6)) `, though I'm not sure if it's correct. And for the ViewUp part I have no idea how I should set it properly...
When using the calculated SliceNormal and some viewUp from the constant orientations, the  view seems like it's correct, except for the camera. It's often positioned wrongly. For example, for these views to be correct, we'd need to: flip the camera horizontally / move the camera to the back of the image instead of the front etc. Cornerstone doesn't give us functionalites like that tho, and even if it did, it'd be hard to dynamically decide what camera operations need to be done to make the specific view correct.

So my question is, how to set the orientation for volumes with custom ImageOrientationPatient metadata?

Volume rendered with Orientation Coronal
[...]

PR #203, opened 2022-09-08
feat: orientation on volumeViewport can be optional

sedghi:
- If not defined, the acquisition axis (scan axis normal) will be used as camera direction
- A new demo that shows the concept using an non-aligned data 
- rename sliceNormal to viewPlaneNormal to be consistent throughout the repo

#160


common/reviews/api/core.api.md
```diff
@@ -1425,6 +1427,19 @@ function loadImageToCanvas(canvas: HTMLCanvasElement, imageId: string, requestTy
 // @public (undocumented)
 function loadVolume(volumeId: string, options?: VolumeLoaderOptions): Promise<Types.IImageVolume>;
 
+// @public (undocumented)
+type Mat3 = [
```
wayfarer3130, line 1431: Can't you write this as number[9]?  Also, the Vec library allows defining both the ArrayBuffer format as well as this format using the | notation - that might be worthwhile, as the underlying implementation is faster.

PR #207, opened 2022-09-09
[DRAFT]: feat: cache now stores image and volume based on URI internally

sedghi:
Previously cache would store images and volumes internally with keys of imageId and volumeId. The image/volumeId has the loader scheme which is totally unnecessary since the whole purpose of cache is to be able to retrieve the image/volume pixelData for an image that is already cached, no matter what loader was used to retrieved it

This PR changes the internal wiring of cache to work with URI. 

In future, it might makes sense to change the surface API to also work with imagerURI, but for now, the API surface has not changed


iconsys:
Can this solve memory exceeded issue? This is frequently coming when we do MPR even on series with less images (300-400) on a PC with 16GB RAM. Once below message comes then all other open browser tabs (also in other application) show same message.

image3

wayfarer3130:
@iconsys - take a look at PR:
https://github.com/OHIF/Viewers/pull/3068
That resolves multiple issues with memory, in particular it prevent a reload of volume data for the previous study if you look at a new study.  If you are seeing the issue on first load, however, then it won't fix things, but it does make the client quite a bit more stable.

packages/core/src/cache/cache.ts
```diff
@@ -286,6 +290,9 @@ class Cache implements ICache {
     // number of bytes
   }
 
+  /**
+   * Puts a n
+   *
```
wayfarer3130, line 295: This bit seems extraneous

wayfarer3130 (approved):
Looks good to me - should make the stack view after viewing a volume faster.

PR #220, opened 2022-09-16
fix(rgba): Handle rgba to rgb conversion based on length

wayfarer3130:
This is a work around for the rgba flag not being set correctly and when we have rgba data but need rgb.


packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -1341,7 +1341,10 @@ class StackViewport extends Viewport implements IStackViewport {
     const scalars = this._imageData.getPointData().getScalars();
     const scalarData = scalars.getData() as Uint8Array | Float32Array;
 
-    if (image.rgba) {
+    if (image.rgba || pixelData.length * 3 == scalarData.byteLength * 4) {
```
wayfarer3130, line 1344: This is a work around for now

sedghi (approved):

PR #224, opened 2022-09-22
feat(voi): added support for sigmoid voiLUTFunction for StackViewport and VolumeViewport

Ouwen:
- CPU fallback is unimplemented

### DIFF
--- a/common/reviews/api/streaming-image-volume-loader.api.md
+++ b/common/reviews/api/streaming-image-volume-loader.api.md
@@ -707,6 +707,7 @@ interface IImage {
         lastRenderTime?: number;
     };
     voiLUT?: CPUFallbackLUT;
+    voiLUTFunction: string;
     width: number;
     windowCenter: number[] | number;
     windowWidth: number[] | number;
@@ -1115,8 +1116,7 @@ interface IVolumeViewport extends IViewport {
     getFrameOfReferenceUID: () => string;
     getImageData(volumeId?: string): IImageData | undefined;
     getIntensityFromWorld(point: Point3): number;
-    // (undocumented)
-    getProperties: () => any;
+    getProperties: () => VolumeViewportProperties;
     getSlabThickness(): number;
     hasImageURI: (imageURI: string) => boolean;
     hasVolumeId: (volumeId: string) => boolean;
@@ -1181,6 +1181,7 @@ type Metadata = {
     Columns: number;
     Rows: number;
     voiLut: Array<VOI>;
+    VOILUTFunction: string;
 };
 
 // @public (undocumented)
@@ -1299,6 +1300,7 @@ type StackViewportNewStackEventDetail = {
 // @public
 type StackViewportProperties = {
     voiRange?: VOIRange;
+    VOILUTFunction?: VOILUTFunctionType;
     invert?: boolean;
     interpolationType?: InterpolationType;
     rotation?: number;
@@ -1412,6 +1414,16 @@ type VOI = {
 };
 
 // @public
+enum VOILUTFunctionType {
+    // (undocumented)
+    EXACT_LINEAR = 'EXACT_LINEAR',
+    // (undocumented)
+    LINEAR = 'LINEAR',
+    // (undocumented)
+    SIGMOID = 'SIGMOID',
+}
+
+// @public
 type VoiModifiedEvent = CustomEvent_2<VoiModifiedEventDetail>;
 
 // @public
@@ -1419,6 +1431,7 @@ type VoiModifiedEventDetail = {
     viewportId: string;
     range: VOIRange;
     volumeId?: string;
+    VOILUTFunction?: VOILUTFunctionType;
 };
 
 // @public (undocumented)
@@ -1495,6 +1508,7 @@ type VolumeNewImageEventDetail = {
 // @public
 type VolumeViewportProperties = {
     voiRange?: VOIRange;
+    VOILUTFunction?: VOILUTFunctionType;
 };
 
 // (No @packageDocumentation comment for this package)
--- a/common/reviews/api/streaming-image-volume-loader.api.md
+++ b/common/reviews/api/streaming-image-volume-loader.api.md
@@ -1416,11 +1416,10 @@ type VOI = {
 // @public
 enum VOILUTFunctionType {
     // (undocumented)
-    EXACT_LINEAR = 'EXACT_LINEAR',
-    // (undocumented)
     LINEAR = 'LINEAR',
     // (undocumented)
-    SIGMOID = 'SIGMOID',
+    SAMPLED_SIGMOID = 'SIGMOID', // SIGMOID is sampled in 1024 even steps so we call it SAMPLED_SIGMOID
+    // EXACT_LINEAR = 'EXACT_LINEAR', TODO: Add EXACT_LINEAR option from DICOM NEMA
 }
 
 // @public
--- a/packages/core/src/RenderingEngine/BaseVolumeViewport.ts
+++ b/packages/core/src/RenderingEngine/BaseVolumeViewport.ts
@@ -18,12 +18,24 @@ import type {
   ActorEntry,
   FlipDirection,
   VolumeViewportProperties,
+  VOIRange,
 } from '../types';
 import type { ViewportInput } from '../types/IViewport';
 import type IVolumeViewport from '../types/IVolumeViewport';
-import { Events, BlendModes, OrientationAxis } from '../enums';
+import {
+  Events,
+  BlendModes,
+  OrientationAxis,
+  VOILUTFunctionType,
+} from '../enums';
 import eventTarget from '../eventTarget';
-import { actorIsA, imageIdToURI, triggerEvent } from '../utilities';
+import {
+  actorIsA,
+  imageIdToURI,
+  triggerEvent,
+  createSigmoidRGBTransferFunction,
+  createLinearRGBTransferFunction,
+} from '../utilities';
 import type { vtkSlabCamera as vtkSlabCameraType } from './vtkClasses/vtkSlabCamera';
 import { VoiModifiedEventDetail } from '../types/EventTypes';
 import { RENDERING_DEFAULTS } from '../constants';
@@ -40,6 +52,10 @@ import { RENDERING_DEFAULTS } from '../constants';
 abstract class BaseVolumeViewport extends Viewport implements IVolumeViewport {
   useCPURendering = false;
   private _FrameOfReferenceUID: string;
+
+  // Viewport Properties
+  private voiRange: VOIRange;
+  private VOILUTFunction: VOILUTFunctionType;
 
   constructor(props: ViewportInput) {
     super(props);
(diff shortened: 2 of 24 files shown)


Ouwen:
@sedghi

packages/core/src/utilities/createSigmoidRGBTransferFunction.ts
```diff
@@ -0,0 +1,48 @@
+import vtkColorTransferFunction from '@kitware/vtk.js/Rendering/Core/ColorTransferFunction';
+import { VOIRange } from '../types';
+
+/**
+ * A utility that can be used to generate an RgbTransferFunction
```
sedghi, line 5: I guess the documentation is wrong here, it should say create Sigmoid RGB? also maybe can you add a little bit of information about why one might prefer this sigmoid one compared to linear?

packages/core/src/utilities/createSigmoidRGBTransferFunction.ts
```diff
@@ -0,0 +1,48 @@
...
+ * @param rgbTransferFunction
+ */
+export default function createSigmoidRGBTransferFunction(
+  windowWidth: number,
+  windowCenter: number,
+  voiRange: VOIRange
+): vtkColorTransferFunction {
+  const cfun = vtkColorTransferFunction.newInstance();
+  const sigmoid = (x: number, wc: number, ww: number) =>
+    1 / (1 + Math.exp((-4 * (x - wc)) / ww));
```
sedghi, line 27: This is not a regular sigmoid equation as I see, where is this -4 coming from? can we add some inline docs?

packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -407,6 +411,7 @@ class StackViewport extends Viewport implements IStackViewport {
       if (Array.isArray(windowCenter)) {
         windowCenter = windowCenter[0];
       }
+      voiLUTFunction = voiLUTFunction ? voiLUTFunction : 'LINEAR';
```
sedghi, line 414: can you use the Enum you created here maybe (instead of 'LINEAR')?

packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -77,6 +79,7 @@ interface ImagePixelModule {
   pixelRepresentation: string;
   windowWidth: number;
   windowCenter: number;
+  voiLUTFunction: string;
```
sedghi, line 82: Can we use the Enum you created instead of `string` here as the type as well? and you handle the metadata reading and assignging the correct Enum?

packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -1755,27 +1778,30 @@ class StackViewport extends Viewport implements IStackViewport {
     }
 
     this.initialVOIRange = voiRange;
+    this.voiFunction = VOILUTFunctionType[voiLUTFunction];
```
sedghi, line 1781: If `imagePixelModule` includes the `voiLUTFunction` as the Enum, we don't need to do this here and can just simply assign

sedghi (commented):
Great contribution as always, please see my comments

packages/core/src/utilities/createSigmoidRGBTransferFunction.ts
Ouwen: https://dicom.nema.org/medical/dicom/2018b/output/chtml/part03/sect_C.11.2.html

Added a reference to dicom nema which uses the -4

Ouwen:
@sedghi added VOI type as an enum and added documentation

sedghi:
All looks good, two last things

1. This only adds it for StackViewport right? maybe we rename the PR to include such information?
2. Can you please post here a before/after image for LINEAR vs SIGMOID voi for reference (since we are not adding new docs)

Thanks

Ouwen:
Sure thing, I think I can also try implementing this in the volume viewer

Ouwen:
Linear
Screen Shot 2022-10-07 at 9 07 54 PM

Sigmoid
Screen Shot 2022-10-07 at 9 05 50 PM

Ouwen:
@sedghi 
I've added the changes you suggested. Also scaffolding for the VolumeViewport. One thing that came to mind while looking at the volume viewport is if the voiLUTfunction should exist at the viewportInput level vs setVOI in the stackviewport.

Thoughts?

packages/streaming-image-volume-loader/src/helpers/makeVolumeMetadata.ts
```diff
@@ -28,8 +29,10 @@ export default function makeVolumeMetadata(
   const voiLutModule = metaData.get('voiLutModule', imageId0);
 
   // voiLutModule is not always present
+  let voiLUTFunction;
   if (voiLutModule) {
     const { windowWidth, windowCenter } = voiLutModule;
+    voiLUTFunction = voiLutModule?.voiLutModule;
```
Ouwen, line 35: I think this should be `voiLutModule?.voiLUTFunction `

common/reviews/api/streaming-image-volume-loader.api.md
```diff
@@ -1251,6 +1254,16 @@ type VOI = {
     windowCenter: number;
 };
 
+// @public
+enum VOILUTFunctionType {
+    // (undocumented)
+    EXACT_LINEAR = 'exactLinear',
```
Ouwen, line 1260: should we follow the same dicom string rather than camel case?

Ouwen:
I noticed that we have the following:

`voiLUTFunction`: seems to refer to cornerstoneWADO image object/metadata properties
`voiLutFunction`: property of viewports
`VOILUTFunction`: dcmjs naming convention

Is this intentional?

packages/streaming-image-volume-loader/src/helpers/makeVolumeMetadata.ts
sedghi: yes i think that was my mistake

sedghi:
I hate this a lot and didn't notice it, good catch

```
voiLUTFunction: seems to refer to cornerstoneWADO image object/metadata properties
voiLutFunction: property of viewports
VOILUTFunction: dcmjs naming convention
```

Can we remove `voiLutFunction` and use either of those? I think we can live with the dcmjs one only

Were you able to run the example i added? It seems to be not working as I expected

```
cd packages/core
yarn run example voiSigmoid
```
```

**Comment by Ouwen** [comment]
@sedghi this PR requires making this change to vtkjs
(https://github.com/gradienthealth/vtk-js/commit/623f5ec9841b207d5d91b754f259872e81b4d249)

VOILUTFunction should be supported in volumes and stack viewport now.

There is some setProperties logic that I believe is still TODO from previous commits, but I think it might be beyond scope of this PR.

**Comment by Ouwen** [comment]
pending https://github.com/Kitware/vtk-js/pull/2603

**Comment by Ouwen** [comment]
@sedghi seems like the vtk.js version would need to be bumped to latest [v25.10.1](https://github.com/Kitware/vtk-js/releases/tag/v25.10.1).

Would this have consequences for other related libraries (OHIF/Viewer)?

**Comment by sedghi** [comment]
That will break a lot of tests, I have fixed them in my other branch, can we wait for my branch (not create PR yet) to be merged before this?

**Comment by Ouwen** [comment]
Sure, no rush

**Comment by Ouwen** [comment]
https://github.com/Kitware/vtk-js/pull/2610, heads up the full fix requires vtk.js 25.11.1

**Comment by Ouwen** [comment]
@sedghi hmmm unsure why these doc issues are popping up. As far as I know, I ran `yarn run build:update-api`

**Comment by Ouwen** [comment]
@sedghi ready for review

**Comment by Ouwen** [comment]
@sedghi just another ping since there are a lot of branches flying around...

File: packages/core/src/RenderingEngine/BaseVolumeViewport.ts [hunk 8]
```diff
@@ -312,6 +409,17 @@ abstract class BaseVolumeViewport extends Viewport implements IVolumeViewport {
...
+      if (i === 0 && !this.voiRange) {
+        const voiRange = actor
+          .getProperty()
+          .getRGBTransferFunction(0)
+          // @ts-ignore: vtk d ts problem
+          .getRange();
+        this.voiRange = {
+          lower: voiRange[0],
+          upper: voiRange[1],
+        };
```
sedghi, line 421: we basically do this inside `createVolumeActor`, how is it different?

packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -78,6 +80,7 @@ interface ImagePixelModule {
   pixelRepresentation: string;
   windowWidth: number;
   windowCenter: number;
+  voiLUTFunction: VOILUTFunctionType;
```
sedghi, line 83: can you please change this to be at least the same as the volume viewport? VOILUTFunction

sedghi: or this is different? i'm confused again :()

packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -896,6 +913,25 @@ class StackViewport extends Viewport implements IStackViewport {
...
+  private setVOILUTFunction(
+    voiLUTFunction: VOILUTFunctionType,
+    suppressEvents?: boolean
+  ): void {
+    if (this.useCPURendering) {
+      throw new Error('VOI LUT function is not supported in CPU rendering');
+    }
+
+    // make sure the VOI LUT function is valid in the VOILUTFunctionType which is enum
+    if (Object.values(VOILUTFunctionType).indexOf(voiLUTFunction) === -1) {
```
sedghi, line 925: This logic is used 3 times `Object.values(VOILUTFunctionType).indexOf(voiLUTFunction)` maybe refactor?

packages/core/src/RenderingEngine/StackViewport.ts
sedghi: I guess this is coming from cswil?

sedghi (commented):
Minor comment about the `if (i === 0 && !this.voiRange) {`

packages/core/src/RenderingEngine/BaseVolumeViewport.ts
Ouwen: I don't think the voiRange is saved to the volume anywhere, here we set the voiRange on the `BaseVolumeViewport` instance. I guess this could be moved into create volume actor, but it would require passing `this` into it.

packages/core/src/RenderingEngine/StackViewport.ts
Ouwen: This should be the same, I try to keep the same convention as the surrounding code. The `voiLUTFunction` is from cswil. Probably makes sense to change the other vars 

```
  pixelRepresentation => PixelRepresentation ;
  windowWidth: number; => I think this is derived dicom tag from voi
  windowCenter: number; =>  I think this is derived dicom tag from voi
```

packages/core/src/RenderingEngine/StackViewport.ts
Ouwen: This check can probably be removed actually since the `VOILUTFunctionType` is forced in the function signature.

Ouwen: The other places are due to CSWIL passing in a string which needs to be properly checked. Once we convert to typed CSWIL these logic can be removed.

packages/core/src/RenderingEngine/BaseVolumeViewport.ts
Ouwen: I guess one way to do this is to have BaseVolumeViewport implement a getter for `voiRange`. This getter takes in an index for the volume actor of interest with default being 0. A VOIRange object is returned.

Thoughts on this?

sedghi: I would say let's take out the `triggerVOIModified` from inside the createVolumeActor, and put it inside the base

packages/core/src/RenderingEngine/StackViewport.ts
sedghi: Ok good

Ouwen:
@sedghi this is ready for review. Instead of moving the triggerEvent function outside of the `createVolumeActor` I decide to remove the need for `this.voiRange` instead this is something that can be derived from the volume actor's look up table. The logic for this is in the `getProperties` section.

sedghi (approved):
Thanks for your contribution as always :D

PR #228, opened 2022-09-28
fix: Add storeAsInitialCamera parameter to StackViewport setCamera

volrath:
While reading the API documentation on [`StackViewport.setZoom`](https://www.cornerstonejs.org/api/core/class/StackViewport#setZoom) I wanted to try out the `storeAsInitialCamera` parameter, but I noticed it wasn't working. Upon further inspection I realized it was due to the fact that, even though `Viewport.setCamera` has said parameter, `StackViewport` doesn't, so the the value is never passed down to the super class.

Please note that I purposely omitted the parameter from the call to `this.setCameraCPU` given that I believe it is not used anywhere. Not sure if this is the correct take though, so let me know if this requires further work.

### DIFF
--- a/common/reviews/api/core.api.md
+++ b/common/reviews/api/core.api.md
@@ -1764,7 +1764,7 @@ export class StackViewport extends Viewport implements IStackViewport {
     // (undocumented)
     setActors(actors: Array<ActorEntry>): void;
     // (undocumented)
-    setCamera(cameraInterface: ICamera): void;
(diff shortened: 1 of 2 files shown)


wayfarer3130:
This is a simple fix, looks correct to me.  @sedghi wasnt sure if I should approve this one, or if you wanted a peak first.

sedghi (approved):

sedghi:
Thanks @volrath , this is a great PR

PR #244, opened 2022-10-10
fix(Length):Throws NPE because length is set to null (race condition)

wayfarer3130:
On loading a DICOM SR, the length can be null, which causes a render issue.
There is still another issue on loading, which seems to render the image partially, and then doesn't complete, but this fix at least allows loading.


wayfarer3130:
@sedghi - this is a NPE fix on loading, doesn't resolve the underlying issue, but at least allows the load to succeed.

sedghi (approved):

PR #247, opened 2022-10-17
feat(Touch): added touch events to tools

Ouwen:
Adds the following cornerstone events:

```
TOUCH_START,
TOUCH_START_ACTIVATE,
MULTI_TOUCH_START,
MULTI_TOUCH_START_ACTIVATE,
TOUCH_DRAG,
MULTI_TOUCH_DRAG,
TOUCH_END
```

Currently only uses native browser events (touchstart, touchmove, touchend)


Ouwen:
@sedghi let me know what you think

packages/tools/src/store/removeEnabledElement.ts
```diff
@@ -46,6 +47,7 @@ function removeEnabledElement(
   // Listeners
   mouseEventListeners.disable(element);
   wheelEventListener.disable(element);
+  touchEventListeners.disable(element);
```
sedghi, line 50: why aren't we adding the disable for the touchEventDispatcher below?

packages/tools/src/eventListeners/touch/index.ts
```diff
@@ -0,0 +1,40 @@
...
+ * @private
+ * @param element - The DOM element to remove event listeners from.
+ */
+function disable(element: HTMLDivElement): void {
+  preventGhostClick.disable(element);
+  element.removeEventListener('touchstart', touchStartListener);
+}
+
+/**
+ * Registers mouse event listeners for native mouse event. Enables
```
sedghi, line 21: doc need update, mouse -> touch

packages/tools/examples/stackManipulationToolsTouch/index.ts
```diff
@@ -0,0 +1,164 @@
+import { RenderingEngine, Types, Enums } from '@cornerstonejs/core';
```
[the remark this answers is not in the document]
Ouwen: Mainly just small html changes to make it more mobile friendly

packages/tools/src/store/removeEnabledElement.ts
Ouwen: good catch will add

Ouwen:
I think either hammerjs or contactjs will be needed to trigger the pinch/pan functionality

So I had some questions regarding hammerjs usage in the cornerstoneTools v2.

I noticed that hammerjs has a press and tap functionality, but cornerstone touch eventlistener seem to internally implement the press/tap. Was there a reason for this? https://github.com/cornerstonejs/cornerstoneTools/pull/120

Last question in cornerstoneTools v2, there were some touch specific tool copies (pan vs panmultitouch) would it make more sense to have the pan tool have configuration which enables multitooltouch instead of these being two separate tools?

sedghi:
Hammer.js would add MULTI_TOUCH_PAN, PINCH, and ROTATE

sedghi:
Example to get shown need to be added [here](https://github.com/cornerstonejs/cornerstone3D-beta/find/main)

sedghi:
In case of changing serve, we need to update the docs

sedghi:
Is this ready to go?

Ouwen:
Not yet, am in the process of adding the multitouch and gestures. I think we can avoid hammerjs dependency as well

Ouwen:
@sedghi this is almost ready for your review. This is becoming a larger change than I initially intended.

Highlights
 - We don't use the hammerjs dependency
 - Multi touch is "reduced" to single touch via taking means.
     - Distances are calculated as the mean distance of unique unordered points
     - In the case of a drag event where the user lifts a finger for example, the deltas of 3 last touch => 2 current touch are calculated as the mean point of the 3touch case minus the mean  point of the 2touch case
     - Rotation is still todo, but following the case above we can take the mean of the 3 touch case and the mean of the 2 touch case. We can first translate the the 2 touch points by looking at the difference in mean points from 2points => 3points. Then for each matched point identifier we can calculate the theta and take the mean of all thetas.

TODO:
 - Since touch handles an array of touch points we have an array of IPoints[] as the currentPoints, lastPoints, and startPoints. This doesn't play nice with the mouse interface since IPoints is not an array.
 - I'll probably keep the IPoints[] under a different key (currentPointsList?). Then I'll reserve the original key with a datatype compatible with IPoints

packages/tools/src/types/EventTypes.ts
```diff
@@ -20,6 +23,24 @@ type NormalizedMouseEventDetail = {
   element: HTMLDivElement;
 };
 
+/**
+ * The normalized mouse event detail
+ */
+type NormalizedTouchEventDetail = {
```
wayfarer3130, line 29: This doesn't have anything different from the normalized mouse event detail - please re-use the existing type (you can rename it if you want, but only have one instance of it)

packages/tools/src/types/EventTypes.ts
```diff
@@ -201,6 +222,28 @@ type MouseDownEventDetail = NormalizedMouseEventDetail & {
   deltaPoints: IPoints;
 };
 
+/**
+ * EventDetail for touchstart event
+ */
+type TouchStartEventDetail = NormalizedTouchEventDetail & {
```
wayfarer3130, line 228: Please start with the MouseStartEventDetail, and then add specific behaviour that identifies this as a touch event.  That allows leaving most of the tools classes alone, and only tweaking them if needed.  There should/must be an identifier that identifies this as a touch event, and probably something that identifies what "drag" type to listen to.

packages/tools/src/types/EventTypes.ts
```diff
@@ -201,6 +222,28 @@ type MouseDownEventDetail = NormalizedMouseEventDetail & {
...
+  /** The starting points of the touch event. */
+  startPoints: ITouchPoints;
```
wayfarer3130, line 230: Let the values that can be inheritted from MouseStartEventDetail be inheritted, and only add the extras as required.

packages/tools/src/tools/ZoomTool.ts
```diff
@@ -12,7 +11,7 @@ import { EventTypes, PublicToolProps, ToolProps } from '../types';
  */
 class ZoomTool extends BaseTool {
   static toolName;
-  touchDragCallback: (evt: EventTypes.MouseDragEventType) => void;
+  touchDragCallback: (evt: EventTypes.TouchDragEventType) => void;
```
wayfarer3130, line 14: This is a place where it does need to distinguish between touch drag and mouse drag - if it can automatically get the touch drag callback here, that would be really good.

common/reviews/api/tools.api.md
```diff
@@ -105,7 +105,7 @@ export class AngleTool extends AnnotationTool {
     // (undocumented)
     _activateModify: (element: HTMLDivElement) => void;
     // (undocumented)
-    addNewAnnotation: (evt: EventTypes_2.MouseDownActivateEventType) => AngleAnnotation;
+    addNewAnnotation: (evt: EventTypes_2.MouseDownActivateEventType | EventTypes_2.TouchStartActivateEventType) => AngleAnnotation;
```
wayfarer3130, line 108: These changes make the 95% case hard, which is that mouse and touch handling are identical.  It also makes adding other custom types hard as well, because they need changes everywhere to handle the default case, which is "mouse like" events.  What you want to do is make everything handle it as though it was a mouse event, EXCEPT:
1. If the tool needs special handling that is different for a specific controller (eg touch controller)
2. Registering subsequent event handlers need to register the right type

For the second item, I mean if you touch down, then you do NOT want to listen to "mouse move", if you add a "mouse move" listener, but what you want is to listen to the move listener appropriate to the touch event.

BardiaKh:
Hey all,

Is there any news on this PR? This is a very handy feature.

Ouwen:
@BardiaKh currently in progress!

Ouwen:
@sedghi @wayfarer3130 this should be ready for review. Main changes to note:

 - TouchEvents and MouseEvents Types have been better merged to be InteractionEvent types. Tools that are agnostic to Touch and Mouse can use this event type instead. Touch and Mouse events then inherit from the interaction event in case there are specific needs (mouseButton, touch pressure, etc).
 
 - We no longer use hammerjs for touch compared to v2

packages/docs/CHANGELOG.md
```diff
@@ -53,1216 +53,620 @@ See [Conventional Commits](https://conventionalcommits.org) for commit guideline
 
 ## [0.9.1](https://github.com/cornerstonejs/cornerstone3D-beta/compare/docs@0.9.0...docs@0.9.1) (2023-01-10)
 
-
 ### Bug Fixes
 
-* no need for wadors header provider in the demo ([#356](https://github.com/cornerstonejs/cornerstone3D-beta/issues/356)) ([92891cf](https://github.com/cornerstonejs/cornerstone3D-beta/commit/92891cf4fd8f502b1dd0908702e46fb3556bacd7))
-
-
+- no need for wadors header provider in the demo ([#356](https://github.com/cornerstonejs/cornerstone3D-beta/issues/356)) ([92891cf](https://github.com/cornerstonejs/cornerstone3D-beta/commit/92891cf4fd8f502b1dd0908702e46fb3556bacd7))
```
sedghi, line 58: why this file is changing?

packages/docs/docs/concepts/cornerstone-tools/touch.md
```diff
@@ -0,0 +1,156 @@
...
+5. optional: `TOUCH_SWIPE`
+6. `TOUCH_END`
+
+Every time a user places a finger down and lifts it up, the touch order flow will always follow the above. Touch events are mutually exclusive from click events.
+
+Other touch events that can occur are the `TOUCH_TAP` event. A `TOUCH_TAP` will not trigger a `TOUCH_START` event flow. If the user taps successively, only one `TOUCH_TAP` event will fire with the count of how many times the user tapped.
+
+| EVENT                  | Description                                                                                                                                                                                      |
+| ---------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| `TOUCH_START`          | Triggers if the user places their touchpoint down for > 50ms.                                                                                                                                    |
```
sedghi, line 23: is 50 ms standard?

packages/docs/docs/concepts/cornerstone-tools/touch.md
```diff
@@ -0,0 +1,156 @@
...
+| `TOUCH_START_ACTIVATE` | Triggers only if no tools decided to stop propagation from the `TOUCH_START` event. It is useful to differentiate between touching an existing annotaton, vs needing to create a new annotaiton. |
+| `TOUCH_PRESS`          | Triggers if the user places their touchpoint down and does not move it for > 700ms                                                                                                               |
```
sedghi, line 25: same for 700 ms

packages/docs/docs/concepts/cornerstone-tools/touch.md
```diff
@@ -0,0 +1,156 @@
...
+| `TOUCH_END`            | Triggers when the user lifts one or more of their touchpoints.                                                                                                                                   |
+| `TOUCH_TAP`            | Triggers when the user makes contact with screen for less than 50ms - 10ms (buffer from the `TOUCH_START` event.)                                                                                |
+
+## Multitouch
+
+Touch events natively support multitouch which is provided as a list of [`ITouchPoints[]`](api/tools/namespace/Types#ITouchPoints).
+In order for touch events to be compatiable with mouse events, these `ITouchPoints[]` need to be reduced into a single
+`ITouchPoint`. The current strategy for array reduction is taking the mean coordinate values. Other strategies can be
+implemented such as first point, median point, etc. This can be implemented in the
+[`touch` utilities codebase](https://github.com/cornerstonejs/cornerstone3D-beta/blob/gradienthealth/main/packages/tools/src/utilities/touch/index.ts)
```
sedghi, line 37: can we use main branch of the main repo instead of gradient health remote?

packages/docs/docs/concepts/cornerstone-tools/touch.md
```diff
@@ -0,0 +1,156 @@
...
+
+```js
+import Events from '@cornerstonejs/tools/enums/Events';
+// element is the cornerstone viewport element
+element.addEventListener(Events.TOUCH_DRAG, (evt) => {
+  // my function on drag
+  console.log(evt);
+});
+
+element.addEventListener('CORNERSTONE_TOOLS_SWIPE', (evt) => {
```
sedghi, line 103: it would be great if these events are also `EVENTS.`

packages/docs/docs/concepts/cornerstone-tools/touch.md
```diff
@@ -0,0 +1,156 @@
...
+const toolGroup = ToolGroupManager.createToolGroup(toolGroupId);
+
+// Add tools to the tool group
+toolGroup.addTool(WindowLevelTool.toolName);
+toolGroup.addTool(PanTool.toolName);
+toolGroup.addTool(ZoomTool.toolName);
+toolGroup.addTool(StackScrollTool.toolName);
+
+// Set the initial state of the tools, here all tools are active and bound to
+// Different mouse inputs
```
sedghi, line 142: I don't understand the following code. So the window level is both mouse and touch?

packages/docs/docs/concepts/cornerstone-tools/touch.md
```diff
@@ -0,0 +1,156 @@
...
+toolGroup.setToolActive(ZoomTool.toolName, {
```
sedghi, line 143: and how many touchpoints we support? (given that the device supports it), can someone register a 5 touch point bingind?

packages/streaming-image-volume-loader/src/sharedArrayBufferImageLoader.ts
```diff
@@ -0,0 +1,155 @@
+import { Enums } from '@cornerstonejs/core';
```
sedghi, line 1: You should rebase on main, this is removed

packages/core/src/Settings.ts
```diff
@@ -292,3 +292,4 @@ function deepSet(context, key, value) {
  * Initial Settings for the repository
  */
 Settings.getDefaultSettings().set('useCursors', true);
+Settings.getDefaultSettings().set('debug', false);
```
sedghi, line 295: what is this

packages/tools/src/eventDispatchers/shared/getActiveToolForTouchEvent.ts
```diff
@@ -0,0 +1,55 @@
...
+    viewportId,
+    renderingEngineId
+  );
+
+  const toolGroupToolNames = Object.keys(toolGroup.toolOptions);
+
+  const numTouchPoints = Object.keys(touchEvent.touches).length;
+
+  // If any keyboard modifier key is also pressed
+  const modifierKey = getMouseModifier(touchEvent);
```
sedghi, line 34: The comment above this line says if `keyboard modifier` is pressed, but the function is `getMouseModifier` is this correct?

packages/tools/src/eventDispatchers/shared/getActiveToolForTouchEvent.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  for (let j = 0; j < toolGroupToolNames.length; j++) {
+    const toolName = toolGroupToolNames[j];
+    const toolOptions = toolGroup.toolOptions[toolName];
+
```
sedghi, line 43: I'm confused about the numTouchPoints === 1 and mix of it here with mouse button. Can you please clarify? and also mention it in the doc for touch? what is the relation of mouse and touch for the case of numTouchPoints === 1. I think it is a special case?

packages/tools/src/eventDispatchers/shared/getToolsWithModesForTouchEvent.ts
```diff
@@ -0,0 +1,54 @@
...
+  for (let j = 0; j < toolGroupToolNames.length; j++) {
+    const toolName = toolGroupToolNames[j];
+    const tool = toolGroup.toolOptions[toolName];
+
+    const correctBinding =
+      numTouchPoints != null &&
+      tool.bindings.length &&
+      tool.bindings.some(
+        (binding) => binding.numTouchPoints === numTouchPoints
+      );
```
sedghi, line 42: For instance here, I expected to see the `numTouchPoints ===1 and mouse` thing in the other file

packages/tools/src/eventListeners/touch/touchStartListener.ts
```diff
@@ -0,0 +1,441 @@
...
+      touch: null,
+    },
+  ],
+  // TODO, these default values may need to optimized
+  touchStartTimeout: null,
+  touchStartDelay: 100,
+  isTouchStart: false,
+
+  pressTimeout: null,
+  pressDelay: 700,
```
sedghi, line 86: oh I see, so these are configurable

packages/tools/src/store/ToolGroupManager/ToolGroup.ts
```diff
@@ -326,8 +326,12 @@ export default class ToolGroup implements IToolGroup {
...
       (unique, binding) => {
         if (
-          !unique.some((obj) => obj.mouseButton === binding.mouseButton) &&
-          binding.mouseButton !== undefined
+          (!unique.some((obj) => obj.mouseButton === binding.mouseButton) &&
+            binding.mouseButton !== undefined) ||
+          (!unique.some(
+            (obj) => obj.numTouchPoints === binding.numTouchPoints
+          ) &&
+            binding.numTouchPoints !== undefined)
```
sedghi, line 334: This has become very cumbersome, can we break it into two checks? `hasMouseBinding` and `hasTouchBinding` (if I got it correctly), or similar naming

packages/tools/src/tools/annotation/AngleTool.ts
```diff
@@ -241,9 +242,8 @@ class AngleTool extends AnnotationTool {
   };
 
   toolSelectedCallback = (
-    evt: EventTypes.MouseDownEventType,
-    annotation: AngleAnnotation,
-    interactionType: InteractionTypes
+    evt: EventTypes.InteractionEventType,
```
sedghi, line 245: I don't understand why our `MouseDownEventType` is changed to `InteractionEventType`? I expected it be `InteractionDownEventType` or similar

packages/tools/src/tools/CrosshairsTool.ts
```diff
@@ -469,7 +471,7 @@ class CrosshairsTool extends AnnotationTool {
   }
 
   handleSelectedCallback = (
-    evt: EventTypes.MouseDownEventType,
+    evt: EventTypes.MouseDownEventType | EventTypes.TouchStartEventType,
```
sedghi, line 474: I'm not sure I follow. For a lot of tools `MouseDownEventType` is changed to `InteractionEventType` but here it is `MouseDownEventType` and `TouchStartEventType` which actually I prefer !

sedghi (commented):
Looks awesome as always. Really appreciate the time you have put into this. Put couple of minor comments. My major comment is that I was confused with the even types for tools: 

The `Mouse[Drag/Down/...]EventTypes` where changed sometimes to `InteractionEventTypes` and sometimes to `Mouse[Drag/Down/...]EventTypes | Touch[Start/Drag/...]EventTypes]`. And the latter I prefer much better than the former!

packages/tools/src/tools/annotation/AngleTool.ts
Ouwen: Since the function doesn't use anything specific to the `DownEventType` we adopt the generic `InteractionEventType`

packages/tools/src/store/ToolGroupManager/ToolGroup.ts
Ouwen: Can do

packages/core/src/Settings.ts
Ouwen: This should probably be removed, I was turning this on when debugging touch events.

packages/streaming-image-volume-loader/src/sharedArrayBufferImageLoader.ts
Ouwen: Yeah I think this was accidentally kept during the rebase.

packages/docs/docs/concepts/cornerstone-tools/touch.md
Ouwen: We can support as many touchpoints as the hardware device so 5 touch points are possible. Fast follows: we can filter on radius / force for touch pen devices.

packages/docs/docs/concepts/cornerstone-tools/touch.md
Ouwen: WindowLevel has an internal `_drag` function take in a generic `InteractionEvent` so both mouse and touch can reuse the function. WindowLevel also exposes a public `mouseDrag` and `touchDrag` for integration into the event dispatcher system.

packages/docs/docs/concepts/cornerstone-tools/touch.md
Ouwen: I think these are included as `EVENTS.` but for doc purposes I just use the string for simplicity. Can def change if needed.

packages/docs/docs/concepts/cornerstone-tools/touch.md
Ouwen: yeah gradient health remote main should mirror cornerstonejs main. There were some weird rebase issues previously which I think are fixed now.

packages/docs/docs/concepts/cornerstone-tools/touch.md
Ouwen: I believe it was what was used in v2 hammerJS also configurable

packages/docs/docs/concepts/cornerstone-tools/touch.md
Ouwen: I believe it was what was used in v2 hammerJS also configurable

packages/docs/CHANGELOG.md
Ouwen: must have been the automated linter...

packages/docs/docs/concepts/cornerstone-tools/touch.md
sedghi: Yes please, I highly prefer consistency in all places

packages/docs/docs/concepts/cornerstone-tools/touch.md
sedghi: can we please modify and put that info in the .md doc then?

packages/tools/src/tools/CrosshairsTool.ts
sedghi: CC @wayfarer3130 what do you prefer?

sedghi:
@diegohennrich @Ouwen 
Let me know when this is ready for review
Also you need to remove the type, we don't need that anymore

@diegohennrich @Ouwen 
Can you please

packages/core/src/Settings.ts
Ouwen, line 295: remove

packages/docs/docs/concepts/cornerstone-tools/touch.md
Ouwen: Rename: 
`https://github.com/cornerstonejs/cornerstone3D-beta/blob/gradienthealth/main/packages/tools/src/utilities/touch/index.ts` => `https://github.com/cornerstonejs/cornerstone3D-beta/main/packages/tools/src/utilities/touch/index.ts`

packages/docs/docs/concepts/cornerstone-tools/touch.md
Ouwen: Update to `CORNERSTONE_TOOLS_SWIPE`  => `Events.[CORNERSTONE_TOOLS_SWIPE]`

packages/docs/docs/concepts/cornerstone-tools/touch.md
```diff
@@ -0,0 +1,156 @@
...
+
+element.addEventListener('CORNERSTONE_TOOLS_SWIPE', (evt) => {
+  // my function on swipe
+  console.log(evt);
+});
+
+// within the chrome console in a deployed OHIF application
+cornerstone
+  .getEnabledElements()[0]
+  .viewport.element.addEventListener('CORNERSTONE_TOOLS_SWIPE', (evt) => {
```
Ouwen, line 111: Update to `CORNERSTONE_TOOLS_SWIPE`  => `Events.TOUCH_SWIPE `

packages/docs/docs/concepts/cornerstone-tools/touch.md
Ouwen: yeah we can add that to the conceptual doc.

packages/tools/src/eventDispatchers/shared/getActiveToolForTouchEvent.ts
Ouwen: Yeah numTouchPoints = 1 is a special case where we also treat it as a "Primary Mouse" event. This way existing modes which just bind a primary mouse get the free touch win without additional explicit config.

Ouwen: We can make a note in the doc

packages/tools/src/eventDispatchers/mouseEventHandlers/mouseMove.ts
```diff
@@ -17,7 +17,7 @@ const { Active, Passive } = ToolModes;
  *
  * @param evt - The normalized mouseDown event.
  */
-export default function mouseMove(evt: MouseMoveEventType) {
+export default function mouseMove(evt: InteractionEventType) {
```
Ouwen, line 20: Since this is a mouse event, we should use the mouse move event.

Ouwen: The dispatchers will still use the mouse event or touch events to be specific. However the handlers will use the general event to be as applicable as needed. We can add this philosophy to the doc too.

packages/docs/docs/concepts/cornerstone-tools/touch.md
```diff
@@ -100,15 +100,15 @@ element.addEventListener(Events.TOUCH_DRAG, (evt) => {
   console.log(evt);
 });
 
-element.addEventListener('CORNERSTONE_TOOLS_SWIPE', (evt) => {
+element.addEventListener(Events.[CORNERSTONE_TOOLS_SWIPE], (evt) => {
```
Ouwen, line 103: This should be `Events.TOUCH_SWIPE`

diegohennrich:
Could you please review it? @Ouwen @sedghi

packages/tools/src/eventDispatchers/shared/getActiveToolForTouchEvent.ts
Ouwen: yes the mousemodifer function is general keyboard related. I think we may need to change the function name but this PR is getting pretty large + there are other tool related PRs coming.

packages/tools/src/eventDispatchers/shared/getToolsWithModesForTouchEvent.ts
Ouwen: In this case the behavior is to only return tools which match numTouchPoints, or if there are no touchpoints we include it in the list of tools. The numTouchPoints is analogous to mouse button.

https://github.com/cornerstonejs/cornerstone3D-beta/blob/87db14afede7def4588bb17f28d66963086e9410/packages/tools/src/eventDispatchers/shared/getToolsWithModesForMouseEvent.ts#L39-L49

The mouse primary logic would get superseeded by the !numTouchPoints

packages/tools/src/eventDispatchers/shared/getToolsWithModesForTouchEvent.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    if (
+      modesFilter.includes(tool.mode) &&
+      (!numTouchPoints || correctBinding)
```
Ouwen, line 46: !numTouchPoints here

packages/tools/src/tools/CrosshairsTool.ts
```diff
@@ -367,8 +367,8 @@ class CrosshairsTool extends AnnotationTool {
    * @returns Crosshairs annotation
    */
   addNewAnnotation = (
-    evt: EventTypes.MouseDownActivateEventType,
-    interactionType: string
+    evt: EventTypes.InteractionEventType,
+    interactionType = 'mouse'
```
Ouwen, line 371: I don't think this is used in this function. Minor nit that might be better to do in a separate PR.

packages/tools/src/tools/CrosshairsTool.ts
```diff
@@ -469,7 +469,7 @@ class CrosshairsTool extends AnnotationTool {
   }
 
   handleSelectedCallback = (
-    evt: EventTypes.MouseDownEventType,
+    evt: EventTypes.InteractionEventType,
     annotation: Annotation,
     handle: ToolHandle,
     interactionType = 'mouse'
```
Ouwen, line 475: Same here

packages/tools/src/tools/annotation/ArrowAnnotateTool.ts
```diff
@@ -409,13 +403,16 @@ class ArrowAnnotateTool extends AnnotationTool {
     triggerAnnotationRenderForViewportIds(renderingEngine, viewportIdsToRender);
   };
 
-  doubleClickCallback = (evt: EventTypes.MouseUpEventType) => {
+  touchTapCallback = (evt: EventTypes.TouchTapEventType) => {
+    if (evt.detail.taps == 2) {
+      this.doubleClickCallback(evt as unknown as EventTypes.MouseUpEventType);
```
Ouwen, line 408: this can pass evt without `as unknown as EventTypes...` with the `InteractionEventType` change we made

packages/tools/src/tools/annotation/DragProbeTool.ts
```diff
@@ -51,7 +51,7 @@ class DragProbeTool extends ProbeTool {
   }
 
   postMouseDownCallback = (
```
Ouwen, line 53: Needs a `postTouchDownCallback` exposed to have touch events dispatched to this tool.

packages/tools/src/tools/base/AnnotationTool.ts
```diff
@@ -172,7 +172,8 @@ abstract class AnnotationTool extends AnnotationDisplayTool {
     element: HTMLDivElement,
     annotation: Annotation,
     canvasCoords: Types.Point2,
-    proximity: number
+    proximity: number,
+    interactionType = 'mouse'
```
Ouwen, line 176: This should be removed

packages/tools/src/tools/segmentation/CircleScissorsTool.ts
```diff
@@ -272,24 +268,24 @@ class CircleScissorsTool extends BaseTool {
...
-    element.addEventListener(Events.MOUSE_DRAG, this._mouseDragCallback);
-    element.addEventListener(Events.MOUSE_CLICK, this._mouseUpCallback);
+    element.addEventListener(Events.MOUSE_UP, this._endCallback);
+    element.addEventListener(Events.MOUSE_DRAG, this._dragCallback);
+    element.addEventListener(Events.MOUSE_CLICK, this._endCallback);
 
-    // element.addEventListener(Events.TOUCH_END, this._mouseUpCallback)
-    // element.addEventListener(Events.TOUCH_DRAG, this._mouseDragCallback)
+    // element.addEventListener(Events.TOUCH_END, this._endCallback)
+    // element.addEventListener(Events.TOUCH_DRAG, this._dragCallback)
```
Ouwen, line 276: This is wrongly commented out. It should also add the `Events.TOUCH_TAP` event

packages/tools/src/tools/segmentation/CircleScissorsTool.ts
```diff
@@ -272,24 +268,24 @@ class CircleScissorsTool extends BaseTool {
...
-    element.removeEventListener(Events.MOUSE_DRAG, this._mouseDragCallback);
-    element.removeEventListener(Events.MOUSE_CLICK, this._mouseUpCallback);
+    element.removeEventListener(Events.MOUSE_UP, this._endCallback);
+    element.removeEventListener(Events.MOUSE_DRAG, this._dragCallback);
+    element.removeEventListener(Events.MOUSE_CLICK, this._endCallback);
 
-    // element.removeEventListener(Events.TOUCH_END, this._mouseUpCallback)
-    // element.removeEventListener(Events.TOUCH_DRAG, this._mouseDragCallback)
+    // element.removeEventListener(Events.TOUCH_END, this._endCallback)
+    // element.removeEventListener(Events.TOUCH_DRAG, this._dragCallback)
```
Ouwen, line 288: Wrongly commented out same as above.

Ouwen: also removeEventListener to the Events.TOUCH_TAP event

packages/tools/src/tools/segmentation/RectangleScissorsTool.ts
```diff
@@ -307,24 +303,24 @@ class RectangleScissorsTool extends BaseTool {
...
-    element.addEventListener(Events.MOUSE_DRAG, this._mouseDragCallback);
-    element.addEventListener(Events.MOUSE_CLICK, this._mouseUpCallback);
+    element.addEventListener(Events.MOUSE_UP, this._endCallback);
+    element.addEventListener(Events.MOUSE_DRAG, this._dragCallback);
+    element.addEventListener(Events.MOUSE_CLICK, this._endCallback);
 
-    // element.addEventListener(Events.TOUCH_END, this._mouseUpCallback)
-    // element.addEventListener(Events.TOUCH_DRAG, this._mouseDragCallback)
+    // element.addEventListener(Events.TOUCH_END, this._endCallback)
+    // element.addEventListener(Events.TOUCH_DRAG, this._dragCallback)
```
Ouwen, line 311: This should be enabled and include

`element.addEventListener(Events.TOUCH_TAP, this._endCallback)`

packages/tools/src/tools/segmentation/RectangleScissorsTool.ts
```diff
@@ -307,24 +303,24 @@ class RectangleScissorsTool extends BaseTool {
...
-    element.removeEventListener(Events.MOUSE_DRAG, this._mouseDragCallback);
-    element.removeEventListener(Events.MOUSE_CLICK, this._mouseUpCallback);
+    element.removeEventListener(Events.MOUSE_UP, this._endCallback);
+    element.removeEventListener(Events.MOUSE_DRAG, this._dragCallback);
+    element.removeEventListener(Events.MOUSE_CLICK, this._endCallback);
 
-    // element.removeEventListener(Events.TOUCH_END, this._mouseUpCallback)
-    // element.removeEventListener(Events.TOUCH_DRAG, this._mouseDragCallback)
+    // element.removeEventListener(Events.TOUCH_END, this._endCallback)
+    // element.removeEventListener(Events.TOUCH_DRAG, this._dragCallback)
```
Ouwen, line 323: Same here
This should be enabled and include

`element. removeEventListener(Events.TOUCH_TAP, this._endCallback)`

packages/tools/src/tools/segmentation/SphereScissorsTool.ts
```diff
@@ -273,24 +271,24 @@ class SphereScissorsTool extends BaseTool {
...
-    element.addEventListener(Events.MOUSE_DRAG, this._mouseDragCallback);
-    element.addEventListener(Events.MOUSE_CLICK, this._mouseUpCallback);
+    element.addEventListener(Events.MOUSE_UP, this._endCallback);
+    element.addEventListener(Events.MOUSE_DRAG, this._dragCallback);
+    element.addEventListener(Events.MOUSE_CLICK, this._endCallback);
 
-    // element.addEventListener(Events.TOUCH_END, this._mouseUpCallback)
-    // element.addEventListener(Events.TOUCH_DRAG, this._mouseDragCallback)
+    // element.addEventListener(Events.TOUCH_END, this._endCallback)
+    // element.addEventListener(Events.TOUCH_DRAG, this._dragCallback)
```
Ouwen, line 279: Same as other segmentation tools.

packages/tools/src/tools/segmentation/SphereScissorsTool.ts
```diff
@@ -273,24 +271,24 @@ class SphereScissorsTool extends BaseTool {
...
-    element.removeEventListener(Events.MOUSE_DRAG, this._mouseDragCallback);
-    element.removeEventListener(Events.MOUSE_CLICK, this._mouseUpCallback);
+    element.removeEventListener(Events.MOUSE_UP, this._endCallback);
+    element.removeEventListener(Events.MOUSE_DRAG, this._dragCallback);
+    element.removeEventListener(Events.MOUSE_CLICK, this._endCallback);
 
-    // element.removeEventListener(Events.TOUCH_END, this._mouseUpCallback)
-    // element.removeEventListener(Events.TOUCH_DRAG, this._mouseDragCallback)
+    // element.removeEventListener(Events.TOUCH_END, this._endCallback)
+    // element.removeEventListener(Events.TOUCH_DRAG, this._dragCallback)
```
Ouwen, line 291: Same as other segmentation tools

Ouwen (commented):
I did a pretty thorough review for each file. I want to be mindful that this is a very large PR and would prefer it to go in before other tool related PRs if possible @sedghi. Otherwise subsequent changes will need to rebased and force push might sneak diffs.

These were the main changes I found, but many of them are cosmetic which might be better handled in another (smaller) PR.

Ouwen:
@sedghi lmk if there are any outstanding changes required aside from the above... thanks again for taking the time to review this large diff

sedghi:
@Ouwen probably yes, but I think I would like to do one last review. Just let me know when it is ready for it

packages/tools/src/tools/annotation/ArrowAnnotateTool.ts
diegohennrich: Actually, I think we need to use the specific TouchTapEventDetail here because we don't have the "taps" parameter from InteractionEventType. So "TouchTapEventDetail" is inherited from the interaction.

```
type TouchTapEventDetail = NormalizedInteractionEventDetail &
  TouchCustomEventDetail & {
    currentPointsList: ITouchPoints[];
    currentPoints: ITouchPoints;
    taps: number; <----
  };
```

diegohennrich:
Could you review it, please? @Ouwen @sedghi

packages/tools/src/tools/annotation/DragProbeTool.ts
diegohennrich: "postTouchDownCallback" or "postTouchStartCallback"?

I believe we're using the "postTouchStartCallback", see below:

packages/tools/src/eventDispatchers/touchEventHandler/touchstart.ts

```
if (activeTool && typeof activeTool.postTouchStartCallback === 'function') {
    const consumedEvent = activeTool.postTouchStartCallback(evt);

    if (consumedEvent) {
      // If the tool claims it consumed the event, prevent further checks.
      return;
    }
  }
```

packages/tools/src/tools/annotation/ArrowAnnotateTool.ts
```diff
@@ -405,11 +405,11 @@ class ArrowAnnotateTool extends AnnotationTool {
 
   touchTapCallback = (evt: EventTypes.TouchTapEventType) => {
     if (evt.detail.taps == 2) {
-      this.doubleClickCallback(evt as unknown as EventTypes.MouseUpEventType);
+      this.doubleClickCallback(evt as EventTypes.TouchTapEventType);
     }
   };
 
-  doubleClickCallback = (evt: EventTypes.InteractionEventType): void => {
+  doubleClickCallback = (evt: EventTypes.TouchTapEventType): void => {
```
Ouwen, line 412: should stay InteractionEventType

packages/tools/src/tools/annotation/ArrowAnnotateTool.ts
```diff
@@ -405,11 +405,11 @@ class ArrowAnnotateTool extends AnnotationTool {
 
   touchTapCallback = (evt: EventTypes.TouchTapEventType) => {
     if (evt.detail.taps == 2) {
-      this.doubleClickCallback(evt as unknown as EventTypes.MouseUpEventType);
+      this.doubleClickCallback(evt as EventTypes.TouchTapEventType);
```
Ouwen, line 408: should just be `evt`

packages/tools/src/tools/segmentation/CircleScissorsTool.ts
Ouwen:
```
    // element.addEventListener(Events.TOUCH_END, this._endCallback)
    // element.addEventListener(Events.TOUCH_DRAG, this._dragCallback)
```

Needed as well

packages/tools/src/tools/annotation/DragProbeTool.ts
Ouwen: should be postTouchStart callback you are correct. Rather than copying the mouse function you can just call the postMouseDownCallback function inside of postTouchStart callback

packages/tools/src/tools/segmentation/RectangleScissorsTool.ts
```diff
@@ -321,8 +321,8 @@ class RectangleScissorsTool extends BaseTool {
     element.removeEventListener(Events.MOUSE_CLICK, this._endCallback);
     element.removeEventListener(Events.TOUCH_TAP, this._endCallback);
```
Ouwen, line 322: let's move the touch_tap back down to after line 325 to match mouse formatting

packages/tools/src/tools/annotation/ArrowAnnotateTool.ts
Ouwen: should be evt since `touchTapCallback = (evt: EventTypes.TouchTapEventType) => {`

packages/tools/src/tools/annotation/ArrowAnnotateTool.ts
diegohennrich: As we discussed yesterday, this is a special case where we need to use the taps parameter and we don't have it inherited from the InteractionType. So, make sense to use the TouchTapEventDetail at this specific point.

diegohennrich:
I think everything was done. Could you review it, please? @Ouwen @sedghi

Ouwen:
@sedghi it is ready for review.

packages/docs/docs/concepts/cornerstone-tools/touch.md
```diff
@@ -0,0 +1,218 @@
...
+  bindings: [
+    {
+      mouseButton: MouseBindings.Primary, // special condition for one finger touch
+    },
+  ],
+});
+```
+
+The `MouseBindings.Primary` is a special binding type which will
+automatically bind single finger touch.
```
sedghi, line 162: I don't understand, why the API here is different, is this supposed to do a mouse binding or touch bindings? If it is touch, I expected the following for consistensy

```
toolGroup.setToolActive(WindowLevelTool.toolName, {
  bindings: [{ numTouchPoints: 1 }],
});
```

packages/tools/examples/stackManipulationToolsTouch/index.ts
```diff
@@ -0,0 +1,143 @@
...
+  toolGroup.setToolActive(StackScrollTool.toolName, {
+    bindings: [{ numTouchPoints: 3 }],
+  });
+  toolGroup.setToolActive(WindowLevelTool.toolName, {
+    bindings: [
+      {
+        mouseButton: MouseBindings.Primary, // Left Click
+      },
+    ],
+  });
```
sedghi, line 97: Same here

packages/tools/src/eventDispatchers/mouseEventHandlers/mouseMove.ts
```diff
@@ -6,7 +6,10 @@ import { ToolModes } from '../../enums';
 import filterToolsWithAnnotationsForElement from '../../store/filterToolsWithAnnotationsForElement';
 import getToolsWithModesForMouseEvent from '../shared/getToolsWithModesForMouseEvent';
 import triggerAnnotationRender from '../../utilities/triggerAnnotationRender';
-import { MouseMoveEventType } from '../../types/EventTypes';
+import {
+  InteractionEventType,
```
sedghi, line 10: Not needed import for InteractionEventType

packages/tools/examples/stackManipulationToolsTouch/index.ts
sedghi: I remember you said this 

```
Yeah numTouchPoints = 1 is a special case where we also treat it as a "Primary Mouse" event. This way existing modes which just bind a primary mouse get the free touch win without additional explicit config.
```

But don't think we should treat the numPoint ===1 special, why not for the others? why just 1 ?

sedghi (approved):
The work done on this PR is truly exceptional and greatly appreciated. Kudos to the Gradient Health team as always 🎉 💯

PR #258, opened 2022-10-24
fix logic for flip, remove useless member variable ( panCache )

bionoone:


packages/core/src/RenderingEngine/Viewport.ts
```diff
@@ -252,20 +252,38 @@ class Viewport implements IViewport {
     vec3.scaleAndAdd(center, center, jVector, (size[1] / 2.0) * spacing[1]);
     vec3.scaleAndAdd(center, center, kVector, (size[2] / 2.0) * spacing[2]);
 
+    const imageToWorld: mat4 = [
```
sedghi, line 255: This is not imageToWorld (since it does not have Origin in the 4th column). Maybe call it `directionMatrix`?

packages/core/src/RenderingEngine/Viewport.ts
```diff
@@ -252,20 +252,38 @@ class Viewport implements IViewport {
...
+      direction[7],
+      direction[8],
+      0, //
+      0,
+      0,
+      0,
+      1, //
+    ];
+    const worldToImage: mat4 = mat4.transpose(
+      [1, 2, 3, 4, 5, 6, 7, 8, 9, 0, 1, 2, 3, 4, 5, 6],
```
sedghi, line 274: I think you are creating an output array here, but can you just use api from `gl-matrix` ? or just put everything as 0

sedghi (commented):
I'm not sure I see what is fixed, the demos work like before, but I wanted the flip to be based on direction which your PR does too

packages/core/src/RenderingEngine/Viewport.ts
bionoone: you are right, we can simplify greatly the code

bionoone:
> 

Original Unflip :
 
original

Original – horizontal flip:
original-hflip

 
Original vertical flip:
 
original-vflip

sedghi:
I did some digging into this, I guess the correct implementation should do it based on `viewRight` and `viewUp`, and not the direction of the image itself, since that is basically what we mean by `flipHorizontal` and `flipVertical`. I tried it 

```js
 const { focalPoint, viewPlaneNormal, viewUp } = this.getCamera();

    const viewRight = vec3.create();
    vec3.cross(viewRight, viewPlaneNormal, viewUp);

    const center = focalPoint;
    let flipHTx, flipVTx;

    const transformToOriginTx = vtkMatrixBuilder
      .buildFromRadian()
      .identity()
      .translate(-center[0], -center[1], -center[2])
      .rotateFromDirections(viewRight, [1, 0, 0])
      .rotateFromDirections(viewUp, [0, 1, 0]);

    const transformBackFromOriginTx = vtkMatrixBuilder
      .buildFromRadian()
      .identity()
      .rotateFromDirections([0, 1, 0], viewUp)
      .rotateFromDirections([1, 0, 0], viewRight)
      .translate(center[0], center[1], center[2]);

    if (flipH) {
      this.flipHorizontal = flipHorizontal;
      flipHTx = vtkMatrixBuilder
        .buildFromRadian()
        .multiply(transformToOriginTx.getMatrix())
        .scale(-1, 1, 1)
        .multiply(transformBackFromOriginTx.getMatrix());
    }

    if (flipV) {
      this.flipVertical = flipVertical;
      flipVTx = vtkMatrixBuilder
        .buildFromRadian()
        .multiply(transformToOriginTx.getMatrix())
        .scale(1, -1, 1)
        .multiply(transformBackFromOriginTx.getMatrix());
    }
``` 

It works for flipHorizontal for all views (except oblique) but not for vertical. 

I guess the main issue is that flipping happens at the actor matrix level, and actor matrix is agnostic to the camera view, so that is why some orientations it works and some don't (?)

I can dig more later

bionoone:
there is a feature definition issue : the flip is about flipping the original orientation.
typical use case is for XRay where depending on what you want to do, you may want to see left & right hand in the same orientation or opposite. ( for CT / MR, the use case, is the patient is in supine/prone position. but this resolve using the rotate tool - all acquisition I ever seen are head first. no feet first ).

this is the current implementation. so if your acquisition is AXIAL, and the view is SAGITAL, horizontal flip, exchanging left and right has no effect on the (sagital) view.

Using viewup + normal would be required if you want to achieve an engineer feature : flip view in any orientation. but in this case, there is a need for a large refactor to handle it correctly. for instance you flip image, then using the crosshair you tilt the view, and then if you flip, it's along another axis. so in the end, you cannot have a "flip" indicator.

sedghi:
https://github.com/cornerstonejs/cornerstone3D-beta/pull/271 draft

sedghi:
can we close this in favor of recent merged PRs?

sedghi:
@bionoone Can we close this?

sedghi:
I'm closing this, please re-open if needed

PR #259, opened 2022-10-24
fix(annotation):The wrong annotation was being selected for common frameUID

wayfarer3130:
When the frame of reference UID is common between series, and two markups happen to coincide on different images, the first one found will be select for nearby tool data, thus updating and setting values for the wrong annotation.

### DIFF
--- a/packages/tools/src/utilities/getAnnotationNearPoint.ts
+++ b/packages/tools/src/utilities/getAnnotationNearPoint.ts
@@ -106,7 +106,8 @@ function findAnnotationNearPointByTool(
   if (annotations?.length) {
     const { element } = enabledElement.viewport;
     for (const annotation of annotations) {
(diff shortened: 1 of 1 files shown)


packages/tools/src/utilities/getAnnotationNearPoint.ts
```diff
@@ -102,9 +102,15 @@ function findAnnotationNearPointByTool(
     enabledElement.viewport.element,
     (tool.constructor as typeof BaseTool).toolName
   );
+  const currentId = enabledElement.viewport?.getCurrentImageId();
```
sedghi, line 105: Interesting bug, but this only exists for StackViewport, VolumeViewport don't have `getCurrentImageId`.  We are using `element` here for `isPointNearTool`, I don't understand why the wrong annotation is selected?

PR #265, opened 2022-10-26
Fix issue when getting an index that equals 0.

gabriellebaudy:
When using this method while being on the first image, the index returns 0, and this method would return `undefined`. This is unwanted since the first image obviously has a 0 index which is usable. My fix simply consists of ensuring the returned value is a number (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isNaN).


sedghi:
Great catch, thanks for your contribution

sedghi (approved):

PR #268, opened 2022-10-27
fix: Use queryselector instead of firstChild to get svg-layer

lambacini:
Related issue : https://github.com/cornerstonejs/cornerstone3D-beta/issues/252#issue-1416549311

if getSvgDrawingHelper is not a generic and will not be used for any other purpose, this code will solve the problem.


sedghi (approved):
Thanks! Looks great

PR #269, opened 2022-10-28
feat: reset to center option for reset camera

sedghi:
- Refactored reset camera to use the setCamera (previous bug where resetting volumes where not updating camera because of clipping planes)
- Added resetToCenter parameter to enable optional keeping the slice and just reseting the slice for volumes in any planes

Try example: https://deploy-preview-269--cornerstone-3d-docs.netlify.app/live-examples/resetcamera

### DIFF
--- a/packages/core/src/RenderingEngine/Viewport.ts
+++ b/packages/core/src/RenderingEngine/Viewport.ts
@@ -573,7 +573,6 @@ class Viewport implements IViewport {
     // Reset the perspective zoom factors, otherwise subsequent zooms will cause
     // the view angle to become very small and cause bad depth sorting.
     // todo: parallel projection only
-    activeCamera.setViewAngle(90.0);
 
     focalPoint[0] = (bounds[0] + bounds[1]) / 2.0;
     focalPoint[1] = (bounds[2] + bounds[3]) / 2.0;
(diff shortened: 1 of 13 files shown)


packages/core/src/utilities/hasNaNValues.ts
```diff
@@ -0,0 +1,12 @@
+/**
+ * A function that checks if there is a value in the array that is NaN.
+ * or if the input is a number it just checks if it is NaN.
+ * @param input - The input to check if it is NaN.
+ * @returns - True if the input is NaN, false otherwise.
+ */
+export default function hasNaNValues(input: number[] | number): boolean {
```
swederik, line 7: does not need to be public API

packages/core/src/types/ICamera.ts
```diff
@@ -28,6 +29,12 @@ interface ICamera {
   flipHorizontal?: boolean;
   /** flip Vertical */
   flipVertical?: boolean;
+  /** physical scale */
```
swederik, line 32: if we have no intention for our end users to ever use this, we should leave it out

packages/core/src/RenderingEngine/Viewport.ts
```diff
@@ -624,53 +624,48 @@ class Viewport implements IViewport {
...
+    this.setCamera({
+      parallelScale: resetZoom ? parallelScale : previousCamera.parallelScale,
+      focalPoint: focalPointToSet,
+      position: positionToSet,
+      viewAngle: 90,
+      viewUp: viewUpToSet,
+      physicalScale: radius,
+      physicalTranslation: focalPointToSet.map((v) => -v) as Point3,
+      clippingRange: clippingRangeToUse,
+      flipHorizontal: this.flipHorizontal ? false : undefined,
```
swederik, line 661: I am confused why this is here

sedghi: I will address it in the other pr

PR #275, opened 2022-11-01
feat: add referenceCursors tool

doepnern:
Hey, I really love this repository and especially the improved API in the new version.
I created this tool for my personal use, its purpose is mostly to act like a linking of camera position and showing a reference to what your cursor is currently hovering over. It also helps a lot when you have multiple series which are not in the same Orientation, to have a feel for what you are looking at in different orientations. This is somewhat similar to the crosshair that exists, but it also works with Stack viewports, which I use it mostly for. 
Would there be an interest in adding this tool, I would take the time to write tests as well if that is the case.
Thank you for your great library anyway!
Niclas

p.s. here is a small video 
https://user-images.githubusercontent.com/32524613/199220760-a20ea8e2-9efb-4335-a763-0325ecba5f43.mp4


sedghi:
@doepnern Hey 
Thanks for your contribution. First of all, this is really impressive PR. 
Yes, we have seen this tool on other viewers and would love to also include it in Cornerstone3D and eventually on OHIF. 

I will go ahead and review this now

doepnern:
I am not completely sure if I'm happy with how this works right now. Maybe it would be easier to not use the annotation tool at all as this doesnt use any of the features where having the worldPosition of the annotation is useful and the easiest way to render the crosshair would be to just have the current enabled element where the mouse is over and the canvas position and then just calculate the world position for every viewport on render.

packages/tools/src/tools/CursorCrosshairSync.ts
```diff
@@ -0,0 +1,581 @@
...
+
+import triggerAnnotationRenderForViewportIds from '../utilities/triggerAnnotationRenderForViewportIds';
+import { StyleSpecifier } from '../types/AnnotationStyle';
+import { vec3 } from 'gl-matrix';
+
+/**
+ * CursorCrosshairSyncTool is a tool that will show your cursors position in all other elements in the toolGroup if they have a matching FrameOfReference relative to its position in world space.
+ * Also when positionSync is enabled, it will try to sync viewports so that the cursor can be displayed in the correct position in all viewports.
+ */
+class CursorCrosshairSyncTool extends AnnotationTool {
```
sedghi, line 37: So I talked internally, and we think maybe `referenceCursors` would be a better name, since it aligns with the `referenceLines` which is basically the same concept to show lines on viewports calculated based on their positioning with each other. What do you think?

packages/tools/src/tools/CursorCrosshairSync.ts
```diff
@@ -0,0 +1,581 @@
...
+    triggerAnnotationRenderForViewportIds(renderingEngine, viewportIdsToRender);
+  }
+
+  /**
+   *
+   * @param evt -  EventTypes.NormalizedMouseEventType
+   * @returns The annotation object.
+   *
+   */
+  addNewAnnotation = (
```
sedghi, line 192: can you move this to the top of class as first method, to be similar to our other tools, I started reading the code and saw it is not there, that raised a lot of questions until I scrolled down

packages/tools/src/tools/CursorCrosshairSync.ts
```diff
@@ -0,0 +1,581 @@
...
+  ): CursorCrosshairSync => {
+    const eventDetail = evt.detail;
+    const { currentPoints, element } = eventDetail;
+    const worldPos = currentPoints.world;
+
+    const enabledElement = getEnabledElement(element);
+    if (!enabledElement) throw new Error('No enabled element found');
+    const { viewport, renderingEngine } = enabledElement;
+
+    this.isDrawing = true;
```
sedghi, line 203: I looked at your demo, the active viewport (the one controlling the cursor) has actually the pointer cursor and others have a nice + (good job on the design of that). I think it makes sense all of them have the +

sedghi: Oh I see why, since you overrode the mouseMouve hmm

packages/tools/src/tools/CursorCrosshairSync.ts
```diff
@@ -0,0 +1,581 @@
...
+        this.updateAnnotationPosition(element, activeAnnotation);
+      },
+      {
+        once: true,
+      }
+    );
+  };
+
+  //when the mouse is moved over a div, attach an event listener to this div to update the world position of the annotation when stack is scrolled
+  setMouseOverElement(element: HTMLDivElement | null): void {
```
sedghi, line 121: can we rename this to `attachElementEventListeners` or similar?

packages/tools/src/tools/CursorCrosshairSync.ts
```diff
@@ -0,0 +1,581 @@
...
+    if (element === this._mouseOverElement) return;
```
sedghi, line 122: also this, can we rename it from `mouseOverElement` to `elementWithCursor` or similar?

packages/tools/src/tools/CursorCrosshairSync.ts
```diff
@@ -0,0 +1,581 @@
...
+        Enums.Events.VOLUME_NEW_IMAGE,
+        this.handleImageChange as EventListener
+      );
+      previousElement.removeEventListener(
+        Enums.Events.STACK_VIEWPORT_SCROLL,
+        this.handleImageChange as EventListener
+      );
+    }
+    if (element) {
+      this._mouseOverElement = element;
```
sedghi, line 136: isn't this repetitive?

packages/tools/src/tools/CursorCrosshairSync.ts
```diff
@@ -0,0 +1,581 @@
...
+    const annotations = getAnnotations(element, this.getToolName());
+    if (annotations === undefined || annotations.length === 0) {
+      return null;
+    }
+    const targetAnnotation = annotations[0];
+    if (annotations.length > 1) {
+      const wrongAnnotations = annotations.slice(1);
+      wrongAnnotations.forEach((annotation) => {
+        if (annotation.annotationUID)
+          removeAnnotation(annotation.annotationUID, element);
```
sedghi, line 158: I guess you have to removeAnnotation since each click will create a new one (via addNewAnnotation)?

sedghi: Do we want to add a early check on `addAnnotation` that check if there is one annotation, return early so that we don't add annotation on each click?

doepnern:
Okay, so I hopefully applied most of the changes you suggested. Thanks a lot for the extensive feedback.

**The reason I was working with imageChange events:**
Currently on every mouseMove the annotations world position is changed which works fine.  However if the viewport over which the cursor is currently positioned is for example scrolled via StackScrollMouseWheelTool, then no mouseMove event is triggered but the world position of the annotation should change. If all viewports would be rerendered we could just calculate it based on the canvasPosition but only the Viewport being scrolled gets rerended.

This leads to the cursor being stale until the next mouseMove is triggered. Also this will prevent the stacks to sync during scrolling for the same reason.

The only way I found around adding the event listeners now is: 
When a Stack change happens, it will cause the annotation on the viewport which was scrolled to be rendered
- check during render of annotation if canvasPosition still corresponds to saved WorldPosition (this wont be true if a StackChange happened)
- if stackChange happened, update worldPosition and trigger rerender for all viewports

There after some longer debugging I noticed that calling triggerAnnotationRerender during an annotationRender will cause the whole tool to break (even when excluding the currently rendering viewport). For that reason, I had to wrap the call into a setTimeout. Feel kind of "hacky" but I just don't see a completely easy way to do this. 

You can test the whole problem by commenting out the block in the renderAnnotation checking for the stackChange. When using StackScrollWheel the cursor won't be synced and the stack states won't be synced until a mouseMove is triggered.

**About the cursor being rendered in all viewports:**
I added it to the viewport with the cursor as well now. 
I'm not sure what to do with the mouse though. Removing it completely might be confusing as all cursors will look the same then. Maybe the main cursor should have a different design to not confuse the user. (I thought about a different color, but then how do you know which one is the main on 2 viewports and how would you handle custom color choices).  I'm very open to suggestions.

sedghi:
@doepnern I just merged a AnnoationDisplayTool since we needed that too (https://github.com/cornerstonejs/cornerstone3D-beta/pull/283) You can go ahead and pull the changes. Now

- You don't need to implement various methods that you don't use
- You don't need the addNewAnnotation since that gets called on mouse click. Maybe rename it something else and then mouseMove will handle creation and addition to state if necessary. 

I see why you where you were implementing the subscriptions. I might have an idea, tools can implement a `onCameraModified` too. That will capture the scroll behaviour and you can triggerAnnotationRender after that, but also `onCameraModified` will get triggered on pan and zoom, so you need to differentiate between these scenarios and only take into account the focal point change in the direction of normal (since zoom don't touch the focal point, and pan change focal point perpendicular to the normal). You might want to look at crosshairs here that might give you some ideas [here](https://github.com/cornerstonejs/cornerstone3D-beta/blob/main/packages/tools/src/tools/CrosshairsTool.ts#L477), more specifically [here](https://github.com/cornerstonejs/cornerstone3D-beta/blob/main/packages/tools/src/tools/CrosshairsTool.ts#L557)

packages/tools/src/tools/ReferenceCursors.ts
```diff
@@ -0,0 +1,572 @@
...
+    return distance;
+  };
+
+  const closestStack = {
+    distance: getDistance(imageIds[currentImageIdIndex]) ?? Infinity,
+    index: currentImageIdIndex,
+  };
+
+  //check higher indices
+  const higherImageIds = imageIds.slice(currentImageIdIndex + 1);
```
sedghi, line 493: why are we splitting the stack into lower and upper sections? shouldn't we just do the search on all imageIds here? Also I believe this function is a useful function for other purposes (stack synchronizer tool), can we refactor it into the utils inside `@core/utilities/getClosestImageForPoint` or similar?

Also plane distance to point can also be pushed inside `@core/utilities/planar`

packages/tools/src/tools/ReferenceCursors.ts
```diff
@@ -0,0 +1,572 @@
...
+        [x - centerSpace / 2, y],
+        { color, lineDash, lineWidth }
+      );
+      renderStatus = true;
+    }
+
+    return renderStatus;
+  };
+
+  updateStackPosition(
```
sedghi, line 410: This does not represent the underlying logic, `updateStackPosition` makes me think this is only for stackViewport, how about `updateViewportImage` or  `updateViewport`?

doepnern:
Thanks for all the suggestions. `onCameraModified` is actually a perfect way. It would be really helpful if you would have a small overview of functions a tool can implement at some point. What I noticed is, that when you set the tool to passive there are errors in the console when clicking on the screen for functions not being implemented even though I'm now extending the   `AnnotationDisplayTool`. You can try it out by setting the tool to passive in the example

packages/tools/src/tools/ReferenceCursors.ts
```diff
@@ -0,0 +1,424 @@
...
+              topRight: <Types.Point3>[0, 0, 0],
+              bottomLeft: <Types.Point3>[0, 0, 0],
+              bottomRight: <Types.Point3>[0, 0, 0],
+            },
+          },
+        },
+      },
+    };
+
+    const annotationId = this._addAnnotation(element, annotation);
```
sedghi, line 139: This check which is basically checking annotation manager if there is one exists can be done very first thing at the top of this method and avoid all the object and logics up here

sedghi:
Another comment after thinking, your tool has active state, meaning that one should activate it, so I expect the demo makes `setToolActive` which changes the cursor (right now it shows pan)

doepnern:
I actually like the option of still having the cursor of another tool. In actual usage this tool would mostly be used while still having a tool active, so I dont like hiding the primary tool that much. I changed the look of the tool in the element with the cursor to hopefully not conflict with another tool and added a configuration to disable the other tools cursor. I was having a lot of issues with curors being replaced depending on the order in which tools are enabled. Would it maybe be useful to only change the cursor to the default one if there is no active primary cursor(I added a check for that in the ToolGroup). Feel free to remove that if you dont feel like this is wanted.

packages/tools/src/tools/ReferenceCursors.ts
```diff
@@ -0,0 +1,471 @@
...
+  constructor(
+    toolProps: PublicToolProps = {},
+    defaultToolProps: ToolProps = {
+      supportedInteractionTypes: ['Mouse', 'Touch'],
+      configuration: {
+        shadow: true,
+        preventHandleOutsideImage: false,
+        displayThreshold: 5,
+        positionSync: true,
+        disableCursor: false,
```
sedghi, line 70: default for disableCursor to be false hmmm, going back to the discussion on the cursors for this tool, I don't understand why one needs to have the pan cursor on and also the cursor for this tool at the same time. It is kind of not in harmony with the rest of the library in which the active tool on the mouse primary claims the cursor and not other tools. Can we set this tool to behave the same way by default? that it overwrite its cursor by default?

packages/tools/src/tools/ReferenceCursors.ts
```diff
@@ -0,0 +1,471 @@
...
+      },
+    }
+  ) {
+    super(toolProps, defaultToolProps);
+    this._disableCursorEnabled = this.configuration.disableCursor;
```
sedghi, line 75: This might cause a bug when configuration changes, since you are only assigning it on the construtor, maybe always use `this.configuration.disableCursor`?

packages/tools/src/tools/ReferenceCursors.ts
```diff
@@ -0,0 +1,471 @@
...
+      const newFocalPoint = vec3.add(
+        vec3.create(),
+        vec3.fromValues(...focalPoint),
+        scaledPlaneNormal
+      ) as Types.Point3;
+      //TODO: make check if new focal point is within bounds of volume
+      const isInBounds = true;
+      if (isInBounds) {
+        viewport.setCamera({ focalPoint: newFocalPoint });
+        const renderingeEngine = viewport.getRenderingEngine();
```
sedghi, line 463: typo

packages/tools/src/tools/ReferenceCursors.ts
sedghi: Ignore this comment

sedghi (commented):
Almost there, see my comments

1. Is there any usecase for having positionSync off? I feel like user wants it always on, and if it is off, it will cause confusion when other viewports not showing the same image

2. The demo doesn't work when switching between disable cursor on and off (probably related to my comment for the constructor)

3. There is a one pixel offset between the pan cursor and your tool cursor, do you think that is easy to fix?

doepnern:
Good points,
1. I think mostly the sync is wanted, however it might interfere in some cases with some other tools or syncing mechanisms. Also the cursor is only displayed in the other viewport if the closest distance is smaller than displayThreshold, so there will be no cursor displayed if it doesn't make sense to display it.
2. This is a more complicated point and I decided for this implementation because the logic of hiding the cursor is pretty impractical to work with, namely:
https://github.com/cornerstonejs/cornerstone3D-beta/blob/a923b6888e80838edc2253fb6a82f5827b90724e/packages/tools/src/cursors/elementCursor.ts#L17-L37 
Problems I had:
- Calling hideElementCursor twice without a resetElementCursor will set cursors[0] and cursors [1] to "none" -> This will lead to resetElementCursor to never work again until some other tool calls setElementCursor

This will also be a problem if 2 tools want to call hideElementCursor

So one would need to be sure to always clean up hideElemetCursor by calling resetElementCursor. As there is no onToolConfigurationChange Event the easiest way I found to do that was to set the internal _disableCursorEnabled every time the tool is activated and call the cleanup when the tool is disabled depending on that state. If you want to change the setting you have to deactivate and activate the tool once.

3. No idea on this one, I think it might have something to do with how the svgCursors are drawn in differentce to the Annotations

sedghi:
Yeah the cursor API is probably the worst API in the whole repo, and I tried to fix it couple of times but initial implementation does not makes sense at all (and not my fault :D) 

I think this is good to go 

Thanks again! and looking forward to your next contribution!

sedghi (approved):
Great work

PR #279, opened 2022-11-04
fix: adjust canvas, not only off screen renderer on resize

Ouwen:
Fix for [#2947 ](https://github.com/OHIF/Viewers/issues/2947)

### DIFF
--- a/packages/core/src/RenderingEngine/RenderingEngine.ts
+++ b/packages/core/src/RenderingEngine/RenderingEngine.ts
@@ -577,6 +577,12 @@ class RenderingEngine implements IRenderingEngine {
 
     // 3. Reset viewport cameras
     vtkDrivenViewports.forEach((vp: IStackViewport | IVolumeViewport) => {
+      const canvas = getOrCreateCanvas(vp.element);
+      const rect = canvas.getBoundingClientRect();
+      const devicePixelRatio = window.devicePixelRatio || 1;
+      canvas.width = rect.width * devicePixelRatio;
+      canvas.height = rect.height * devicePixelRatio;
+
       const prevCamera = vp.getCamera();
       vp.resetCamera();
 


sedghi:
Interesting, 
So changing the canvas width and height will force a clear on it (https://www.w3.org/TR/2010/WD-html5-20100624/the-canvas-element.html#the-canvas-element)

```
When the [canvas](https://www.w3.org/TR/2010/WD-html5-20100624/the-canvas-element.html#the-canvas-element) element is created, and subsequently whenever the [width](https://www.w3.org/TR/2010/WD-html5-20100624/the-canvas-element.html#attr-canvas-width) and [height](https://www.w3.org/TR/2010/WD-html5-20100624/the-canvas-element.html#attr-canvas-height) attributes are set (whether to a new value or to the previous value), the bitmap and any associated contexts must be cleared back to their initial state and reinitialized with the newly specified coordinate space dimensions.
``` 

So I'm pretty sure in OHIF or consuming apps, collapsing the side panel would show a huge ugly white canvas for each viewport.

Have you tried it in OHIF while collapsing sidebars and see what happens?

Ouwen:
Hmmmm, this wasn't the behavior I saw when resizing window. But I can try a side panel resize to see what happens.

Ouwen:
Here is with and without resize videos:

[OHIF Viewer without resize.webm](https://user-images.githubusercontent.com/5455421/200096547-ae9d0cb2-f917-48de-b6fd-efd0e1000db2.webm)
[OHIF Viewer with resize.webm](https://user-images.githubusercontent.com/5455421/200096550-32ca5dc4-aa24-43d4-8d43-7e45ebbd6e22.webm)

sedghi (approved):
I tested this and It works great. Thanks

PR #280, opened 2022-11-04
feat: added display area to viewport

Ouwen:
Adds initial display area to viewport.

`imageArea` takes a `imageX` and `imageY` which should be a number (0,1]  which determines the % of image to display.
`imageFocalPoint` takes a `focalX` and `focalY` which determines the camera focal point. `focalX=1` and `focalY=1` means the image is in the bottom right corner offscreen.  `focalX=0` and `focalY=0` refers to the upper left corner offscreen and `focalX=0.5` and `focalY=0.5` is the default center image. These numbers represent the XY translation amount as a % of the image width/height

The displayArea will save the stored camera and functions properly for stack viewport. The display area can also be programmatically set with or without saving the stored camera. One use case for programmatic display area setting is to automatically zoom in on a ROI.

Implementation note: setPan is not available despite the existence of imageData and vtkRenderer. It seems the off screen renderer requires some initial imageData render before the `displayToWorld` function is available to the viewer instance. Since the setPan function requires `canvasToWorld` which requires `displayToWorld` we do not use `setPan` in the implementation of setDisplayArea. Instead we use the imageData.indexToWorld function is used instead to calculate the delta vector which the camera should move. This delta vector is doubled so the image can translate completely out of the viewport rather than just 50% (since radius is calculated to fit image).

### DIFF
--- a/common/reviews/api/streaming-image-volume-loader.api.md
+++ b/common/reviews/api/streaming-image-volume-loader.api.md
@@ -403,6 +403,27 @@ interface CustomEvent_2<T = any> extends Event {
     ): void;
 }
 
+// @public (undocumented)
+type DisplayArea = {
+    imageArea: [number, number]; // areaX, areaY
+    imageCanvasPoint: {
+        imagePoint: [number, number]; // imageX, imageY
+        canvasPoint: [number, number]; // canvasX, canvasY
+    };
+    storeAsInitialCamera: boolean;
+};
+
+// @public
+type DisplayAreaModifiedEvent = CustomEvent_2<DisplayAreaModifiedEventDetail>;
+
+// @public
+type DisplayAreaModifiedEventDetail = {
+    viewportId: string;
+    displayArea: DisplayArea;
+    volumeId?: string;
+    storeAsInitialCamera?: boolean;
+};
+
 // @public
 enum DynamicOperatorType {
     AVERAGE = 'AVERAGE',
@@ -436,6 +457,7 @@ enum Events {
     CAMERA_MODIFIED = 'CORNERSTONE_CAMERA_MODIFIED',
 
     CAMERA_RESET = 'CORNERSTONE_CAMERA_RESET',
+    DISPLAY_AREA_MODIFIED = 'CORNERSTONE_DISPLAY_AREA_MODIFIED',
     ELEMENT_DISABLED = 'CORNERSTONE_ELEMENT_DISABLED',
     ELEMENT_ENABLED = 'CORNERSTONE_ELEMENT_ENABLED',
     GEOMETRY_CACHE_GEOMETRY_ADDED = 'CORNERSTONE_GEOMETRY_CACHE_GEOMETRY_ADDED',
@@ -478,6 +500,8 @@ declare namespace EventTypes {
         CameraModifiedEvent,
         VoiModifiedEvent,
         VoiModifiedEventDetail,
+        DisplayAreaModifiedEvent,
+        DisplayAreaModifiedEventDetail,
         ElementDisabledEvent,
         ElementDisabledEventDetail,
         ElementEnabledEvent,
@@ -1078,6 +1102,7 @@ interface IViewport {
     // (undocumented)
     _getCorners(bounds: Array<number>): Array<number>[];
     getDefaultActor(): ActorEntry;
+    getDisplayArea(): DisplayArea | undefined;
     getFrameOfReferenceUID: () => string;
     getPan(): Point2;
     getRenderer(): void;
@@ -1094,6 +1119,11 @@ interface IViewport {
     reset(immediate: boolean): void;
     setActors(actors: Array<ActorEntry>): void;
     setCamera(cameraInterface: ICamera, storeAsInitialCamera?: boolean): void;
+    setDisplayArea(
+    displayArea: DisplayArea,
+    callResetCamera?: boolean,
+    suppressEvents?: boolean
+    );
     setOptions(options: ViewportInputOptions, immediate: boolean): void;
     setPan(pan: Point2, storeAsInitialCamera?: boolean);
     setZoom(zoom: number, storeAsInitialCamera?: boolean);
@@ -1438,6 +1468,7 @@ type TransformMatrix2D = [number, number, number, number, number, number];
 type ViewportInputOptions = {
     background?: [number, number, number];
     orientation?: OrientationAxis | OrientationVectors;
+    displayArea?: DisplayArea;
     suppressEvents?: boolean;
     parallelProjection?: boolean;
 };
--- a/packages/core/examples/programaticPanZoom/index.ts
+++ b/packages/core/examples/programaticPanZoom/index.ts
@@ -21,6 +21,9 @@ const { ViewportType } = Enums;
 const renderingEngineId = 'myRenderingEngine';
 const viewportId = 'CT_STACK';
 
+function getRand(min, max) {
+  return Math.random() * (max - min) + min;
+}
 // ======== Set up page ======== //
 setTitleAndDescription(
   'Programmatic Pan and Zoom with initial pan and zoom',
@@ -30,6 +33,7 @@ setTitleAndDescription(
 const content = document.getElementById('content');
 const element = document.createElement('div');
 element.id = 'cornerstone-element';
+
 element.style.width = '500px';
 element.style.height = '500px';
 
@@ -65,7 +69,7 @@ addButtonToToolbar({
     );
 
     const zoom = viewport.getZoom();
-    console.log('Current zoom', zoom);
+
     viewport.setZoom(zoom * 1.05);
     viewport.render();
   },
@@ -81,9 +85,7 @@ addButtonToToolbar({
     const viewport = <Types.IVolumeViewport>(
       renderingEngine.getViewport(viewportId)
     );
-    //viewport.resetCamera();
-    viewport.setZoom(1);
-    viewport.setPan([0, 0]);
+    viewport.resetCamera();
     viewport.render();
   },
 });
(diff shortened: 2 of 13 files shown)


packages/core/src/RenderingEngine/Viewport.ts
```diff
@@ -147,12 +152,16 @@ class Viewport implements IViewport {
    * @param options - The viewport options to set.
    * @param immediate - If `true`, renders the viewport after the options are set.
    */
-  public setOptions(options: ViewportInputOptions, immediate = false): void {
+  public setOptions(
+    options: ViewportInputOptions,
+    immediate = false,
+    storeAsInitialCamera = false
```
sedghi, line 158: can we avoid having `storeAsInitialCamera` here?

packages/core/src/RenderingEngine/Viewport.ts
```diff
@@ -697,6 +796,16 @@ class Viewport implements IViewport {
...
+  /**
+   * Sets the provided camera as the displayArea camera.
+   * This allows computing differences applied later as compared to the initial
+   * position, for things like zoom and pan.
+   * @param camera - to store as the initial value.
+   */
+  protected setInitialDisplayAreaCamera(camera: ICamera): void {
+    this.initialDisplayAreaCamera = camera;
+  }
+
```
sedghi, line 808: We already have the `initialCamera` that I hate very much, why we are adding this new initialDisplayCamera too?

sedghi:
1. `focalX=1 and focalY=1 means the image is in the bottom right corner offscreen. ` what do you mean? which part of image is in the right corner? center of the image is in the bottom right?

2. Is there any reason you are not using just a flat object for displayedArea? {imageX, imageY, focalX, focalY}

3. Having focal point at the surface api mean that we are assuming the user knows the focal point is the center of the image, can we rename it to something else from `focalX` to something else (like `offsetX` or etc.)? Also I guess `imageX` does not represent that is %. We used to have topLeftHandCorner but I guess the benefit of % is more.

Let me know what you think, then I can review another round

Ouwen:
Hey Alireza, thanks for the review! 

1. focalX=1, focalY=1 will set the top left corner of the image at the bottom right of the screen,and the image will not be visible on canvas. focalX=0, focalY=0 will set the bottom right corner of the image to the top left of the screen, and the image will not be visible on canvas.

2. We can make it flat

3. Yeah there might be a better name possible, effectively it is what % the center of the image will be on screen. Ideally we can use the pan functionality, but it seems like at constructor time the vtkoffsetrenderer is not initialized.

packages/core/src/RenderingEngine/Viewport.ts
```diff
@@ -524,6 +533,85 @@ class Viewport implements IViewport {
...
+    storeAsInitialCamera = false,
+    suppressEvents = false
+  ): void {
+    this.setCamera(this.initialDisplayAreaCamera, false); // ensure displayArea is absolute
+
+    const {
+      imageArea = { areaX: 1, areaY: 1 },
+      imageFocalPoint = { focalX: 0.5, focalY: 0.5 },
+    } = displayArea;
+    // setPan functions requires offsetviewer to be initialized, which does not seem to be the case at constructor time
```
Ouwen, line 557: The most ideal is if we can start this initial display logic when the vtkoffset renderer is able to perform its canvas to world calculations. For some reason these pop up as NaN during constructor time

packages/core/src/RenderingEngine/Viewport.ts
Ouwen: The initialCamera is the position the user will reset to
The initialDisplayCamera is the position prior to applying the initialDisplay (this allows intialDisplay settings to be absolute, otherwise subsequent applications will cause another zoom and another pan on top of what has already applied)

packages/core/src/RenderingEngine/Viewport.ts
Ouwen: Yes, this could move into the options.DisplayArea

sedghi:
> focalX=1, focalY=1 will set the top left corner of the image at the bottom right of the screen,and the image will not be visible on canvas. focalX=0, focalY=0 will set the bottom right corner of the image to the top left of the screen, and the image will not be visible on canvas.

This is too confusing for an api. Why it switches the definition from top left (in 1, 1) to bottom right in (0,0)?

Ouwen:
After debugging, I found the issue with the canvasToWorld. It seems reset camera is called on `addVtkjsDrivenViewport`
Then another reset camera is called on `addActors` since vtkImageData is not present. On the third call which reuses the vtkImageData object in stackviewport, is `canvasToWorld` functional

packages/core/src/types/displayArea.ts
```diff
@@ -0,0 +1,16 @@
...
+};
+
+type ImageFocalPoint = {
+  focalX: number;
+  focalY: number;
+};
+
+type DisplayArea = {
+  imageArea: InitialDisplayArea;
+  imageFocalPoint: ImageFocalPoint;
```
wayfarer3130, line 13: This doesn't quite work without a second focal point -  you need to be able to handle the following cases:
1. Center of the image as a center point - WITHOUT knowing about the size of the display area OR the size of the image
2. Left image aligned or Right image aligned
3. Zoomed to specific point in image as center of displayed area.

For #1, using fractional points as the DICOMweb requests do for displayed area makes sense.
For #2, having both an image focal point and a display focal point makes sense, where the display focal point is optional, and default to the center of the image area.
For #3, using both 1,2 you get exactly what you want.

So, what you need is to have the initial display area as a fraction between 0 and 1 (but do NOT constrain it to 1 at the top end, but strictly larger than 0 at the bottom end - in fact larger than 1% makes sense)

The ImageFocalPoint should be renamed FocalPoint, and there should be two of them (could also call it ViewFocalPoint, or if you leave it ImageFocalPoint, then realize it is image area related).
Change DisplayArea to be:
// The focus point of interest for this image
imageFocalPoint?: ImageFocalPoint
// The focus point of interest for this display area.  Defaults if undefined to the center of the display area
displayFocalPoint?: ImageFocalPoint

You might also want to have a units setting - so that you can provide exact zoom levels - the three units of interest would specify the display area in any of:
Scale to Fit -  as a fraction of the original image size
Zoom - as a zoom scale (independent x,y) - 1.0 means 1 image pixel displays as 1 screen pixel
TrueSize - as a multiple of the true size, displayed on screen, eg 1.0 means 1 mm in body displays as 1 mm on screen

AND, either areaX or areaY should be nullable so that you can specify one exactly, without consideration for how the OTHER one fits.

I'm well aware this is much harder, but it does a lot of good for how to apply settings consistently, before the image area is actually known.   If you KNOW the actual sizes of images, it is easy to set them in fractions, but if you don't know, then there isn't a good way to know what values to use up front.

packages/core/src/RenderingEngine/Viewport.ts
```diff
@@ -557,40 +557,11 @@ class Viewport implements IViewport {
...
-          ...previousCamera,
-          focalPoint: newFocal as Point3,
-          position: newPosition as Point3,
-        },
-        storeAsInitialCamera
-      );
-    }
+
+    this.setPan(
+      [imageFocalPoint.focalX, imageFocalPoint.focalY],
```
Ouwen, line 562: @diegohennrich, this will require a diff calculation. We need to get the current camera focalX and focalY and subtract it from the imageFocalPoint X and Y. I believe the setPan requires a delta input, not an absolute input value.

Ouwen:
@sedghi @wayfarer3130 the PR is ready for review

common/reviews/api/streaming-image-volume-loader.api.md
```diff
@@ -426,6 +447,7 @@ enum Events {
     CAMERA_MODIFIED = 'CORNERSTONE_CAMERA_MODIFIED',
 
     CAMERA_RESET = 'CORNERSTONE_CAMERA_RESET',
+    DISPLAY_AREA_MODIFIED = 'CORNERSTONE_DISPLAY_AREA_MODIFIED',
```
wayfarer3130, line 450: I don't think this should have it's own even - there should be a way to get properties changed, or this should trigger camera modified events.  @sedghi - any thoughts about the event?  Other properties don't have their own events.

common/reviews/api/streaming-image-volume-loader.api.md
```diff
@@ -1078,6 +1103,11 @@ interface IViewport {
     reset(immediate: boolean): void;
     setActors(actors: Array<ActorEntry>): void;
     setCamera(cameraInterface: ICamera, storeAsInitialCamera?: boolean): void;
+    setDisplayArea(
```
wayfarer3130, line 1106: When there are more than one option, the options should go into a typed options property so that the number of arguments doesn't change.

packages/core/src/RenderingEngine/Viewport.ts
```diff
@@ -527,6 +532,78 @@ class Viewport implements IViewport {
...
+   * the volume (or slice); otherwise, only the camera zoom and camera Pan or Zoom
+   * is reset for the current view.
+   * @param displayArea - The display area of interest.
+   * @param suppressEvents - If true, don't fire displayArea event.
+   */
+  public setDisplayArea(
+    displayArea: DisplayArea,
+    suppressEvents = false
+  ): void {
+    const storeAsInitialCamera = displayArea.storeAsInitialCamera;
```
wayfarer3130, line 547: const { storeAsInitialCamera } = displayArea;

packages/core/src/RenderingEngine/Viewport.ts
```diff
@@ -527,6 +532,78 @@ class Viewport implements IViewport {
...
+    this.setCamera(this.fitToCanvasCamera, false);
+
+    const { imageArea, imageCanvasPoint } = displayArea;
+
+    if (imageArea) {
+      const { areaX, areaY } = imageArea;
+      const zoom = Math.min(this.getZoom() / areaX, this.getZoom() / areaY);
```
wayfarer3130, line 554: The correct calculation for this is:
zoomWidth = areaX * canvas.width / image.width
zoomHeight = areaY * canvas.width / image.height
setZoom( getZoom() / min(zoomWidth,zoomHeight) ) 
assuming zoom fits the larger of the width and height.

packages/core/src/enums/Events.ts
```diff
@@ -38,6 +38,13 @@ enum Events {
    * and see what event detail is included in {@link EventTypes.VoiModifiedEventDetail | VoiModified Event Detail }
    */
   VOI_MODIFIED = 'CORNERSTONE_VOI_MODIFIED',
+  /**
+   * Triggers on the HTML element when viewport modifies its display area
+   *
+   * Make use of {@link EventTypes.DisplayAreaModifiedEvent | DisplayAreaModified Event Type } for typing your event listeners for DISPLAY_AREA_MODIFIED event,
+   * and see what event detail is included in {@link EventTypes.DisplayAreaModifiedEventDetail | DisplayAreaModified Event Detail }
+   */
+  DISPLAY_AREA_MODIFIED = 'CORNERSTONE_DISPLAY_AREA_MODIFIED',
```
wayfarer3130, line 47: What about calling it PROPERTIES_MODIFIED, and fire it when any property is modified, including the properties updated?

packages/core/src/types/displayArea.ts
```diff
@@ -0,0 +1,25 @@
+type InitialDisplayArea = {
```
wayfarer3130, line 1: These should all be two element arrays for consistency.

wayfarer3130 (changes requested):
There are a few issues to update here.

packages/core/examples/programaticPanZoom/index.ts
```diff
@@ -104,6 +106,161 @@ addButtonToToolbar({
   },
 });
 
+addButtonToToolbar({
```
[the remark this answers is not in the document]
Ouwen: thanks!

packages/core/src/enums/Events.ts
Ouwen: Maybe this can be a separate PR to also group up camera, WW/WL events?

packages/core/src/types/displayArea.ts
Ouwen: this change is now made

packages/core/src/RenderingEngine/Viewport.ts
Ouwen: So what I do is I treat the fitToCanvas camera as the zoom=1 case for the DisplayArea API. Then I increase or decrease the factor relative to the fitToCanvas camera. The actual zoom stored is still absolute (since the fitToCanvas camera calculates absolute camera). However, the DisplayArea API is all relative to what "fitToCanvas" is defined by, not an absolute coordinate system.

Ouwen:
@wayfarer3130, let me know what you think of these changes!

sedghi:
Could we add some documentation on what is what exactly with the examples you have on the demo (Top, etc.) in here as well please? https://www.cornerstonejs.org/docs/concepts/cornerstone-core/viewports

Ouwen:
Sure I can add some documentation on this

packages/docs/docs/concepts/cornerstone-core/viewports.md
```diff
@@ -31,16 +31,52 @@ viewed via 4 viewports in a “4-up” view: Axial MPR, Sagittal MPR, Coronal MP
...
+  },
+  storeAsInitialCamera: boolean,
+};
+```
+
+Zoom and pan are all relative to the initial "fit to screen" view.
+
+In order to zoom into the image 200%, we would set the `imageArea` to [0.5, 0.5].
+
+Panning is controlled by a provided `imagePoint` and a provided `canvasPoint`. You can imagine the canvas as a sheet of white paper and the image as another sheet of paper like a chest x-ray. Mark a point in the canvas paper with a pen and then mark another point on your chest x-ray image. Now try to "pan" your image so the point so the `imagePoint` matches the
```
sedghi, line 67: Great docs indeed

sedghi (approved):

wayfarer3130 (approved):

PR #288, opened 2022-11-09
fix: ZoomTool fix for imageData spacing for PolyData files

NeilMacPhee:
### Fix

-  Added spacing parameter presets of [0.5, 0.5, 0.5] for cases where the viewport does not have imageData, such as PolyData where actors are added directly to viewport

### DIFF
--- a/packages/core/CHANGELOG.md
+++ b/packages/core/CHANGELOG.md
@@ -3,6 +3,17 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+# [0.22.0](https://github.com/cornerstonejs/cornerstone3D-beta/compare/@cornerstonejs/core@0.21.5...@cornerstonejs/core@0.22.0) (2022-11-23)
+
+
+### Features
+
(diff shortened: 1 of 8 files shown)


packages/tools/src/tools/ZoomTool.ts
```diff
@@ -151,9 +151,13 @@ class ZoomTool extends BaseTool {
 
     // If it is a regular GPU accelerated viewport, then parallel scale
     // has a physical meaning and we can use that to determine the threshold
-    const imageData = viewport.getImageData();
+    // Added spacing preset in case there is no imageData on viewport
+    let spacing = [0.5, 0.5, 0.5];
+    if (viewport.getImageData()) {
```
sedghi, line 156: You can say `const imageData = viewport.getImageData()` and check on imageData instead of twice calling the `viewport.geteImageData()`

sedghi (commented):
Why [0.5, 0.5, 0.5]? Why not 1, 1, 1 

There is a weird jump after I start the drag for the zoom see below


[5eddcf58-9d0d-4848-bafc-fa70d6029b80.webm](https://user-images.githubusercontent.com/7490180/200948470-16cc73d5-24cb-450e-bf28-28e8f6dfd12a.webm)

PR #291, opened 2022-11-10
feat: add stack synchronization within or across studies

sedghi:
Add stack synchronization for stack viewports. This works for viewports in the same frame of reference and viewports that are not in the same frame of reference 

Make the synchronizer callbacks async

### DIFF
--- a/packages/tools/src/synchronizers/callbacks/stackImageSyncCallback.ts
+++ b/packages/tools/src/synchronizers/callbacks/stackImageSyncCallback.ts
@@ -1,0 +1,153 @@
+import { vec3 } from 'gl-matrix';
+import {
+  getRenderingEngine,
+  Types,
+  metaData,
+  utilities,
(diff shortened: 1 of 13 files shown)


packages/tools/src/synchronizers/callbacks/stackImageSyncCallback.ts
```diff
@@ -0,0 +1,153 @@
...
+    const closestImageIdIndex = _getClosestImageIdIndex(
+      sourceImagePositionPatient,
+      targetImageIds
+    );
+
+    if (
+      closestImageIdIndex.index !== -1 &&
+      tViewport.getCurrentImageIdIndex() !== closestImageIdIndex.index
+    ) {
+      // await tViewport.setImageIdIndex(closestImageIdIndex.index);
```
swederik, line 81: Remove

swederik (approved):
Lgtm

PR #294, opened 2022-11-13
fix: get correct imageData with targetId in BaseTool

DMIAOCHEN:
when multiple viewport display same stack data at the same time,   `getTargetIdImage` method get viewport by targetId  without judging whether the first viewport is what we want. then i found `calculateCachedStats` in `EllipticalROITool` will failed.


sedghi (approved):
This looks good to me, thanks

PR #296, opened 2022-11-14
feat: trigger annotation modified event on textbox moving

### LINKED ISSUE #274: Measurement textbox update event missing
In the old `cornerstoneTools`, when a measurement textbox was moved, a `MEASUREMENT_MODIFIED` event would be emitted. However, it seems that the current logic fails to emit an `ANNOTATION_MODIFIED` event when a measurement textbox is updated

m00n620:
This should close #274 , set invalidated attribute as true on textbox moving

### DIFF
--- a/packages/tools/src/tools/annotation/AngleTool.ts
+++ b/packages/tools/src/tools/annotation/AngleTool.ts
@@ -400,6 +400,7 @@ class AngleTool extends AnnotationTool {
       worldPosition[2] += worldPosDelta[2];
 
       textBox.hasMoved = true;
+      annotation.invalidated = true;
     } else if (handleIndex === undefined) {
       // Drag mode - moving handle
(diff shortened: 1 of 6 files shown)


sedghi:
Looking quickly at this, we don't want annotation calculateCachedStats to be re-calculated everytime textBox moves, that is not efficient at all

m00n620:
> Looking quickly at this, we don't want annotation calculateCachedStats to be re-calculated everytime textBox moves, that is not efficient at all

Is it fine for you to do it on mouse up event?

m00n620:
@sedghi, please take a look again. Events are being triggered on mouse up callback right now.

sedghi:
I don't understand how does `movingTextBox` is true when mouse up. I don't think that fixes your problem

PR #298, opened 2022-11-15
add ScaleOverlayTool

tineke01:


sedghi:
I thought you said you have the old implementation

tineke01:
> I thought you said you have the old implementation

[https://github.com/cornerstonejs/cornerstoneTools/blob/master/src/tools/ScaleOverlayTool.js](url)

sedghi:
The most important part is the renderToolData. It needs to look at the viewport current camera, and calculates the scale and render it. I recommend start by comparing the implementation of a legacy tool (Length) and the Length tool in 3D

PR #301, opened 2022-11-17
don't reset display pipeline when spacing is missing

bionoone:
handle the case when image has no pixel spacing and  image data use default spacing.


sedghi (approved):
Looks great thanks

sedghi (approved):

PR #302, opened 2022-11-17
fix(conversion):World to image throws an exception on outside image

wayfarer3130:
There are lots of cases where world to image should allow producing coordinates outside the image.  Make the exception optional.

### DIFF
--- a/packages/core/src/utilities/worldToImageCoords.ts
+++ b/packages/core/src/utilities/worldToImageCoords.ts
@@ -58,17 +58,6 @@ function worldToImageCoords(
     columnDistance / columnPixelSpacing,
   ];
 
-  if (
-    imageCoords[0] < 0 ||
-    imageCoords[0] >= columns ||
-    imageCoords[1] < 0 ||
-    imageCoords[1] >= rows
-  ) {
-    throw new Error(
-      `The image coordinates are outside of the image, imageCoords: ${imageCoords}`
-    );
-  }
-
   return imageCoords as Point2;
 }
 


packages/core/src/utilities/worldToImageCoords.ts
```diff
@@ -59,10 +63,11 @@ function worldToImageCoords(
   ];
 
   if (
-    imageCoords[0] < 0 ||
-    imageCoords[0] >= columns ||
-    imageCoords[1] < 0 ||
-    imageCoords[1] >= rows
+    checkCoordinateSpace &&
+    (imageCoords[0] < 0 ||
+      imageCoords[0] > columns ||
```
wayfarer3130, line 68: The image coordinates have to be strictly larger than columns,rows as the last pixel is columns-1...columns inclusive.

packages/core/src/utilities/worldToImageCoords.ts
```diff
@@ -59,10 +63,11 @@ function worldToImageCoords(
   ];
 
   if (
-    imageCoords[0] < 0 ||
-    imageCoords[0] >= columns ||
-    imageCoords[1] < 0 ||
-    imageCoords[1] >= rows
+    checkCoordinateSpace &&
```
wayfarer3130, line 66: Only check coordinates if requested to do so - normally that isn't a useful thing to do inside the method, as there are quite a few cases where at least one coordinate can usefully be outside by nearby the image.

sedghi:
Hmmm, looking at it again, I guess it shouldn't throw an error at all. Maybe we just remove that check fully?

sedghi (approved):

PR #303, opened 2022-11-19
Fix/crosshairs annotations

sedghi:
- Make Crosshairs to remove their annotations upon disabled
- Make annotation rendering engine to reset its rendering flags when a viewport is removed

### DIFF
--- a/packages/core/src/RenderingEngine/RenderingEngine.ts
+++ b/packages/core/src/RenderingEngine/RenderingEngine.ts
@@ -210,6 +210,7 @@ class RenderingEngine implements IRenderingEngine {
 
     // 5. Remove the requested viewport from the rendering engine
     this._removeViewport(viewportId);
+    viewport.hasBeenDisabled = true;
 
     // 6. Avoid rendering for the disabled viewport
     this._needsRender.delete(viewportId);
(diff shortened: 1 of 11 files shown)


packages/core/src/RenderingEngine/RenderingEngine.ts
```diff
@@ -210,6 +210,7 @@ class RenderingEngine implements IRenderingEngine {
 
     // 5. Remove the requested viewport from the rendering engine
     this._removeViewport(viewportId);
+    viewport.hasBeenDisabled = true;
```
wayfarer3130, line 213: The naming here is very odd, and I think it ends up not being quite right.  hasBeenDisabled doesn't mean it is currently disabled, but suggests that at some point in the past it got disabled.   Is it safe to say:
viewport.isDisabled = true 
or
viewport.disabled = true
Otherwise, what about making it more prescriptive, something like disableOverlayDrawing, so that it is specifying that no overlays are to be drawn this viewport, and that would include things like the crosshairs, but may also include things like segmentation overlays etc.

sedghi: Makes sense

wayfarer3130 (commented):
Looks good now.

PR #304, opened 2022-11-23
Fix/stack flip

sedghi:


packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -1666,9 +1664,11 @@ class StackViewport extends Viewport implements IStackViewport {
       // it in the space 3) restore the pan, zoom props.
       const cameraProps = this.getCamera();
 
-      this.panCache[0] = this.cameraPosOnRender[0] - cameraProps.position[0];
-      this.panCache[1] = this.cameraPosOnRender[1] - cameraProps.position[1];
-      this.panCache[2] = this.cameraPosOnRender[2] - cameraProps.position[2];
+      const panCache: Point3 = [
```
wayfarer3130, line 1667: This should be doable with a single line Mat3 operation - I think
const panCache = Mat3.subtract(Mat3.create(), this.cameraFocalPointOnRender, cameraProps.focalPoint)
It is too easy to get wrong when writing it all out by hand every time, would prefer to see it use the methods.

packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -1977,26 +1984,29 @@ class StackViewport extends Viewport implements IStackViewport {
...
-    previousCamera: ICamera
+    previousCamera: ICamera,
+    panCache: Point3
   ): void {
     const renderer = this.getRenderer();
 
     // get the focalPoint and position after the reset
     const { position, focalPoint } = this.getCamera();
 
     const newPosition = <Point3>[
```
wayfarer3130, line 1995: Ditto, use the vec3 methods

packages/core/src/RenderingEngine/Viewport.ts
```diff
@@ -208,17 +209,19 @@ class Viewport implements IViewport {
...
     const dimensions = imageData.getDimensions();
     const middleIJK = dimensions.map((d) => Math.floor(d / 2));
 
+    const centeredFocalPoint = vec3.create();
     const idx = [middleIJK[0], middleIJK[1], middleIJK[2]];
-    imageData.indexToWorld(idx, resetFocalPoint);
+    imageData.indexToWorld(idx, centeredFocalPoint);
+
+    const resetFocalPoint = this._getFocalPointForResetCamera(
+      centeredFocalPoint as Point3,
```
wayfarer3130, line 220: You don't need the extra variable - just use vec3.create() inline here.
Also, consider making the definition for Point3 as being identical to the vec3 one so that no casting is required.  I know that means you can inadvertently pass a wrong length FloatArray32, but anything using regular arrays will be type checked, and the others are mostly created by the vec3 functions and as cast, so it doesn't decrease type testing.

wayfarer3130 (changes requested):
Few small changes

packages/core/src/RenderingEngine/Viewport.ts
sedghi: I always forget that :D

PR #313, opened 2022-11-25
Fix/htj2k and keymodifier

wayfarer3130:
Fix HTJ2K loading in the streaming volume image loader.
Also, fix the key modifier so that it actually works on mouse event information

### DIFF
--- a/packages/tools/src/eventDispatchers/shared/getMouseModifier.ts
+++ b/packages/tools/src/eventDispatchers/shared/getMouseModifier.ts
@@ -5,14 +5,26 @@ import { KeyboardBindings as kb } from '../../enums';
  * Supports Shift, Ctrl, Alt, in singly and in combinations of 2
  * Supports Meta singly.
  */
-const getMouseModifierKey = (evt) =>
-  (evt.shiftKey &&
-    ((evt.ctrlKey && kb.ShiftCtrl) ||
-      (evt.altKey && kb.ShiftAlt) ||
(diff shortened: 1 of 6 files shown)


packages/tools/src/eventDispatchers/shared/getMouseModifier.ts
```diff
@@ -0,0 +1,18 @@
+import { KeyboardBindings as kb } from '../../enums';
+
+/**
+ * Gets the mouse modifier key from a mouse event.
+ * Supports Shift, Ctrl, Alt, in singly and in combinations of 2
+ * Supports Meta singly.
+ */
+const getMouseModifierKey = (evt) =>
```
sedghi, line 8: The logic below is too cumbersome and I don't understand what it is doing. Can you please add more docs? So the evt is mouseEvent, it sees if the shiftKey is pressed with mouse? while ctrl and shiftCtrol is there OR alt or shiftAlt, I'm just lost

sedghi (commented):
Can you please add more docs to the getMouseModifier, I don't understand a bit of that function. Otherwise looks good

packages/tools/src/eventDispatchers/shared/getMouseModifier.ts
wayfarer3130: Redid the logic - it is the same logic, just easier to understand as I broke it out.

sedghi (approved):
Looks great

PR #315, opened 2022-11-26
Filter PlanarFreeHandROI annotations based on parallel normals instead of equal normals.

### LINKED ISSUE #314: PlanarFreeHandROI incorrect filtering of annotations within slice
As discussed on slack, it seems that the `PlanarFreeHandROI` tool has its own `filterAnnotationsWithinSlice` which is only keeping annotations that have the same `viewPlaneNormal`. The correct implementation is to filter annotations whose view plane normal is not parallel (as implemented in `utilities/planar/filterAnnotationsWithinSlice`).

ramonemiliani93:
Closes #314.


sedghi (approved):
Thanks!

PR #323, opened 2022-12-01
fix(volumeViewport): Add optional scaling as the volume can be undefined

gabriellebaudy:
While trying to get the volume from the cache, it can be undefined so getting the scaling attribute would throw an error in that case. This is a quick fix


sedghi (approved):
Thanks

PR #327, opened 2022-12-05
feat: enable having multiple instances of the same tool and add more seg tools

sedghi:
## Try
https://deploy-preview-327--cornerstone-3d-docs.netlify.app/live-examples/basicsegmentationtools

## Multiple Tool Instances
There are various scenarios that we want to use the same tool but with different strategy (segmentation brush can be threshold brush, eraser, simple brush, sphere brush etc.). 

```
public addToolInstance(
    toolName: string,
    parentClassName: string,
    configuration = {}
  ): void {
```

## New Strategies
- Erase Circle 
- Erase Sphere

## New Tool
- PaintFill Tool

PS: Kudos to @JamesAPetts for this awesome piece of work

### DIFF
--- a/packages/tools/examples/basicSegmentationTools/index.ts
+++ b/packages/tools/examples/basicSegmentationTools/index.ts
@@ -10,6 +10,7 @@ import {
   createImageIdsAndCacheMetaData,
   setTitleAndDescription,
   addDropdownToToolbar,
+  addSliderToToolbar,
   setCtTransferFunctionForVolumeActor,
 } from '../../../../utils/demo/helpers';
 import * as cornerstoneTools from '@cornerstonejs/tools';
@@ -28,13 +29,16 @@ const {
   SphereScissorsTool,
   CircleScissorsTool,
   BrushTool,
+  PaintFillTool,
   PanTool,
   ZoomTool,
   StackScrollMouseWheelTool,
+  utilities: cstUtils,
 } = cornerstoneTools;
 
 const { MouseBindings } = csToolsEnums;
 const { ViewportType } = Enums;
+const { segmentation: segmentationUtils } = cstUtils;
 
 // Define a unique id for the volume
 const volumeName = 'CT_VOLUME_ID'; // Id of the volume less loader prefix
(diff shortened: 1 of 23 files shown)


packages/tools/src/eventDispatchers/keyboardEventHandlers/keyDown.ts
```diff
@@ -23,7 +23,8 @@ export default function keyDown(evt: KeyDownEventType): void {
   );
 
   const toolName = activeTool.getToolName();
+  const toolInstanceName = activeTool.instanceName;
```
wayfarer3130, line 26: Is it necessary to pass both attributes?  The viewports cursor should have all cursors necessary, registered multiple times if needed.

packages/tools/src/eventDispatchers/shared/customCallbackHandler.ts
```diff
@@ -37,37 +38,43 @@ export default function customCallbackHandler(
...
   }
 
   activeTool[customFunction](evt);
 }
+
+// TODO: Filter tools by interaction type?
+/**
+ * Iterate tool group tools until we find a tool that is:
+ * - active
+ * - has the custom callback function
```
wayfarer3130, line 54: How about:
- has an attribute with key 'customFunctionName' and whose value is a function.

I just was expecting customFunction to be a function itself, and it was weird that it wasn't.  Renaming it to customFunctionName and indicating how it works is useful.

packages/tools/src/eventDispatchers/shared/getActiveToolForKeyboardEvent.ts
```diff
@@ -37,22 +37,29 @@ export default function getActiveToolForKeyboardEvent(
...
-      toolOptions.bindings.length &&
-      toolOptions.bindings.some(
-        (binding) =>
-          binding.mouseButton === (mouseButton ?? MouseBindings.Primary) &&
-          binding.modifierKey === modifierKey
-      );
-
-    if (toolOptions.mode === Active && correctBinding) {
-      return toolGroup.getToolInstance(toolName);
+  for (let i = 0; i < toolGroupToolNames.length; i++) {
```
wayfarer3130, line 40: Consider refactoring this into a method on ToolGroupManager that finds a tool having a given function return true, and have it take as parameters the tool name, the instance name and the options for that tool.  Then, just use that for the getActiveTool...  - could make it either as a finder or a getAll method.  Then, the actual check method becomes fairly easy to write.

sedghi:
@wayfarer3130 I applied your review

packages/tools/examples/basicSegmentationTools/index.ts
```diff
@@ -170,10 +237,45 @@ async function run() {
   toolGroup.addTool(RectangleScissorsTool.toolName);
   toolGroup.addTool(CircleScissorsTool.toolName);
   toolGroup.addTool(SphereScissorsTool.toolName);
-  toolGroup.addTool(BrushTool.toolName);
+  toolGroup.addTool(PaintFillTool.toolName);
+  toolGroup.addToolInstance(
+    brushInstanceNames.CircularBrush,
```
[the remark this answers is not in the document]
sedghi: Thank you for the great suggestion

PR #335, opened 2022-12-14
Add coplanar check in stackImageSync callback

rodrigobasilio2022:
In order to protect the stackImageSyncCallback from image sync problems, I suggest add an extra test that checks if the two viewports being considered has the same orientation. 
I add an extra function coPlanarViewports that checks if the viewPlaneNormal from the two viewports are parallel, by checking if the dot product is greater than 0.9

### DIFF
--- a/packages/tools/src/synchronizers/callbacks/stackImageSyncCallback.ts
+++ b/packages/tools/src/synchronizers/callbacks/stackImageSyncCallback.ts
@@ -7,7 +7,7 @@ import {
 } from '@cornerstonejs/core';
 import { Synchronizer } from '../../store';
 import { jumpToSlice } from '../../utilities';
-
+import coPlanarViewports from './coPlanarViewports';
 /**
  * Synchronizer callback to synchronize the source viewport image to the
  * target viewports closest image in its stack. There are two scenarios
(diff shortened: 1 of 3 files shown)


packages/tools/src/synchronizers/callbacks/stackImageSyncCallback.ts
```diff
@@ -69,21 +69,23 @@ export default async function stackImageSyncCallback(
   if (frameOfReferenceUID1 === frameOfReferenceUID2) {
     // if frames of references are the same we can use the absolute
     // imagePositionPatient to find the closest image in the target viewport's stack
-    const closestImageIdIndex = _getClosestImageIdIndex(
-      sourceImagePositionPatient,
-      targetImageIds
-    );
+    if (coPlanarViewports(sViewport, tViewport)) {
```
wayfarer3130, line 72: @sedghi - the reason we are adding the changes is that the synchronizer can be added in other situations than the tool, such as being useful in a hanging protocol.  As well, if the tool is active, and one switches to a different set of series, then the coplanar check needs to be in the tool itself.

sedghi (approved):

PR #337, opened 2022-12-17
fix: could not access 'index' before initialization

mmcardle:
Hi,

Working on using cornerstone3d in a [Vite](https://vitejs.dev/) project, ran into an issue when running the project.
I think I have narrowed it down to a possibly circular dependency in [packages/tools/src/tools/CrosshairsTool.ts](https://github.com/cornerstonejs/cornerstone3D-beta/blob/main/packages/tools/src/tools/CrosshairsTool.ts) when importing `liangBarksyClip`
Happy to provide an example minimal repository with the error to verify.

Console error
```
cornerstonejs.bd843ac7.js:18089 Uncaught ReferenceError: Cannot access 'index' before initialization
    at cornerstonejs.bd843ac7.js:18089:29
```
with `index` being 
```
...
const { RENDERING_DEFAULTS } = CONSTANTS;
const { liangBarksyClip } = index;
function defaultReferenceLineColor() {
  return "rgb(0, 200, 0)";
}
...
```

### DIFF
--- a/packages/tools/src/tools/CrosshairsTool.ts
+++ b/packages/tools/src/tools/CrosshairsTool.ts
@@ -51,7 +51,6 @@ import { MouseDragEventType } from '../types/EventTypes';
 import { CONSTANTS } from '@cornerstonejs/core';
 
 const { RENDERING_DEFAULTS } = CONSTANTS;
-const { liangBarksyClip } = math.vec2;
 
 // TODO: nested config is weird
 interface ToolConfiguration {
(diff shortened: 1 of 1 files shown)


santomon:
wow this is incredible! I had the same issue and only just now found the same fix independently;

If only I found your PR a little earlier

mmcardle:
I have created an example app with the failure here https://github.com/mmcardle/cornerstone-vite

mmcardle:
> wow this is incredible! I had the same issue and only just now found the same fix independently;
> 
> If only I found your PR a little earlier...
> 
> I wonder what's causing this behaviour? the import itself seems logical to me;
> 
> On a different note, how did you find this out?
> 
> I am just happy to be able to finally work with cornerstone3D in my astro build

I suspect the issue is with Vite and not with cornerstone3d.

To figure out the problem and fix  i used the [Contributing Guide on Linking](https://www.cornerstonejs.org/docs/contribute/linking) to build my own version of cornerstonejs/tools with the fix in this PR.

When i used this my custom build of cornerstone/tools in my Vite project I stopped getting the error.

mmcardle:
> So is this PR fixing your issue with vite?

Yes it does, although I would still like to find the underlying Vite issue as this fix feels like a superficial workaround to delay the import of `liangBarksyClip`.

sedghi:
can you try direct import in that file? `import liangBarksyClip from '....../utilities/math/vec2/liangBarksyClip`

mmcardle:
> can you try direct import in that file? `import liangBarksyClip from '....../utilities/math/vec2/liangBarksyClip`

Yes that also works, probably a better fix to do the import like that?

sedghi (approved):

PR #340, opened 2022-12-22
feat(volumeLoader): no need for streaming-wadors imageLoader anymore since streaming volume loader will use cswil wadors image loader

sedghi:
Depending on https://github.com/cornerstonejs/cornerstoneWADOImageLoader/pull/504

Previously we had a whole new image loader called `SharedArrayBufferImageLoader` for the Streaming Volume Loader, which whole purpose was to not create the [image object](https://github.com/cornerstonejs/cornerstoneWADOImageLoader/blob/master/src/imageLoader/createImage.js#L290-L323) since it don't use it, it only uses the pixel data to fill and replace the texture in GPU. 

Based on the initial profiling `createImage` takes 4ms for each image which is not much but for large number of images it can be a bottleneck

image

having a new option to `skipCreateImage` will make the it 0.17 ms on average; we can achieve the same functionality without any added complexity of new image loaders.


packages/streaming-image-volume-loader/src/StreamingImageVolume.ts
```diff
@@ -251,6 +251,7 @@ export default class StreamingImageVolume extends ImageVolume {
       imageId: string,
       scalingParameters
     ) => {
+      debugger;
```
wayfarer3130, line 254: Please remove

wayfarer3130:
Once the debugger line is removed, LGTM

packages/streaming-image-volume-loader/src/StreamingImageVolume.ts
sedghi: Good catch

PR #342, opened 2022-12-30
fix: AngleTool not working after cancellation

BardiaKh:
AngleTool will be broken after annotation cancellation, because it tries to get the `annotationId` of a non-existing annotation.

Fixed this issue by setting `angleStartedNotYetCompleted` to false upon cancellation.

### DIFF
--- a/packages/tools/src/tools/annotation/AngleTool.ts
+++ b/packages/tools/src/tools/annotation/AngleTool.ts
@@ -462,6 +462,7 @@ class AngleTool extends AnnotationTool {
       }
 
       this.editData = null;
+      this.angleStartedNotYetCompleted = false;
       return annotation.annotationUID;
     }
   };


packages/tools/src/tools/annotation/AngleTool.ts
```diff
@@ -462,6 +462,7 @@ class AngleTool extends AnnotationTool {
       }
 
       this.editData = null;
+      this.angleStartedNotYetCompleted = false;
```
sedghi, line 465: I don't see how this line only is solving the bug you found, maybe not all your source changes are pushed?

BardiaKh: I only meant to change this single line. 
Basically, after calling the cancel event, the user should be able to click on another point in the viewport to start annotating an angle, calling the `addNewAnnotation` method. In line 99 of this script, the tool checks if there are some unfinished angles:

```javascript
if (this.angleStartedNotYetCompleted) {
      return;
    }
```

This check prevents adding new annotations as `this.angleStartedNotYetCompleted` is set to `true` after the drawing starts (line 103). So, we have to set it to `false` in the cancel method to allow for new annotations to be created after that.

sedghi (approved):

sedghi:
Thanks for your contribution

Issue #338, opened 2022-12-20
8-bit colored images are not displayed properly 


Zaid-Safadi:
When using wado-uri or file system to load an ultrasound 8-bit image with with "Photometric Interpretation (0028,0004)" "RGB" and "Samples Per Pixel (0028,0002)" "3" the image rendering is corrupted. 

There are different rendering results using CPU or WebGL, You can try this in the Examples:
1. DICOM P10 from local file system (CPU): image is completely corrupted 
2. DICOM P10 from local file system: Image is rendered correctly but the W/L values are off. Compare this when applying the same WW/WC on other DICOM viewer (desktop or OHIF WADO-RS) 

CPU Example Result:
![image](https://user-images.githubusercontent.com/10813109/208570262-6e3796b4-568d-482c-b24f-fda6cf65abfb.png)

WebGL Example Result:
![image](https://user-images.githubusercontent.com/10813109/208570558-b54fec05-0563-4246-b621-a7afd0e94a9f.png)

Desktop Viewer (notice the non-organ part is black):
![image](https://user-images.githubusercontent.com/10813109/208570654-a0d3d7c7-d29e-41a6-b9f5-c25e600b9b00.png)

Zaid-Safadi:
This is an anonymized dataset that reproduces the issue: https://1drv.ms/u/s!AluyBah_ueAcg8EHRae6kDbLtkR5sA?e=c3kRnR

I checked the cornerstone3d example and it configures cornerstoneWADOImageLoader as recommended in this issue https://github.com/cornerstonejs/cornerstoneWADOImageLoader/issues/498: 
  ```js
cornerstoneWADOImageLoader.configure({
    useWebWorkers: true,
    decodeConfig: {
      convertFloatPixelDataToInt: false,
    },
  });
```

Zaid-Safadi:
I was able to track down the corrupted display in:

> DICOM P10 from local file system (CPU): image is completely corrupted

to the StackViewport useRGBA flag here:
https://github.com/cornerstonejs/cornerstone3D-beta/blob/5605a39b17749f4f1d0bf6f3ee6f5ee9be492be8/packages/core/src/RenderingEngine/StackViewport.ts#L1520

Once I set the useRGBA to true, the image is rendered to match the way it is rendered in point 2:

> DICOM P10 from local file system: Image is rendered correctly but the W/L values are off

@sedghi any idea what could be the problem with RGB <--> RGBA?

[...]

PR #345, opened 2023-01-05
fix: colored images using rgba in CPU rendering #338

Zaid-Safadi:
flip the useRGBA in StackViewport to true in the CPU path as discussed here: https://github.com/cornerstonejs/cornerstone3D-beta/issues/338

### DIFF
--- a/packages/core/src/RenderingEngine/StackViewport.ts
+++ b/packages/core/src/RenderingEngine/StackViewport.ts
@@ -1517,7 +1517,7 @@ class StackViewport extends Viewport implements IStackViewport {
         preScale: {
           enabled: true,
         },
-        useRGBA: false,
+        useRGBA: true,
       };
 
       imageLoadPoolManager.addRequest(


sedghi (changes requested):
default is true, I don't think we need to specify here (please check wado).

Zaid-Safadi:
> default is true, I don't think we need to specify here (please check wado).

I think I see where you default it to true [here ](https://github.com/cornerstonejs/cornerstoneWADOImageLoader/blob/master/src/imageLoader/createImage.js#L100)

Isn't the StackViewport in cs3d? The comment says `true` only required for cs-legacy, sorry maybe I am not understanding the comment very well but it sounds like it should have worked if we set it to false in the StackViewport.

yeah, I can remove it from the options!

```
function createImage(imageId, pixelData, transferSyntax, options = {}) {
  // whether to use RGBA for color images, default true as cs-legacy uses RGBA
  // but we don't need RGBA in cs3d, and it's faster, and memory-efficient
  // in cs3d
  let useRGBA = true;

  if (options.useRGBA !== undefined) {
    useRGBA = options.useRGBA;
  }


```

sedghi:
My comment was you don't need to have `useRGBA:true` in your changed files, since the default is true, you can just remove it

Zaid-Safadi:
> My comment was you don't need to have `useRGBA:true` in your changed files, since the default is true, you can just remove it

Lets keep it explicitly set to `true` so we are always sure it will work even if the wado library changes

PR #353, opened 2023-01-09
feat(CobbAngle): Add CobbAngle tool

wayfarer3130:
Add a Cobb Angle implementation for CS3D.  
Added to the examples that have AngleTool so that it can be compared.


sedghi (commented):
Great stuff

Two things
1. Trying the example [here](https://deploy-preview-353--cornerstone-3d-docs.netlify.app/live-examples/stackannotationtools), user should be able to drag and draw both lines. I can drag and draw the first one, but for some reason I can't drag and draw the second line. I only can click and draw second one. Maybe there is some limitation in our toolings and it is not this PR

2. For ui/ux I would like to see some linking between the pairs, as it gets out of hand which one is paired with which one. This tool is often used in spine and you can guess what happens.
image

I looked into other viewers, and there are some sort of linkings between them. 

ClearCanvas
image

Meddream
image

Another one here
https://youtu.be/7XpYu-AjW0c?t=99

I honestly think doing something similar to meddream is like 10 mins of work (finding the middle of both lines and draw a dashed line). Can we have that please?

sedghi (changes requested):
This is certainly not the correct base for commits. It says 377 commits 

image

PR #354, opened 2023-01-09
feat: add support for WADO-URI Streaming Volume Loading 

jmhmd:
### Description
This PR adds support for streaming images into a volume with wadouri, i.e. a static set of dicom part 10 files on a server with a list of URIs pointing to those files. 

### Motivation
This may be helpful in the case where you have not extracted the series level metadata ahead of time, and you do not (or cannot) set up a dicomweb wado-rs server.

### Overview of changes
The main difference is that with wadouri, you do not have a way to pre-fetch the metadata for the entire series/volume ahead of time in order to construct the volume, calculate slice spacing, reorder the images, etc. 

So for this to work, we:

- Have to assume the array of imageIds provided to the volume loader is pre-sorted by location
- Pre-fetch the first, middle and last images of the series
- The loader uses the first image to get series and study level metadata, uses the first and middle images to calculate slice spacing (assuming a regular spacing across the volume), uses the first and middle images to determine if the stack needs to be reversed in order, and uses the last image in the event the stack is reversed, so the correct origin is used.

### Limitations

- This will always be slower I think than using wado-rs, because each dicom file needs to be parsed instead of streaming in the pixel data directly.
- If the correct sort order of the files is not known, the loaded images may be out of order.

This is a draft, and I have confirmed that loading works against the dcmjs.org dicomweb server in the provided example file but have not done testing with other more involved examples. I wanted to see if this approach is reasonable or if there are suggestions for improvements before adding more examples/tests.


jmhmd:
Of note, the image loading using wado-uri from dcmjs.org is way slower than wado-rs, slower I would think than could be explained by just the larger file size and parsing time. I think at least some of this is related to the server but not sure. It seems to be faster to load directly from and s3 bucket but I haven't done extensive testing here either.

packages/core/examples/volumeBasicWadoUri/index.ts
```diff
@@ -0,0 +1,150 @@
...
+  Enums,
+  volumeLoader,
+  imageLoader,
+} from '@cornerstonejs/core';
+import {
+  initDemo,
+  setTitleAndDescription,
+  setCtTransferFunctionForVolumeActor,
+} from '../../../../utils/demo/helpers';
+import instanceUIDs from './instanceUIDs.js';
```
sedghi, line 13: It was giving me error in compile for this `instanceUIDs.js`, so I had to convert that to .ts

packages/core/examples/volumeBasicWadoUri/index.ts
```diff
@@ -0,0 +1,150 @@
...
+        middleImageDataSet
+      ) {
+        return metaData.metaDataProvider(type, middleImageId);
+      } else {
+        console.warn(`Dataset for imageId ${imageId} not available`);
+        return;
+      }
+    }
+  }
+  cornerstone.metaData.addProvider(streamingMetaDataProvider);
```
sedghi, line 85: I guess later you should add this provider with a higher priority in the list of providers

packages/core/examples/volumeBasicWadoUri/index.ts
```diff
@@ -0,0 +1,150 @@
...
+
+  /**
+   * Preload first, middle, and last image metadata as these are the images the
+   * current streaming image loader may explicitly request metadata from. The
+   * last image metadata would only be specifically requested if the imageId
+   * array order is reversed in the `sortImageIdsAndGetSpacing.ts` file.
+   */
+  const indexesToPrefetch = [0, middleImageIndex, imageIds.length - 1];
+  for (let i of indexesToPrefetch) {
+    await imageLoader.loadImage(imageIds[i], { skipCreateImage: true }).then(
```
sedghi, line 95: I see what you are doing here. Just an FYI we have shifted away from using `loadImage` directly to using `imageLoadPoolManager` to handle it (to respect the priority of the queue). You can see an example in the Streaming WADO RS loader. Basically we create a callback (that uses loadImage), but the job is run through `imageLoadPoolManager`

packages/streaming-image-volume-loader/src/helpers/sortImageIdsAndGetSpacing.ts
```diff
@@ -45,28 +55,80 @@ export default function sortImageIdsAndGetSpacing(
...
+        referenceImagePositionPatient,
+        metadataForMiddleImage.imagePositionPatient
+      );
+      const distanceBetweenFirstAndMiddleImages = vec3.dot(
+        positionVector,
+        scanAxisNormal
+      );
+      zSpacing =
+        Math.abs(distanceBetweenFirstAndMiddleImages) /
+        Math.floor(imageIds.length / 2);
```
sedghi, line 124: why you are flooring here? I guess this result in inaccurate zspacing for odd numbers?

packages/streaming-image-volume-loader/src/helpers/sortImageIdsAndGetSpacing.ts
```diff
@@ -45,28 +55,80 @@ export default function sortImageIdsAndGetSpacing(
...
+    } else {
+      ({ spacingBetweenSlices: zSpacing } = metaData.get(
```
sedghi, line 126: I don't think we can rely on `spacingBetweenSlices` at all since it is optional and might be wrong. I would prefer to through here.

sedghi (commented):
I run this and it worked (it gave error for the instanceUIDs, but I got it fixed), the rendering worked fine. So great first iteration I would say

I think since it is streaming into the volume, it belongs to the same package `streaming-image-volume-loader`. I really like the idea that you are using the same loader scheme `cornerstoneStreamingImageVolume`, and we don't really care about wadoURI vs wadoRS at volume loader level. 

The parts that we need to decide where to go are 

1. `imageLoader.loadImage` for the three imageIds: I guess the `cornerstoneStreamingImageVolumeLoader` can identify the wadoURI scheme and call the load via imageLoadPoolManager before calling the `sortImageIdsAndGetSpacing`

2. Custom extended metadata provider: This one is tricky. I would like to get @wayfarer3130 view on this

packages/core/examples/volumeBasicWadoUri/index.ts
jmhmd: So when adding requests to the `imageLoadPoolManager`, is there a way to await those prefetched images? The reason is we have to have the metadata for those images cached before going on to construct the volume.

packages/streaming-image-volume-loader/src/helpers/sortImageIdsAndGetSpacing.ts
jmhmd: That is just the way I am getting the middle imageId index so the average is calculated against the right number of images.

packages/streaming-image-volume-loader/src/helpers/sortImageIdsAndGetSpacing.ts
jmhmd: That's reasonable, we can just fail if the middle image metadata is not prefetched.

jmhmd:
As for the metadata provider - I am not 100% sure it is necessary, it would only be needed if there is a code path that requests an image's metadata before that image has loaded into the volume. If it only requests metadata from the 1st, middle, and possibly last images prior to loading, those will be prefetched before volume construction and the built-in metadata provider in cswid should be sufficient.

jmhmd:
It looks like `StreamingImageVolume.ts` calls for each image's metadata to get scaling parameters, but at least in the example case, these never get used as `successCallback` only applies scaling if the image was already cached by another scheme it appears. So removing the custom extended metadata-provider still works to initially load via wadouri but I'm not sure if there will be issues with more complicated setups. I'm not familiar with the use case for that rescaling of cached images logic.

packages/core/examples/volumeBasicWadoUri/index.ts
jmhmd: I just pushed a commit that moves this prefetching logic to `cornerstoneStreamingImageVolumeLoader.ts` and detects the scheme used. Also now using `imageLoadPoolManager` as suggested. This seems much cleaner, the only thing is that I had to move all the logic in that file into an async function so that I could await the file prefetches before constructing the volume. The function already returned a promise, so that's ok I think, but now the `.decache()` and `.cancel()` functions are also async instead of sync.

packages/streaming-image-volume-loader/src/cornerstoneStreamingImageVolumeLoader.ts
```diff
@@ -36,140 +42,186 @@ function cornerstoneStreamingImageVolumeLoader(
...
-        loading: false,
-        cachedFrames: [],
-        callbacks: [],
-      },
-    }
-  );
-
   return {
-    promise: Promise.resolve(streamingImageVolume),
+    promise: getStreamingImageVolume(),
```
sedghi, line 213: can't we say before return something like `const streamingImageVolume = await getStreaming .... ` and then use it below directly like before?

jmhmd: Yes -- how about `const streamingImageVolumePromise = getStreaming...` so that it's explicit? Will defer to your preference.

sedghi (approved):

sedghi (approved):

sedghi:
@jmhmd Awesome contribution!

jmhmd:
Thanks for merging and the great feedback! Very thankful for the amazing work you all are doing with this library.

PR #358, opened 2023-01-11
feat: option to use Arraybuffer for volume loader instead of sharedArrayBuffer

Ouwen:
This feature allows for volumes to be loaded even if the page is not cross origin isolated.
For some use cases such as running cornerstone3D in cordova or mobile webview this is required.

### DIFF
--- a/packages/core/src/utilities/createFloat32SharedArray.ts
+++ b/packages/core/src/utilities/createFloat32SharedArray.ts
@@ -1,4 +1,5 @@
 import global from '../global';
+import { getShouldUseSharedArrayBuffer } from '../init';
 
 /**
  * A helper function that creates a new Float32Array that utilized a shared
@@ -25,7 +26,7 @@ import global from '../global';
  * @public
  */
 function createFloat32SharedArray(length: number): Float32Array {
-  if (!window.crossOriginIsolated) {
+  if (!getShouldUseSharedArrayBuffer()) {
     throw new Error(
       'Your page is NOT cross-origin isolated, see https://developer.mozilla.org/en-US/docs/Web/API/crossOriginIsolated'
     );
--- a/packages/streaming-image-volume-loader/src/StreamingImageVolume.ts
+++ b/packages/streaming-image-volume-loader/src/StreamingImageVolume.ts
@@ -384,6 +384,33 @@ export default class StreamingImageVolume extends ImageVolume {
       triggerEvent(eventTarget, Enums.Events.IMAGE_LOAD_ERROR, eventDetail);
     }
 
+    function handleArrayBufferLoad(scalarData, image, options) {
+      if (scalarData.buffer instanceof ArrayBuffer) {
+        const offset = options.targetBuffer.offset; // in bytes
+        const length = options.targetBuffer.length; // in frames
+        try {
+          if (scalarData instanceof Float32Array) {
+            const bytesInFloat = 4;
+            const floatView = new Float32Array(image.pixelData);
+            if (floatView.length !== length) {
+              throw 'Error pixelData length does not match frame length';
+            }
+            scalarData.set(floatView, offset / bytesInFloat);
+          }
+          if (scalarData instanceof Uint8Array) {
+            const bytesInUint8 = 1;
+            const intView = new Uint8Array(image.pixelData);
+            if (intView.length !== length) {
+              throw 'Error pixelData length does not match frame length';
+            }
+            scalarData.set(intView, offset / bytesInUint8);
+          }
+        } catch (e) {
+          console.error(e);
+        }
+      }
+    }
+
     const requests = imageIds.map((imageId, imageIdIndex) => {
       if (cachedFrames[imageIdIndex]) {
         framesLoaded++;
@@ -415,7 +442,13 @@ export default class StreamingImageVolume extends ImageVolume {
       const options = {
         // WADO Image Loader
         targetBuffer: {
-          arrayBuffer,
+          // keeping this in the options means a large empty volume array buffer
+          // will be transferred to the worker. This is undesirable for streaming
+          // volume without shared array buffer because the target is now an empty
+          // 300-500MB volume array buffer. Instead the volume should be progressively
+          // set in the main thread.
+          arrayBuffer:
+            arrayBuffer instanceof ArrayBuffer ? undefined : arrayBuffer,
           offset: imageIdIndex * lengthInBytes,
           length,
           type,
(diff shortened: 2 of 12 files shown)


Ouwen:
@sedghi what are your thoughts on this?

packages/core/src/utilities/createFloat32SharedArray.ts
```diff
@@ -24,21 +24,26 @@ import global from '../global';
...
     );
   }
 
-  const sharedArrayBuffer = new SharedArrayBuffer(length * 4);
+  const buffer = useSharedArrayBuffer
+    ? new SharedArrayBuffer(length * 4)
+    : new ArrayBuffer(length * 4);
 
-  return new Float32Array(sharedArrayBuffer);
+  return new Float32Array(buffer);
```
sedghi, line 46: I don't think you should do it here. This file should do what it says (creating sharedArray). Maybe in the cornerstoneStreamingImageVolume you branch to use non-shared

packages/streaming-image-volume-loader/src/StreamingImageVolume.ts
```diff
@@ -415,12 +414,14 @@ export default class StreamingImageVolume extends ImageVolume {
...
         targetBuffer: {
-          arrayBuffer,
+          arrayBuffer:
+            arrayBuffer instanceof ArrayBuffer ? undefined : arrayBuffer,
           offset: imageIdIndex * lengthInBytes,
           length,
           type,
         },
         skipCreateImage: true,
+        transferPixelData: arrayBuffer instanceof ArrayBuffer,
```
sedghi, line 424: I was about to fix this in another PR. Do you need it here? since I prefer it to be a separate `fix: broken transferPixelData` for both volume and stack viewport. 

Does it hurt to set `transferPixelData` to true for SharedArrayBuffer too? although I guess it is shared by construct ..?

packages/streaming-image-volume-loader/src/StreamingImageVolume.ts
```diff
@@ -415,12 +414,14 @@ export default class StreamingImageVolume extends ImageVolume {
       const options = {
         // WADO Image Loader
         targetBuffer: {
-          arrayBuffer,
+          arrayBuffer:
+            arrayBuffer instanceof ArrayBuffer ? undefined : arrayBuffer,
```
sedghi, line 418: Can you please add a comment to explain why we need it for sharedArrayBuffer (since the .set is happening inside the worker in cswil), but not for the arrayBuffer since it happens down below?

packages/streaming-image-volume-loader/src/StreamingImageVolume.ts
```diff
@@ -434,7 +435,30 @@ export default class StreamingImageVolume extends ImageVolume {
...
+                  const intView = new Uint8Array(image.pixelData);
+                  if (intView.length !== length) {
+                    throw 'Error pixelData length does not match frame length';
+                  }
+                  scalarData.set(intView, offset);
+                }
+              } catch (e) {
+                console.error(e);
+              }
+            }
```
sedghi, line 461: These all better be refactor into its own function e.g., `handleArrayBufferLoad` or somthing similar

sedghi (changes requested):
This looks great as first iteration. 
The only part I'm not certain right now is the global setting for `sharedArrayBuffer` usage in the `init`. Hmmmm, Do we want it there? I'm not really sure I like the idea of people fallbacking to ArrayBuffer (instead of more performant sharedArrayBuffer) if they have not set the headers correctly. They can just debug the error for cross origin and fix their headers and enjoy the performant path. 

Do you think we put it at the volume loader options?

CC @wayfarer3130

utils/ExampleRunner/example-runner-cli.js
```diff
@@ -179,7 +179,7 @@ if (configuration.examples) {
     // console.log('conf', conf);
     shell.ShellString(conf).to(webpackConfigPath);
     shell.cd(exBasePath);
-    shell.exec(`webpack serve --progress --config ${webpackConfigPath}`);
+    shell.exec(`webpack serve --host 0.0.0.0 --progress --config ${webpackConfigPath}`);
```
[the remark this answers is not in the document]
Ouwen: It will run on localhost and bind ip address so mobile devices on the same network can access.

packages/streaming-image-volume-loader/src/StreamingImageVolume.ts
Ouwen: I can remove this for your fix. It shouldn’t cause any issue to copy rather than transfer the encoded pixel array data

packages/streaming-image-volume-loader/src/StreamingImageVolume.ts
Ouwen: Sure, I’ll add something like: keeping this in the options means a large empty volume array buffer will be transferred to the worker. 

This is undesirable for streaming volume without shared array buffer because the target is now an empty 300-500MB volume array buffer. Instead the volume should be progressively set in the main thread.

Ouwen:
After some more investigation. It seems like the iOS safari behavior is due to something with web worker garbage collection. Hard refreshing iOS will have 1 web worker properly spawn; however, follow on web workers will fail with a generic "Range Error: Out of Memory".

With just one web worker, we can refresh the page within safari and have the volume properly load. However, the web worker will occasionally run into some error and crash (not loading any slices). A hard refresh (hard close safari and refresh) causes the page to function again.

The consistent finding is that the first load after a hard refresh will always work.

abluobo:
@Ouwen Will the memory occupied by the browser increase after the Arraybuffer replaces the sharedArrayBuffer?

Ouwen:
The browser memory creates an empty volume array. Let's say it is 100MB or so.
The workers will decode image pixel data and set a 1MB slice inside the 100MB volume.

The main difference is that without sharedArrayBuffer the workers have to transfer the 1MB decoded slice back into the main thread. The main thread then performs the set.

With sharedArrayBuffer, the workers can set directly into the 100MB sharedArrayBuffer

sedghi:
@Ouwen I talked to Bill, and we think `useSharedArrayBuffer` should have three options `true, false, auto` with default of `true`. What do you think?

Ouwen:
Yeah I think that makes sense.

Ouwen:
> After some more investigation. It seems like the iOS safari behavior is due to something with web worker garbage collection. Hard refreshing iOS will have 1 web worker properly spawn; however, follow on web workers will fail with a generic "Range Error: Out of Memory".
> 
> With just one web worker, we can refresh the page within safari and have the volume properly load. However, the web worker will occasionally run into some error and crash (not loading any slices). A hard refresh (hard close safari and refresh) causes the page to function again.
> 
> The consistent finding is that the first load after a hard refresh will always work.

This is unrelated to the non shared array buffer  implementation and is an issue with loading codecs on iOS safari.

https://github.com/cornerstonejs/cornerstoneWADOImageLoader/issues/506

Ouwen:
@sedghi, ready for review. I'm unsure why after running the api-update there are so many diffs but let me know if there is a different command I should run.

sedghi:
@Ouwen One of the tests are failing. You can force run it via changing `it` to `fit` and `describe` to `fdescribe` just FYI. Hopefully it will fail once you run it in isolation, if not there is some leaking happening between tests (or in source)


image

sedghi (approved):
Looks great !

Ouwen:
Thanks @sedghi

PR #369, opened 2023-01-16
feat(adapters): Import the adapters for cornerstone 3D to DICOM SR for annotations

wayfarer3130:
This is a move from dcmjs to change ownership of the cornerstone adapters from dcmjs into CS3D


cs3dLink.bat
```diff
@@ -0,0 +1,46 @@
+## cornerstone 3d / core
```
sedghi, line 1: can you remove these?

PR #374, opened 2023-01-18
fix(StackViewport): Reset camera bug when rotation happens on StackViewport

jbocce:
For a GPU reset, start by resetting the flips and rotation (via viewUp) because the ordering of the various flips and rotations that have been applied is not known.

### DIFF
--- a/packages/core/src/RenderingEngine/StackViewport.ts
+++ b/packages/core/src/RenderingEngine/StackViewport.ts
@@ -181,14 +181,14 @@ class StackViewport extends Viewport implements IStackViewport {
       renderer.setActiveCamera(camera);
 
       const viewPlaneNormal = <Point3>[0, 0, -1];
-      const viewUp = <Point3>[0, -1, 0];
+      this.initialViewUp = <Point3>[0, -1, 0];
 
       camera.setDirectionOfProjection(
         -viewPlaneNormal[0],
         -viewPlaneNormal[1],
         -viewPlaneNormal[2]
       );
-      camera.setViewUp(...viewUp);
+      camera.setViewUp(...this.initialViewUp);
       camera.setParallelProjection(true);
       camera.setThicknessFromFocalPoint(0.1);
       // @ts-ignore: vtkjs incorrect typing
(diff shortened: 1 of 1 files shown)


packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -155,6 +155,9 @@ class StackViewport extends Viewport implements IStackViewport {
   public modality: string; // this is needed for tools
   public scaling: Scaling;
 
+  // Camera properties
+  private initialViewUp: Point3;
```
jbocce, line 159: I had to introduce this instance variable because initialCamera might still be pointing to the previous image's camera (e.g. in _updateActorToDisplayImageId).

packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -1748,6 +1751,9 @@ class StackViewport extends Viewport implements IStackViewport {
 
     this.setCameraNoEvent({ viewUp, viewPlaneNormal });
 
+    // Setting this makes the following comment about resetCameraNoEvent not modifying viewUp true.
+    this.initialViewUp = viewUp;
+
     // Reset the camera to point to the new slice location, reset camera doesn't
```
jbocce, line 1757: Prior to my changes, this comment was not true. viewUp was getting changed via a call to camera.roll.

packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -155,6 +155,9 @@ class StackViewport extends Viewport implements IStackViewport {
   public modality: string; // this is needed for tools
   public scaling: Scaling;
 
+  // Camera properties
+  private initialViewUp: Point3;
+
```
sedghi, line 160: I think we should `this.initialViewUp=` inside the constructor as well

packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -1883,7 +1889,14 @@ class StackViewport extends Viewport implements IStackViewport {
     // Todo: we need to make the rotation a camera properties so that
     // we can reset it there, right now it is not possible to reset the rotation
     // without this
-    this.getVtkActiveCamera().roll(this.rotationCache);
+    if (this.initialViewUp) {
```
sedghi, line 1892: if you set it in the constructor you don't need this if here, since it will always be defined

sedghi (changes requested):
Great work. See my comments please

packages/core/src/RenderingEngine/StackViewport.ts
jbocce: agreed

jbocce:
> Great work. See my comments please

Thank you.

PR #377, opened 2023-01-19
fix: tool bindings with different modifier keys

matthis-d:
It was not possible to add bindings on a tool with same mouseBinging but different modifier keys. You can now create the follwing tool configuration:
```js
toolGroup.setToolActive(toolName, {
  bindings: [
    {
      mouseButton: MouseBindings.Primary, // Left Click
    },
    {
      mouseButton: MouseBindings.Primary,
      modifierKey: KeyboardBindings.Shift,
    },
  ],
});
```

### DIFF
--- a/packages/tools/src/store/ToolGroupManager/ToolGroup.ts
+++ b/packages/tools/src/store/ToolGroupManager/ToolGroup.ts
@@ -10,6 +10,7 @@ import {
 import type { Types } from '@cornerstonejs/core';
 import { state } from '../index';
 import {
+  IToolBinding,
   IToolClassReference,
   IToolGroup,
   SetToolBindingsType,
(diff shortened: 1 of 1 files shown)


packages/tools/src/store/ToolGroupManager/ToolGroup.ts
```diff
@@ -326,7 +327,7 @@ export default class ToolGroup implements IToolGroup {
     const bindingsToUse = [...prevBindings, ...newBindings].reduce(
       (unique, binding) => {
         if (
-          !unique.some((obj) => obj.mouseButton === binding.mouseButton) &&
+          !unique.some((obj) => hasSameBinding(obj, binding)) &&
```
sedghi, line 330: Thanks for the contribution.
Can you please edit the comment above this reducer to say it allows duplicated mouse button but not both mouse and modifier?

matthis-d: Thank.
I added the comment you asked for.

sedghi (approved):
Thanks for your contribution!

PR #381, opened 2023-01-21
added drawOnMouseDrag option for referenceCursors

ssokmen:


sedghi:
Thanks for the contribution (CI is failing, please read this https://www.cornerstonejs.org/docs/contribute/update-api) 
Overall looks fine, but I'm not sure about adding all the callbacks for the mouseClick on a tool that is extending the `AnnotationDisplay` for a purpose which means does not need to have a mouse binding. I think it makes sense to have two separate tools 

RefereceCursorBase (not a tool, the logic)
 |
 |--- RefernceCursor (as we have right now)
 |--- ReferenceCursorDrag (the one that you are adding)

What do you think?

ssokmen:
Sorry, I have read the document. I will fix the pr problem. I can review and rewrite the logic as you mentioned. But, If you can investigate dicom viewers and you will see generally reference cursor using down and drag. Because If I change my display screen I can't see my last location review in detail.

I will proceed according to your answer.

heyflynn:
This is a great feature. Users have been asking for it some time (similar to crosshairs but for StackViewport).

I created my own version using AnnotationTool and it works great except when dealing with localizers containing different planes. The cursor locks up when one of the images changes planes mid drag. Once Its resolved I can come help with this resolution. The conflict is as Sedghi mentions, AnnotationDisplayTool doesn't have callbacks for mouse down etc..

Issue #375, opened 2023-01-18
Double click detection


sedghi:
Currently the double click on a viewport when an annotation tool is active will start the annotation/rendering loop which is not great since that means to App can run callbacks on double click.

There should be a timeout for the [mouseDownlistener](https://github.com/cornerstonejs/cornerstone3D-beta/blob/feat/sphere-surface-cut/packages/tools/src/eventListeners/mouse/mouseDownListener.ts#L86-L141) set so that a double click is distinguished from a single click. Since otherwise, the double click will be triggered as : single click (first click) + double click (when second click happens)

jbocce:
Related PR: https://github.com/cornerstonejs/cornerstone3D-beta/pull/382

jbocce:
Subsequent PRs to fix bugs:

- https://github.com/cornerstonejs/cornerstone3D-beta/pull/417
- https://github.com/cornerstonejs/cornerstone3D-beta/pull/416

PR #382, opened 2023-01-23
feat(doubleClick): Add Double click detection (#375)

jbocce:
Added a timeout in mouseDownListener to determine if a double click occurs. If a double click occurs during the timeout,
then any mouse down and up events that occurred during the timeout are suppressed. If the timeout expires, then the first mouse down and up events during the timeout are fired and any browser double click event that might occur is suppressed/ignored. Double clicks that occur during tool interaction are suppressed Double click event capture phase listeners are used to more reliably suppress double clicks.

The related issue is: https://github.com/cornerstonejs/cornerstone3D-beta/issues/375


packages/tools/src/eventListeners/mouse/mouseDownListener.ts
```diff
@@ -9,6 +9,8 @@ import getMouseEventPoints from './getMouseEventPoints';
 const { MOUSE_DOWN, MOUSE_DOWN_ACTIVATE, MOUSE_CLICK, MOUSE_UP, MOUSE_DRAG } =
   Events;
 
+const DOUBLE_CLICK_TOLERANCE_MS = 400;
```
jbocce, line 12: What is the best way to make this configurable?

sedghi: I guess we can worry about this in another PR, since we are not configuring [any other suff](https://github.com/cornerstonejs/cornerstone3D-beta/blob/feat/sphere-surface-cut/packages/tools/src/init.ts#L26)

utils/test/testUtilsMouseEvents.ts
```diff
@@ -44,4 +47,57 @@ function createNormalizedMouseEvent(
...
+ * down is triggered but before the mouse up. An optional callback is invoked
+ * after the mouse up has fired.
+ *
+ * @param element the element to dispatch to
+ * @param mouseDownEvent the mouse down event to dispatch
+ * @param mouseUpEvent the mouse up event to dispatch
+ * @param betweenDownAndUpCallback optional callback between the down and up
+ * @param afterDownAndUpCallback optional callback after the up
+ * @returns a Promise for the eventual completion of the mouse down and up
+ */
```
sedghi, line 64: Love this doc, thanks

utils/test/testUtilsMouseEvents.ts
```diff
@@ -44,4 +47,57 @@ function createNormalizedMouseEvent(
...
+function doMouseDownAndUp(
```
sedghi, line 65: can we rename this function to `performMouseDownAndUp`?

packages/tools/test/BidirectionalTool_test.js
```diff
@@ -1,6 +1,8 @@
 import * as cornerstone3D from '@cornerstonejs/core';
 import * as csTools3d from '../src/index';
 import * as testUtils from '../../../utils/test/testUtils';
+import * as cornerstoneTools from '@cornerstonejs/tools';
```
sedghi, line 4: not a big deal, but just saying for future, let's order the library imports at top and the local ones ('../../etc') after

packages/tools/test/LengthTool_test.js
```diff
@@ -586,13 +587,15 @@ describe('LengthTool:', () => {
...
         // Just grab and don't really move it
-        evt = new MouseEvent('mouseup');
-
-        addEventListenerForAnnotationRendered();
-        document.dispatchEvent(evt);
+        const mouseUpEvt = new MouseEvent('mouseup');
+
+        doMouseDownAndUp(
+          element,
+          evt,
```
sedghi, line 595: I know evt, is mouseDown, but renaming the variable to `mouseDownEvt` would match the `mouseUpEvt` nicelu

packages/tools/src/tools/annotation/ArrowAnnotateTool.ts
```diff
@@ -446,6 +450,10 @@ class ArrowAnnotateTool extends AnnotationTool {
 
     this.editData = null;
     this.isDrawing = false;
+
+    // This double click was handled and the dialogue was displayed. No need for anyone else to handle it too.
+    evt.stopImmediatePropagation();
```
sedghi, line 455: why `stopImmediatePropagation` as opposed to `stopPropagation`? not immediately clear to me so I appreciate an inline comment above it

packages/tools/src/eventListeners/mouse/mouseDoubleClickListener.ts
```diff
@@ -35,7 +35,21 @@ function mouseDoubleClickListener(evt: MouseEvent): void {
     deltaPoints,
   };
 
-  triggerEvent(element, Events.MOUSE_DOUBLE_CLICK, eventDetail);
+  const preventDefault = !triggerEvent(
```
sedghi, line 38: can we rename this to `consumed` or similar?

packages/tools/src/eventListeners/mouse/mouseDownListener.ts
```diff
@@ -102,8 +134,30 @@ function mouseDownListener(evt: MouseEvent) {
...
+  state.lastPoints = _copyPoints(state.startPoints);
+
+  // First mouse down of a potential double click. So save it and start
+  // a timeout to determine a double click.
+  doubleClickState.mouseDownEvent = evt;
+
+  doubleClickState.doubleClickTimeout = setTimeout(
+    _doubleClickTimeoutHandler,
+    DOUBLE_CLICK_TOLERANCE_MS
+  );
```
sedghi, line 147: shouldn't this be the first thing right after the following?

```
if (doubleClickState.doubleClickTimeout) {
    return;
  }
```

packages/tools/src/eventListeners/mouse/mouseDoubleClickListener.ts
sedghi: or `eventDidNotPropagate`

sedghi: or even better `const eventDidPropagate = trigger ...`

packages/tools/src/eventListeners/mouse/mouseDownListener.ts
```diff
@@ -102,8 +134,30 @@ function mouseDownListener(evt: MouseEvent) {
   // Prevent CornerstoneToolsMouseMove while mouse is down
   state.element.removeEventListener('mousemove', mouseMoveListener);
 
-  const startPoints = getMouseEventPoints(evt, state.element);
-  const deltaPoints = _getDeltaPoints(startPoints, startPoints);
+  state.startPoints = getMouseEventPoints(evt, state.element);
```
sedghi, line 137: we were doing `copyPoints` previously, so I'm wondering not copying will result some bug due to reference manipulation

packages/tools/src/eventListeners/mouse/mouseDownListener.ts
```diff
@@ -176,52 +237,186 @@ function _onMouseDrag(evt: MouseEvent) {
...
+ * @private
+ */
+function _doubleClickTimeoutHandler() {
+  // The time out means that this latest sequence of mouse down(s) and up(s) is NOT a double click.
+  // Set a flag in case our timeout value is inaccurate and we do indeed get a double click event later.
+  // The flag will be cleared in the mouseDoubleClickIgnoreListener should a double click event get fired.
+  // If there is no eventual double click for this latest sequence, the flag spills into the next
+  // sequence where it will either get set again (here) or cleared in _onMouseUp
+  // if an actual double click is detected. Leaving the value true is perfectly safe since
+  // any future double click must start with a mouse down.
```
sedghi, line 384: great job explaining this

packages/tools/src/eventListeners/mouse/mouseDownListener.ts
```diff
@@ -176,52 +237,186 @@ function _onMouseDrag(evt: MouseEvent) {
...
+}
+
+/**
+ * Determines if the given delta is past the double click, drag distance tolerance.
+ * @param delta the delta
+ * @returns true iff the delta is past the tolerance
+ */
+function _isDragPastDoubleClickTolerance(delta: Types.Point2): boolean {
+  return (
+    Math.sqrt(Math.pow(delta[0], 2) + Math.pow(delta[1], 2)) >
```
sedghi, line 362: This is too much to get calculated for each mouse move, a lot of computations. Can we make it simpler to not rely on sqrt and pow?

utils/test/testUtilsMouseEvents.ts
jbocce: thanks

utils/test/testUtilsMouseEvents.ts
jbocce: as you wish

packages/tools/test/BidirectionalTool_test.js
jbocce: Ok, I will try to be more careful next time. Thanks for the 'heads up'. I am not 100% certain, but I think it might have been VS Code that did this automatically. Funny enough, that import is no longer needed.

packages/tools/test/LengthTool_test.js
jbocce: Sure. I will go through the various unit test files and make updates.

packages/tools/src/eventListeners/mouse/mouseDoubleClickListener.ts
jbocce: Note that `triggerEvent` in turn calls `dispatchEvent`. I checked the mdn docs for `dispatchEvent` (https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/dispatchEvent) and its return value has NOTHING to do with propagation. In fact it is related to `preventDefault`. So I still prefer the name I had. However of your choices I think `consumed` is best.

packages/tools/src/eventListeners/mouse/mouseDownListener.ts
jbocce: No I don't understand why it should be. The listeners that follow have to be added. That said I can move the `addListener` calls before the `setTimeout` call and the method will return right away regardless. However, I am still wondering why the suggestion.

packages/tools/src/eventListeners/mouse/mouseDownListener.ts
jbocce: thanks

packages/tools/src/eventListeners/mouse/mouseDownListener.ts
jbocce: I understand and yes I can change it. However note that this is only for every move during the timeout. The timeout is relatively short. Also, once a drag/move goes past the tolerance the timeout is cleared. I would argue that yes it is expensive but for a very short time - likely less than the timeout if the mouse is moving any significant amount. Would you still like me to change it?

packages/tools/src/eventListeners/mouse/index.ts
```diff
@@ -29,6 +34,10 @@ function enable(element: HTMLDivElement): void {
   disable(element);
 
   element.addEventListener('dblclick', mouseDoubleClickListener);
+  element.addEventListener('dblclick', mouseDoubleClickIgnoreListener, {
+    capture: true, // capture phase is the best way to ignore double clicks
+  });
```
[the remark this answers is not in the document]
jbocce: Yes there are two reasons...

- If the logic of `mouseDoubleClickIgnoreListener` is moved into `mouseDoubleClickListener`, then the `ignoreDoubleClick` flag in `mouseDownListener` has to be exposed. Furthermore its mutation from `true` to `false` also has to be exposed. I do not particularly like that.
- Notice that a capture phase listener is used which `mouseDoubleClickListener` is not. To best prevent listeners of `dblclick` to NOT get an ignored double click, a capture phase listener is best.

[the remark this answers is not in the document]
sedghi: Then can we put them as inline comment?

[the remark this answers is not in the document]
jbocce: Most certainly.

packages/tools/src/eventListeners/mouse/mouseDownListener.ts
sedghi: This means that if in case the browser sends it (not the user)

packages/tools/src/eventListeners/mouse/mouseDownListener.ts
```diff
@@ -176,52 +237,186 @@ function _onMouseDrag(evt: MouseEvent) {
...
+/**
+ * Do a mouse down and potential mouse up using each of the events in the double click state.
+ * The events were stored in the state during the timeout to determine a double click.
+ * @private
+ */
+function _doStateMouseDownAndUp() {
+  const mouseDownEvent = doubleClickState.mouseDownEvent;
+  const mouseUpEvent = doubleClickState.mouseUpEvent;
+
+  _clearDoubleClickTimeoutAndEvents();
```
sedghi, line 399: maybe we can put ignore here

packages/tools/src/eventListeners/mouse/mouseDownListener.ts
```diff
@@ -176,52 +237,186 @@ function _onMouseDrag(evt: MouseEvent) {
...
+
+      // ingore any mouse down and up events captured and let the double click happen
+      _clearDoubleClickTimeoutAndEvents();
+
+      doubleClickState.ignoreDoubleClick = false;
+
+      state = JSON.parse(JSON.stringify(defaultState));
+    } else {
+      // this is the first mouse up during the double click timeout; we'll need it later if the timeout expires
+      doubleClickState.mouseUpEvent = evt;
```
sedghi, line 275: switch order

packages/tools/src/eventListeners/mouse/mouseDownListener.ts
jbocce: I moved the setting of the doubleClickTimeout to near the top of the method.

packages/tools/src/eventListeners/mouse/index.ts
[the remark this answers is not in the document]
jbocce: Actually I will move this logic code out of this index.ts and put it in mouseDownListener because none of it needs to be exposed outside that module.

packages/tools/src/eventListeners/mouse/mouseDownListener.ts
jbocce: Yes we can and as such I will be removing the `doubleClickTimeoutHandler` function entirely since all it does is call `_doStateMouseDownAndUp`. Great suggestion.

packages/tools/src/eventListeners/mouse/mouseDownListener.ts
jbocce: will do

packages/tools/src/tools/annotation/ArrowAnnotateTool.ts
jbocce: will do

packages/tools/src/eventListeners/mouse/mouseDownListener.ts
```diff
@@ -9,6 +9,25 @@ import getMouseEventPoints from './getMouseEventPoints';
...
+// TODO This module should detect and fire 'dblclick' events at its discretion and
+// ignore all those generated by the browser.
+//
+const DOUBLE_CLICK_TOLERANCE_MS = 400;
+
+// A drag (distance) during the double click timeout that is greater than this
+// value will cancel the timeout and suppress any double click that might occur.
+// This tolerance is particularly important on touch devices where some movement
+// might occur between the two clicks.
+const DOUBLE_CLICK_DRAG_TOLERANCE = 3;
```
jbocce, line 29: The last commit had this as 2. I did some further testing on my phone and thought 3 was better. I will add a TODO to revisit this value for touch devices.

packages/tools/src/eventListeners/mouse/index.ts
```diff
@@ -12,6 +14,14 @@ import mouseMoveListener from './mouseMoveListener';
  */
 function disable(element: HTMLDivElement): void {
   element.removeEventListener('dblclick', mouseDoubleClickListener);
+
+  // A separate double click listener for the element. Separate because...
+  // - it listens on the capture phase (and not the typical bubble phase)
+  // - the data used to ignore the double click is private to mouseDoubleClickIgnoreListener
+  element.removeEventListener('dblclick', mouseDoubleClickIgnoreListener, {
```
jbocce, line 21: This adds the listener for every enabled viewport. It is a bit of a waste since only one (global) listener is required.

PR #386, opened 2023-01-26
feat(ScaleOverlayTool): Add scale overlay tool

NeilMacPhee:
### Summary

Displays a scale overlay on images to show the real world size. Scale can be positioned at the bottom, top, left or right side of the canvas. Scale size has been verified by using the length tool.

### Screenshot

scaleoverlaytool-demo

### DIFF
--- a/packages/tools/src/tools/ScaleOverlayTool.ts
+++ b/packages/tools/src/tools/ScaleOverlayTool.ts
@@ -126,7 +126,7 @@ class ScaleOverlayTool extends AnnotationDisplayTool {
   onCameraModified = (evt: Types.EventTypes.CameraModifiedEvent): void => {
     // If the camera is modified, we need to update the viewport
     // that the camera was modified on
-    this.configuration.viewportId = evt.target.dataset.viewportUid;
+    this.configuration.viewportId = evt.detail.viewportId;
     this._init();
   };
 
--- a/packages/tools/src/tools/ScaleOverlayTool.ts
+++ b/packages/tools/src/tools/ScaleOverlayTool.ts
@@ -1,7 +1,7 @@
 import AnnotationDisplayTool from './base/AnnotationDisplayTool';
 import { vec3 } from 'gl-matrix';
 import {
-  metaData,
+  getEnabledElementByIds,
   getRenderingEngines,
   utilities as csUtils,
 } from '@cornerstonejs/core';
@@ -20,6 +20,7 @@ import {
   SVGDrawingHelper,
 } from '../types';
 import { StyleSpecifier } from '../types/AnnotationStyle';
+import { getToolGroup } from '../store/ToolGroupManager';
 
 const SCALEOVERLAYTOOL_ID = 'scaleoverlay-viewport';
 
@@ -66,15 +67,22 @@ class ScaleOverlayTool extends AnnotationDisplayTool {
       return;
     }
 
-    let viewports = renderingEngine.getViewports();
-    viewports = filterViewportsWithToolEnabled(viewports, this.getToolName());
+    const viewportIds = getToolGroup(this.toolGroupId).viewportsInfo;
 
-    let viewport = viewports[0];
+    if (!viewportIds) return;
+
+    const enabledElements = viewportIds.map((e) =>
+      getEnabledElementByIds(e.viewportId, e.renderingEngineId)
+    );
+
+    let viewport = enabledElements[0].viewport;
 
     if (this.configuration.viewportId) {
-      viewport = viewports.find(
-        (viewportId) => viewportId.id === this.configuration.viewportId
-      );
+      enabledElements.forEach((element) => {
+        if (element.viewport.id === this.configuration.viewportId) {
+          viewport = element.viewport;
+        }
+      });
     }
 
     if (!viewport) {
@@ -177,7 +185,7 @@ class ScaleOverlayTool extends AnnotationDisplayTool {
     const worldWidthViewport = vec3.distance(bottomLeft, bottomRight);
     const worldHeightViewport = vec3.distance(topLeft, bottomLeft);
 
-    // 0.05 gives margin to horizontal and vertical lines
+    // hscaleBounds and vscaleBounds compute the max bound for scales on the image
     const hscaleBounds = this.computeScaleBounds(
       canvasSize,
       0.05,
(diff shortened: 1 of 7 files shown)


sedghi:
great job, you need to update the api

`yarn build:update-api` 

then commit all the changed markdowns

sedghi:
also to add the example to the list of examples on the website, you need to add it to the example json file

NeilMacPhee:
@sedghi Added the example to the list of examples for the website and updated the api

packages/tools/src/tools/ScaleOverlayTool.ts
```diff
@@ -0,0 +1,666 @@
...
+    annotation: ScaleOverlayAnnotation;
+  } | null = {} as any;
+  isDrawing: boolean;
+  isHandleOutsideImage: boolean;
+
+  constructor(
+    toolProps: PublicToolProps = {},
+    defaultToolProps: ToolProps = {
+      configuration: {
+        viewportId: '',
```
sedghi, line 52: why you need viewportId in the confi? this doesn't seem right to me

packages/tools/src/tools/ScaleOverlayTool.ts
```diff
@@ -0,0 +1,666 @@
...
+        scaleLocation: 'bottom',
+        scaleColor: 'yellow',
```
sedghi, line 54: can this be rgb instead of names?

sedghi: Follow `packages/tools/src/stateManagement/annotation/config/ToolStyle.ts`

packages/tools/src/tools/ScaleOverlayTool.ts
```diff
@@ -0,0 +1,666 @@
...
+  _init = (): void => {
+    const renderingEngines = getRenderingEngines();
+    const renderingEngine = renderingEngines[0];
+
+    if (!renderingEngine) {
+      return;
+    }
+
+    let viewports = renderingEngine.getViewports();
+    viewports = filterViewportsWithToolEnabled(viewports, this.getToolName());
```
sedghi, line 70: instead of this you should do 

`const viewportIds = getToolGroup(this.toolGroupId).viewportsInfo;` 

you can grab the viewport from the toolGroupId, see the refernceCursors implementation
packages/tools/src/tools/ReferenceCursors.ts

packages/tools/src/tools/ScaleOverlayTool.ts
```diff
@@ -0,0 +1,666 @@
...
+    enabledElement: Types.IEnabledElement,
+    svgDrawingHelper: SVGDrawingHelper
+  ) {
+    if (!this.editData.viewport) {
+      return;
+    }
+    const location = this.configuration.scaleLocation;
+    const { annotation, viewport } = this.editData;
+    const image = viewport.getImageData();
+    const imageId = viewport.getCurrentImageId();
```
sedghi, line 152: if we are using imageId, that means only stackViewport is supported. Can you at least mention it in the documentation? and in the demo doc?

packages/tools/src/tools/ScaleOverlayTool.ts
```diff
@@ -0,0 +1,666 @@
...
+      0.05,
+      0.05,
+      location
+    );
+
+    const scaleSize = this.computeScaleSize(
+      worldWidthViewport,
+      worldHeightViewport,
+      location
+    );
```
sedghi, line 221: These require documentation as a comment above each. Not sure what each function is doing

packages/tools/src/tools/ScaleOverlayTool.ts
```diff
@@ -0,0 +1,666 @@
...
+        imagePlane.columnPixelSpacing || imagePlane.colImagePixelSpacing;
+    }
+
+    // Check whether pixel spacing is defined
+    if (!rowPixelSpacing || !colPixelSpacing) {
+      console.warn(
+        `Unable to define rowPixelSpacing or colPixelSpacing from data on ScaleOverlayTool's renderAnnotation`
+      );
+      return;
+    }
```
sedghi, line 185: imageId is limited to stackViewport, you are getting imagePlaenModule for the imageId and then row/col pixel spacing. What are you using row/col pixel spacing for? I see it is only used for checking they exists. 

So basically you are limitting this tool for stack viewport for no reason?

sedghi (changes requested):
See my comments please

packages/tools/src/tools/ScaleOverlayTool.ts
NeilMacPhee: the viewportId is used to pass the current enabled viewport back into the innit function

packages/tools/src/tools/ScaleOverlayTool.ts
NeilMacPhee: Removed this section from the code, the tool works on both stack and volume viewports as shown below.

scaleoverlay-volume

packages/tools/src/tools/ScaleOverlayTool.ts
NeilMacPhee: Added documentation

packages/tools/src/tools/ScaleOverlayTool.ts
NeilMacPhee: scaleColor has been removed, the annotation color is now determined by the StyleSpecifier

packages/tools/src/tools/ScaleOverlayTool.ts
NeilMacPhee: This section has been removed, I was originally using pixel spacing in calculating the scale size but this section is not of use anymore.

NeilMacPhee:
The requested changes have been made

sedghi (approved):
Great work

tineke01:
In the example, I can only see 'top' & 'left','bottom' and 'right' is not displayed

tineke01:
image
image

NeilMacPhee:
I just tested the example file and everything worked fine for me. Are you getting any errors in the console when you run it?

**Screenshots**
Screenshot 2023-02-09 at 09 39 39
Screenshot 2023-02-09 at 09 39 51
Screenshot 2023-02-09 at 09 40 09
Screenshot 2023-02-09 at 09 40 22

tineke01:
image
image

tineke01:
I tried to view the sample on multiple monitors and found different results

tineke01:
I find that when the value of 'window.devicePixelRatio' is not equal to 1, it is incorrectly displayed
image
image
image

NeilMacPhee:
@tineke01 Interesting, it appears that changing the device pixel ratio affects all annotation tools, not just the scale overlay tool. If you are looking to resize the viewport I would recommend changing the element's width and height in the example file rather than zooming in on the browser window.

tineke01:
I'm just illustrating the problem. In fact, there are many computers with the default zoom ratio of 125% or 200%, in this case, even if the browser zoom ratio is 100%, 'window.devicePixelRatio' will not equal to 1, an error display.

Issue #252, opened 2022-10-20
Can use getElementByClassName instead of firstChild ?


lambacini:
https://github.com/cornerstonejs/cornerstone3D-beta/blob/dfdc4bfbf331da40368a4976f3dc199bd355864a/packages/tools/src/drawingSvg/getSvgDrawingHelper.ts#L37

in angularjs, we use nativeElement for rendering element (enabledElement), but nativeElement contains overlayboxes and some other ui components. in this case, cannot add drawings because viewport-element it is not the first element of enabledElement.
if we use getElementByClassName(VIEWPORT_ELEMENT) instead of firstChild, i think it will work.

thank you for your understanding.

![image](https://user-images.githubusercontent.com/943188/196950278-15cadee9-f016-4f55-86da-29b3ce09c958.png)
![image](https://user-images.githubusercontent.com/943188/196950336-91c27742-6b70-4cc5-87ab-88a267bc8962.png)

sedghi:
Interesting bug. Thanks for the report

sedghi:
Do you mind creating a PR for this testing it on your framwork?

lambacini:
i create pr and tested it in my own app and cornerstone examples, it works correctly. but this function will not work correctly if used elsewhere instead of viewport-element.

lambacini:
https://github.com/cornerstonejs/cornerstone3D-beta/pull/268

PR #387, opened 2023-01-26
fix(svg): find and reset svg-layer within the correct element

ssokmen:
We use nativeElement for rendering an element (enabledElement) that contains overlayboxes as well as some other components. While querying for the svg-layer during the remove element, we are having problems using these properties because it searches under the wrong element. I investigated our repository issues and found it looks like #252.

removeChild

### DIFF
--- a/packages/tools/src/store/removeEnabledElement.ts
+++ b/packages/tools/src/store/removeEnabledElement.ts
@@ -32,13 +32,7 @@ function removeEnabledElement(
   const { element, viewportId } = elementDisabledEvt.detail;
 
   _resetSvgNodeCache(element);
-
-  const viewportNode = element;
-  const svgLayer = viewportNode.querySelector('svg');
-  const internalViewportNode = element.querySelector(`div.${VIEWPORT_ELEMENT}`);
-  if (svgLayer) {
-    internalViewportNode.removeChild(svgLayer);
-  }
+  _removeSvgNode(element);
 
   // Remove this element from the annotation rendering engine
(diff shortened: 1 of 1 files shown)


ssokmen:
Hello, did you have any chance to review the PR? 
I have trouble changing the viewport type. :( For example, if I have any element unless "viewport-element". And then, If I want to convert Stack viewport to Volume viewport. RenderingEngine tries to _resetviewport() and it can't Because can't find to remove svg-layer. The svg-layer is unlikely to be outside the viewport-element. It would make more sense to look for this svg-layer with in the viewport-element rather than the cornerstone-element.

sedghi:
Hi, thanks for the PR

So

> I have trouble changing the viewport type. :( For example, if I have any element unless "viewport-element". And then, If I want to convert Stack viewport to Volume viewport. RenderingEngine tries to _resetviewport() and it can't Because can't find to remove svg-layer. The svg-layer is unlikely to be outside the viewport-element. It would make more sense to look for this svg-layer with in the viewport-element rather than the cornerstone-element.

Does this mean this PR will fix that behavior for changing viewport type?

ssokmen:
Yes, this will fix it. It can be used to fix the changing viewport or any other features related to this area. Here, I have given an example of changing the viewport type. In this case, we are searching for the svg-layer under the incorrect element. That is why it gives an error when attempting to remove it. Under ViewportNode our search would be more accurate. Because svg-layer always to be in viewportNode.

image

Issue #372, opened 2023-01-17
Reset camera bug when rotation happens on StackViewport


sedghi:
New small bug with latest version : https://www.cornerstonejs.org/live-examples/stackapi

Reproduce via
- Rotate
- Flip
- Reset => does not reset

( note : reset work on rotate, flip, flip + rotate, but not on rotate + flip

funnier :  rotate, flip H, flip V, reset : works)

jbocce:
Another issue found...

rotate + flip + next image is NOT is the correct orientation

jbocce:
PR: https://github.com/cornerstonejs/cornerstone3D-beta/pull/388

jbocce:
Various scenarios to test - also try permutations of each:

rotation + flip + next image
rotation + flip + reset
rotation + flip + next image + absolute rotation
rotation + flip + next image + delta rotation
rotation + flip + delta rotation

PR #388, opened 2023-01-27
feat(StackViewport): rework rotation for stack viewport

jbocce:
fix(StackViewport): #372 Removed both the Viewport.rotation and StackViewport.rotationCache properties.
    Introduced an IViewport.getRotation for those viewports supporting rotation.
    StackViewport.getRotation calculates the rotation from the camera properties.
    Changed the StackAPI example to add buttons for constant absolute and delta rotations.
    The StackAPI example now reports the current rotation and flips.

Various scenarios to test - also try permutations of each:

- rotation + flip + next image
- rotation + flip + reset
- rotation + flip + next image + absolute rotation
- rotation + flip + next image + delta rotation
- rotation + flip + delta rotation

### DIFF
--- a/packages/core/examples/stackAPI/index.ts
+++ b/packages/core/examples/stackAPI/index.ts
@@ -18,7 +18,7 @@ console.warn(
   'Click on index.ts to open source code for this example --------->'
 );
 
-const { ViewportType } = Enums;
+const { ViewportType, Events } = Enums;
 
 // ======== Constants ======= //
 const renderingEngineId = 'myRenderingEngine';
@@ -37,6 +37,35 @@ element.style.width = '500px';
 element.style.height = '500px';
 
 content.appendChild(element);
+
+const info = document.createElement('div');
+content.appendChild(info);
+
+const rotationInfo = document.createElement('div');
+info.appendChild(rotationInfo);
+
+const flipHorizontalInfo = document.createElement('div');
+info.appendChild(flipHorizontalInfo);
+
+const flipVerticalInfo = document.createElement('div');
+info.appendChild(flipVerticalInfo);
+
+element.addEventListener(Events.CAMERA_MODIFIED, (_) => {
+  // Get the rendering engine
+  const renderingEngine = getRenderingEngine(renderingEngineId);
+
+  // Get the stack viewport
+  const viewport = <Types.IStackViewport>(
+    renderingEngine.getViewport(viewportId)
+  );
+
+  const { flipHorizontal, flipVertical } = viewport.getCamera();
+  const { rotation } = viewport.getProperties();
+
+  rotationInfo.innerText = `Rotation: ${rotation}`;
+  flipHorizontalInfo.innerText = `Flip horizontal: ${flipHorizontal}`;
+  flipVerticalInfo.innerText = `Flip vertical: ${flipVertical}`;
+});
 
 addButtonToToolbar({
   title: 'Set VOI Range',
@@ -143,7 +172,7 @@ addButtonToToolbar({
 });
 
 addButtonToToolbar({
-  title: 'Rotate',
+  title: 'Rotate Random',
   onClick: () => {
     // Get the rendering engine
     const renderingEngine = getRenderingEngine(renderingEngineId);
@@ -156,6 +185,41 @@ addButtonToToolbar({
     const rotation = Math.random() * 360;
 
     viewport.setProperties({ rotation });
+
+    viewport.render();
+  },
+});
+
+addButtonToToolbar({
+  title: 'Rotate Absolute 150',
+  onClick: () => {
+    // Get the rendering engine
+    const renderingEngine = getRenderingEngine(renderingEngineId);
+
+    // Get the stack viewport
+    const viewport = <Types.IStackViewport>(
+      renderingEngine.getViewport(viewportId)
+    );
+
+    viewport.setProperties({ rotation: 150 });
+
+    viewport.render();
+  },
+});
+
+addButtonToToolbar({
+  title: 'Rotate Delta 30',
+  onClick: () => {
+    // Get the rendering engine
+    const renderingEngine = getRenderingEngine(renderingEngineId);
+
+    // Get the stack viewport
+    const viewport = <Types.IStackViewport>(
+      renderingEngine.getViewport(viewportId)
+    );
+
+    const { rotation } = viewport.getProperties();
+    viewport.setProperties({ rotation: rotation + 30 });
 
     viewport.render();
   },
--- a/packages/core/examples/stackAPI/index.ts
+++ b/packages/core/examples/stackAPI/index.ts
@@ -59,10 +59,14 @@ element.addEventListener(Events.CAMERA_MODIFIED, (_) => {
     renderingEngine.getViewport(viewportId)
   );
 
+  if (!viewport) {
+    return;
+  }
+
   const { flipHorizontal, flipVertical } = viewport.getCamera();
   const { rotation } = viewport.getProperties();
 
-  rotationInfo.innerText = `Rotation: ${rotation}`;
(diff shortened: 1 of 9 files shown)


packages/core/examples/stackAPI/index.ts
```diff
@@ -161,6 +161,41 @@ addButtonToToolbar({
...
+    );
+
+    viewport.setProperties({ rotation: 150 });
+
+    viewport.render();
+  },
+});
+
+addButtonToToolbar({
+  title: 'Rotate Delta 30',
```
jbocce, line 182: I added this button because it is most likely to be used by a third party app that just does incremental rotation.

packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -1706,8 +1730,6 @@ class StackViewport extends Viewport implements IStackViewport {
         panCache as Point3
       );
 
-      // Restore rotation for the new slice of the image
-      this.rotationCache = 0;
```
jbocce, line 1710: I have no idea why this was being done in the past, but I am pretty sure it is not correct.

packages/core/src/RenderingEngine/Viewport.ts
```diff
@@ -990,13 +989,17 @@ class Viewport implements IViewport {
...
+        rotation: this.getRotation(),
       };
 
       triggerEvent(this.element, Events.CAMERA_MODIFIED, eventDetail);
     }
   }
 
+  protected getRotation(): number {
+    return 0;
+  }
```
sedghi, line 1001: I'm lost, what is going on here?

packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -1674,20 +1697,21 @@ class StackViewport extends Viewport implements IStackViewport {
...
-      const rotationCache = this.rotationCache;
+      const rotation = this.rotationCache;
 
       // Reset the camera to point to the new slice location, reset camera doesn't
       // modify the direction of projection and viewUp
       this.resetCameraNoEvent();
 
       // restore the rotation cache for the new slice
-      this.setRotation(rotationCache, rotationCache);
+      this.rotationCache = rotation;
```
sedghi, line 1707: is `resetCameraNoEvent` changing the `this.rotationCache`? why we are again re assignging the rotation cache in the next line?

packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -816,13 +816,38 @@ class StackViewport extends Viewport implements IStackViewport {
...
+    // A similar thing holds true for vertical flips below.
+    //
+    if (flipHorizontal) {
+      this.rotationCache = -this.rotationCache;
+    }
+
+    if (flipVertical) {
+      this.rotationCache = -this.rotationCache;
+    }
+  }
```
sedghi, line 838: This is the only tricky part ....
So if it is flipped, and someone do the `getProperties` the rotation will be negative?

sedghi (commented):
see my comments

packages/core/src/RenderingEngine/Viewport.ts
jbocce: Viewport.rotation is not really needed. However, `triggerCameraModifiedEventIfNecessary` needs a rotation and gets it from here. Subclasses that incorporate rotation can override this hook method to provide the rotation.

packages/core/src/RenderingEngine/StackViewport.ts
jbocce: Yes `resetCameraNoEvent` does change `this.rotationCache` as is mentioned in the comment on line 1699. That is why we are restoring it here.

packages/core/src/RenderingEngine/StackViewport.ts
jbocce: That is correct. However we could make it positive if that is desirable.

packages/core/src/RenderingEngine/Viewport.ts
jbocce: Notice that `StackViewport.getRotation` overrides this and returns `this.rotatiocCache`.

packages/core/src/RenderingEngine/StackViewport.ts
jbocce: Hmm.. I should change the `const` to be `rotationCache`.

packages/core/src/RenderingEngine/Viewport.ts
jbocce: I suppose it is also worth mentioning here that in the base `Viewport` class, `this.rotation` was always zero and that is why `getRotation` in the base `Viewport` class returns zero by default.

packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -817,13 +815,69 @@ class StackViewport extends Viewport implements IStackViewport {
...
+    // The start view up vector without any rotation, but incorporating any vertical flip.
+    const startViewUp = flipVertical
+      ? vec3.negate(vec3.create(), this.initialViewUp)
+      : this.initialViewUp;
+
+    // The dot product of the start and end view up vectors.
+    const startToEndViewUpDot = vec3.dot(startViewUp, endViewUp);
+
+    // The angle between the start and end view up vectors.
+    const startToEndViewUpAngle =
```
jbocce, line 854: Should we do some rounding here? What would you recommend?

packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -817,13 +815,69 @@ class StackViewport extends Viewport implements IStackViewport {
...
-  private setRotation(rotationCache: number, rotation: number): void {
+  getRotation = (): number => {
+    if (this.useCPURendering) {
+      return this.getRotationCPU();
+    } else {
+      return this.getRotationGPU();
+    }
+  };
+
+  private getRotationCPU = (): number => {
```
jbocce, line 826: I suspect CPU rendering has issues with flip and rotate as well which I did not address. Should I?

packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -859,22 +913,26 @@ class StackViewport extends Viewport implements IStackViewport {
...
-    this.rotation = rotation;
   }
 
-  private setRotationGPU(rotationCache: number, rotation: number): void {
+  private setRotationGPU(rotation: number): void {
+    const { flipVertical } = this.getCamera();
+
     // Moving back to zero rotation, for new scrolled slice rotation is 0 after camera reset
-    this.getVtkActiveCamera().roll(rotationCache);
+    const initialViewUp = flipVertical
```
jbocce, line 926: A nice way to undo the rotation/roll without having to know the existing roll/rotation.

packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -1256,7 +1314,6 @@ class StackViewport extends Viewport implements IStackViewport {
     this.currentImageIdIndex = currentImageIdIndex;
     this.targetImageIdIndex = currentImageIdIndex;
     this.stackInvalidated = true;
-    this.rotationCache = 0;
```
jbocce, line 1259: _resetProperties on line 1321 takes care of this since rotation is now solely fetched from the camera.

packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -650,7 +649,6 @@ class StackViewport extends Viewport implements IStackViewport {
     this.setProperties(
       {
         voiRange: this.voiRange,
-        rotation: this.rotation,
```
jbocce, line 653: There is no longer a cached rotation so no need to reset it.

sedghi (approved):
Great work!

PR #417, opened 2023-02-02
fix(doubleClick): mouse up events are not being propagated at all

jbocce:
moved the mouse click/down timeout detection back into mouseDownListener and the timeout is now cleared on the subsequent 'mouseup' event.

### DIFF
--- a/packages/tools/src/eventListeners/mouse/mouseDownListener.ts
+++ b/packages/tools/src/eventListeners/mouse/mouseDownListener.ts
@@ -149,6 +149,11 @@ function mouseDownListener(evt: MouseEvent) {
   state.renderingEngineId = renderingEngineId;
   state.viewportId = viewportId;
 
+  state.preventClickTimeout = setTimeout(
+    _preventClickHandler,
+    state.clickDelay
+  );
+
   // Prevent CornerstoneToolsMouseMove while mouse is down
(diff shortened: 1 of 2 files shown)


packages/tools/src/eventListeners/mouse/mouseDownListener.ts
```diff
@@ -149,6 +149,11 @@ function mouseDownListener(evt: MouseEvent) {
   state.renderingEngineId = renderingEngineId;
   state.viewportId = viewportId;
 
+  state.preventClickTimeout = setTimeout(
```
jbocce, line 152: Start the timeout as soon as the mouse goes down.

packages/tools/src/eventListeners/mouse/mouseDownListener.ts
```diff
@@ -267,6 +267,9 @@ function _onMouseDrag(evt: MouseEvent) {
  * @param evt - The mouse event.
  */
 function _onMouseUp(evt: MouseEvent): void {
+  // Cancel the timeout preventing the click event from triggering
+  clearTimeout(state.preventClickTimeout);
```
jbocce, line 271: Clear the timeout as soon as the mouse goes up.

wayfarer3130 (approved):

PR #420, opened 2023-02-04
feat(rendering): 16 bit texture support with flag

Ouwen:
Adds 16 bit texture support to cornerstone3D
  - CSWIL will decide on array type to match native storage format. In the case native format is unsigned int, if the scaling params are negative, prescale = true will return back a signed int.
  - Streaming Volume does require the array buffer type to be predetermined (targetBuffer must be specified). This is because all incoming slices need to adhere to one data type.
  - Block Height is set to 1 in the case norm16 is used due to this chromium bug: https://bugs.chromium.org/p/chromium/issues/detail?id=1408247. In the case preferSizeOverAccuracy is used block height for texture transfers can be calculated as usual.
  
 External PR (https://github.com/cornerstonejs/cornerstoneWADOImageLoader/pull/509): 
  - CSWIL needs to implement scaling logic. When a buffer type is not provided native is used (using signed versions of array buffers in the case scale is negative). When a buffer type is provided scaling will occur with possible over/underflows.

### DIFF
--- a/packages/core/src/RenderingEngine/StackViewport.ts
+++ b/packages/core/src/RenderingEngine/StackViewport.ts
@@ -6,6 +6,7 @@ import vtkCamera from '@kitware/vtk.js/Rendering/Core/Camera';
 import { vec2, vec3, mat4 } from 'gl-matrix';
 import vtkImageMapper from '@kitware/vtk.js/Rendering/Core/ImageMapper';
 import vtkImageSlice from '@kitware/vtk.js/Rendering/Core/ImageSlice';
+import { getConfiguration } from '../init';
 import * as metaData from '../metaData';
 import Viewport from './Viewport';
 import eventTarget from '../eventTarget';
@@ -1419,15 +1420,23 @@ class StackViewport extends Viewport implements IStackViewport {
     bitsAllocated,
     numComps,
     numVoxels,
+    TypedArray,
   }): void {
     let pixelArray;
+    const { useNorm16Texture, preferSizeOverAccuracy } =
+      getConfiguration().rendering;
+    const use16BitDataType = useNorm16Texture || preferSizeOverAccuracy;
+
     switch (bitsAllocated) {
       case 8:
         pixelArray = new Uint8Array(numVoxels * numComps);
         break;
-
       case 16:
-        pixelArray = new Float32Array(numVoxels * numComps);
+        if (use16BitDataType) {
+          pixelArray = new TypedArray(numVoxels * numComps);
+        } else {
+          pixelArray = new Float32Array(numVoxels * numComps);
+        }
 
         break;
       case 24:
@@ -1516,13 +1525,13 @@ class StackViewport extends Viewport implements IStackViewport {
     if (!imageData) {
       return false;
     }
-
     const [xSpacing, ySpacing] = imageData.getSpacing();
     const [xVoxels, yVoxels] = imageData.getDimensions();
     const imagePlaneModule = this._getImagePlaneModule(image.imageId);
     const direction = imageData.getDirection();
     const rowCosines = direction.slice(0, 3);
     const columnCosines = direction.slice(3, 6);
+    const dataType = imageData.getPointData().getScalars().getDataType();
 
     // using spacing, size, and direction only for now
     return (
@@ -1533,7 +1542,8 @@ class StackViewport extends Viewport implements IStackViewport {
       xVoxels === image.columns &&
       yVoxels === image.rows &&
       isEqual(imagePlaneModule.rowCosines, <Point3>rowCosines) &&
-      isEqual(imagePlaneModule.columnCosines, <Point3>columnCosines)
+      isEqual(imagePlaneModule.columnCosines, <Point3>columnCosines) &&
+      dataType === image.getPixelData().constructor.name
     );
   }
 
@@ -1557,7 +1567,11 @@ class StackViewport extends Viewport implements IStackViewport {
     //    from the loaded Cornerstone image
     const pixelData = image.getPixelData();
     const scalars = this._imageData.getPointData().getScalars();
-    const scalarData = scalars.getData() as Uint8Array | Float32Array;
+    const scalarData = scalars.getData() as
+      | Uint8Array
+      | Float32Array
+      | Uint16Array
+      | Int16Array;
 
     if (image.rgba || isRgbaSourceRgbDest(pixelData, scalarData)) {
       if (!image.rgba) {
@@ -1715,19 +1729,10 @@ class StackViewport extends Viewport implements IStackViewport {
         );
       }
 
-      // Todo: Note that eventually all viewport data is converted into Float32Array,
-      // we use it here for the purpose of scaling for now.
-      const type = 'Float32Array';
-
       const priority = -5;
       const requestType = RequestType.Interaction;
       const additionalDetails = { imageId };
       const options = {
-        targetBuffer: {
-          type,
-          offset: null,
-          length: null,
-        },
         preScale: {
           enabled: true,
         },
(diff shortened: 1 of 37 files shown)


Ouwen:
Credit goes to @sedghi for initial commits here: #336, commit should be coauthored by him.

Ouwen:
pending vtk 26.3.3 release

Ouwen:
@sedghi ready for review

packages/core/src/init.ts
```diff
@@ -1,24 +1,61 @@
...
+  // ...
+};
+
+let config = {
+  detectGPU: {},
+  rendering: {
+    preferSizeOverAccuracy: true,
+    useCPURendering: false,
+    hasNorm16TextureSupport: _hasNorm16TextureSupport(),
+  },
```
Ouwen, line 27: @sedghi based on our discussion, this would default to false for now for both preferSizeOverAccuracy and hasNorm16TextureSupport

packages/tools/src/utilities/index.ts
```diff
@@ -5,7 +5,6 @@ import {
 
 // Lodash/common JS functionality
 import debounce from './debounce';
-import deepMerge from './deepMerge';
 import throttle from './throttle';
```
sedghi, line 8: so you are moving the deepMerge into the core (which is fine), but removing it here, but not removing it from the src/utilities/deepMerge? I'm confusied

packages/streaming-image-volume-loader/src/helpers/scaleArray.ts
```diff
@@ -8,9 +8,9 @@ import type { Types } from '@cornerstonejs/core';
  * @returns The array is being scaled
  */
 export default function scaleArray(
-  array: Float32Array | Uint8Array,
+  array: Float32Array | Uint8Array | Uint16Array | Int16Array,
   scalingParameters: Types.ScalingParameters
-): Float32Array | Uint8Array {
+): Float32Array | Uint8Array | Uint16Array | Int16Array | undefined {
```
sedghi, line 13: there will be no undefined, each return has a type, why undefined?

packages/streaming-image-volume-loader/src/cornerstoneStreamingImageVolumeLoader.ts
```diff
@@ -127,60 +151,72 @@ function cornerstoneStreamingImageVolumeLoader(
...
+          : new Uint8Array(length);
         break;
 
       case 16:
+        sizeInBytes = length * 2;
+        if (use16BitDataType && (signed || hasNegativeRescale)) {
+          if (!cache.isCacheable(sizeInBytes)) {
+            throw new Error(Enums.Events.CACHE_SIZE_EXCEEDED);
+          }
+          cache.decacheIfNecessaryUntilBytesAvailable(sizeInBytes);
```
sedghi, line 182: can you refactor this?

packages/core/src/init.ts
```diff
@@ -1,24 +1,61 @@
...
+    hasNorm16TextureSupport: false, // _hasNorm16TextureSupport(),
+  },
+  // cache
+  // ...
+};
+
+let config = {
+  detectGPU: {},
+  rendering: {
+    preferSizeOverAccuracy: true,
```
sedghi, line 24: this should be default false

packages/core/src/index.ts
```diff
@@ -58,8 +62,14 @@ import {
...
 
 export {
+  // init
   init,
+  hasNorm16TextureSupport,
+  hasActiveWebGLContext,
   isCornerstoneInitialized,
+  // configs
+  getConfiguration,
+  setPreferSizeOverAccuracy,
```
sedghi, line 72: These are getting out of hand, can we put them under the configUtils?

packages/core/src/types/Cornerstone3DConfig.ts
```diff
@@ -0,0 +1,27 @@
...
+    // Read more in the following Pull Request:
+    // 1. HalfFloat: https://github.com/Kitware/vtk-js/pull/2046
+    // 2. Norm16: https://github.com/Kitware/vtk-js/pull/2058
+    preferSizeOverAccuracy: boolean;
+    // Whether the EXT_texture_norm16 extension is supported by the browser.
+    // WebGL 2 report (link: https://webglreport.com/?v=2) can be used to check
+    // if the browser supports this extension.
+    // In case the browser supports this extension, instead of using 32bit float
+    // textures, 16bit float textures will be used to reduce the memory usage where
+    // possible.
```
sedghi, line 21: Thanks a lot for the detailed explanation

sedghi: Could we add the chromium and webkit links to mention there are bugs still, and might not be a garantee yet that it works?

packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -1263,15 +1264,27 @@ class StackViewport extends Viewport implements IStackViewport {
...
-        pixelArray = new Float32Array(numVoxels * numComps);
+        if (use16BitDataType) {
+          if (isSigned) {
+            pixelArray = new Int16Array(numVoxels * numComps);
+          } else {
+            pixelArray = new Uint16Array(numVoxels * numComps);
+          }
+        } else {
+          pixelArray = new Float32Array(numVoxels * numComps);
+        }
```
sedghi, line 1288: I'm confused, why we don't have the complex logic which involved isSigned and isNegative inside the volume loader here?

packages/core/src/RenderingEngine/vtkClasses/vtkStreamingOpenGLTexture.js
```diff
@@ -142,13 +155,25 @@ function vtkStreamingOpenGLTexture(publicAPI, model) {
...
+        (TypedArrayConstructor === Uint16Array ||
+          TypedArrayConstructor === Int16Array)
+      ) {
+        for (let idx = 0; idx < dataView.length; idx++) {
+          dataView[idx] = HalfFloat.toHalf(dataView[idx]);
+        }
+        if (TypedArrayConstructor === Int16Array) {
+          dataView = new Uint16Array(dataView);
+        }
+      }
```
sedghi, line 175: can we have some comment above this of what it is doing?

sedghi (commented):
Great work! Thanks, it is getting there

packages/tools/src/utilities/index.ts
Ouwen: I think this was actually from your branch, I can remove the actual file too

Ouwen: actually i'm not sure what you mean by the above, the `src/utilities/deepMerge` is indeed moved?

packages/streaming-image-volume-loader/src/StreamingImageVolume.ts
```diff
@@ -391,18 +381,38 @@ export default class StreamingImageVolume extends ImageVolume {
...
           }
           scalarData.set(floatView, offset / bytesInFloat);
         }
+        if (scalarData instanceof Int16Array) {
+          const bytesInInt16 = 2;
+          const intView = new Int16Array(pixelData);
+          if (intView.length !== length) {
+            throw 'Error pixelData length does not match frame length';
+          }
+          scalarData.set(intView, offset / bytesInInt16);
```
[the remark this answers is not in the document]
Ouwen: yep it is similar to the way Float32 is handled in upstream, I will add a comment.

packages/core/src/init.ts
sedghi: I think we should have the hasNorm16TextureSupport settable, which make me think we should rename it to `useNorm16Texture` since it has bugs in browsers and user can set it as necessary

packages/core/src/RenderingEngine/StackViewport.ts
Ouwen: Revisiting, the `_checkVTKImageDataMatchesCornerstoneImage` should not need to check length and checking type is correct. The isSigned logic is removed in favor of just checking the dtype in the case use16bit is true.

Ouwen:
@sedghi the above is ready for review with previous comments addressed. Let me know what you think!

PR #427, opened 2023-02-08
feat(adapters): Add adapters for Rectangle, Angle and fix generate DICOM

wayfarer3130:
Add additional adapters for Rectangle, Angle, PlanarFreehand
Fix the generated DICOM so it included all the patient/study attributes so it doesn't fail to send against a validating PACS.  This is a critical bug in that the generated DICOM was missing the Patient ID attribute, and thus would fail to store on any DICOM compliant PACS back end.

### DIFF
--- a/packages/adapters/rollup.config.mjs
+++ b/packages/adapters/rollup.config.mjs
@@ -8,10 +8,7 @@ import { readFileSync } from "fs";
 const pkg = JSON.parse(readFileSync("package.json", { encoding: "utf8" }));
 
 export default {
-    external: ["dcmjs"],
-    globals: {
-        dcmjs: "dcmjs"
-    },
+    external: ["dcmjs", "gl-matrix", "lodash.clonedeep", "ndarray"],
     input: pkg.src || "src/index.ts",
     output: [
         // {
--- a/packages/adapters/rollup.config.mjs
+++ b/packages/adapters/rollup.config.mjs
@@ -5,35 +5,38 @@ import commonjs from "@rollup/plugin-commonjs";
 import json from "@rollup/plugin-json";
 import { readFileSync } from "fs";
 
-const pkg = JSON.parse(readFileSync('package.json', {encoding: 'utf8'}));
-
+const pkg = JSON.parse(readFileSync("package.json", { encoding: "utf8" }));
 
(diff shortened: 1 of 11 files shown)


packages/adapters/src/adapters/Cornerstone3D/MeasurementReport.ts
```diff
@@ -132,25 +136,21 @@ export default class MeasurementReport {
         return _meta;
     }
 
-    static generateDerivationSourceDataset(
-        StudyInstanceUID,
-        SeriesInstanceUID
-    ) {
+    static generateDerivationSourceDataset = instance => {
```
wayfarer3130, line 139: Use the actual instance data for the derived data, as it is used for generating the overall report.

packages/tools/src/tools/annotation/PlanarFreehandROITool.ts
```diff
@@ -582,9 +582,11 @@ class PlanarFreehandROITool extends AnnotationTool {
     if (!(isDrawing || isEditingOpen || isEditingClosed)) {
       // No annotations are currently being modified, so we can just use the
       // render contour method to render all of them
-      annotations.forEach((annotation) =>
-        this.renderContour(enabledElement, svgDrawingHelper, annotation)
-      );
+      annotations.forEach((annotation) => {
+        console.log('annotation render', annotation);
```
sedghi, line 586: remove this log please

sedghi (changes requested):
Please remove the unnecessarsy console log.

packages/adapters/rollup.config.mjs
```diff
@@ -5,35 +5,38 @@ import commonjs from "@rollup/plugin-commonjs";
...
-    commonjs(),
-    typescript(),
-//    globals(),
-    // builtins(),
-    babel({
-      exclude: "node_modules/**"
-    }),
-    json()
-  ]
+    external: ["dcmjs"],
```
wayfarer3130, line 11: Sorry about the refactor - the externals library is this line here - it also needs the globals, but I don't quite know how to figure out the global name for the other externals.

sedghi (approved):

PR #435, opened 2023-02-15
feat(VolumeViewport): Add optional flag to assume fallbacks for ZSpacing for volumes

rodrigobasilio2022:
This PR deals with the possibility of having multiframe images without metadata position inforamtion for each frame. For that we simply use the sliceThickness as the zSpacing


packages/streaming-image-volume-loader/src/helpers/sortImageIdsAndGetSpacing.ts
```diff
@@ -126,11 +126,22 @@ export default function sortImageIdsAndGetSpacing(
...
   );
 
+  // if zspacing = 0, use sliceThickness if possible
+  if (zSpacing === 0) {
+    if (sliceThickness) {
+      console.log('Could not calculate zSpacing. Using sliceThickness');
+      zSpacing = sliceThickness;
+    } else {
+      console.log(
+        'Could not calculate zSpacing. The VolumeViewport is compromised.'
```
wayfarer3130, line 141: Set zSpacing to 1 - not a great option here, but at least it will render.

wayfarer3130:
Could you please also fix the frame extraction in line 31 of metaDataManager.js

packages/streaming-image-volume-loader/src/helpers/sortImageIdsAndGetSpacing.ts
sedghi: agree

packages/streaming-image-volume-loader/src/helpers/sortImageIdsAndGetSpacing.ts
```diff
@@ -126,11 +126,22 @@ export default function sortImageIdsAndGetSpacing(
...
+      );
+    }
+  }
```
sedghi, line 144: This should only be for the `NM`, put a comment above it please and also explain the weirdness for NM

packages/streaming-image-volume-loader/src/helpers/sortImageIdsAndGetSpacing.ts
rodrigobasilio2022: Done

packages/streaming-image-volume-loader/src/helpers/sortImageIdsAndGetSpacing.ts
rodrigobasilio2022: I dont agree. We dont know for sure that this only happens in NM

rodrigobasilio2022:
I updated the code. Please see if fits.

packages/streaming-image-volume-loader/src/helpers/sortImageIdsAndGetSpacing.ts
sedghi: @wayfarer3130 what do you think?

wayfarer3130: The NM behaviour is quite complex, and in fact you can't assume that the slice thickness even corresponds to positioning of elements.  There is a whole module in DICOM dedicated to how the slices are ordered/managed - it could be multiple volumes, with several at the same position, differing in time.
I don't know enough offhand to understand what the spacing is for Tomo MG, but those also have different setup for positioning.

sedghi (approved):

PR #436, opened 2023-02-16
Fix stack rotate tool for volume viewport

sharon-shrits:
Was written with the help of @wayfarer3130

### DIFF
--- a/packages/tools/src/tools/StackRotateTool.ts
+++ b/packages/tools/src/tools/StackRotateTool.ts
@@ -1,5 +1,9 @@
-import { getEnabledElement, Types } from '@cornerstonejs/core';
-import { vec3 } from 'gl-matrix';
+import {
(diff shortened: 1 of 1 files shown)


sedghi (approved):

sedghi (changes requested):
The tool name is `stackRotate` so we need to change its name it is supporting volume as well

wayfarer3130:
@sharon-shrits  - could you rename the tool to PlanarRotateTool so that it clearly describes what it does?

sharon-shrits:
@sedghi @wayfarer3130 Renamed to PlanarRotateTool

sharon-shrits:
@sedghi @wayfarer3130 Renamed to PlanarRotateTool

sedghi (approved):
Looks great, thanks a lot @sharon-shrits and Bill

PR #438, opened 2023-02-16
feat(4D): added support for 4D data rendering

lscoder:
cornerstone now loads 4D volumes when using `cornerstoneStreamingDynamicImageVolume` schema. Currently it is implemented only for (0054,1300) FrameReferenceTime tag. Minor changes are needed to add support for some other 4D tags as follow:

- (0018,1060) Trigger Time
- (0018,0081) Echo Time
- (0018,0086) Echo Number
- (0020,0100) Temporal Position Identifier

# changes:
- `scalarData` is now private to `ImageVolume` and need to be accessed via `getScalarData()`


packages/core/src/types/IImageVolume.ts
```diff
@@ -1,6 +1,17 @@
 import type { vtkImageData } from '@kitware/vtk.js/Common/DataModel/ImageData';
 import { Metadata, Point3, IImageLoadObject, Mat3 } from '../types';
 
+type VolumeTimePoints = {
+  timePoints: {
+    /** imageIds of each timepoint  */
+    imageIds: Array<string>;
+    /** volume scalar data  */
+    scalarData: Float32Array | Uint8Array;
+  }[];
```
sedghi, line 10: so the volume has a root `imageIds` right now, you are saying we add another one `timePoints.imageIds`? What do you think about using the same `imageIds` and `scalarData` and make them [] for 3D and [][] for 4D?

packages/core/src/types/IVolume.ts
```diff
@@ -20,7 +20,7 @@ interface IVolume {
   /** volume direction */
   direction: Mat3;
   /** volume scalarData */
-  scalarData: Float32Array | Uint8Array;
+  scalarData?: Float32Array | Uint8Array;
```
sedghi, line 23: I don't like how we are making this optional, since you are gathering the scalarData inside `timePointsData`. To me we should have the same interface, but change `imageIds` to `[] | [][]` and same for `scalarData`

packages/core/src/loaders/volumeLoader.ts
```diff
@@ -34,34 +34,35 @@ interface LocalVolumeOptions {
...
 
   imageData.setDimensions(dimensions);
   imageData.setSpacing(spacing);
   imageData.setDirection(direction);
   imageData.setOrigin(origin);
-  imageData.getPointData().setScalars(scalarArray);
+
+  scalarDataArrays.forEach((scalarDataArray, i) => {
+    const vtkScalarArray = vtkDataArray.newInstance({
+      name: `timePoint-${i}`,
```
sedghi, line 57: why are we creating `vtkDataArray` with `timePoint-0` even for 3D? let's just branch out for 3D and 4D and have separate logics

packages/core/src/loaders/volumeLoader.ts
```diff
@@ -34,34 +34,35 @@ interface LocalVolumeOptions {
...
+      numberOfComponents: numComponents,
+      values: scalarDataArray,
+    });
+
+    imageData.getPointData().addArray(vtkScalarArray);
+  });
+
+  imageData.getPointData().setActiveScalars('timePoint-0');
```
sedghi, line 65: same here, `timePoint` in the context of 3D doesn't makes sense

packages/core/src/requestPool/requestPoolManager.ts
```diff
@@ -289,9 +289,12 @@ class RequestPoolManager {
     }
 
     if (this.grabDelay !== undefined) {
-      this.timeoutHandle = window.setTimeout(() => {
-        this.startGrabbing();
-      }, this.grabDelay);
+      if (!this.timeoutHandle) {
+        this.timeoutHandle = window.setTimeout(() => {
+          this.timeoutHandle = 0;
```
sedghi, line 294: null?

packages/streaming-image-volume-loader/src/StreamingImageVolume.ts
```diff
@@ -37,17 +39,24 @@ export default class StreamingImageVolume extends ImageVolume {
     super(imageVolumeProperties);
     this.imageIds = streamingProperties.imageIds;
     this.loadStatus = streamingProperties.loadStatus;
+    this.cornerstoneImageMetaData = null;
```
sedghi, line 42: maybe initialize it above if it is null.

packages/streaming-image-volume-loader/src/StreamingImageVolume.ts
```diff
@@ -83,18 +92,49 @@ export default class StreamingImageVolume extends ImageVolume {
...
+
+  public get numFrames(): number {
+    return this.imageIds.length;
+  }
+
+  public getScalarDataArrays(): Array<Float32Array | Uint8Array> {
+    return [this.scalarData];
+  }
+
+  public getActiveScalarData(): Float32Array | Uint8Array {
```
sedghi, line 113: `getActiveScalarData` is weird name for 3D, can we just call it `getScalarData` and then for 4D it becomes `getScalarData(timepointIndex)` ?

packages/streaming-image-volume-loader/src/StreamingImageVolume.ts
```diff
@@ -83,18 +92,49 @@ export default class StreamingImageVolume extends ImageVolume {
...
+
+  public getScalarDataByImageIdIndex(
+    imageIdIndex: number
+  ): Float32Array | Uint8Array {
+    if (imageIdIndex < 0 || imageIdIndex >= this.imageIds.length) {
+      throw new Error('imageIdIndex out of range');
+    }
+
+    return this.scalarData;
+  }
```
sedghi, line 125: This is not needed in 3D, just have it for 4D please

packages/streaming-image-volume-loader/src/StreamingImageVolume.ts
```diff
@@ -83,18 +92,49 @@ export default class StreamingImageVolume extends ImageVolume {
...
-        return false;
-      }
+      vtkOpenGLTexture.setUpdatedFrame(i);
     }
 
-    return true;
-  };
+    imageData.modified();
+    autoLoad(this.volumeId);
+  }
```
sedghi, line 137: this should accept a param as `immediate` or something similar to decide whether to call `autoLoad` or not

packages/streaming-image-volume-loader/src/StreamingImageVolume.ts
```diff
@@ -83,18 +92,49 @@ export default class StreamingImageVolume extends ImageVolume {
...
-    const numFrames = imageIds.length;
+  protected initialize(): void {
+    this._createCornerstoneImageMetaData();
+  }
+
+  // The frameIndex is equal to imageIdIndex for 3D datasets
+  // but they are different for 4D datasets
+  protected imageIdIndexToFrameIndex(imageIdIndex: number): number {
+    return imageIdIndex % this.numFrames;
+  }
```
sedghi, line 103: again this does not need to be implemented for 3d

packages/streaming-image-volume-loader/src/StreamingImageVolume.ts
```diff
@@ -412,10 +452,19 @@ export default class StreamingImageVolume extends ImageVolume {
...
+    const requests = imageIds.map((imageId, frameIndex) => {
+      const imageIdIndex = this.imageIds.indexOf(imageId);
+
       if (cachedFrames[imageIdIndex]) {
-        framesLoaded++;
-        framesProcessed++;
+        this.framesLoaded++;
+        this.framesProcessed++;
         return;
       }
```
sedghi, line 469: same here, we don't need to explain what 4d does in the base class. This should be inside the 4d

sedghi (changes requested):
Good start, there is a lot of methods for the 3D which doesn't make sense such as `imageIdIndexToFrameIndex` , `getScalarDataByImageIdIndex` , `getActiveScalarData`. I guess the design of the inheritance hierarchy is not optimal. Maybe we need a `BaseStreamingImageVolume` and then `StreamingImageVolume3D` and `StreamingImageVolume4D`. What do you think

packages/core/src/loaders/volumeLoader.ts
lscoder: There are two new methods to create `vtkDataArrays`

packages/core/src/loaders/volumeLoader.ts
lscoder: It's back to `Pixels` for 3D volumes

packages/core/src/types/IImageVolume.ts
lscoder: Removed! `timePoints` is now private to 4D volumes only

packages/core/src/types/IVolume.ts
lscoder: I tried changing imageIds from `string[]` to `string[][]` but I started facing issues outside of the volume context because some functions access `imageIds` directly (eg: [cache](https://github.com/cornerstonejs/cornerstone3D-beta/blob/main/packages/core/src/cache/cache.ts#L463-L475))

packages/core/src/cache/classes/ImageVolume.ts
```diff
@@ -16,7 +29,7 @@ export class ImageVolume implements IImageVolume {
   /** volume origin, Note this is an opinionated origin for the volume */
   origin: Point3;
   /** volume scalar data  */
-  scalarData: Float32Array | Uint8Array;
+  scalarData: VolumeScalarData | Array<VolumeScalarData>;
```
sedghi, line 32: add a comment on top saying 3D or 4D

packages/core/src/cache/cache.ts
```diff
@@ -455,22 +455,16 @@ class Cache implements ICache {
...
       const cachedVolume = this._volumeCache.get(volumeId);
+      const { volume } = cachedVolume;
 
-      if (!cachedVolume.volume) {
+      if (!volume?.imageIds?.length) {
         return;
       }
 
-      let { imageIds } = cachedVolume.volume;
+      const imageIdIndex = volume.getImageURIIndex(imageIdToUse);
```
sedghi, line 464: nice

packages/core/src/requestPool/requestPoolManager.ts
```diff
@@ -289,9 +289,12 @@ class RequestPoolManager {
     }
 
     if (this.grabDelay !== undefined) {
-      this.timeoutHandle = window.setTimeout(() => {
-        this.startGrabbing();
-      }, this.grabDelay);
+      if (!this.timeoutHandle) {
+        this.timeoutHandle = window.setTimeout(() => {
+          this.timeoutHandle = null;
```
sedghi, line 294: add a comment why

packages/core/src/cache/classes/ImageVolume.ts
sedghi: should'nt scalarData also be private? since they should be accessed via `.getScalarData`?

utils/demo/helpers/createImageIdsAndCacheMetaData.js
```diff
@@ -64,13 +65,18 @@ export default async function createImageIdsAndCacheMetaData({
...
   // if not multiframe data available, just returns the same list given
   imageIds = convertMultiframeImageIds(imageIds);
+
   imageIds.forEach((imageId) => {
     let instanceMetaData =
       cornerstoneWADOImageLoader.wadors.metaDataManager.get(imageId);
-    instanceMetaData = JSON.parse(JSON.stringify(instanceMetaData));
+
+    // It was using JSON.parse(JSON.stringify(...)) before but it is 8x slower
+    instanceMetaData = removeInvalidTags(instanceMetaData);
```
sedghi, line 78: i'm not sure why we were doing the json strigify either, do we really need this? does it really throw error if we have null or undefiend?

packages/core/src/loaders/volumeLoader.ts
```diff
@@ -34,34 +34,70 @@ interface LocalVolumeOptions {
...
   }
 
-  const scalarArray = vtkDataArray.newInstance({
-    name: 'Pixels',
-    numberOfComponents: numComponents,
-    values: scalarData,
-  });
-
   const imageData = vtkImageData.newInstance();
+  const scalarData = volume.scalarData;
```
sedghi, line 86: shoudn't this be `getScalarData` too?

packages/streaming-image-volume-loader/src/helpers/splitImageIdsBy4DTags.ts
```diff
@@ -0,0 +1,68 @@
...
+    .sort((a, b) => a - b);
+
+  const imageIdsGroups = sortedFrameReferenceTimes.map((key) =>
+    framesGroups[key].map((item) => item.imageId)
+  );
+
+  return imageIdsGroups;
+}
+
+function splitImageIdsBy4DTags(imageIds: string[]): string[][] {
```
sedghi, line 43: needs doc

packages/streaming-image-volume-loader/src/helpers/splitImageIdsBy4DTags.ts
```diff
@@ -0,0 +1,68 @@
...
+  return array.reduce((rv, x) => {
+    (rv[x[key]] = rv[x[key]] || []).push(x);
+    return rv;
+  }, {});
+};
+
+function splitFramesByFrameReferenceTime(imageIds: string[]): string[][] {
+  const framesMetadata: Array<MappedFrameReferenceTime> = imageIds.map(
+    (imageId: string): MappedFrameReferenceTime => {
+      const petImageModule = metaData.get('petImageModule', imageId);
```
sedghi, line 25: should we add a todo for other tags other than pt? e.g., mr, fmri? etc.?

sedghi (commented):
almost there, see my comments

utils/demo/helpers/createImageIdsAndCacheMetaData.js
lscoder: Yes, it does fail when there are null/undefined values but at least it is taking 110ms vs 936ms before for the PT/CT demo dataset we are using to test it.

The only issue that I see after removing `JSON.stringify` is related to this [change](https://github.com/cornerstonejs/cornerstoneWADOImageLoader/blob/master/src/imageLoader/wadors/metaDataManager.js#L10) because it adds `isMultiFrame: undefined`. We would have to remove it from the json and keep it private to the `metadataManager` context or make sure it's set to `false` instead of `undefined`. But even after making this change I would not be 100% sure if it would work for all scenarios because `JSON.stringify` was added one year ago while that `isMultiFrame` property was added only one month ago. It may have another reason for calling it.

packages/core/src/cache/classes/ImageVolume.ts
lscoder: Yes and I totally agree but making that change will break any code from other devs that are already working on cs3d. May I remove it from the interface?

packages/streaming-image-volume-loader/src/helpers/splitImageIdsBy4DTags.ts
lscoder: Updated

packages/core/src/requestPool/requestPoolManager.ts
lscoder: Done

packages/streaming-image-volume-loader/src/helpers/splitImageIdsBy4DTags.ts
lscoder: Added a  "TODO" at the top of this file.

packages/core/src/cache/classes/ImageVolume.ts
lscoder: Updated the '.scalarData` from `public` to `protected` because children classes access it and fixed some other codes that were accessing it directly.

packages/tools/src/tools/annotation/ProbeTool.ts
```diff
@@ -608,7 +608,9 @@ class ProbeTool extends AnnotationTool {
         continue;
       }
 
-      const { dimensions, scalarData, imageData, metadata } = image;
+      const { dimensions, imageData, metadata } = image;
+      const scalarData =
+        'getScalarData' in image ? image.getScalarData() : image.scalarData;
```
lscoder, line 613: `image` is returned by `getTargetIdImage(...)`. This function returns `IImageData | CPUIImageData | IImageVolume` but only IImageData has `getScalarData()`.

packages/core/src/loaders/volumeLoader.ts
lscoder: Updated since `.scalarData` is now protected

sedghi (approved):
Looks great, thanks a lot !
Can you please update the PR with listing the breaking changes? 
- scalarData is now private and need to be accessed via getScalarData
- ....

PR #441, opened 2023-02-21
feat(dicomImageLoader types): Add types to the dicom image loader

sedghi:
Shout-out to @jmannau and his team at [Pliny](http://pliny.com.au/) for their awesome work on the TypeScript addition to the cornerstoneWADOImageLoader. They were a crucial part of this PR for adding types to this project and their hard work made my job so much easier, which was mostly copy pasting.

---

# Todo
Seems like the ESM build does not work properly with the codecs "exports". 
As per my investigation, the `moduleResolution` in the tsConfig is "node" which is basically "node10" and not newer node versions that supports "export". 

See these references 
- https://github.com/microsoft/TypeScript/issues/51862
- https://github.com/microsoft/TypeScript/issues/50152


packages/dicomImageLoader/package.json
```diff
@@ -21,7 +21,12 @@
     "access": "public"
   },
   "scripts": {
-    "build": "echo 'build not implemented yet'",
+    "build:umd": "cross-env NODE_ENV=production webpack --config .webpack/webpack-dynamic-import.js",
```
wayfarer3130, line 24: Please up above on the "main" import list as the:
module - the typescript file import name
exports - both the inline bundle and the external/JS bundle.

and then build both the inline bundle and the worker bundle.  If the worker is just a simple wrapper for the inline bundle, then it is easy to just ship a single build that has both - no custom compilation/setup required.

packages/dicomImageLoader/src/externalModules.ts
```diff
@@ -13,13 +12,13 @@ const external = {
   },
   get cornerstone() {
     if (!cornerstone) {
-      if (window && window.cornerstone) {
-        cornerstone = window.cornerstone;
+      if (window && (window as any).cornerstone) {
```
wayfarer3130, line 15: Just delete the if condition and put it into the assignment
cornerstone = window && (window as any).cornerstone

packages/dicomImageLoader/examples/customWebWorkerTask/index.html
```diff
@@ -126,20 +126,20 @@ <h1>Example of using a custom web worker</h1>
...
 <script src="../dicomParser.min.js"></script>
 
-<!-- include the cornerstoneWADOImageLoader library -->
-<script src="../dist/cornerstoneWADOImageLoader.bundle.min.js"></script>
+<!-- include the dicomImageLoader library -->
+<script src="../dist/dicomImageLoader.bundle.min.js"></script>
 
 <!-- Lines ONLY required for this example to run without building the project -->
-<script>window.cornerstoneWADOImageLoader || document.write('<script src="https://unpkg.com/cornerstone-wado-image-loader">\x3C/script>')</script>
+<script>window.dicomImageLoader || document.write('<script src="https://unpkg.com/cornerstone-wado-image-loader">\x3C/script>')</script>
```
wayfarer3130, line 133: The unpkg is probably wrong

packages/dicomImageLoader/examples/dicomfile/index.html
```diff
@@ -108,18 +108,18 @@ <h1>Example of displaying a DICOM P10 from the local file system</h1>
...
-<script src="../../dist/cornerstoneWADOImageLoader.bundle.min.js"></script>
+<!-- include the dicomImageLoader library -->
+<script src="../../dist/dicomImageLoader.bundle.min.js"></script>
 
 <!-- uids -->
 <script src="uids.js"></script>
 
 <!-- Lines ONLY required for this example to run without building the project -->
-<script>window.cornerstoneWADOImageLoader || document.write('<script src="https://unpkg.com/cornerstone-wado-image-loader">\x3C/script>')</script>
+<script>window.dicomImageLoader || document.write('<script src="https://unpkg.com/cornerstone-wado-image-loader">\x3C/script>')</script>
```
wayfarer3130, line 118: unpkg again

packages/dicomImageLoader/src/imageLoader/wadors/metaData/getNumberValue.ts
```diff
@@ -1,13 +1,17 @@
...
+  index?: number
+): number {
   const value = getValue(element, index);
 
   if (value === undefined) {
     return;
   }
 
-  return parseFloat(value);
+  return parseFloat(value as string);
```
wayfarer3130, line 14: This isn't a good assumption here - parseFloat will handle string, but also will handle number natively, and some of the big number formats.  Just let typescript complain about it if it isn't happy.

packages/dicomImageLoader/src/imageLoader/wadors/metaData/getValue.ts
```diff
@@ -6,7 +8,11 @@
  * @param [defaultValue] - The default value to return if the element does not exist
  * @returns {*}
  */
-function getValue(element, index, defaultValue) {
+function getValue(
+  element: WADORSMetaDataElement,
+  index?: number,
+  defaultValue?: number | string
+): number[] | string[] | string | number {
```
wayfarer3130, line 15: This is insufficient - there is also boolean, and there is no guarantee you won't get other types.  This should be unknown as a return type because it is hard to guess.  defaultValue should also be unknown, and should allow for the array types just as much as the simple types.

packages/dicomImageLoader/src/imageLoader/wadors/metaData/metaDataProvider.ts
```diff
@@ -99,14 +99,14 @@ function metaDataProvider(type, imageId) {
 
     if (imageOrientationPatient) {
       rowCosines = [
-        parseFloat(imageOrientationPatient[0]),
-        parseFloat(imageOrientationPatient[1]),
-        parseFloat(imageOrientationPatient[2]),
+        parseFloat(imageOrientationPatient[0] as any),
```
wayfarer3130, line 102: getNumberValues already returns an array of number[], so you shouldn't need the parsefloat any longer, nor the cast as any.  If you do need the case, as any is a better cast than as string.  I suggest dropping the cast and letting typescript complain until the response type from getNumberValues is properly typed.

packages/dicomImageLoader/src/imageLoader/wadouri/dataset-from-partial-content.ts
```diff
@@ -52,7 +57,7 @@ function parsePartialByteArray(byteArray) {
     // metadata header and regular datasets, so transfer syntax and
     // other metadata headers aren't included.
     pixelDataSet = dicomParser.parseDicom(byteArray);
-  } catch (err) {
+  } catch (err: any) {
```
wayfarer3130, line 60: This is probably invalid handling - it expects the only reason to fail is a partial dataset.  That being said, it isn't part of this change, so it should be handled later.

packages/dicomImageLoader/src/shared/decoders/decodeJPEG2000.ts
```diff
@@ -1,20 +1,22 @@
 // https://emscripten.org/docs/api_reference/module.html
-import openJpegFactory from '@cornerstonejs/codec-openjpeg/dist/openjpegwasm_decode.js';
+import openJpegFactory from '@cornerstonejs/codec-openjpeg/dist/openjpegwasm_decode';
```
wayfarer3130, line 2: Ideally these should use the exports key value rather than assuming a dist/... path as that allows the paths to be updated/changed, whereas this is a fixed path.

packages/dicomImageLoader/src/types/ImageFrame.ts
```diff
@@ -0,0 +1,32 @@
...
+  columns: number;
+  bitsAllocated: number;
+  bitsStored: number;
+  pixelRepresentation: number;
+  smallestPixelValue: number;
+  largestPixelValue: number;
+  redPaletteColorLookupTableDescriptor: number[];
+  greenPaletteColorLookupTableDescriptor: number[];
+  bluePaletteColorLookupTableDescriptor: number[];
+  redPaletteColorLookupTableData: number[];
```
wayfarer3130, line 15: These can also be bulkdata URIs which need to be loaded, although it looks like that has gotten broken at some point.

packages/dicomImageLoader/src/types/MetadataModules.ts
```diff
@@ -0,0 +1,81 @@
+export type MetaDataTypes =
+  | 'generalSeriesModule'
+  | 'patientStudyModule'
+  | 'imagePlaneModule'
+  | 'imagePixelModule'
+  | 'transferSyntax'
+  | 'sopCommonModule'
+  | string;
+
+export interface DicomDateObject {
```
wayfarer3130, line 10: If this is getting retrieved/parsed, we should always get it as a combined date/time object.  Splitting it up causes all sorts of issues.  If we are going to not parse, then just leaving the value as a string is preferred.

wayfarer3130 (commented):
Needs some updates still, but quite close.

packages/dicomImageLoader/src/imageLoader/wadors/metaData/getValue.ts
jmannau: We could add a generic type to the function rather than return unknown.

```
function getValue<ReturnType = unknown>(
   element: WADORSMetaDataElement,
   index?: number,
   defaultValue?: number | string
 ): ReturnType
```

packages/dicomImageLoader/src/types/ImageFrame.ts
sedghi: So you are saying this can be `number[] | string`?

wayfarer3130 (approved):

sedghi:
Examples don't run in the deploy preview, need to investigate that https://deploy-preview-441--cornerstone-3d-docs.netlify.app/

TardigradeX:
I just tried to use cornestoneDicomImageLoader in my project, but there are no type files available. The compiler does not find them, and I also cant find them in the node_modules folder.

I see here, that in the last commit  [0192389129](https://github.com/cornerstonejs/cornerstone3D/pull/441/commits/0192389129414364bf39afa06edb14483ccf1af5)

everything was reverted to the old cornerstoneWADOImageLoader without TS support. 
Is this the case? does the cornestoneDicomImageLoader not work at the moment? If yes the documentation should be updated.

justinsolomon:
> I just tried to use cornestoneDicomImageLoader in my project, but there are no type files available. The compiler does not find them, and I also cant find them in the node_modules folder.
> 
> I see here, that in the last commit [0192389129](https://github.com/cornerstonejs/cornerstone3D/pull/441/commits/0192389129414364bf39afa06edb14483ccf1af5)
> 
> everything was reverted to the old cornerstoneWADOImageLoader without TS support. Is this the case? does the cornestoneDicomImageLoader not work at the moment? If yes the documentation should be updated.

I am having a similar issue. Trying to use the cornerstone suite of tools in an angular application to load + view local DICOM files. I was able to install the core and tools packages and they seem to be working. But my IDE is giving an error when trying to import dicom-image-loader. Does this package work with typescript?

`Could not find a declaration file for module '@cornerstonejs/dicom-image-loader'. '.../node_modules/@cornerstonejs/dicom-image-loader/dist/cornerstoneDICOMImageLoader.bundle.min.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/cornerstonejs__dicom-image-loader` if it exists or add a new declaration (.d.ts) file containing `declare module '@cornerstonejs/dicom-image-loader';`

sedghi:
Yes i guess, would be great if you add a PR

PR #442, opened 2023-02-21
feat(annotations): rework annotation manager api and enable multi-manager setup

sedghi:
# Changes to Annotation Manager

## Annotation state 
we can now switch between different annotation managers via `csTools.annotations.state.setAnnotationManager(manager)` to set the annotation manager to get used to filter/add/show/save annotations (by default it is our FrameOfReferenceSpecificAnnotationManager) that stores and group annotations based on the `FrameOfReferenceUID`

Some methods in the `annotation state` has been modified to add `annotationGroupSelector` which can be `element` or a string that is passed inside the annotation manager and let the manager _map_ those to a key (`GroupKey`) that is used to group annotations together. In the default annotation manager, the `groupKey` is the `FrameOfReferenceUID`; however, you can write custom managers that store annotations separately per `viewportId`, or `timepoints` (prior vs current reading) or `readerId`, etc.

Here are the changes
- `getAnnotations(element, toolName)` is now `getAnnotations(toolName, annotationGroupSelector)` 
- `addAnnotation(element, annotation)` is now `addAnnotation(annotation, annotationGroupSelector)`
- `getNumberOfAnnotations(toolName, frameOfReferenceUID)` is now `getNumberOfAnnotations(toolName, annotationGroupSelector)`
- `removeAnnotation(annotationUID, element)` is now `removeAnnotation(annotationUID)`
-  `getAnnotation(annotationUID, element?)` is now `getAnnotation(annotationUID)`
-  `removeAllAnnotations(element?)` is now `removeAllAnnotations()` without argument

## FrameOfReferenceSpecificAnnotationManager

- annotationManagers should implement a `getGroupKey` that receives the annotationGroupSelector and returns the key
- old `get(FrameOfReferenceUID, toolName)` has been replaced by `getAnnotations(groupKey, toolName?)`
- old `getAnnotation` now only needs `annotationUID` and not more arguments
- old `getNumberOfAnnotations(toolName, frameOfReferenceUID)` is now `getNumberOfAnnotations(groupKey, toolName?)`
- old `addAnnotation(annotation)` is now `addAnnotation(annotation, groupKey)`
- old `removeAnnotation` now only requires annotationUID
- old `saveAnnotations(FrameOfReferenceUID, toolName?)` is now `saveAnnotations(groupKey?, toolName?)`
- new method `removeAnnotations(groupKey, toolName?)`


common/reviews/api/tools.api.md
```diff
@@ -41,7 +41,7 @@ type ActorSliceRange = {
 };
 
 // @public (undocumented)
-function addAnnotation(element: HTMLDivElement, annotation: Annotation): string;
+function addAnnotation(annotation: Annotation, element?: HTMLDivElement): string;
```
wayfarer3130, line 44: I agree this is probably a good change, but that is definitely a breaking change.  Is there a second name we could add/implement to make it backwards compatible for a while or is it the time still for breaking changes?
Is there a good reason to include the element even optionally?  If it is just for frameOfReferenceUID, then that could be applied directly, or maybe should already be inside annotation?  My vote is to remove it entirely, assuming it has to be in annotation directly.

packages/core/test/stackViewport_cpu_render_test.js
```diff
@@ -23,6 +23,7 @@ const {
   setUseCPURendering,
   resetUseCPURendering,
   CONSTANTS,
+  getEnabledElement,
```
wayfarer3130, line 26: Why is this added without a use below?

packages/core/test/volumeViewport_gpu_render_test.js
```diff
@@ -30,6 +30,7 @@ const {
   volumeLoader,
   utilities,
   setVolumesForViewports,
+  getEnabledElement,
```
wayfarer3130, line 33: Ditto

packages/tools/src/stateManagement/annotation/annotationState.ts
```diff
@@ -40,58 +39,59 @@ function getViewportSpecificAnnotationManager(
   return defaultFrameOfReferenceSpecificAnnotationManager;
 }
 
+function getAnnotationManager(element?: HTMLDivElement) {
```
wayfarer3130, line 42: Is there a good reason not to just use the annotation manager - if they need different managers, then probably the annotation itself should specify which manager, and not the viewport - otherwise, it just isn't clear how one would select the right manager.

packages/tools/src/stateManagement/annotation/annotationState.ts
```diff
@@ -40,58 +39,59 @@ function getViewportSpecificAnnotationManager(
...
  * @param toolName - The name of the tool.
+ * @param FrameOfReferenceUID - The Frame of Reference UID.
+ * @param element - The HTML element.
  * @returns The annotations corresponding to the Frame of Reference and the toolName.
  */
 function getAnnotations(
-  element: HTMLDivElement,
-  toolName: string
+  toolName: string,
+  FrameOfReferenceUID: string,
```
wayfarer3130, line 62: What FOR is used if not specified by an instance?

packages/tools/src/stateManagement/annotation/annotationState.ts
```diff
@@ -40,58 +39,59 @@ function getViewportSpecificAnnotationManager(
...
  *
- * @param element - HTMLDivElement
  * @param annotation - The annotation that is being added to the annotations manager.
+ * @param element - HTMLDivElement
  */
 function addAnnotation(
-  element: HTMLDivElement,
-  annotation: Annotation
+  annotation: Annotation,
+  element?: HTMLDivElement
```
wayfarer3130, line 78: Suggest a parameter "managerSelector" that has various types associated with it that we can apply later to figure out the right manager.

packages/tools/src/utilities/getAnnotationNearPoint.ts
```diff
@@ -99,8 +99,8 @@ function findAnnotationNearPointByTool(
   // the first annotation that is found in the proximity. BUT, we are not using
   // the function anywhere.
   const annotations = getAnnotations(
-    enabledElement.viewport.element,
-    (tool.constructor as typeof BaseTool).toolName
+    (tool.constructor as typeof BaseTool).toolName,
+    enabledElement.FrameOfReferenceUID
```
wayfarer3130, line 103: We should have an exported function that provides the key to which state manager to use - that way one could have several different types of state manager for different purposes:
default - no FOR
volume - uses FOR and shows everywhere
volume - custom - uses FOR and display set key to key images to specific volumes
comparison - uses an alternate from default
If that is set on the element itself as a simple getter, then one passes in the key for what state manager to use, rather than assuming it will be FOR or nothing.

wayfarer3130 (changes requested):
Please add a function that takes an element and returns an annotation manager id to use rather than assuming that it is FOR or default.  That really gives you the ability to add other annotation manager types in the future, or even switch between them on a viewport, eg add some FOR annotations, and then some volume specific annotations.  Could then have a combined one for viewing/rendering, and a specific one depending on the state for adding/udpating.  Or, two of them for two different clinicians interpretations for comparison purposes.

common/reviews/api/tools.api.md
sedghi: Yes, this is a breaking change and I want it to be before 1.x. 
Element is optional since we can have element specific annotation manager (which we haven't implemented yet)

packages/tools/src/stateManagement/annotation/annotationState.ts
sedghi: `getAnnotations` and `getAnnotation` both can provide the `element` optionally, which then here if we have a element (viewport)-specific one, it can grab from there. Identifying the manager inside the annotation doesn't seems right to me. It should be separate. The same annotation can live in five different managers IMO

packages/tools/src/stateManagement/annotation/annotationState.ts
sedghi: previously it was getting the FOR from the enabledElement, so nothing is changed here except providing the FOR itself instead of `const {FOR} = getEnabledElement(elemnet)` 

SeriesUID does not seem to be right since a CT and PT can have different UID but still in the same FOR. I don't think you can even use cs3d without FOR

packages/tools/test/cpu_RectangleROI_test.js
```diff
@@ -370,8 +365,8 @@ describe('RectangleROITool (CPU):', () => {
     const addEventListenerForAnnotationRendered = () => {
       element.addEventListener(csToolsEvents.ANNOTATION_RENDERED, () => {
         const rectangleAnnotations = annotation.state.getAnnotations(
-          element,
-          RectangleROITool.toolName
+          RectangleROITool.toolName,
+          getEnabledElement(element).FrameOfReferenceUID
```
wayfarer3130, line 369: I would rather this have been a fixed function, eg:
getAnnotationManagerId: (element):string => getEnabledElement(element).FrameOfReferenceUID
as that provides a single place to change everything once in order to add more tyeps of annotations managers.

wayfarer3130 (changes requested):
Almost there - just add the function
getAnnotationManagerId: (element) => getEnabledElement(element).FrameOfReferenceUID
 in a single file, and export it/use it.
The idea is to avoid exposing any knowledge of what the ID is in all these files, so that it can be changed in a single place, or even injected as a custom function at some point externally

wayfarer3130 (approved):

packages/tools/src/stateManagement/annotation/annotationState.ts
```diff
@@ -1,151 +1,146 @@
...
 ): Annotations {
-  const enabledElement = getEnabledElement(element);
-  const annotationManager =
-    getViewportSpecificAnnotationManager(enabledElement);
-  const { FrameOfReferenceUID } = enabledElement;
-
-  return annotationManager.get(FrameOfReferenceUID, toolName);
+  const manager = getAnnotationManager();
+  const groupKey = manager.getGroupKey(options);
+  return manager.getAnnotations(groupKey, toolName) as Annotations;
```
Gabsha, line 57: It seems here that the output of type of `manager.getAnnotations` could be `undefined`, but is being coerced into `Annotations`. Would it make sense to return an empty Array if annotations are undefined, to respect to return type?

sedghi: Good catch, thanks!

PR #450, opened 2023-02-23
add types for wasm codec modules

jmannau:


sedghi (approved):
I guess we need to put these at the codecs repo https://github.com/cornerstonejs/codecs, but we can use them here for now

sedghi (approved):

PR #454, opened 2023-03-02
fix(voi): stack viewport should prioritize image metadata for windowlevel and not persist

sedghi:
Stack viewport was not updating the `voi` if the new image's `voi` was different. This PR will

- let stack viewport each image set the `voiRange` if it is in the metadata
- if the `voi` is set via outside set (e.g., `setProperties`) then the `voi` becomes a global voi for the whole stack
- a `resetProperties` will make it back to the per image `voi`

**Ground truth Test images are changed since now the metadata has more priority**

address https://github.com/OHIF/Viewers/issues/3190


packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -1487,11 +1482,9 @@ class StackViewport extends Viewport implements IStackViewport {
     imageId: string,
     imageIdIndex: number
   ): Promise<string> {
-    if (this.useCPURendering) {
-      await this._loadAndDisplayImageCPU(imageId, imageIdIndex);
-    } else {
-      await this._loadAndDisplayImageGPU(imageId, imageIdIndex);
-    }
+    await (this.useCPURendering
```
wayfarer3130, line 1485: You might consider assign to _loadAndDisplayImage from one of the CPU or GPU functions, and then just call it as
this._loadAndDisplayImage(iamgeId, imageIdIndex)  That way you avoid the if condition everywhere, and if in the future there is another alternate, well, it becomes easy to add.

packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -2545,22 +2556,16 @@ class StackViewport extends Viewport implements IStackViewport {
    * @param colormap - The colormap data to use.
    */
   public setColormap(colormap: CPUFallbackColormapData): void {
-    if (this.useCPURendering) {
-      this.setColormapCPU(colormap);
-    } else {
-      this.setColormapGPU(colormap);
-    }
+    this.useCPURendering
```
wayfarer3130, line 2559: A better implementation is where the useCPURendering is called, just assign this.setColorMap = thisuseCPURendering ? this.setColormapCPU : this.setColormapGPU
Then you don't need the double indirection.  You can still declare setColormap as a public function, just assign the value once you know cpu/gpu usage.

wayfarer3130 (commented):
Would like to see the change to remove CPU/GPU test everywhere.

packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -2636,6 +2584,95 @@ class StackViewport extends Viewport implements IStackViewport {
 
     return newImagePlaneModule;
   }
+
+  private renderingPipelineFunctions = {
```
sedghi, line 2588: here @wayfarer3130

wayfarer3130 (approved):

PR #458, opened 2023-03-03
fix(adapters): Measurement reports can throw exceptions that prevent loading

wayfarer3130:
Simple fix to allow loading measurements even if it throws an exception on one measurement.


sedghi (approved):

PR #459, opened 2023-03-03
[Feat] CircleROI tool

md-prog:
### Feature

- Implementation of CircleROI tool, we had for `cornerstonejs`

### bugfix

- EllipticalROI tool - when setting the `centerPointRadius` the drawing gets corrupted
- RectangleROI, EllipticalROI tool's `area` sometimes becomes negative

### notes

The CircleROI tool can be drawn by pointing center and another point.


wayfarer3130 (approved):
This looks good to me - it looks like it should be able to save and load and display.
I was able to edit the circle, both in radius and position.
Needs an update to match CS3D, but I don't think that will affect any code.

PR #462, opened 2023-03-05
fix(windowLevelTool): WWWL multipler too high when burned in pixels are present

Ouwen:
Burned in Pixels often use pixel values above the BitsStored. This results in a multiplier which is way higher than what you would want in practice.


Ouwen:
@sedghi added the inline comment ready for merge!

sedghi (approved):

PR #470, opened 2023-03-09
fix(scroll): Scrolling failed to find the volume with segmentation

wayfarer3130:
On attempting to scroll a second frame of reference series against the original series, the scrolling failed to display.
Also fix a NPE on turning off CINE related to segmentation.

### DIFF
--- a/packages/core/src/utilities/getTargetVolumeAndSpacingInNormalDir.ts
+++ b/packages/core/src/utilities/getTargetVolumeAndSpacingInNormalDir.ts
@@ -25,12 +25,13 @@ export default function getTargetVolumeAndSpacingInNormalDir(
 ): {
   imageVolume: IImageVolume;
   spacingInNormalDirection: number;
+  uid: string;
 } {
   const { viewPlaneNormal } = camera;
   const volumeActors = viewport.getActors();
 
   if (!volumeActors || !volumeActors.length) {
-    return { spacingInNormalDirection: null, imageVolume: null };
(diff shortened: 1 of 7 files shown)


packages/core/src/utilities/getTargetVolumeAndSpacingInNormalDir.ts
```diff
@@ -74,9 +78,12 @@ export default function getTargetVolumeAndSpacingInNormalDir(
       viewPlaneNormal
     );
 
-    if (spacingInNormalDirection < smallest.spacingInNormalDirection) {
+    // Allow for 1/1000 part larger requirement to prefer earlier volumes
+    // when the spacing is all essentially the same.
+    if (spacingInNormalDirection * 1.001 < smallest.spacingInNormalDirection) {
```
sedghi, line 83: How about 

```
if (Math.abs(spacingInNormalDirection - smallest.spacingInNormalDirection) > CORE.Constant.Epsilon)
```

wayfarer3130: No, I don't think that is as good, this spacing check is related to the size of the actual spacing, and is multiplicative rather than absolute, and suppose we had extremely fine spacing, then using a fixed constant will either be too big or too small.  I'm ok with  (spacingInNormalDirection * (1+CORE.Constant.Epsilon)) if you want?

sedghi (approved):

PR #471, opened 2023-03-09
feat(cine): added support for 4D volumes

lscoder:
Changed CINE to make it possible to play 4D volumes. A new options `dynamicCineEnabled` was added to `PlayClipOptions` which allow us to play CINE on 4D volumes in two different ways: position or time. That variable is always set to `true` for the example I added (`dynamicCINETool`) to make it simple.


packages/core/src/utilities/getVolumeSliceRangeInfo.ts
```diff
@@ -0,0 +1,43 @@
+import { Types, utilities as csUtils } from '@cornerstonejs/core';
+import { ActorSliceRange } from '../types';
+
+function getVolumeSliceRangeInfo(
+  viewport: Types.IVolumeViewport,
+  volumeId: string
+): {
```
sedghi, line 7: add docs please

packages/core/src/utilities/getVolumeViewportScrollInfo.ts
```diff
@@ -0,0 +1,23 @@
+import { Types, utilities as csUtils } from '@cornerstonejs/core';
+
+function getVolumeViewportScrollInfo(
```
sedghi, line 3: docs

packages/tools/examples/CINETool/index.ts
```diff
@@ -147,31 +212,97 @@ async function run() {
...
+    const { viewportId, type } = viewportInput;
+
+    if (type === ViewportType.STACK) {
+      const viewport = <Types.IStackViewport>(
+        renderingEngine.getViewport(viewportId)
+      );
+      viewport.setStack(imageIds);
+      viewport.render();
+    } else if (type === ViewportType.ORTHOGRAPHIC) {
+      setVolumesForViewports(renderingEngine, [{ volumeId }], [viewportId]);
```
sedghi, line 297: we only need to setVolumesForVIewports once, you don't need to do it 3 times, since it is [viewportIds]

packages/tools/examples/dynamicCINETool/index.ts
```diff
@@ -0,0 +1,479 @@
...
+  });
+
+  return container;
+}
+
+function initLayout() {
+  const content = document.getElementById('content');
+  const stagesContainer = document.createElement('div');
+  const firstStageContainer = createFirstStageLayout();
+  const secondStageContainer = createSecondStageLayout();
```
sedghi, line 189: this file has 70% UI setup, can we refactor the createFirst/Seconf functions outside maybe?

packages/tools/src/utilities/cine/playClip.ts
```diff
@@ -37,20 +45,23 @@ function playClip(
     );
   }
 
-  const { viewport } = enabledElement;
+  // 4D Cine is enabled by default
+  playClipOptions.dynamicCineEnabled =
+    playClipOptions.dynamicCineEnabled ?? true;
```
sedghi, line 50: it should be default false, 4D cine is maybe 5% of use cases, even less

packages/tools/src/utilities/cine/playClip.ts
```diff
@@ -262,4 +296,143 @@ function _stopClipWithData(playClipData) {
...
+    },
+    get frameTimeVectorEnabled(): boolean {
+      // It is always in acquired orientation
+      return true;
+    },
+    scroll(delta: number): void {
+      scroll(viewport, { delta, debounceLoading: debounced });
+    },
+  };
+}
```
sedghi, line 330: I like these contexts a lot, great job

packages/tools/src/utilities/cine/playClip.ts
```diff
@@ -262,4 +296,143 @@ function _stopClipWithData(playClipData) {
...
+    },
+    get currentStepIndex(): number {
+      return volume.timePointIndex;
+    },
+    get frameTimeVectorEnabled(): boolean {
+      // Looping throught time does not uses frameTimeVector
+      return false;
+    },
+    scroll(delta: number): void {
+      volume.timePointIndex += delta;
```
sedghi, line 410: maybe put a comment that this is a setter.

packages/tools/src/utilities/cine/playClip.ts
```diff
@@ -262,4 +296,143 @@ function _stopClipWithData(playClipData) {
...
+}
+
+function _createVolumeViewportCinePlayContext(
+  viewport: VolumeViewport,
+  volume: Types.IImageVolume
+): CINETypes.CinePlayContext {
+  const actorEntry = viewport.getDefaultActor();
+
+  if (!actorEntry) {
+    console.warn('No actor found');
```
sedghi, line 339: shouldn't it return or something?

packages/tools/src/utilities/cine/playClip.ts
```diff
@@ -262,4 +296,143 @@ function _stopClipWithData(playClipData) {
...
+      cachedScrollInfo.viewPlaneNormal = camera.viewPlaneNormal;
+      cachedScrollInfo.scrollInfo = scrollInfo;
+    }
+
+    return cachedScrollInfo.scrollInfo;
+  };
+
+  return {
+    get numScrollSteps(): number {
+      return getScrollInfo().numScrollSteps;
```
sedghi, line 372: should we call `getScrollInfo` couple of times here and below? or we can just call that it once and re-use it ?

sedghi (commented):
One of the greatest PR I've seen recently in terms of quality and clarity. Great job indeed

packages/tools/src/utilities/cine/playClip.ts
lscoder: if `dynamicCineEnabled` but it is a 3D volume nothing will change. It will work as before. But if it is a 4D volume it will always scroll through time.

packages/tools/src/utilities/cine/playClip.ts
lscoder: `numScrollSteps` and `currentStepIndex` are accessed on every frame. I thought about creating a method to return both and reduce the number of calls to `getScrollInfo` but after caching the data that solved the problem because the only thing it does is compare the `viewPlanesNormal`s which are a very cheap operation. If you want I can change but that will not make any different in terms of performance (nanoseconds range).

packages/tools/src/utilities/cine/playClip.ts
lscoder: Done

packages/tools/src/utilities/cine/playClip.ts
lscoder: Changed it to return `undefined`. Is there a way to have an `enabledElement` with no volume/stack?

packages/core/src/utilities/getVolumeViewportScrollInfo.ts
lscoder: Updated

packages/tools/src/utilities/cine/playClip.ts
lscoder: Thank you!

packages/tools/examples/dynamicCINETool/index.ts
lscoder: Updated

packages/core/src/utilities/getVolumeSliceRangeInfo.ts
lscoder: Done

sedghi (approved):

PR #476, opened 2023-03-13
feat: added out of bounds scroll

Ouwen:
This adds an event for when user stack scroll goes out of bounds


packages/tools/src/utilities/scroll.ts
```diff
@@ -75,4 +97,24 @@ export function scrollVolume(
...
+    // One common use case of this trigger might be to load the next
+    // volume in a time series or the next segment of a partially loaded volume.
+    csUtils.triggerEvent(eventTarget, EVENTS.VOLUME_SCROLL_OUT_OF_BOUNDS, {
+      volumeId: volumeId,
+      viewport: viewport,
+      delta,
+      desiredFrameIndex,
+      maxFrameIndex,
+      currentImageId: viewport.getCurrentImageId(),
+      currentFrameIndex: desiredFrameIndex - delta,
```
sedghi, line 117: can you add these to the EventTypes too?

packages/tools/src/utilities/scroll.ts
```diff
@@ -31,6 +33,26 @@ export default function scroll(
...
 }
 
+/**
+ * _getDeltaFrameIndex is a subset of functionality copied from csUtils.snapFocalPointToSlice
+ * @param delta
+ * @param sliceRange
+ * @param spacingInNormalDirection
+ * @returns { int, int }
+ */
+function _getDeltaFrameIndex(delta, sliceRange, spacingInNormalDirection) {
```
sedghi, line 43: can you add the exports you want from the `getVolumeViewportScrollInfo` so that we don't recomute these again?

sedghi (commented):
see my comments

Ouwen:
@sedghi this is ready for your review

packages/tools/src/utilities/scroll.ts
```diff
@@ -59,4 +61,32 @@ export function scrollVolume(
     position: newPosition,
   });
   viewport.render();
+
+  const { numScrollSteps, currentStepIndex } =
+    csUtils.getVolumeViewportScrollInfo(viewport, volumeId);
```
sedghi, line 66: I"m a little bit confused, isn't it the fact that this function relies on the `sliceRangeInfo` that is calculated above? So can we refactor its logic so that we don't do `sliceRangeInfo` getter again?

```
function getVolumeViewportScrollInfo(
  viewport: IVolumeViewport,
  volumeId: string
) {
  const { sliceRange, spacingInNormalDirection } = getVolumeSliceRangeInfo(
    viewport,
    volumeId
  );

  const { min, max, current } = sliceRange;

  // Now we can see how many steps there are in this direction
  const numScrollSteps = Math.round((max - min) / spacingInNormalDirection);

  // Find out current frameIndex
  const fraction = (current - min) / (max - min);
  const floatingStepNumber = fraction * numScrollSteps;
  const currentStepIndex = Math.round(floatingStepNumber);

  return { numScrollSteps, currentStepIndex };
}
```

to 

```
function getVolumeViewportScrollInfoFromSliceRange(
  sliceRange,
  spacingInNormalDirection
) {
  const { min, max, current } = sliceRange;

  // Now we can see how many steps there are in this direction
  const numScrollSteps = Math.round((max - min) / spacingInNormalDirection);

  // Find out current frameIndex
  const fraction = (current - min) / (max - min);
  const floatingStepNumber = fraction * numScrollSteps;
  const currentStepIndex = Math.round(floatingStepNumber);

  return { numScrollSteps, currentStepIndex };
}
```

sedghi: API check failed

Ouwen: Ah yeah I need to build updateapi

sedghi (approved):

PR #477, opened 2023-03-14
fix(segmentation): segmentation could not render segment while invisible

sedghi:
Previously when segmentation was set as hidden the actor was set as hidden and as a result you could not do use cases such as hide all, then unhide one. 

The problem was as a result of having two states (visibility and segmentsHidden), this PR will make them as one (only segmentsHidden) and visibility becomes a getter only

### DIFF
--- a/packages/tools/src/stateManagement/segmentation/config/segmentationVisibility.ts
+++ b/packages/tools/src/stateManagement/segmentation/config/segmentationVisibility.ts
@@ -15,10 +15,11 @@ function getSegmentationIndices(segmentationId) {
 
 /**
  * Set the visibility of a segmentation representation for a given tool group. It fires
- * a SEGMENTATION_REPRESENTATION_MODIFIED event. If the segmentation had some segments
- * hidden, it will remove them from the hidden set and make them all visible. On the
- * other hand, if the segmentation had all segments hidden, it will make all segments
- * visible.
+ * a SEGMENTATION_REPRESENTATION_MODIFIED event. If the visibility is set to true, and
+ * the segmentation representation had some segments hidden, it will show all the segments.
+ * And if the visibility is set to false, and it had some segments shown, it will hide all
+ * the segments.
+ *
  *
  * @triggers SEGMENTATION_REPRESENTATION_MODIFIED
  * @param toolGroupId - The Id of the tool group that contains the segmentation.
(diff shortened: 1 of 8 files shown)


packages/tools/src/tools/displayTools/SegmentationDisplayTool.ts
```diff
@@ -65,13 +64,15 @@ class SegmentationDisplayTool extends BaseTool {
...
 
     // for each segmentationData, make the visibility false
-    for (const segmentationRepresentation of toolGroupSegmentationRepresentations) {
-      segmentationRepresentation.visibility = true;
-      triggerSegmentationRepresentationModified(
-        toolGroupId,
-        segmentationRepresentation.segmentationRepresentationUID
-      );
-    }
+    toolGroupSegmentationRepresentations.forEach(
```
jbocce, line 67: I think this chunk of code is setting the visibility to `true`, yet the comment on line 66 says `false`.

packages/tools/src/stateManagement/segmentation/config/segmentationVisibility.ts
```diff
@@ -53,26 +81,40 @@ function getSegmentationVisibility(
...
 
-  if (!segmentationData) {
+  if (!representation) {
     return;
   }
 
-  return segmentationData.visibility;
+  const { segmentsHidden } = representation;
+
+  return segmentsHidden.size === 0;
```
jbocce, line 99: I thought I read in `SegmentationStateType` that if `segmentsHidden` contains `-1` then all segments are hidden. So shouldn't that need to be accounted for here?

jbocce (commented):
@sedghi, overall the changes seem fine - some comments for you to look at. Please update the PR with details how your changes can be tested.

packages/tools/src/stateManagement/segmentation/config/segmentationVisibility.ts
sedghi: I will remove it that was wrong old implementation

packages/tools/src/stateManagement/segmentation/config/segmentationVisibility.ts
```diff
@@ -15,10 +15,11 @@ function getSegmentationIndices(segmentationId) {
 
 /**
  * Set the visibility of a segmentation representation for a given tool group. It fires
- * a SEGMENTATION_REPRESENTATION_MODIFIED event. If the segmentation had some segments
- * hidden, it will remove them from the hidden set and make them all visible. On the
- * other hand, if the segmentation had all segments hidden, it will make all segments
- * visible.
+ * a SEGMENTATION_REPRESENTATION_MODIFIED event. If the visibility is set to true, and
```
jbocce, line 18: There is something perhaps I don't understand here, why do we mention things like "had some segments hidden/shown"... does this even have to be said? Is it not correct to say "visibility true will show all segments" and "visibility false will hide all segments"?

sedghi: ok changed it

jbocce (approved):
Thanks for your diligence. Approved.

PR #479, opened 2023-03-14
fix for elliptical roi

IbrahimCSAE:
This fixes the elliptical roi tool not rendering properly when rotation happens, I added an initial rotation attribute to help determine where the bottom/left/top/right attributes are located in the canvasCoordinates array.

[bottom, top, left, right] would become [left, right, bottom, top] when rotated

```JS
function getCanvasEllipseCorners(ellipseCanvasPoints) {
    const [bottom, top, left, right] = ellipseCanvasPoints;
    const topLeft = [left[0], top[1]];
    const bottomRight = [right[0], bottom[1]];
    return [topLeft, bottomRight];
}
```
this function expects the order to be [bottom. top, left. right], so we need to re-order the canvas coordinates for it if there's a rotation,  we can't pass it the array as it is when it's rotated.

```JS
getCanvasEllipseCorners([
          canvasCoordinates[2], // bottom
          canvasCoordinates[3], // top
          canvasCoordinates[0], // left
          canvasCoordinates[1], // right
        ]);
```

this is a more simple fix than the one I proposed previously in my other pull request, let me know if there's a better approach of doing this.

### DIFF
--- a/packages/tools/src/tools/annotation/EllipticalROITool.ts
+++ b/packages/tools/src/tools/annotation/EllipticalROITool.ts
@@ -214,6 +214,7 @@ class EllipticalROITool extends AnnotationTool {
           activeHandleIndex: null,
         },
         cachedStats: {},
+        initialRotation: viewport.getRotation(),
       },
     };
 
@@ -765,9 +766,22 @@ class EllipticalROITool extends AnnotationTool {
       const canvasCoordinates = points.map((p) =>
         viewport.worldToCanvas(p)
       ) as [Types.Point2, Types.Point2, Types.Point2, Types.Point2];
-      const canvasCorners = <Array<Types.Point2>>(
-        getCanvasEllipseCorners(canvasCoordinates)
-      );
+
+      const rotation = Math.abs(viewport.getRotation() - data.initialRotation);
+      let canvasCorners;
+
+      if (rotation == 90 || rotation == 270) {
+        canvasCorners = <Array<Types.Point2>>getCanvasEllipseCorners([
+          canvasCoordinates[2], // bottom
+          canvasCoordinates[3], // top
+          canvasCoordinates[0], // left
+          canvasCoordinates[1], // right
+        ]);
+      } else {
+        canvasCorners = <Array<Types.Point2>>(
+          getCanvasEllipseCorners(canvasCoordinates) // bottom, top, left, right, keep as is
+        );
+      }
 
       const { centerPointRadius } = this.configuration;
 


packages/tools/src/tools/annotation/EllipticalROITool.ts
```diff
@@ -765,9 +766,22 @@ class EllipticalROITool extends AnnotationTool {
...
         viewport.worldToCanvas(p)
       ) as [Types.Point2, Types.Point2, Types.Point2, Types.Point2];
-      const canvasCorners = <Array<Types.Point2>>(
-        getCanvasEllipseCorners(canvasCoordinates)
-      );
+
+      const rotation = Math.abs(viewport.getRotation() - data.initialRotation);
+      let canvasCorners;
+
+      if (rotation == 90 || rotation == 270) {
```
wayfarer3130, line 773: Can you check the save/restore in the adapters please?

IbrahimCSAE: Is this what you are referring to?

https://github.com/cornerstonejs/cornerstone3D-beta/blob/b989d505098a2d1dcdcb23c9af7ace719f07dcbd/packages/adapters/src/adapters/Cornerstone3D/EllipticalROI.ts#L96

would you like me to follow this approach here too?

wayfarer3130: No, I want you to run OHIF, create some markup with different initial rotations, then save it, and see if it still displays the same object after it loads back in.  You may need to do something special in the adapter to make sure it gets saved/restored correctly.

IbrahimCSAE: Yeah it doesn’t save correctly, I’ll work on this and update you.

IbrahimCSAE: It works as expected now, before saving, I convert so the rotation is 0, so when its restored we won't need the rotation value since it will always be 0.

FYI the DICOMSRDisplayTool in OHIF needs to change so it renders properly, only the cornerstone native tool renders properly now.

It renders fine in my GIF because I modified the DICOMSRDisplayTool locally to account for the rotation the same way I did it for the native cornerstone tool.

```JS
    const rotation = viewport.getRotation();

      canvasCoordinates = ellipsePointsWorld.map(p =>
        viewport.worldToCanvas(p)
      );
      let canvasCorners;
      if (rotation == 90 || rotation == 270) {
        canvasCorners = <Array<Types.Point2>>(
          utilities.math.ellipse.getCanvasEllipseCorners([
            canvasCoordinates[2],
            canvasCoordinates[3],
            canvasCoordinates[0],
            canvasCoordinates[1],
          ])
        );
      } else {
        canvasCorners = <Array<Types.Point2>>(
          utilities.math.ellipse.getCanvasEllipseCorners(canvasCoordinates)
        );
      }
```

GIF 6

Thank u!

wayfarer3130 (approved):
Tested, and seems to work consistently now.  The fixes for saving look good.

PR #482, opened 2023-03-15
feat(volume3d): add orientation to volume viewport 3d

sedghi:
This will add `Orientation` to volume viewport 3D


packages/core/src/RenderingEngine/VolumeViewport3D.ts
```diff
@@ -13,13 +13,27 @@ class VolumeViewport3D extends BaseVolumeViewport {
...
-    const { parallelProjection } = this.options;
+    const { parallelProjection, orientation } = this.options;
 
     const activeCamera = this.getVtkActiveCamera();
 
     if (parallelProjection != null) {
       activeCamera.setParallelProjection(parallelProjection);
     }
+
+    if (orientation && orientation !== OrientationAxis.ACQUISITION) {
```
jbocce, line 24: Actually... does it make sense to refactor this code and put it in the `BaseVolumeViewport` constructor? That is move it there and remove it from here and in `VolumeViewport`?

jbocce (commented):
One comment that we really should consider.

packages/core/src/RenderingEngine/VolumeViewport3D.ts
sedghi: sure

jbocce: Perhaps I was not clear... apologies. Why not put the whole `if` block (i.e. lines 37-40) in the `BaseVolumeViewport` `constructor`?

jbocce (commented):
Apologies, I just wanted to clarify what I had meant.

packages/core/src/RenderingEngine/VolumeViewport3D.ts
sedghi: because one has a early return but more logic later

jbocce (approved):
Looks good. Approved.

PR #483, opened 2023-03-15
fix(stackViewport): check same image was broken after 16 bit texture

sedghi:
when not using the 16 bit texture, we shouldn't check data type


Ouwen:
LGTM!

PR #485, opened 2023-03-16
fix: Allow direct debugging into typescript, fix exports

wayfarer3130:
This change replaces some of the main/module/export names in order to:
1. Allow direct debugging into typescript when linked
2. Debug into the esm generated output when not linked
3. Use the UMD build for release builds

That is done by having the main be the esm module, the module be the typescript and the export be the umd module


sedghi:
why not streaming loader?

wayfarer3130:
OK, added streaming image volume loader.  Tried adding adapters as well, but something isn't quite right there yet.

package.json
```diff
@@ -141,5 +142,12 @@
     "not ie < 11",
     "not op_mini all"
   ],
-  "dependencies": {}
+  "resolutions": {
```
wayfarer3130, line 145: These are not strictly needed, but fix the build to avoid duplicated package imports with just differing versions.

packages/core/package.json
```diff
@@ -3,8 +3,16 @@
   "version": "0.42.1",
   "description": "",
   "main": "dist/umd/index.js",
+  "module": "src/index.ts",
   "types": "dist/esm/index.d.ts",
```
wayfarer3130, line 7: I'm not sure why you removed the types line - it was the primary issue you were seeing with streaming loader - both the one here and the one in the exports is needed.

sedghi: it is useless when you have `exports`

package.json
sedghi: how did you figure this out?

wayfarer3130: I looked for node_modules directories inside the packages/* containing various packages, and then chose a version of the packages for here.  That forces a unified version.  In the case of uuid I chose a version based on the comments on earlier versions being problematic, while for core-js, I chose a version that allows the istanbul instrumenter to run (which it does not on core-js v3).  Canvas was chosen based on being compatible with node 18 and having pre-built binaries for it.

packages/core/package.json
wayfarer3130: Doesn't build without it

packages/adapters/package.json
```diff
@@ -13,9 +13,12 @@
         "build": "dist"
     },
     "exports": {
-        ".": "./dist/@cornerstonejs/adapters.es.js"
+        ".": {
+            "types": "./dist/@cornerstonejs/dts/index.d.ts",
+            "import": "./dist/@cornerstonejs/adapters.es.js"
```
sedghi, line 18: it is recommended to use `.mjs` instead of this

wayfarer3130: Can you make the change and test it with OHIF?  It isn't perfect yet, but I know it works right now with OHIF - will see if it still builds, but I think I'd like to stop searching for perfect in favour of good enough.

sedghi: fair enough, let's test these changes against other viewers. Will let you know

packages/core/package.json
```diff
@@ -4,7 +4,15 @@
   "description": "",
   "main": "dist/umd/index.js",
   "types": "dist/esm/index.d.ts",
-  "module": "dist/esm/index.js",
+  "module": "src/index.ts",
+  "exports": {
+    ".": {
+      "import": "./dist/esm/index.js",
+      "require": "./dist/umd/index.js",
```
sedghi, line 11: - try to check if node 16 plus is there
- try to see if require and import be the same to be esm to check the failed import
- try to publish cjs again and put it in the require as the last resort

sedghi:
seems like production debugging is fine, and we should not include src in the module for package json. The local yarn link problem is probably related to the tsconfig extending the base. For those who want to use yarn link in local, you can change the module to be src/index.ts as a solution until we fix the inheritance bug for each tsconfig

PR #486, opened 2023-03-16
fix(SVGFilter) - SVG filters do not have unique ids

jbocce:
## Changes Made

- filters should have unique ids per viewport
- updated doubleClickWithStackAnnotationTools to demonstrate the fix

## Testing

The problem can be demonstrated as such...

1. Navigate to https://www.cornerstonejs.org/live-examples/doubleclickwithstackannotationtools.
2. Open the browser developer tools and add the following snippet of html as the last child of the `<body>` element.

```
<div id="toto" style="border: 1px solid red; height: 800px; width: 800px">
    <svg class="svg-layer" id="svg-layer" xmlns=http://www.w3.org/2000/svg style="width: 100%; height: 100%; pointer-events: none; position: absolute;">
        <defs>
            <filter id="shadow" filterUnits="userSpaceOnUse">
                <feOffset result="offOut" in="SourceGraphic" dx="0.5" dy="0.5"/>
                <feColorMatrix result="matrixOut" in="offOut" in2="matrix" values="0.2 0 0 0 0 0 0.2 0 0 0 0 0 0.2 0 0 0 0 0 1 0"/>
                <feBlend in="SourceGraphic" in2="matrixOut" mode="normal"/>
            </filter>
         </defs>
         <line data-id="677cb2f0-68d0-46cd-9251-afb10258cb17-line" x1="10" y1="10" x2="790" y2="790" stroke="rgb(255, 255, 0)" style="filter:url(#shadow);" stroke-width="1"/>
    </svg>
</div>
```
3. The inserted div SHOULD show a line that approximately goes top-left to bottom-right corner of the div but does NOT.

The corrected behaviour can be seen by running the same example now modified to show two viewports showing the same data. Add a markup and double clicking either should toggle the size of the viewport and the markup extents should not be affected.

### DIFF
--- a/packages/tools/examples/doubleClickWithStackAnnotationTools/index.ts
+++ b/packages/tools/examples/doubleClickWithStackAnnotationTools/index.ts
@@ -34,7 +34,6 @@ const { Events } = cornerstoneTools.Enums;
 const { MouseBindings } = csToolsEnums;
 
 const renderingEngineId = 'myRenderingEngine';
-const viewportId = 'CT_STACK';
 
 document.documentElement.style.userSelect = 'none';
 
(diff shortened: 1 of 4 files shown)


sedghi (commented):
Not sure why my review didn't appear. Great PR!

PR #487, opened 2023-03-16
feat(segmentation): segmentation threshold utility

NeilMacPhee:
**Summary**
This feature provides the ability to select and threshold a segmentation using the values and threshold information from one or more image volumes, even if they have different sizes.

The utility takes the segmentation volume, the index of the segmentation to threshold, the threshold volume information of one or more volumes, and the overlap type as parameters.


packages/tools/examples/segmentationThreshold/index.ts
```diff
@@ -0,0 +1,503 @@
+import {
```
sedghi, line 1: I don't think we need a new public demo, this was good for development

sedghi (approved):

PR #493, opened 2023-03-18
fix(mobile): Crosshairs highlighted for mobile

diegohennrich:
- My idea here is to increase the annotation stroke to a better experience on mobile screens.

### DIFF
--- a/packages/tools/src/tools/CrosshairsTool.ts
+++ b/packages/tools/src/tools/CrosshairsTool.ts
@@ -1108,7 +1108,14 @@ class CrosshairsTool extends AnnotationTool {
 
       let color =
         viewportColor !== undefined ? viewportColor : 'rgb(200, 200, 200)';
+
       let lineWidth = 1;
+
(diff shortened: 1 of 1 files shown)


diegohennrich:
Could you check, please? @Ouwen

packages/tools/src/tools/CrosshairsTool.ts
```diff
@@ -1108,7 +1108,14 @@ class CrosshairsTool extends AnnotationTool {
 
       let color =
         viewportColor !== undefined ? viewportColor : 'rgb(200, 200, 200)';
+
       let lineWidth = 1;
+
+      if (/Mobi|Android/i.test(navigator.userAgent)) {
```
sedghi, line 1114: can we put it in the constructor and assign it to a private property?

diegohennrich: Hey @sedghi great idea. I've changed it. Could you check it? Is there anything else to change? Thanks.

sedghi (approved):

PR #501, opened 2023-03-22
fix(rendering): should still use Float32 when not 16 bit for scaling issues

sedghi:
We introduced optional 16 bit texture support in the previous pr [here](https://github.com/cornerstonejs/cornerstone3D-beta/pull/420), but because of scaling issues in PT, the images were casted to int/uint 16 while they should still be Float. 

These changes will get reverted back once we merge the dicomImageLoader which properly handles scaling

FYI @Ouwen

### DIFF
--- a/common/reviews/api/core.api.md
+++ b/common/reviews/api/core.api.md
@@ -98,6 +98,8 @@ export abstract class BaseVolumeViewport extends Viewport implements IVolumeView
     // (undocumented)
     setVolumes(volumeInputArray: Array<IVolumeInput>, immediate?: boolean, suppressEvents?: boolean): Promise<void>;
     // (undocumented)
+    use16BitTexture: boolean;
+    // (undocumented)
     useCPURendering: boolean;
     // (undocumented)
     static get useCustomRenderingPipeline(): boolean;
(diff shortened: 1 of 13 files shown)


common/reviews/api/core.api.md
```diff
@@ -497,7 +499,7 @@ function createUint16SharedArray(length: number): Uint16Array;
 function createUint8SharedArray(length: number): Uint8Array;
 
 // @public (undocumented)
-export function createVolumeActor(props: createVolumeActorInterface, element: HTMLDivElement, viewportId: string, suppressEvents?: boolean): Promise<VolumeActor>;
+export function createVolumeActor(props: createVolumeActorInterface, element: HTMLDivElement, viewportId: string, suppressEvents?: boolean, use16BitTexture?: boolean): Promise<VolumeActor>;
```
wayfarer3130, line 502: It's probably time to add an options bag here rather than just adding additional parameters.

sedghi: This will get removed soon, so can we not do it in this PR?

packages/core/src/RenderingEngine/helpers/createVolumeActor.ts
```diff
@@ -61,7 +62,7 @@ async function createVolumeActor(
   // types of volumes which might not be composed of imageIds would be e.g., nrrd, nifti
   // format volumes
   if (imageVolume.imageIds) {
-    await setDefaultVolumeVOI(volumeActor, imageVolume);
+    await setDefaultVolumeVOI(volumeActor, imageVolume, use16BitTexture);
```
wayfarer3130, line 65: This option should go in an options object, as there may well be others in the future.

wayfarer3130 (approved):
As a temporary fix to resolve rendering issues this looks ok.  There are comments we will likely need applied eventually, but looks good in the short term.

Ouwen:
@sedghi does this change also solve https://github.com/cornerstonejs/cornerstone3D-beta/issues/481#issuecomment-1478255032

Ouwen:
@sedghi, I was going through a rebase and I think there are some breaking changes here. The main issue is that we allow `undefined` types for the `targetBuffer` object. However, currently CSWIL expects a type if `targetBuffer` object is provided.

We can either fix in CSWIL or fix in CS3D. I'm thinking CS3D and can push up a fix

sedghi:
@Ouwen I guess in the next couple of days i will use dicom Image loader

PR #502, opened 2023-03-22
feat(3d from 4d): 3D image generation from 4D

NeilMacPhee:
### Summary
This feature will allow users to choose between three methods to generate a 3D volume from 4D data: subtract, average, or sum. The "subtract" method works with two time frames and will subtract the values of each voxel in the 4D data from the corresponding voxel in the previous frame. The "average" method will take the average of the values of each voxel in the 4D data over time. The "sum" method will add the values of each voxel in the 4D data over time. The output scalar array can then be used to create a 3D volume.

"generateImageFromTimeData" allows the user to specify which time frames to use for generating the scalar data.

### Demo

3DimageGeneration

### DIFF
--- a/packages/tools/src/utilities/dynamicVolume/generateImageFromTimeData.ts
+++ b/packages/tools/src/utilities/dynamicVolume/generateImageFromTimeData.ts
@@ -1,4 +1,5 @@
-import { Types } from '@cornerstonejs/core';
+import { Enums, Types } from '@cornerstonejs/core';
+import { DynamicOperatorType } from 'core/src/enums';
 
 /**
  * Gets the scalar data for a series of time frames from a 4D volume, returns an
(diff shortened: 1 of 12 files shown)


packages/tools/src/utilities/dynamicVolume/generateImageFromTimeData.ts
```diff
@@ -0,0 +1,67 @@
...
+  const arrayLength = typedArrays[0].length;
+  const finalArray = new Float32Array(arrayLength);
+
+  if (operation === 'SUM') {
+    for (let i = 0; i < numFrames; i++) {
+      const currentArray = typedArrays[frames[i]];
+      for (let j = 0; j < arrayLength; j++) {
+        finalArray[j] += currentArray[j];
+      }
+    }
```
sedghi, line 40: you can return final array at the end of this

packages/tools/src/utilities/dynamicVolume/generateImageFromTimeData.ts
```diff
@@ -0,0 +1,67 @@
...
+    throw new Error('Please provide two or more time points');
+  }
+
+  // Gets scalar data for all time frames
+  const typedArrays = dynamicVolume.getScalarDataArrays();
+
+  const arrayLength = typedArrays[0].length;
+  const finalArray = new Float32Array(arrayLength);
+
+  if (operation === 'SUM') {
```
sedghi, line 34: operations should be Enums that you use, similar to viewportTypes etc

packages/tools/src/utilities/dynamicVolume/generateImageFromTimeData.ts
```diff
@@ -0,0 +1,67 @@
...
+      }
+    }
+  }
+
+  if (operation === 'SUBTRACT') {
+    if (frames.length > 2) {
+      throw new Error('Please provide only 2 time points for subtraction.');
+    }
+    for (let j = 0; j < arrayLength; j++) {
+      finalArray[j] += typedArrays[0][j] - typedArrays[1][j];
```
sedghi, line 48: why subtracting 0 and 1? it should accept frames

sedghi (commented):
see my comments

packages/tools/src/utilities/dynamicVolume/generateImageFromTimeData.ts
NeilMacPhee: Moved 'return finalArray' to the end of each operator function

packages/tools/src/utilities/dynamicVolume/generateImageFromTimeData.ts
NeilMacPhee: Added Enums DynamicOperatorType, changed if statements to use types

packages/tools/src/utilities/dynamicVolume/generateImageFromTimeData.ts
NeilMacPhee: changed to typedArrays[frames[0]][j] - typedArrays[frames[1]][j], was using this during development and forgot to change it back

sedghi (approved):

PR #511, opened 2023-03-23
fix: Segmentation slice range is wrong when nearly orthonormal as well as for segmentation volumes

wayfarer3130:
If a segmentation is created with a much larger range than one of the display sets it applies to, then that display set is shown with large swathes of black, and the scrolling can be wrong because it uses the finer of the two scrolls, which can end up duplicating base images.

This change uses primary volumes, which it determines by those having a volume image loader prefix, for slice range, and only uses secondary volumes when there are no primaries.

### DIFF
--- a/packages/core/src/utilities/getTargetVolumeAndSpacingInNormalDir.ts
+++ b/packages/core/src/utilities/getTargetVolumeAndSpacingInNormalDir.ts
@@ -3,15 +3,21 @@ import { EPSILON } from '../constants';
 // import type { VolumeViewport } from '../RenderingEngine'
 import { ICamera, IImageVolume, IVolumeViewport } from '../types';
 import getSpacingInNormalDirection from './getSpacingInNormalDirection';
+import { getVolumeLoaderSchemes } from '../loaders/volumeLoader';
 
 // One EPSILON part larger multiplier
 const EPSILON_PART = 1 + EPSILON;
+
+const startsWith = (str, starts) =>
+  starts === str.substring(0, Math.min(str.length, starts.length));
 
 // Check if this is a primary volume
 // For now, that means it came from some sort of image loader, but
 // should be specifically designated.
 const isPrimaryVolume = (volume): boolean =>
-  volume.volumeId.indexOf(':') !== -1;
+  !!getVolumeLoaderSchemes().find((scheme) =>
+    startsWith(volume.volumeId, scheme)
+  );
 
 /**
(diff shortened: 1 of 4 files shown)


packages/core/src/utilities/getTargetVolumeAndSpacingInNormalDir.ts
```diff
@@ -7,6 +7,12 @@ import getSpacingInNormalDirection from './getSpacingInNormalDirection';
 // One EPSILON part larger multiplier
 const EPSILON_PART = 1 + EPSILON;
 
+// Check if this is a primary volume
+// For now, that means it came from some sort of image loader, but
+// should be specifically designated.
+const isPrimaryVolume = (volume): boolean =>
+  volume.volumeId.indexOf(':') !== -1;
```
sedghi, line 14: can we add a method to `volumeLoader` to `get` all loaders and just extract the scope and use it here instead?

packages/streaming-image-volume-loader/package.json
```diff
@@ -2,9 +2,12 @@
...
   "version": "0.15.5",
   "description": "",
-  "main": "dist/umd/index.js",
+  "main": "dist/esm/index.js",
   "types": "dist/esm/index.d.ts",
-  "module": "dist/esm/index.js",
+  "module": "src/index.ts",
+  "exports": {
+    ".": "./dist/umd/index.js"
+  },
```
sedghi, line 10: can you remove this for now, we do it in the other PR

packages/streaming-image-volume-loader/api-extractor.json
```diff
@@ -1,7 +1,7 @@
 {
   "extends": "../../api-extractor.json",
   "projectFolder": ".",
-  "mainEntryPointFilePath": "<projectFolder>/dist/cjs/index.d.ts",
+  "mainEntryPointFilePath": "<projectFolder>/dist/esm/index.d.ts",
```
sedghi, line 4: other `api-extractor`s use cjs, why this one is only esm, can we do it in another PR?

sedghi (commented):
see my comments

packages/streaming-image-volume-loader/api-extractor.json
wayfarer3130: Thought I had excluded that

sedghi (approved):

PR #512, opened 2023-03-24
Fixed ToolStyleConfig type error for projects with strictNullChecks

### LINKED ISSUE #505: TypeScript error TS2411: Property 'global' of type 'AnnotationStyle | undefined' is not assignable to 'string' index type 'AnnotationStyle'
When using cornerstone tools I get the following error during build (in strict mode):

image

Since `global` is optional property it can't be added to an interface like this.

Changing the type definition like this resolves the error:
image

StefanNedelchev:
Fixes #505


StefanNedelchev:
@sedghi I linked the PR with the issue it corresponds with and it's ready to review

wayfarer3130 (approved):
Changes LGTM.

Issue #513, opened 2023-03-24
Q: Which ECMA Script version really is the target for Cornerstone 3D?


StefanNedelchev:
By looking at the browserslist configuration in `package.json` I see that one of the conditions is `not ie < 11` which puts IE 11 as desired browser to support. I guess this is the reason why we use ES5 as a target version in the TypeScript configuration?

I think that some features used in cornerstone tools already don't work with IE 11. Additionally in #503  @sedghi shared his vision on the usage of `structuredClone` which is a property only supported by modern browsers.

In this case can we conclude that IE support is not goal of Cornerstone js anymore and does it make sense to change the target ES version to ES6 or even something higher like ES2020 to avoid unnecessary transpiling of ES features that our target browsers already support?

sedghi:
IE is not (intention was not) to be a target ever. So I'm happy to ditch that

StefanNedelchev:
> IE is not (intention was not) to be a target ever. So I'm happy to ditch that

@sedghi  I created a PR for this

StefanNedelchev:
@sedghi maybe this could be closed (now that the PR is merged)?

PR #514, opened 2023-03-24
Drop ES5 support and change target to ES2020

StefanNedelchev:
Related to #513 

Since it has been confirmed that we don't support IE or any old browsers, it makes sense to remove it from the `browserslist` configuration and update the TS configuration.


wayfarer3130 (approved):
This looks good to me, @sedghi  - you probably want to take a look at this and think about it a bit.

tsconfig.base.json
```diff
@@ -15,7 +15,7 @@
     "forceConsistentCasingInFileNames": true,
     "useDefineForClassFields": false,
     "jsx": "react-jsx",
-    "lib": ["ES6", "dom"], // "ES6", "DOM" | "es2015", "dom"
+    "lib": ["ES2020", "dom"],
```
wayfarer3130, line 18: Does ES2021 work here?  If so, please use ES2021 instead.

sedghi: It is more complicated than that, we should test each

sedghi: https://kangax.github.io/compat-table/es2016plus/

StefanNedelchev: > Does ES2021 work here? If so, please use ES2021 instead.

Check my other PR (which is still not quite ready) for dropping lodash.clonedeep - #517 . I changed it to ES2021 there because older versions don't support structuredClone. Once that gets merged we'll target ES2021

sedghi: Based on my research, the only thing that might be tricky is `Function.prototype.toString revision` that doesn't seem to be fully supported on Safari or iOS. What are your take @StefanNedelchev @wayfarer3130 

image

StefanNedelchev: > Based on my research, the only thing that might be tricky is `Function.prototype.toString revision` that doesn't seem to be fully supported on Safari or iOS. What are your take @StefanNedelchev @wayfarer3130
> 
> image

do we rely on this support?

StefanNedelchev:
@sedghi is still there something that's stopping us from this update?

sedghi:
I have not tested the whole thing yet for this

sedghi:
I'm about to create a beta branch and beta release for 2.0 soon, and we will merge this in 2.0beta

sedghi:
I merged your other PR that was targeting ES2021, so I'm gonna close this one I guess

StefanNedelchev:
> I merged your other PR that was targeting ES2021, so I'm gonna close this one I guess

it only depends if you want to pick the browserslist change

PR #515, opened 2023-03-24
Changing the value of the EPSILON in getSliceRange.ts file

rodrigobasilio2022:
The actual EPSILON value didnt resolve the problem. We are using a much lower value so the fix already implemented in this current file could execute


wayfarer3130 (approved):
Changes look good - just makes the comparison size smaller so it has to be nearly exactly 1.

PR #517, opened 2023-03-24
feat(structuredClone): drop lodash.clonedeep in favor of structuredClone

### LINKED ISSUE #503: A suggestion to drop lodash.clonedeep in favour of structuredClone
When I import `@cornerstonejs/core` into an Angular project and try to build it I get the following warning:
image

This is a common warning that appears if pretty much any lodash feature is used (or other commonjs dependencies) and [there is a good reason behind](https://angular.io/guide/build#configuring-commonjs-dependencies). But even if we put Angular aside, it's always a good thing if we can get rid of 3rd party packages when there is an opportunity. Since Cornerstone 3D is meant to be for modern browsers anyway, we could use the [structuredClone()](https://developer.mozilla.org/en-US/docs/Web/API/structuredClone) method for deep cloning of objects instead of relying on `lodash.clonedeep()`. The function is supported by pretty much every modern desktop and web browser with the exception that it's not available in workers on Opera and Safari.

StefanNedelchev:
Resolves #503 

Notes:
- `@types/node` had to be updated to v18 because the minimal version of the type definitions that support `structuredClone` is 17.
- `typescript` had to be updated as well to a version that supports `structuredClone`
- the `lib` option in tsconfig was set to `ES2021` instead of `ES2022` because with `ES2022` I'm getting errors during `build:update-api` (any help with that is welcome)
- the type assertion in `Settings.ts` had to be done due to TypeScript error after updating to the newer version
- the type corrections in `createLabelmapVolumeForViewport.ts` had to be done because the use of `structuredClone` type checks the input (unlike lodash.clonedeep)


StefanNedelchev:
Although the builds and tests passed locally, the tests here failed. From the logs I see there is one of the errors:

```
 An error was thrown in afterAll
  DataCloneError: Failed to execute 'structuredClone' on 'Window': SVGCircleElement object could not be cloned.
  error properties: Object({ code: 25, INDEX_SIZE_ERR: 1, DOMSTRING_SIZE_ERR: 2, HIERARCHY_REQUEST_ERR: 3, WRONG_DOCUMENT_ERR: 4, INVALID_CHARACTER_ERR: 5, NO_DATA_ALLOWED_ERR: 6, NO_MODIFICATION_ALLOWED_ERR: 7, NOT_FOUND_ERR: 8, NOT_SUPPORTED_ERR: 9, INUSE_ATTRIBUTE_ERR: 10, INVALID_STATE_ERR: 11, SYNTAX_ERR: 12, INVALID_MODIFICATION_ERR: 13, NAMESPACE_ERR: 14, INVALID_ACCESS_ERR: 15, VALIDATION_ERR: 16, TYPE_MISMATCH_ERR: 17, SECURITY_ERR: 18, NETWORK_ERR: 19, ABORT_ERR: 20, URL_MISMATCH_ERR: 21, QUOTA_EXCEEDED_ERR: 22, TIMEOUT_ERR: 23, INVALID_NODE_TYPE_ERR: 24, DATA_CLONE_ERR: 25 })
  Error: Failed to execute 'structuredClone' on 'Window': SVGCircleElement object could not be cloned.
      at resetCornerstoneToolsState (webpack-internal:///./packages/tools/src/store/state.ts:8:4094)
      at Module.destroy (webpack-internal:///./packages/tools/src/init.ts:21:14870)
      at UserContext.eval (webpack-internal:///./packages/tools/test/FrameOfReferenceSpecificToolStateManager_test.js:56:45)
      at <Jasmine>
```
I double checked the MDN docs for structuredClone() and by looking at [supported types](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm#supported_types) it seems that HTML and SVG elements are not present. A quick test in the a JS console confirmed that:
image

I'm leaving the draft PR open in case if anyone wants to investigate different solutions to these cases.

StefanNedelchev:
> @StefanNedelchev - could you add the transfer option with a reference to the SVG object contained in it? I believe that should allow removing the SVG object from the old spot and creating it in the new spot. Also, please see if you can merge: #514 to include this.
I was thinking about adding the transfer options, I will try that. #514 is supposed to be ready to merge but I don't seem to have the merge button

tsconfig.base.json
```diff
@@ -37,7 +37,7 @@
     "packages/**/dist",
     "packages/**/lib",
     "packages/**/lib-esm",
-    "packages/adapters",
+    "packages/adapters/**/*.js",
```
StefanNedelchev, line 40: I had to do this because of a new error that started appearing after the latest sync with main:
```
Parsing error: ESLint was configured to run on `<tsconfigRootDir>/packages/adapters/src/adapters/Cornerstone3D/EllipticalROI.ts` using `parserOptions.project`: <tsconfigRootDir>/tsconfig.json
```

StefanNedelchev:
@wayfarer3130 check the solution I did with the structuredClone in the sate store. Passing the svg elements to the transferable items didn't work (it wasn't even accepted because it's for binary objects) so I had to use this funky solution of excluding the element when passing it to the `structuredClone` method and additionally copying the svgElements map with a spread operator.

StefanNedelchev:
@sedghi sorry for leaving this PR behind. I took some time to sync it with the latest cornerstone 3d and remove some leftovers of lodash.cloneDeep. The CI seems successful now but I have to point out that for some reason I had to add 2 rules in the ESLint ignore configuration due to questionable lint errors ("ESLint was configured to run ... However, that TSConfig does not include this file") when trying to commit.

PR #521, opened 2023-03-27
fix(binding): fix this binding

bionoone:
strange binding issue when processing loading failure


sedghi (approved):
This is great, thanks

PR #526, opened 2023-03-28
fix(Synchronizer): Check for null rendering engine when handling disabling of elements.

jbocce:
The one line description says it all.


wayfarer3130 (approved):
LGTM

PR #527, opened 2023-03-28
fix(volumeLoad): should still update texture when loading

sedghi:
we shouldn't hold texture update from happening in load, just in cancel

### DIFF
--- a/packages/streaming-image-volume-loader/src/BaseStreamingImageVolume.ts
+++ b/packages/streaming-image-volume-loader/src/BaseStreamingImageVolume.ts
@@ -317,7 +317,8 @@ export default class BaseStreamingImageVolume extends ImageVolume {
       // data loader scheme)
       const cachedImage = cache.getCachedImageBasedOnImageURI(imageId);
 
-      // check if we are still loading the volume and we have not canceled loading
+      // check if the load was cancelled while we were waiting for the image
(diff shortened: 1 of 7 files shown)


packages/streaming-image-volume-loader/src/BaseStreamingImageVolume.ts
```diff
@@ -318,7 +318,11 @@ export default class BaseStreamingImageVolume extends ImageVolume {
       const cachedImage = cache.getCachedImageBasedOnImageURI(imageId);
 
       // check if we are still loading the volume and we have not canceled loading
```
wayfarer3130, line 320: This comment needs updating

packages/streaming-image-volume-loader/src/BaseStreamingImageVolume.ts
```diff
@@ -251,8 +253,6 @@ export default class BaseStreamingImageVolume extends ImageVolume {
     const { vtkOpenGLTexture, imageData, metadata, volumeId } = this;
     const { FrameOfReferenceUID } = metadata;
 
-    loadStatus.loading = true;
```
wayfarer3130, line 254: Where does loading status get set?

wayfarer3130 (approved):
Just the one comment update.

PR #530, opened 2023-03-29
fix(StackViewport): float number comparison to use epsilon when StackViewport is abou…

### LINKED ISSUE #516: Float comparison inconsistency when checking if vtkimageData is equal to cs image data
https://github.com/cornerstonejs/cornerstone3D-beta/blob/main/packages/core/src/RenderingEngine/StackViewport.ts#L1541

JS has a limitation when comparing float number, so for every place on the project, when comparing float numbers, we should avoid strict comparison and instead use comparison considering a deviation.

At same conditional of Line 1541 there is already a comparison using the utility for that (isEqual). But it does not apply to xspacing or yspacing.

Explaining the problem.

Datasets with multiple slices, because of data error extraction (machine error/deviation) can produce very close xspacing, yspacing values but not the same. So, lets image two frames
A = [xspacing, yspacing] = [0.99999,0.99999]
B= [xspacing, yspacing] = [0.99998,0.99998]
Both A and B belong to same series but having the comparison mentioned, it might consider they belong to different series data set.

To fix that I recommend:
- use isEqual utility
- define if the actual tolerance in isEqual satisfies the case to consider different frames but same series (otherwise the viewport must contain a **public** configuration for the more appropriate value for tolerance and client can set up it as he wants)

ladeirarodolfo:
…t to be updated

fixes #516 

Adding epsilon comparison when StackViewport is about to be update.

Changed isEqual method to also consider number type of params. Backward compatibility is preserved (no loss to api)

Tech note: comparing float number in JS should be always using this approach.
Added some unit tests to validate new changes

Later on we can explore more the tolerances values for different situation and maybe include it as configuration within the lib

### DIFF
--- a/packages/core/src/utilities/isEqual.ts
+++ b/packages/core/src/utilities/isEqual.ts
@@ -1,27 +1,72 @@
-/**
- * returns equal if the two arrays are identical within the
- * given tolerance.
- *
- * @param v1 - The first array of values
- * @param v2 - The second array of values.
- * @param tolerance - The acceptable tolerance, the default is 0.00001
- *
- * @returns True if the two values are within the tolerance levels.
- */
-export default function isEqual(
-  v1: number[] | Float32Array,
-  v2: number[] | Float32Array,
+function areNumbersEqualWithTolerance(
+  num1: number,
+  num2: number,
+  tolerance: number
+): boolean {
+  return Math.abs(num1 - num2) <= tolerance;
+}
+
+function areArraysEqual(
+  arr1: number[] | Float32Array | Float64Array,
+  arr2: number[] | Float32Array | Float64Array,
   tolerance = 1e-5
 ): boolean {
-  if (v1.length !== v2.length) {
+  if (arr1.length !== arr2.length) {
     return false;
   }
 
-  for (let i = 0; i < v1.length; i++) {
-    if (Math.abs(v1[i] - v2[i]) > tolerance) {
+  for (let i = 0; i < arr1.length; i++) {
+    if (!areNumbersEqualWithTolerance(arr1[i], arr2[i], tolerance)) {
       return false;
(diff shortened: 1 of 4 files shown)


packages/core/src/utilities/isEqual.ts
```diff
@@ -1,27 +1,72 @@
...
+}
+
+function isFloatArrayLike(
+  value: any
+): value is number[] | Float32Array | Float64Array {
+  const isArrayLike = typeof value === 'object' || 'length' in value;
+  const floatTypedArrayLike =
+    value instanceof Float32Array || value instanceof Float64Array;
+
+  return isArrayLike || floatTypedArrayLike;
```
sedghi, line 38: so this becomes true if we have Int8Array too right? because of `isArrayLike`, then why the function name suggests it should be float array?

packages/core/src/utilities/isEqual.ts
```diff
@@ -1,27 +1,72 @@
...
+  // values must be the same type
+  if (typeof v1 !== typeof v2) {
     return false;
   }
 
-  for (let i = 0; i < v1.length; i++) {
-    if (Math.abs(v1[i] - v2[i]) > tolerance) {
-      return false;
-    }
+  if (isNumberType(v1) && isNumberType(v2)) {
```
sedghi, line 62: I guess we only need to check one of them, since you return false early in the function right? same for below

ladeirarodolfo: Remember when we were seeing support for es stuff? 
There there was the concept of narrowing.
https://www.typescriptlang.org/docs/handbook/2/narrowing.html

To answer your question narrowing op is not transitive in TS yet. So we need to narrow both params to ensure the if block has the type we want

PR #531, opened 2023-03-30
fix: memory leak on deletion

Ouwen:
vtk.js needs to call delete in order to clear CPU memory.


packages/core/src/cache/cache.ts
```diff
@@ -133,6 +133,7 @@ class Cache implements ICache {
     }
 
     if (volume.imageData) {
+      volume.imageData.delete();
```
sedghi, line 136: shouldn't we delete after decache? decache writes each volume in multiple single slices images

sedghi (commented):
see my one comment

Ouwen:
@sedghi let me know what you think of the changes

packages/core/src/cache/cache.ts
sedghi, line 1: I'm confused, where did `.delete` go?

Ouwen: The delete is moved into the ImageVolume destroy call which is called during a decache()

sedghi: but shouldn't we call delete after decache too? since it doesn't run destroy as I see

Ouwen: Some more exploration... it seems like the `volume.delete()` is not required if the `imageLoadObject.decache()` is called since `imageLoadObject.decache()` internally calls `volume.delete()`

One limitation of cache clearing in general is that it only clears CPU cache. The `this.vtkOpenGLTexture.delete()` does not actually clear the GPU memory.

Calling RenderingEngine.destroy() does not seem to clear the GPU memory either according to the Chrome TaskManager.

sedghi (approved):

PR #536, opened 2023-03-31
fix(suv display): fix scaling of non-SUV PT images

pwespi:
Currently PT images that are **not** in SUV are not rescaled at all. This is problematic because PT images can be in units of e.g. counts or Bq/ml, so it's important that the rescaling is applied in those cases.

### DIFF
--- a/packages/tools/src/tools/annotation/CircleROITool.ts
+++ b/packages/tools/src/tools/annotation/CircleROITool.ts
@@ -6,6 +6,7 @@ import {
   eventTarget,
   triggerEvent,
   utilities as csUtils,
+  metaData,
 } from '@cornerstonejs/core';
 import type { Types } from '@cornerstonejs/core';
 
@@ -855,10 +856,11 @@ class CircleROITool extends AnnotationTool {
       isEmptyArea,
       Modality,
       areaUnit,
+      suvbw,
     } = cachedVolumeStats;
 
     const textLines: string[] = [];
-    const unit = getModalityUnit(Modality, isPreScaled);
+    const unit = getModalityUnit(Modality, isPreScaled, suvbw);
 
     if (radius) {
       const radiusLine = isEmptyArea
@@ -1021,8 +1023,15 @@ class CircleROITool extends AnnotationTool {
         stdDev /= count;
         stdDev = Math.sqrt(stdDev);
 
+        const { suvbw } =
+          metaData.get(
+            'scalingModule',
+            annotation.metadata.referencedImageId
+          ) || {};
+
         cachedStats[targetId] = {
           Modality: metadata.Modality,
+          suvbw,
           area,
           mean,
           max,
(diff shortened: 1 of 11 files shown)


packages/tools/src/utilities/getModalityUnit.ts
```diff
@@ -1,8 +1,6 @@
 function getModalityUnit(modality: string, isPreScaled: boolean): string {
   if (modality === 'CT') {
     return 'HU';
-  } else if (modality === 'PT' && isPreScaled === true) {
```
sedghi, line 4: Why we are removing this unit here?

sedghi: Can we get the metadata and check if suvbw is there in scalingModule, and return 'SUV' to be able to show SUV like before?

pwespi: Good idea. Adapted.

packages/tools/src/tools/annotation/CircleROITool.ts
```diff
@@ -1021,8 +1023,15 @@ class CircleROITool extends AnnotationTool {
         stdDev /= count;
         stdDev = Math.sqrt(stdDev);
 
+        const { suvbw } =
```
sedghi, line 1026: we don't need to do this metadata query on each cachedStats calculation since it doesn't change with annotation data change, it can be done outside, in the render loop, just before _getTextLines calls

pwespi: moved the metadata query

pwespi:
Some tests are currently failing. I assume it's because of the metadata query, but I'm not sure how to fix it.

sedghi:
can you follow this please https://www.cornerstonejs.org/docs/contribute/update-api

sedghi:
seems like tools' tests are failing, maybe look here to how to run tests locally https://www.cornerstonejs.org/docs/contribute/tests

pwespi:
> seems like tools' tests are failing, maybe look here to how to run tests locally https://www.cornerstonejs.org/docs/contribute/tests

It seems like in some tests `annotation.metadata.referencedImageId` is `undefined`, which causes `fakeMetaDataProvider` to throw an error (calling `.split` of undefined).

Do you think calling `const { suvbw } = metaData.get('scalingModule', annotation.metadata.referencedImageId) || {};` is problematic?

If I catch the undefined in `fakeMetaDataProvider`, all tests pass locally, but I'm not sure if this would hide another problem.

pwespi:
Added a check for `referencedImageId` being defined and this fixes the tests.

sedghi:
Hmmm, this will break our volume viewports I guess, since they don't necessarily have referenceImageId, right now they can show suv 

image

I guess the correct logic can be created inside AnnotationTool base class to check if it is a volume or stack. If volume it checks if scaling is presented in the volume (https://github.com/cornerstonejs/cornerstone3D-beta/blob/578830013cb7b9474166dfe1f469f7908b39aabe/packages/core/src/cache/classes/ImageVolume.ts#L37-L46), and for stack it looks at scaling module or seems like we have StackViewport.scaling as well that you can check.

pwespi:
Added `isSuvScaled` to AnnotationTool base class that checks for viewport type.

sedghi (approved):
thanks, this looks great now

PR #538, opened 2023-04-02
feat: added reset option to Crosshairs

Ouwen:
Adds the option to reset the annotation. This can be called manually on the ToolGroup toolInstance for reseting purposes.

### DIFF
--- a/packages/tools/src/tools/CrosshairsTool.ts
+++ b/packages/tools/src/tools/CrosshairsTool.ts
@@ -181,6 +181,28 @@ class CrosshairsTool extends AnnotationTool {
       defaultReferenceLineSlabThicknessControlsOn;
   }
 
+  resetAnnotations = () => {
+    const viewportsInfo = this._getViewportsInfo();
+    viewportsInfo.forEach(({ viewportId, renderingEngineId }) => {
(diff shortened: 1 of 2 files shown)


packages/tools/src/tools/CrosshairsTool.ts
```diff
@@ -181,6 +181,28 @@ class CrosshairsTool extends AnnotationTool {
       defaultReferenceLineSlabThicknessControlsOn;
   }
 
+  resetAnnotations = () => {
```
sedghi, line 184: can't you disable and enable again the tool?

wayfarer3130 (changes requested):
Could you add example buttons to the crosshairs example tools so that we can see the behaviour when you turn the crosshairs off/on again, and a button to specifically reset everything?
It just isn't quite clear how this would get used in practice.

packages/tools/src/tools/CrosshairsTool.ts
wayfarer3130: At least add an example with resetting in the tool samples, but also try enable/disable to see what happens.

sedghi:
We have added this in main, thanks!

PR #545, opened 2023-04-11
fix(segmentation): Do not render inapplicable segmentations

rodrigobasilio2022:
Add a check of same frameOfReferenceUid inside labelMapDisplay to avoid wrongly displaying segmentations in a configuration of multiple studies, each one having its own segmentation.

### DIFF
--- a/packages/tools/src/tools/displayTools/Labelmap/labelmapDisplay.ts
+++ b/packages/tools/src/tools/displayTools/Labelmap/labelmapDisplay.ts
@@ -4,6 +4,7 @@ import vtkColorTransferFunction from '@kitware/vtk.js/Rendering/Core/ColorTransf
 import {
   cache,
   getEnabledElementByIds,
+  metaData,
   Types,
   utilities,
 } from '@cornerstonejs/core';
@@ -135,6 +136,46 @@ function removeSegmentationRepresentation(
 }
 
 /**
+ * Checks if a segmentation data belongs to the same study as the series
+ * displayed in a given viewport
+ * @param viewport
+ * @param segmentationId
+ * @returns
+ */
+function canRenderSegmentationInViewport(viewport, segmentationId) {
+  const defaultActor = viewport.getDefaultActor();
+  if (defaultActor) {
+    const { uid: defaultActorUID } = defaultActor;
+    const volume = cache.getVolume(defaultActorUID);
+    const viewportImageIds = volume.imageIds;
+
+    if (viewportImageIds) {
+      const segmentation = SegmentationState.getSegmentation(segmentationId);
+      const labelmapData =
+        segmentation.representationData[Representations.Labelmap];
+      const { volumeId: labelmapUID } = labelmapData;
+
+      const labelmap = cache.getVolume(labelmapUID);
+      const segmentationImageIds = labelmap.imageIds;
+      if (segmentationImageIds) {
+        const segmentationSeriesInfo = metaData.get(
+          'generalSeriesModule',
+          segmentationImageIds[0]
+        );
+        const viewportSeriesInfo = metaData.get(
+          'generalSeriesModule',
+          viewportImageIds[0]
+        );
+        return (
+          segmentationSeriesInfo.studyInstanceUID ===
+          viewportSeriesInfo.studyInstanceUID
+        );
+      }
+    }
+  }
+  return false;
+}
+/**
  * It takes the enabled element, the segmentation Id, and the configuration, and
  * it sets the segmentation for the enabled element as a labelmap
  * @param enabledElement - The cornerstone enabled element
@@ -155,6 +196,9 @@ async function render(
     config: renderingConfig,
   } = representation;
 
+  if (!canRenderSegmentationInViewport(viewport, segmentationId)) {
+    return;
+  }
   const segmentation = SegmentationState.getSegmentation(segmentationId);
   const labelmapData =
     segmentation.representationData[Representations.Labelmap];
@@ -182,23 +226,25 @@ async function render(
     actorEntry = viewport.getActor(segmentationRepresentationUID);
   }
 
-  const { cfun, ofun } = renderingConfig as LabelmapRenderingConfig;
+  if (actorEntry) {
+    const { cfun, ofun } = renderingConfig as LabelmapRenderingConfig;
 
-  const renderInactiveSegmentations =
-    toolGroupConfig.renderInactiveSegmentations;
+    const renderInactiveSegmentations =
+      toolGroupConfig.renderInactiveSegmentations;
 
-  _setLabelmapColorAndOpacity(
-    viewport.id,
-    actorEntry,
-    cfun,
-    ofun,
-    colorLUTIndex,
-    toolGroupConfig.representations[Representations.Labelmap],
-    representation,
-    active,
-    renderInactiveSegmentations,
-    segmentsHidden
-  );
+    _setLabelmapColorAndOpacity(
+      viewport.id,
+      actorEntry,
+      cfun,
+      ofun,
+      colorLUTIndex,
+      toolGroupConfig.representations[Representations.Labelmap],
+      representation,
+      active,
+      renderInactiveSegmentations,
+      segmentsHidden
+    );
+  }
 }
 
 function _setLabelmapColorAndOpacity(
--- a/packages/tools/src/tools/displayTools/Labelmap/labelmapDisplay.ts
+++ b/packages/tools/src/tools/displayTools/Labelmap/labelmapDisplay.ts
@@ -139,10 +139,10 @@ function removeSegmentationRepresentation(
  * Checks if a segmentation data belongs to the same study as the series
  * displayed in a given viewport
  * @param viewport
- * @param segmentationId
+ * @param referencedVolumeId volume id of the segmentation reference series
  * @returns
  */
-function canRenderSegmentationInViewport(viewport, segmentationId) {
+function canRenderSegmentationInViewport(viewport, referencedVolumeId) {
   const defaultActor = viewport.getDefaultActor();
   if (defaultActor) {
     const { uid: defaultActorUID } = defaultActor;
@@ -150,13 +150,8 @@ function canRenderSegmentationInViewport(viewport, segmentationId) {
     const viewportImageIds = volume.imageIds;
 
     if (viewportImageIds) {
-      const segmentation = SegmentationState.getSegmentation(segmentationId);
-      const labelmapData =
-        segmentation.representationData[Representations.Labelmap];
-      const { volumeId: labelmapUID } = labelmapData;
-
-      const labelmap = cache.getVolume(labelmapUID);
-      const segmentationImageIds = labelmap.imageIds;
+      const referencedVolume = cache.getVolume(referencedVolumeId);
+      const segmentationImageIds = referencedVolume.imageIds;
       if (segmentationImageIds) {
         const segmentationSeriesInfo = metaData.get(
           'generalSeriesModule',
@@ -175,6 +170,7 @@ function canRenderSegmentationInViewport(viewport, segmentationId) {
   }
   return false;
 }
+
 /**
  * It takes the enabled element, the segmentation Id, and the configuration, and
  * it sets the segmentation for the enabled element as a labelmap
@@ -196,9 +192,6 @@ async function render(
     config: renderingConfig,
   } = representation;
 
-  if (!canRenderSegmentationInViewport(viewport, segmentationId)) {
-    return;
-  }
   const segmentation = SegmentationState.getSegmentation(segmentationId);
   const labelmapData =
     segmentation.representationData[Representations.Labelmap];
@@ -210,6 +203,16 @@ async function render(
     throw new Error(`No Labelmap found for volumeId: ${labelmapUID}`);
   }
 
+  if (segmentation.referencedVolumeId) {
+    if (
+      !canRenderSegmentationInViewport(
+        viewport,
+        segmentation.referencedVolumeId
+      )
+    ) {
+      return;
+    }
+  }
   let actorEntry = viewport.getActor(segmentationRepresentationUID);
 
   if (!actorEntry) {
--- a/packages/tools/src/tools/displayTools/Labelmap/labelmapDisplay.ts
+++ b/packages/tools/src/tools/displayTools/Labelmap/labelmapDisplay.ts
@@ -200,7 +200,7 @@ async function render(
     throw new Error(`No Labelmap found for volumeId: ${labelmapUID}`);
   }
 
-  if (!isSameFrameOfReference(viewport, segmentation.referencedVolumeId)) {
+  if (!isSameFrameOfReference(viewport, labelmapData?.referencedVolumeId)) {
     return;
   }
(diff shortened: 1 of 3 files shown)


packages/tools/src/tools/displayTools/Labelmap/labelmapDisplay.ts
```diff
@@ -134,6 +135,61 @@ function removeSegmentationRepresentation(
...
 }
 
+/**
+ * Checks if a segmentation data belongs to the same study as the series
+ * displayed in a given viewport
+ * @param viewport
+ * @param referencedVolumeId volume id of the segmentation reference series
+ * @returns
+ */
+function canRenderSegmentationInViewport(
```
sedghi, line 145: I think toolGroups are the solution for the need to restrict visualization of the segmentations on any viewport. @wayfarer3130 What do you think? We are breaking the contract between toolGroup's viewports and Segmentation by introducing this flag. It is very odd that there will be a viewport in a toolGroup in the same FrameOfReference that don't render the SEG

packages/tools/src/tools/displayTools/Labelmap/labelmapDisplay.ts
```diff
@@ -134,6 +135,61 @@ function removeSegmentationRepresentation(
...
+        const viewportSeriesInfo = metaData.get(
+          'generalSeriesModule',
+          viewportImageIds[0]
+        );
+        const viewportImageInfo = metaData.get(
+          'imagePlaneModule',
+          viewportImageIds[0]
+        );
+        if (
+          segmentationSeriesInfo.studyInstanceUID ===
```
wayfarer3130, line 177: This restricts the viewing to same study instance UID, NOT the same series instance UID as is specified elsewhere.

packages/tools/src/tools/displayTools/Labelmap/labelmapDisplay.ts
```diff
@@ -134,6 +135,61 @@ function removeSegmentationRepresentation(
...
+            viewportSeriesInfo.studyInstanceUID &&
+          viewportImageInfo.frameOfReferenceUID ===
```
wayfarer3130, line 179: The same FOR is an absolute requirement to visualize the segmentation, so that one shouldn't be occurring now.  If it is, then it is an outright defect and needs to be addressed independently of a flag.

packages/tools/src/tools/displayTools/Labelmap/labelmapDisplay.ts
wayfarer3130: That isn't really possible to do because the tool groups aren't defined per study right now in a way that is actually controllable.  It is a nice theory, I just don't see how to actually manage it.  If the tool group had a setting as part of the segmentation visualization that said same study instance UID, then I would agree, it would work and is probably the right approach.

Also, the images should absolutely require the same frame of reference to be displayed, I'd like that to be tested/confirmed that the same FOR is being provided for different studies images, and that it was intentional, and not accidental.  Really, the same FOR means it is actually the same coordinates, so you can actually display it.

sedghi: Something to always consider is that CS3D don't have drag and drop. CS3D is a rendering and tooling library. Any drag and drop related logic is an app concern.

wayfarer3130: It doesn't matter how one chooses to display a new series, or even if you simply load multiple segmentations initially - the segmentation display is just not valid if it isn't the same FOR, so the labels shouldn't display.

packages/tools/src/tools/displayTools/Labelmap/labelmapDisplay.ts
wayfarer3130: This isn't dependent on D&D, but consider a series with images from different FOR's (forget about the fact right now that we can't render on non-volumes), if you navigate through that series, the different segmentations should apply automatically.
Similarly, if you just load 2 segmentations, and then display a single series, if the FOR is not the segmentation one, then the segmentation must not be displayed because the coordinates don't match.  That is exactly the situation:
There is a mirror 3 steps past the doorway at  123 My Street
and then expecting to find a mirror at 3 steps past he doorway at 456 Your Street.
(mirror at is the segmentation, and the FOR is 123 My Street or 456 Your Street).
I may well load information on both places, but if the main images aren't from either segmentation, then neither applies.

packages/tools/src/tools/displayTools/Labelmap/labelmapDisplay.ts
```diff
@@ -134,6 +135,44 @@ function removeSegmentationRepresentation(
...
 }
 
+/**
+ * Checks if a segmentation data belongs to the same study as the series
+ * displayed in a given viewport
+ * @param viewport
+ * @param referencedVolumeId volume id of the segmentation reference series
+ * @returns
+ */
+function canRenderSegmentationInViewport(viewport, referencedVolumeId) {
```
wayfarer3130, line 145: Please call it isSameFrameOfReference

packages/tools/src/types/SegmentationStateTypes.ts
```diff
@@ -77,6 +77,10 @@ export type Segmentation = {
    * only labelmap representation is supported)
    */
   representationData: SegmentationRepresentationData;
+  /**
```
wayfarer3130, line 80: Should this just be FOR?  Is there any need to have the full volume id?

rodrigobasilio2022: Put it inside the function but I added the volumeOfRefrenceId so I dont want the applications to break if they did not set it

packages/tools/src/tools/displayTools/Labelmap/labelmapDisplay.ts
```diff
@@ -182,23 +231,25 @@ async function render(
     actorEntry = viewport.getActor(segmentationRepresentationUID);
   }
 
-  const { cfun, ofun } = renderingConfig as LabelmapRenderingConfig;
+  if (actorEntry) {
```
sedghi, line 234: how come there will be no actor?, this will not happen since we get the actor if not we create it

packages/tools/src/tools/displayTools/Labelmap/labelmapDisplay.ts
```diff
@@ -134,6 +135,51 @@ function removeSegmentationRepresentation(
...
+ * @returns
+ */
+function isSameFrameOfReference(viewport, referencedVolumeId) {
+  // if the referencedVolumeId is not defined, we acted as before to not break
+  // applications as referencedVolumeId is inserted in this change
+  // Can modify that in the future commits
+  if (referencedVolumeId) {
+    const defaultActor = viewport.getDefaultActor();
+    if (defaultActor) {
+      const { uid: defaultActorUID } = defaultActor;
```
sedghi, line 152: return early if not, to reduce indents

packages/tools/src/tools/displayTools/Labelmap/labelmapDisplay.ts
```diff
@@ -134,6 +135,51 @@ function removeSegmentationRepresentation(
...
+      const volume = cache.getVolume(defaultActorUID);
+      const viewportImageIds = volume.imageIds;
+
```
sedghi, line 155: same

packages/tools/src/tools/displayTools/Labelmap/labelmapDisplay.ts
```diff
@@ -134,6 +135,51 @@ function removeSegmentationRepresentation(
...
+ * displayed in a given viewport
+ * @param viewport
+ * @param referencedVolumeId volume id of the segmentation reference series
+ * @returns
+ */
+function isSameFrameOfReference(viewport, referencedVolumeId) {
+  // if the referencedVolumeId is not defined, we acted as before to not break
+  // applications as referencedVolumeId is inserted in this change
+  // Can modify that in the future commits
+  if (referencedVolumeId) {
```
sedghi, line 149: we should add `FrameOfReferenceUID` to the IImageVolume, and just grab it here

sedghi (commented):
see my comments

packages/tools/src/tools/displayTools/Labelmap/labelmapDisplay.ts
rodrigobasilio2022: if the viewport is a StackViewport that could happen

packages/tools/src/tools/displayTools/Labelmap/labelmapDisplay.ts
```diff
@@ -134,6 +135,51 @@ function removeSegmentationRepresentation(
...
+    if (defaultActor) {
+      const { uid: defaultActorUID } = defaultActor;
+      const volume = cache.getVolume(defaultActorUID);
+      const viewportImageIds = volume.imageIds;
+
+      if (viewportImageIds) {
+        const referencedVolume = cache.getVolume(referencedVolumeId);
+        const segmentationImageIds = referencedVolume.imageIds;
+        if (segmentationImageIds) {
+          const segmentationImageInfo = metaData.get(
```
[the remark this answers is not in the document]
rodrigobasilio2022: Getting from imagevolume

packages/tools/src/tools/displayTools/Labelmap/labelmapDisplay.ts
```diff
@@ -134,6 +135,44 @@ function removeSegmentationRepresentation(
   }
 }
 
+/**
+ * Checks if a segmentation data belongs to the same study as the series
```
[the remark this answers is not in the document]
rodrigobasilio2022: ok

packages/tools/src/tools/displayTools/Labelmap/labelmapDisplay.ts
rodrigobasilio2022: ok

packages/tools/src/tools/displayTools/Labelmap/labelmapDisplay.ts
```diff
@@ -166,6 +205,16 @@ async function render(
     throw new Error(`No Labelmap found for volumeId: ${labelmapUID}`);
   }
 
+  if (segmentation.referencedVolumeId) {
+    if (
```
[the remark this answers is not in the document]
rodrigobasilio2022: ok

packages/tools/src/tools/displayTools/Labelmap/labelmapDisplay.ts
rodrigobasilio2022: ok

packages/tools/src/types/SegmentationStateTypes.ts
```diff
@@ -77,6 +77,10 @@ export type Segmentation = {
...
+   * volume id of the segmentation reference series
+   */
+  referencedVolumeId?: string;
```
sedghi, line 83: a Segmenation object should not be genrarlized to be a labelmap. we already have SegmentationRepresentationData in which it can be labelmap or contour. And in the LabelmapSegmentationData we already have the referencedVolumeId

packages/tools/src/tools/displayTools/Labelmap/labelmapDisplay.ts
sedghi: we don't have stackViewport Segmentation ....

packages/tools/src/tools/displayTools/Labelmap/labelmapDisplay.ts
```diff
@@ -134,6 +134,41 @@ function removeSegmentationRepresentation(
...
+ * Checks if a segmentation data have the same frameOfReference as the series
+ * displayed in a given viewport
+ * @param viewport
+ * @param referencedVolumeId volume id of the segmentation reference series
+ * @returns
+ */
+function isSameFrameOfReference(viewport, referencedVolumeId) {
+  // if the referencedVolumeId is not defined, we acted as before to not break
+  // applications as referencedVolumeId is inserted in this change
+  // Can modify that in the future commits
```
sedghi, line 147: This is not true, we already had it at LabelmapSegmentationData

packages/tools/src/tools/displayTools/Labelmap/labelmapDisplay.ts
```diff
@@ -134,6 +134,41 @@ function removeSegmentationRepresentation(
...
+    if (referencedVolume) {
+      if (
+        volume.metadata.FrameOfReferenceUID ===
+        referencedVolume.metadata.FrameOfReferenceUID
+      ) {
+        return true;
+      }
+    }
+  }
+  return false;
```
sedghi, line 169: how about 


```js
if (volume) {
  const referencedVolume = cache.getVolume(referencedVolumeId);
  if (referencedVolume && volume.metadata.FrameOfReferenceUID === referencedVolume.metadata.FrameOfReferenceUID) {
    return true;
  }
}
return false;
```

sedghi (changes requested):
see my comments

packages/tools/src/types/SegmentationStateTypes.ts
rodrigobasilio2022: There is two points here: the referenced series data will always be an dicom image data, where we can get the dimensions and other properties, if any not descriped in representationalData. So, we can call it referencedVolumeId, And it should be valid independent of the segmentation representation, so it would be best if it goes in segmentation level, not in representational level. And although the representationData has the field, currently the value has not been set.

packages/tools/src/tools/displayTools/Labelmap/labelmapDisplay.ts
rodrigobasilio2022: Although the representationData has the field, currently the value has not been set

packages/tools/src/tools/displayTools/Labelmap/labelmapDisplay.ts
rodrigobasilio2022: Done

packages/tools/src/tools/displayTools/Labelmap/labelmapDisplay.ts
rodrigobasilio2022: Removed the comments

packages/tools/src/tools/displayTools/Labelmap/labelmapDisplay.ts
```diff
@@ -166,6 +200,9 @@ async function render(
     throw new Error(`No Labelmap found for volumeId: ${labelmapUID}`);
   }
 
+  if (!isSameFrameOfReference(viewport, labelmapData?.referencedVolumeId)) {
```
sedghi, line 203: is this correct? should it be referencedVolumeId or referenceSeriesInstanceUID?

rodrigobasilio2022: It is faster if use the referencedVolumeId, because I need the FrameOfReference of the referencedSeries

sedghi: then why did you add the referenceSeriesInstanceUID at all? I don't see that being used anywhere

sedghi (approved):

PR #547, opened 2023-04-12
feat(crosshairs): added threshold for jumpWorld

Ouwen:
Adds a configurable threshold for jumpWorld.


packages/tools/src/tools/CrosshairsTool.ts
```diff
@@ -71,6 +71,7 @@ interface ToolConfiguration {
       opacity: number;
       handleRadius: number;
     };
+    jumpThreshold: number;
```
sedghi, line 74: could you please add the documentation for the `jumpThreshold` ? I'm not sure what it does

PR #551, opened 2023-04-13
fix(crosshairs): Autopan causing infinite loop

rogerbramon:
Currently, `autopan` option of the CrosshairsTool is only working when using predefined orientations (Axial, Sagittal, Coronal). As soon as you rotate the axis, autopan may cause an infinite loop. This PR aims to fix that, changing the implementation to use the canvas context.

In case you want to reproduce the issue, you can:
1. Add ZoomTool or PanTool to the crosshairs example
2. Enable autopan to the example
3. On the example, apply pan or zoom to one of the non main views
4. Rotate the axis 
5. Select a point in the main view that's out of the viewpoint in one of the other views.


sedghi (approved):

PR #552, opened 2023-04-13
fix(crosshairs): Reference lines are wrongly clipped

rogerbramon:
Currently, the reference lines of the Crosshairs Tool may be wrongly clipped when the other viewports have a smaller size.
This PR aims to fix this issue by using the canvasBox of the current viewport instead of the other viewport for clipping lines.

wrong-clipping


sedghi (approved):

PR #553, opened 2023-04-14
feat(dicomImageLoader): make cornerstone to use new dicom image loader and handle scaling correctly

sedghi:
Blocked by https://github.com/Kitware/vtk-js/pull/2786 to get merged first 

--- 

Tried on various scenarios, multiple transfer syntaxes grayscale and color

- GPU
  - StackViewport -> with and without preferSizeOverAccuracy
  - Volume Viewport -> SAB or not

- CPU
  - StackViewport 

---

- StackViewport properly implements the preferSizeOverAccuracy
- Delegate data type decision to DICOMImageLoader based on the whether scaling should be needed or not 
- Fixed the libjpeg-turbo 8-bit
- Fixed the YBR JPEG images rendering
- Fixes the CPU rendering with Float results
- Fixed the CPU and window levelling
- Add new dicomImageLoader example loading various data 
- Changed examples in the dicom image loader -> examplesOld (just to keep them around for now)
- Make color images to use Uint8 instead of Uint8ClampedArray to avoid vtk that was converting it to float
- Calculate min and max inside the worker prior to deciding on the data type


sedghi:
@Ouwen Could you please review this code as well?

packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -1427,33 +1433,16 @@ class StackViewport extends Viewport implements IStackViewport {
     bitsAllocated,
     numComps,
     numVoxels,
-    TypedArray,
+    typedArray,
```
wayfarer3130, line 1436: typedArray is the type of the thing, not what it is in terms of usage - please make this one pixelArray, as it is an array of pixels that happens to be of type TypedArray for general usage.

packages/dicomImageLoader/src/imageLoader/createImage.ts
```diff
@@ -283,7 +234,7 @@ function createImage(
           // If we don't need the RGBA but the decoding is done with RGBA (the case
           // for JPEG Baseline 8 bit color), AND the option specifies to use RGB (no RGBA)
           // we need to remove the A channel from pixel data
-          const colorBuffer = new Uint8ClampedArray(
+          const colorBuffer = new Uint8Array(
```
wayfarer3130, line 237: Please add a comment about vtk needing the Uint8Array here - it wouldn't be good to change that back unknowing.

packages/dicomImageLoader/src/imageLoader/createImage.ts
```diff
@@ -394,6 +324,14 @@ function createImage(
           canvas.width = image.columns;
           const context = canvas.getContext('2d');
 
+          // Create an ImageData object from the Uint8ClampedArray
+          const imageData = new ImageData(
```
wayfarer3130, line 328: Why is this one clamped while the other one isn't?

packages/dicomImageLoader/src/shared/getTypedArrayFromMinMax.ts
```diff
@@ -0,0 +1,23 @@
+export default function getTypedArrayFromMinMax(min, max) {
+  let TypedArray;
```
wayfarer3130, line 2: This function can return undefined - please have this start with:
let pixelArrayType = Float32Array;
Also, the function name implies you are getting a typed array, but what you are getting is the type of the pixel array, so please call it something like  getPixelArrayTypeFromMinMax   I got confused looking at the function expecting it to return a typed array instance.

packages/dicomImageLoader/src/shared/getTypedArrayFromMinMax.ts
```diff
@@ -0,0 +1,23 @@
+export default function getTypedArrayFromMinMax(min, max) {
```
wayfarer3130, line 1: Shouldn't the type of this by:
getPixelDataTypeFromMinMax(min: number, max: number): (() => TypedArray)) { ....
I think that will declare it as a funtion returning a constructor for TypedArray

wayfarer3130 (changes requested):
There are a number of smallish changes that I'd like to see before approving/merging.

packages/dicomImageLoader/src/shared/getTypedArrayFromMinMax.ts
sedghi: There is an else statement which makes it Float32, typescript will yell if I do `let pixelArrayType = Float32Array;`

packages/dicomImageLoader/src/imageLoader/createImage.ts
```diff
@@ -187,66 +150,56 @@ function createImage(
...
         }
 
-        // If need to scale, need to scale correct array.
         imageFrame.pixelData = typedArray;
         alreadyTyped = true;
       }
 
       if (!alreadyTyped) {
-        setPixelDataType(imageFrame, imageFrame.preScale);
+        setPixelDataType(imageFrame);
```
Ouwen, line 202: is this done after scaling?

packages/dicomImageLoader/src/shared/decodeImageFrame.ts
```diff
@@ -186,108 +189,184 @@ function postProcessDecodedPixels(
...
-    let offset, length;
-    // If we have a target buffer, write to that instead. This helps reduce memory duplication.
+  const typedArrayConstructors = {
+    Uint8Array,
+    Uint16Array: use16BitDataType ? Uint16Array : undefined,
+    Int16Array: use16BitDataType ? Int16Array : undefined,
+    Float32Array,
+  };
+
+  const isColoImage = isColorImage(imageFrame.photometricInterpretation);
```
Ouwen, line 204: `isColorImage`

packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -1573,32 +1579,44 @@ class StackViewport extends Viewport implements IStackViewport {
...
+    // with RGBA, and here we need to remove the A channel from the
+    // pixel data.
+    if (image.color && image.rgba) {
+      const newPixelData = new Uint8Array(image.columns * image.rows * 3);
+      for (let i = 0; i < image.columns * image.rows; i++) {
+        newPixelData[i * 3] = pixelData[i * 4];
+        newPixelData[i * 3 + 1] = pixelData[i * 4 + 1];
+        newPixelData[i * 3 + 2] = pixelData[i * 4 + 2];
       }
+      pixelData = newPixelData;
```
wayfarer3130, line 1600: This line is meaningless - you never access pixelData after this point again.

packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -1573,32 +1579,44 @@ class StackViewport extends Viewport implements IStackViewport {
 
     // 1. Update the pixel data in the vtkImageData object with the pixelData
     //    from the loaded Cornerstone image
-    const pixelData = image.getPixelData();
```
wayfarer3130, line 1576: Please undo this line - it isn't needed as the assignment to pixelData is never used.

packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -1573,32 +1579,44 @@ class StackViewport extends Viewport implements IStackViewport {
...
+      //  for (let i = 0; i < numPixels; i++) {
+      //    pixelDataWithoutAlpha[rgbIndex++] = pixelData[rgbaIndex++]; // red
+      //    pixelDataWithoutAlpha[rgbIndex++] = pixelData[rgbaIndex++]; // green
+      //    pixelDataWithoutAlpha[rgbIndex++] = pixelData[rgbaIndex++]; // blue
+      //    rgbaIndex++; // skip alpha
+      //  }
+
+      // modify the image object to have the correct pixel data for later
+      // use.
+      image.rgba = false;
```
wayfarer3130, line 1617: Any way the image could be replaced rather than being modified in place?  Will this ever cause issues with other display environments needing rgba such as the thumbnails?

packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -1645,6 +1663,50 @@ class StackViewport extends Viewport implements IStackViewport {
           return;
         }
 
+        const pixelData = image.getPixelData();
+
+        // handle the case where the pixelData is a Float32Array
+        // CPU path cannot handle it, it should be converted to Uint16Array
+        // and via the Modality LUT we can display it properly
+        if (pixelData instanceof Float32Array) {
```
wayfarer3130, line 1671: This feels like a change that could be directly added to the Image class, rather than being in StackViewport, and potentially required for other Viewport types?

packages/core/src/RenderingEngine/StackViewport.ts
sedghi: This is an edge case + optimization where image is rendered with rgba first, I guess I can make it not replace at all just uses the new pixelData, but every other call will do the same operation. Plus, I have a PR to make thumbnails use GPU soon

packages/core/src/RenderingEngine/StackViewport.ts
sedghi: This is for CPU, image class doesn't care about GPU or CPU

packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -1667,6 +1729,11 @@ class StackViewport extends Viewport implements IStackViewport {
           this._cpuFallbackEnabledElement.viewport.colormap
         );
 
+        const { windowCenter, windowWidth } = viewport.voi;
+        this.voiRange = {
```
wayfarer3130, line 1733: There is no need to create a secondary object to hold this return as toLowHighRange returns a new object every time.

packages/core/src/init.ts
```diff
@@ -68,21 +71,21 @@ function hasSharedArrayBuffer() {
   }
 }
 
-function _hasNorm16TextureSupport() {
-  const gl = _getGLContext();
+// function _hasNorm16TextureSupport() {
```
wayfarer3130, line 74: Can you add a reason for commenting this out rather than removing it?

wayfarer3130 (approved):

Ouwen:
🎆 🍾 🎉

PR #556, opened 2023-04-17
Improve performance for `getSegmentationIndices`

### LINKED ISSUE #546: Drop in performance when showing/hiding a segmentation representation
After updating to the latest version, I noticed that showing or hiding a segmentation representation (tested only on labelmap) takes more than a second to react when it used to be instantaneous.

After running a profile, I suspect that the problem was introduced on #477 from @sedghi, however I was not able to check it further because of an error when using the prior version.

Evaluating the profile, it's seems that the cause is in `getSegmentationIndices` and the amount of times that `scalarData.forEach` is called. (`setToolDisabled` happens twice because I have two ToolGroups)

image

This is the time spent on `@ segmentationVisibility.js:12:1`:
image

Let me know if you need any more info. Thanks

rogerbramon:
As described in #546, toggling the visibility of a segmentation could be quite slow for large volumes. This PR aims to improve the performance of this action. This fixes #546.

### DIFF
--- a/packages/tools/src/stateManagement/segmentation/config/segmentationVisibility.ts
+++ b/packages/tools/src/stateManagement/segmentation/config/segmentationVisibility.ts
@@ -13,7 +13,12 @@ function getSegmentationIndices(segmentationId) {
     const scalarData = volume.getScalarData();
 
     const keySet = {};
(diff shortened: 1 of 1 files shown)


rogerbramon:
Thanks @sedghi, feel free to merge it.

PR #557, opened 2023-04-17
feat(crosshairs): Make the reference lines gap configurable

rogerbramon:
Currently, the gap between the reference lines depends on the size of the canvas, causing it to be variable. This means that when working on large viewports, the gap could be too large. Moreover, there's no consistency if all viewports don't have the same size.

This PR proposes to have a fixed and configurable gap size. Moreover, it aims to improve position of the controls to prevent them from being out of view depending on the proportion of the canvas. As you can see, depending on the size of the canvas, the controls may be out of view. 

Current:
image

Proposal:
image

You can check the proposal on the `crosshairs` example. Let me know your feedback.

### DIFF
--- a/packages/tools/src/tools/CrosshairsTool.ts
+++ b/packages/tools/src/tools/CrosshairsTool.ts
@@ -775,6 +775,7 @@ class CrosshairsTool extends AnnotationTool {
     const canvasDiagonalLength = Math.sqrt(
       clientWidth * clientWidth + clientHeight * clientHeight
     );
+    const canvasMinDimensionLength = Math.min(clientWidth, clientHeight);
 
     const data = viewportAnnotation.data;
     const crosshairCenterCanvas = viewport.worldToCanvas(this.toolCenter);
@@ -883,7 +884,6 @@ class CrosshairsTool extends AnnotationTool {
       //                           Long
       const canvasVectorFromCenterLong = vec2.create();
 
-      // Todo: configuration should provide constants below (100, 0.25, 0.15, 0.04)
       vec2.scale(
         canvasVectorFromCenterLong,
         canvasUnitVectorFromCenter,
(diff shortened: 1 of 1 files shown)


packages/tools/src/tools/CrosshairsTool.ts
```diff
@@ -888,20 +893,21 @@ class CrosshairsTool extends AnnotationTool {
       vec2.scale(
         canvasVectorFromCenterMid,
         canvasUnitVectorFromCenter,
-        canvasDiagonalLength * 0.25
+        canvasMinDimensionLength * 0.4
```
sedghi, line 896: I'm not sure why it was 0.25 and not you are suggesting 0.4, same for below

rogerbramon: Let's see if I can explain myself...

Notice that I'm not using the diagonal, but the shorter side of the canvas because currently it can happen that they are out of canvas. In my screenshot of the description, you can see that only the controls of the slab thickness are visible, the rotation controls are out. And, it could also happen that all are out depeding on the canvas dimensions ratio, forcing you to pan the image or change the crosshair center point to be able to use the controls.

To ensure that at least the controls of one side are visible, the outer ones (rotation) should be placed at half the length of the shorter side. **Thus, should be 0.5 at most, and I decided to use 0.4 to have some margin to the edge. 0.2 is because it's in the middle of 0 and 0.4.**

sedghi: This is great explanation, thanks!
Can you put this comment in the code as well then we can merge

rogerbramon: Great. Done!

sedghi (approved):

PR #560, opened 2023-04-17
fix(mouse): Avoid the delay on double click checking for right click

wayfarer3130:
The double click check was occurring for right click, delaying context menus appearing.


jbocce (approved):
Thanks for the change. It looks good!

wayfarer3130:
> Thanks for the change. It looks good!

Can you hit approved, then I will merge it.

jbocce:
> > Thanks for the change. It looks good!
> 
> Can you hit approved, then I will merge it.

I did hit approve yesterday.

jbocce (approved):

PR #561, opened 2023-04-17
fix(stackviewport): swap image row and column pixel spacing

sedghi:
This commit swaps image row and column pixel spacing of `StackViewport` in `RenderingEngine` while applying an epsilon comparison for float numbers comparison. It also adds Zoom tool to the `StackScrollToolMouseWheelTool_test.js` test script to scroll through stacks of images. Finally, it adjusts the test image data and updates the `yarn.lock` file to include codec-openjph 2.4.2 package.

fixes https://github.com/cornerstonejs/cornerstone3D-beta/issues/548


wayfarer3130 (approved):
LGTM

sedghi:
you are right

PR #562, opened 2023-04-18
feat(4D): fixed cine play issue and added getDynamicVolumeInfo method

lscoder:
- cine is checking if there are any 4D volume loaded on a viewport instead of checking only if the default actor is a 4D series because when there are PT (4D) + CT (3D) it has to play as 4D series otherwise it can get into a state where some PT viewports plays as 4D and PT/CT viewports plays as 3D moving the 4D series in time and space at the same time.
- added getDynamicVolumeInfo that is being used by the Viewer to check if a series is a 4D series. That is the same method used when a 4D series is loaded by Cornerstone.

### DIFF
--- a/packages/tools/src/utilities/cine/playClip.ts
+++ b/packages/tools/src/utilities/cine/playClip.ts
@@ -298,16 +298,15 @@ function _stopClipWithData(playClipData) {
   }
 }
 
+function _getVolumesFromViewport(viewport): Types.IImageVolume[] {
+  return viewport.getActors().map((actor) => cache.getVolume(actor.uid));
+}
+
 function _getVolumeFromViewport(viewport): Types.IImageVolume {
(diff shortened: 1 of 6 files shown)


packages/tools/src/utilities/cine/playClip.ts
```diff
@@ -298,16 +298,15 @@ function _stopClipWithData(playClipData) {
   }
 }
 
-function _getVolumeFromViewport(viewport): Types.IImageVolume {
-  const actorEntry = viewport.getDefaultActor();
+function _getVolumesFromViewport(viewport): Types.IImageVolume[] {
+  return viewport.getActors().map((actor) => cache.getVolume(actor.uid));
```
sedghi, line 302: there might be non volume actors, this should have a `.Boolean()` at the end to filter them

sedghi (approved):

PR #571, opened 2023-04-19
fix: Double click and multi-key bindings

wayfarer3130:
Changed the double click handling to correctly handle the button click for double click selection as well as for context menu click selection.
Made the wheel scrolling go through the existing event handling by adding a new
namedEvent property to allow binding on the wheel specifically in the normal way rather than needing custom handling to control the binding.

Test by running the modifierKey example, and the doubleClickWithStackAnnotationTools examples.
`yarn example doubleClickWithStackAnnotationTools`
http://localhost:3000

### DIFF
--- a/packages/tools/examples/modifierKeys/index.ts
+++ b/packages/tools/examples/modifierKeys/index.ts
@@ -16,6 +16,9 @@ const {
   LengthTool,
   RectangleROITool,
   BidirectionalTool,
+  ZoomTool,
+  PanTool,
+  StackScrollTool,
   ToolGroupManager,
   Enums: csToolsEnums,
 } = cornerstoneTools;
@@ -45,10 +48,14 @@ content.appendChild(element);
 
 const instructions = document.createElement('p');
 instructions.innerText = `
-- Left click to use the Window/Level tool.
-- Shift + Left click to use the Length tool.
-- Ctrl + Left click to use the Bidirectional tool.
-- Alt/Option + Left click to use the RectangleROI tool.
+- Single touch is equivalent to left click.
+- Left or Meta+Left for stack scroll.
+- Right or Option to use the Window/Level tool.
+- Center or Ctrl to Pan.
+- Shift to Zoom.
+- Shift/Ctrl click to use the Length tool.
+- Ctrl/Alt click to use the Bidirectional tool.
+- Shift/Alt + Left click to use the RectangleROI tool.
 `;
 
 content.append(instructions);
@@ -68,6 +75,9 @@ async function run() {
   cornerstoneTools.addTool(LengthTool);
   cornerstoneTools.addTool(RectangleROITool);
   cornerstoneTools.addTool(BidirectionalTool);
+  cornerstoneTools.addTool(StackScrollTool);
+  cornerstoneTools.addTool(PanTool);
+  cornerstoneTools.addTool(ZoomTool);
 
   // Define a tool group, which defines how mouse events map to tool commands for
   // Any viewport using the group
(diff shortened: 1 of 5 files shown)


packages/tools/src/eventListeners/mouse/mouseDownListener.ts
```diff
@@ -144,7 +144,7 @@ function mouseDownListener(evt: MouseEvent) {
 
   state.element = <HTMLDivElement>evt.currentTarget;
 
-  state.mouseButton = evt.button;
+  state.mouseButton = evt.buttons;
```
ladeirarodolfo, line 147: this is something that might be used to get multi buttons at the same time. But depending on how user press them we might have different situations. I.e 
1. left then right then move
1. right then left then move
1. right then move a bit then left then move

wayfarer3130: There absolutely are different setups, so it only works for interactive tools like Pan, and not click based tools like annotations.  There is no distinction between the various click options if you only consider tools that listen to move.

packages/tools/src/eventListeners/wheel/wheelListener.ts
```diff
@@ -33,19 +37,37 @@ function wheelListener(evt: WheelEvent) {
...
+
+  const activeTool = getActiveToolForNamedEvent('wheel', event);
+  if (activeTool) {
+    if (!activeTool.mouseWheelCallback) {
+      throw new Error(`Tool ${activeTool.toolName} missing mouseWheelCallback`);
+    }
+    activeTool.mouseWheelCallback(eventDetail);
+  } else {
+    // TODO - deprecate this handling entirely in favour of the active handling
+    triggerEvent(element, Events.MOUSE_WHEEL, eventDetail);
```
jbocce, line 69: Do we really want to deprecate `Events.MOUSE_WHEEL`?

packages/tools/src/types/EventTypes.ts
```diff
@@ -376,6 +376,10 @@ type MouseWheelEventDetail = NormalizedInteractionEventDetail &
       pixelY: number;
       direction: number;
     };
+    /** Pretend to be a drag */
```
jbocce, line 379: Is mouse wheel really like a drag?

packages/tools/src/eventDispatchers/shared/getActiveToolForMouseEvent.ts
```diff
@@ -22,10 +22,11 @@ export default function getActiveToolForMouseEvent(
   const { renderingEngineId, viewportId } = evt.detail;
   const mouseEvent = evt.detail.event;
 
-  // If any keyboard modifier key is also pressed
-  // Use the actual key if set, otherwise get the key from the mouse event.
+  // If any keyboard modifier key is also pressed - get the mouse version
+  // first since it handle combinations, while the key event handles non-modifier
```
jbocce, line 26: Typo "... since it handles*..."

jbocce (changes requested):
IMHO, we should split this into two or more PRs. One to fix the broken double click and others to address mouse wheel and maybe modifier keys. 

I think have a closer look at mouse wheel on its own would be best.

packages/tools/examples/modifierKeys/index.ts
wayfarer3130, line 1: The modifierKeys example is a primary test case for this, so I have extended it with various handling to enable testing all the different modifier keys and combinations.
Note that two button mouse combinations which add annotations are not tested, as that is not currently or previously supported - a keyboard combination + mouse click works, because the timing there is very specific, but not two mouse buttons.

wayfarer3130:
@jbocce and @sedghi - removed all enhancements and only made fixes to double click and dispatch of events, plus added a complete set of (working) examples.  Removed the comment about the example not working, as it does work correctly now.

packages/tools/src/eventListeners/mouse/mouseDownListener.ts
```diff
@@ -144,7 +144,7 @@ function mouseDownListener(evt: MouseEvent) {
...
```
jbocce, line 148: Changing this from `button` to `buttons` (on the surface) feels problematic because don't all of the `MouseBindings` enum values relate to the old `button` values? If so, please revert this back to `button`.

jbocce: nevermind... I was wrong

jbocce: Despite the fact I was wrong, I do not understand how the bindings were actually functional before this change... likely because the original event is stored in `eventDetail`. That said, it bothers me that `buttons` is now being assigned to the singular `mouseButton`. @sedghi any thoughts?

jbocce: I just checked and the `modifierkeys` example works even with the former/old code (and the rest of you other changes of course), so what is this particular change for actually? I apologize if these are stupid questions.

wayfarer3130: The modifierKeys only worked for single modifiers, not combined modifiers, so this fix addresses that issue.
The bindings is somewhat problematic - I'm not sure why the state setup is there exactly.

jbocce: To clarify, the `modifierkeys` example still works if I revert line 147. So my question is why do we need the change for line 147. I understand we need all the other changes for combined modifiers.

wayfarer3130: That allows the event to support multiple buttons, whereas using the .button doesn't allow for that.

jbocce: I understand. To clarify again, `state.mouseButton` appears (historically) to be for 'button` and the `buttons` value can be fetched from `eventDetail.event.buttons`. So either do not make this change, or rename `state.mouseButton` to `state.mouseButtons` so that there is no ambiguity. I recommend NOT making the change in case somebody, somewhere in the world is relying on it being `button`.

sedghi (approved):

packages/tools/src/eventListeners/mouse/mouseDownListener.ts
sedghi: I guess it is not that bad, adding another property to state is an overkill for this imo

PR #574, opened 2023-04-20
feat(tools): Add invert zoom option

wayfarer3130:
This change adds the ability to invert the zoom direction as an option.


wayfarer3130:
This change mirrors equivalent changes in other tools.

PR #577, opened 2023-04-25
fix(calibration): Apply the update only once

wayfarer3130:
The calibration was being applied multiple times on each view.  Now apply it just a single time.


sedghi (approved):
Thanks

Issue #570, opened 2023-04-19
Annotation: duplicate event "annotation added"?


jlopes90:
When drawing annotation it fires "annotation added" event but then "triggerAnnotationAddedForFOR" means get all viewports and fire "annotation added" event again.

https://github.com/cornerstonejs/cornerstone3D-beta/blob/cc3ce18f1c601fc911d8866987ae369a8f4c28df/packages/tools/src/stateManagement/annotation/annotationState.ts#L87-L96

iamiota:
same +1, putting `triggerAnnotationAddedForFOR` in `else` can solve it.

PR #584, opened 2023-04-28
fix triggering of 'ANNOTATION_ADDED' event multiple times (#570)

iamiota:
Fix triggering of 'ANNOTATION_ADDED' event multiple times (#570)


sedghi (approved):

sedghi:
Thanks for your contribution

PR #586, opened 2023-04-28
feat/render to canvas gpu

sedghi:
renderToCanvas which was used in OHIF for thumbnails was using CPU before, and caused so many discrepancies. This PR add the GPU rendering to it by default

You can try it here https://deploy-preview-586--cornerstone-3d-docs.netlify.app/live-examples/rendertocanvas


packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -1924,6 +1921,22 @@ class StackViewport extends Viewport implements IStackViewport {
     });
   }
 
+  /**
+   * Given the cornerstone image object it will render it in the viewport
+   * This method is used by some utilities, for loading an actual image use
+   * the setStack method and not this one.
```
wayfarer3130, line 1927: Could you add a comment about where to look for the rendered content?  It feels like this should be able to render to a canvas or the like, but it isn't obvious that one can't do that.

packages/core/src/RenderingEngine/renderingEngineCache.ts
```diff
@@ -36,6 +36,18 @@ const renderingEngineCache = {
     const renderingEngineIds = Object.keys(cache);
     const renderingEngines = renderingEngineIds.map((id) => cache[id]);
 
+    // sort the rendering engines so that the ones that start with _
```
wayfarer3130, line 39: Please add an explanation of why the _ ones are at the end, they are .... and applied as ... because ...

packages/core/src/types/ViewportInputOptions.ts
```diff
@@ -1,12 +1,14 @@
...
 import DisplayArea from './displayArea';
+import Point3 from './Point3';
+
 /**
  * This type defines the shape of viewport input options, so we can throw when it is incorrect.
  */
 type ViewportInputOptions = {
   /** background color */
-  background?: [number, number, number];
+  background?: Point3;
```
wayfarer3130, line 11: I thought we had a color type?

wayfarer3130:
Could you add a runnable example of this, that gets the canvas rendered and then gets the data url for that and displays the data url, maybe dropping it into a copyable text box that the user can test with?

packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -1924,6 +1921,22 @@ class StackViewport extends Viewport implements IStackViewport {
     });
   }
 
+  /**
+   * Given the cornerstone image object it will render it in the viewport
+   * This method is used by some utilities, for loading an actual image use
```
jbocce, line 1926: I do not really understand this sentence. Perhaps it should be broken into two sentences? Maybe a period after `utilities`?

packages/core/src/RenderingEngine/renderingEngineCache.ts
jbocce: I second this.

packages/core/src/utilities/renderToCanvasGPU.ts
```diff
@@ -0,0 +1,120 @@
...
+  image: IImage,
+  modality = undefined,
+  renderingEngineId = '_thumbnails'
+): Promise<string> {
+  if (!canvas || !(canvas instanceof HTMLCanvasElement)) {
+    throw new Error('canvas element is required');
+  }
+
+  const imageIdToPrint = image.imageId;
+  const viewportId = `thumbnailViewport-${imageIdToPrint}`;
```
jbocce, line 38: Why this viewport id? Why not something more generic like `renderGPUViewport-${imageIdToPring}`? Pardon me if this is a stupid question.

packages/core/src/utilities/renderToCanvasGPU.ts
```diff
@@ -0,0 +1,120 @@
...
+  const imageId = image.imageId;
+  const element = document.createElement('div');
+  element.style.width = `${canvas.width}px`;
+  element.style.height = `${canvas.height}px`;
+
+  // Todo: we should be able to use the temporary element without appending
+  // it to the DOM
+  element.style.visibility = 'hidden';
+  document.body.appendChild(element);
```
jbocce, line 47: Hmm, simply adding it to the body with visibility hidden might disrupt the layout of the body. Perhaps consider setting `element.style.position='absolute'`. But I could be wrong here.

jbocce (commented):
The changes look good. Just some comments and please add testing instructions.

packages/core/src/RenderingEngine/renderingEngineCache.ts
jbocce: Thanks for filling in the comment nicely. Perhaps also mention what the underscore signifies?

packages/core/test/renderToCanvas_gpu_test.js
```diff
@@ -0,0 +1,162 @@
...
+          vp.setProperties({ interpolationType: InterpolationType.NEAREST });
+          vp.render();
+        });
+      });
+    } catch (e) {
+      done.fail(e);
+    }
+  });
+
+  it('Should render one stack viewport of square size properly: nearest', function (done) {
```
jbocce, line 115: This test name is the same as the previous. What can we do to distinguish it?

packages/core/test/renderToCanvas_gpu_test.js
```diff
@@ -0,0 +1,162 @@
...
+    metaData.removeProvider(fakeMetaDataProvider);
+    imageLoader.unregisterAllImageLoaders();
+    this.DOMElements.forEach((el) => {
+      if (el.parentNode) {
+        el.parentNode.removeChild(el);
+      }
+    });
+  });
+
+  it('Should render one stack viewport of square size properly: nearest', function (done) {
```
jbocce, line 67: Ok stupid question time, but shouldn't the tests be marked async or something considering that there is a promise to `loadImageToCanvas` or maybe there is a magical `await` somewhere and I am simply none the wiser. 🤣

packages/dicomImageLoader/README.md
```diff
@@ -115,6 +115,9 @@ path is relative to the `node_modules` folder in the OHIF Viewer.
 Note 2: For other http servers like IIS, you need to configure it to serve WASM
 files with the correct MIME type.
 
+## Loading
+
+![Alt text](assets/load.png)
```
jbocce, line 120: very nice addition!

jbocce (commented):
Almost there... stay on target (yesterday was May the 4th) :lol. Minor comments.

packages/core/examples/renderToCanvas/index.ts
```diff
@@ -0,0 +1,132 @@
...
+    renderingEngine.setViewports(viewportInputArray);
+
+    // Get the stack viewport that was created
+    const viewport1 = <Types.IStackViewport>(
+      renderingEngine.getViewport(viewportIds[0])
+    );
+
+    const imageId = imageIds[100];
+
+    utilities.loadImageToCanvas({ canvas, imageId, useCPURendering });
```
wayfarer3130, line 121: Thank you - it wasn't clear how this was happening

packages/core/test/renderToCanvas_gpu_test.js
sedghi: No, the done does the same thing

jbocce (approved):
Approved.

PR #588, opened 2023-05-01
chore(lerna): upgrade lerna to version 6

sedghi:


wayfarer3130 (approved):
Looks good to me, and since it seems to be running in the test environment it should be fine.

PR #590, opened 2023-05-02
fix(planarFreehandROITool): proper handling of pure movements on y-axis

### LINKED ISSUE #589: planarFreehandROITool: pure movements on y-axis are not handled properly
The planarFreehandROITool on a VolumeViewport seems to "lag", when the mouse movement is purely on the y-axis (up/down, no movement in x).

Actually, I found out that it doesn't lag, but that such movements are not handled properly and in such cases, new points are not added to the polygon  and drawn to the viewport (that's why users have the impression of "lagging").

The reason, why such movements are rejected, is that the two dot products in https://github.com/cornerstonejs/cornerstone3D-beta/blob/main/packages/tools/src/tools/annotation/planarFreehandROITool/drawLoop.ts#L111-L112 always evaluate to Zero in case of pure y-movements.

And the reason therefor seems to be that the wrong normal vector is used here: https://github.com/cornerstonejs/cornerstone3D-beta/blob/main/packages/tools/src/utilities/math/polyline/getSubPixelSpacingAndXYDirections.ts#L69-L71

I could fix the problem with the following diff:

```Diff
index da1f6d11..708dbd10 100644
--- a/packages/tools/src/utilities/math/polyline/getSubPixelSpacingAndXYDirections.ts
+++ b/packages/tools/src/utilities/math/polyline/getSubPixelSpacingAndXYDirections.ts
@@ -66,19 +66,19 @@ const getSubPixelSpacingAndXYDirections = (
       throw new Error('No support yet for oblique plane planar contours');
     }
 
-    const absViewPlaneNormalDotI = Math.abs(vec3.dot(viewPlaneNormal, iVector));
-    const absViewPlaneNormalDotJ = Math.abs(vec3.dot(viewPlaneNormal, jVector));
-    const absViewPlaneNormalDotK = Math.abs(vec3.dot(viewPlaneNormal, kVector));
+    const absViewUpDotI = Math.abs(vec3.dot(viewUp, iVector));
+    const absViewUpDotJ = Math.abs(vec3.dot(viewUp, jVector));
+    const absViewUpDotK = Math.abs(vec3.dot(viewUp, kVector));
 
     // Get Y spacing
     let ySpacing;
-    if (Math.abs(1 - absViewPlaneNormalDotI) < EPSILON) {
+    if (Math.abs(1 - absViewUpDotI) < EPSILON) {
       ySpacing = volumeSpacing[0];
       yDir = iVector;
-    } else if (Math.abs(1 - absViewPlaneNormalDotJ) < EPSILON) {
+    } else if (Math.abs(1 - absViewUpDotJ) < EPSILON) {
       ySpacing = volumeSpacing[1];
       yDir = jVector;
-    } else if (Math.abs(1 - absViewPlaneNormalDotK) < EPSILON) {
+    } else if (Math.abs(1 - absViewUpDotK) < EPSILON) {
       ySpacing = volumeSpacing[2];
       yDir = kVector;
     } else {
```  
  
(I would have created a PR, but I don't have permissions)

m-barthel:
fixes #589   
  
@JamesAPetts FYI you changed these lines the last time in https://github.com/cornerstonejs/cornerstone3D-beta/commit/7c33fcc529a20fd5cbf2e35e0ad995ba79acd834

### DIFF
--- a/packages/tools/src/utilities/math/polyline/getSubPixelSpacingAndXYDirections.ts
+++ b/packages/tools/src/utilities/math/polyline/getSubPixelSpacingAndXYDirections.ts
@@ -66,19 +66,19 @@ const getSubPixelSpacingAndXYDirections = (
       throw new Error('No support yet for oblique plane planar contours');
     }
 
-    const absViewPlaneNormalDotI = Math.abs(vec3.dot(viewPlaneNormal, iVector));
-    const absViewPlaneNormalDotJ = Math.abs(vec3.dot(viewPlaneNormal, jVector));
-    const absViewPlaneNormalDotK = Math.abs(vec3.dot(viewPlaneNormal, kVector));
+    const absViewUpDotI = Math.abs(vec3.dot(viewUp, iVector));
+    const absViewUpDotJ = Math.abs(vec3.dot(viewUp, jVector));
+    const absViewUpDotK = Math.abs(vec3.dot(viewUp, kVector));
 
     // Get Y spacing
     let ySpacing;
-    if (Math.abs(1 - absViewPlaneNormalDotI) < EPSILON) {
+    if (Math.abs(1 - absViewUpDotI) < EPSILON) {
       ySpacing = volumeSpacing[0];
       yDir = iVector;
-    } else if (Math.abs(1 - absViewPlaneNormalDotJ) < EPSILON) {
+    } else if (Math.abs(1 - absViewUpDotJ) < EPSILON) {
       ySpacing = volumeSpacing[1];
       yDir = jVector;
-    } else if (Math.abs(1 - absViewPlaneNormalDotK) < EPSILON) {
+    } else if (Math.abs(1 - absViewUpDotK) < EPSILON) {
       ySpacing = volumeSpacing[2];
       yDir = kVector;
     } else {


m-barthel:
@rodrigobasilio2022 Thanks for your hints. All checks are green now.

@andrebot @rkapitan @a-fouda FYI

sedghi:
when I try it it still lags, are you saying it is less now?

m-barthel:
> when I try it it still lags, are you saying it is less now?

Yes, exactly.  @sedghi, thanks for your reply.
On main, without this fix, it's hard to draw (almost) straight lines like this:

image  
   
With this fix, the behavior is the same as for drawing other directions.
By the way, I think this issue is only happening on VolumeViewports and not on StackViewports.

I'd like to extend the tests accordingly, but I couldn't find a test for the `palanrFreehandROITool`.

m-barthel:
> when I try it it still lags, are you saying it is less now?  
  
I have double-checked and included this fix to our application, and yes, it seems to fix our problem with the 'lagging' freehand tool.

sedghi (approved):
I think your changes makes sense. Thanks

m-barthel:
> I think your changes makes sense. Thanks

Cool. Thanks for reviewing and merging.

Issue #225, opened 2022-09-26
CT Image GPU render issue adds lot of white noise  


toropok:
Hi,

For our custom DICOM viewer we use following libs:

```
"@cornerstonejs/core": "^0.16.3",
"@kitware/vtk.js": "24.18.7",
"wslink": "^1.8.2",
"gl-matrix": "^3.4.3”,
"cornerstone-wado-image-loader": "^4.2.1"
```

Unexpetedly cornerstone adds a lot of white to images while rendering (except first image in series):

https://user-images.githubusercontent.com/14867014/192189820-f2121204-16e2-46d6-aac9-f1b9c6cae864.mov

But if you're setting CPU render only this "white" effect dissappears
```
cornerstone.setUseCPURendering(true)
```

https://user-images.githubusercontent.com/14867014/192190164-878fa145-f29c-4782-9235-47687f553efd.mov

What may cause that?

regards,
--leo

md-prog:
@toropok This might have been fixed by 0.16.8

toropok:
@md-prog Thanks! will try it asap

toropok:
> 



> @toropok This might have been fixed by 0.16.8

unfortunately it doesn't help

md-prog:
@toropok  Indeed, it's unfortunate. It resolved mine.

md-prog:
@toropok Is there any chance you can post the anonymized DICOM file for us to test and fix?

medicalconnections:
I am having the same issue and could provide an anonymized study that reproduces it. I think it has to do with the way the windowing values are used but unsure. I tried the CPU rendering and the images will render correctly (but breaks some other things so I would like to stick to GPU rendering).

The odd thing is that when I load only 1 canvas, I am able to scroll through the images and it gets rendered correctly.
![image](https://user-images.githubusercontent.com/11630793/234561847-08f98b60-1f6f-49c2-8e1c-c43ef897227b.png)

NOTE: The thumbnails are bitmaps so they are not cornerstone canvases.

 As soon as another canvas comes in play, it will become white like this:
![image](https://user-images.githubusercontent.com/11630793/234562173-9677ec7b-6569-458d-a880-628603a01bfb.png)

The cached images (when it was only 1 canvas active and I scrolled through few images)  are actually loading fine and only the newly rendered images become white:
[...]

PR #598, opened 2023-05-05
bugfix - GPU rendering - inheriting voiRange when recreating the default actor

md-prog:
`StackViewport.ts`
Method: `_updateActorToDisplayImageId()`

Analysis:

In this method, we check if the image has been changed to decide whether to reuse the existing actor or create new one.
So when a new actor is created, it was not setting the `voiRange` property correctly, particularly when it was applied from the previous rendering. Therefore the images were rendering as whitened.

Related issues:

Potentially related to #225


md-prog:
image

The screenshot shows the reason why the test truth image has been changed

sedghi (approved):

md-prog:
Thank you, @sedghi

PR #602, opened 2023-05-08
feat: add colormap preset and invert to volume viewport

sedghi:
Add the following through setProperties for volume viewports
- Colormap
- Preset
- Invert

Fix the NAC PT getting considered as scaled with suvbw


common/reviews/api/core.api.md
```diff
@@ -2751,6 +2788,9 @@ export class VolumeViewport3D extends BaseVolumeViewport {
 type VolumeViewportProperties = {
```
wayfarer3130, line 2788: Can you define a common class for shared properties between stack and volume?  Things like invert and colormap probably should be defined once rather than multiple times.

common/reviews/api/streaming-image-volume-loader.api.md
```diff
@@ -53,6 +53,19 @@ type CameraModifiedEventDetail = {
...
+type ColormapPublic = {
+    name: string;
+    opacityMapping?: number[][];
+};
+
+// @public (undocumented)
+type ColormapRegistration = {
+    ColorSpace: string;
+    Name: string;
+    RGBPoints: number[];
```
wayfarer3130, line 66: Can we use RGB type here?

packages/core/test/volumeViewport_gpu_setProperties_test.js
```diff
@@ -0,0 +1,133 @@
...
+  return element;
+}
+
+describe('Volume Viewport SetProperties -- ', () => {
+  beforeAll(() => {
+    window.devicePixelRatio = 1;
+    cornerstone3D.setUseCPURendering(false);
+  });
+
+  describe('should do something--- ', function () {
```
wayfarer3130, line 62: Could we get real descriptions here?  Like 'should register volume loader'

packages/core/test/volumeViewport_gpu_setProperties_test.js
```diff
@@ -0,0 +1,133 @@
...
+      metaData.removeProvider(fakeMetaDataProvider);
+      unregisterAllImageLoaders();
+      this.DOMElements.forEach((el) => {
+        if (el.parentNode) {
+          el.parentNode.removeChild(el);
+        }
+      });
+    });
+
+    it('should successfully do something ', function (done) {
```
wayfarer3130, line 86: 'should create volume instance'

wayfarer3130:
Can we have the colormap included in one of the test suites?  If it already is, just let us know which one so we can play with this item before final approval.

packages/core/src/types/Colormap.ts
```diff
@@ -0,0 +1,24 @@
+import RGB from './RGB';
+
+type ColormapRegistration = {
+  ColorSpace: string;
+  Name: string;
+  RGBPoints: RGB[];
```
jbocce, line 6: Do we still want this as an array?

sedghi: yeah it is an array

jbocce (approved):

jbocce (approved):

packages/core/src/types/Colormap.ts
```diff
@@ -0,0 +1,24 @@
...
+  /** midpoint mapping between values to opacity if the colormap
+   * is getting used for fusion, this is an array of arrays which
+   * each array containing 2 values, the first value is the value
+   * to map to opacity and the second value is the opacity value.
+   * By default, the minimum value is mapped to 0 opacity and the
+   * maximum value is mapped to 1 opacity, but you can configure
+   * the points in the middle to be mapped to different opacities
+   * instead of a linear mapping from 0 to 1.
+   */
+  opacityMapping?: number[][];
```
wayfarer3130, line 21: I like the array of objects rather than the double array as that is then self explanatory, and is more self documenting.

wayfarer3130 (commented):
I'm ok with this, but I prefer the array of objects.

packages/core/src/types/Colormap.ts
```diff
@@ -0,0 +1,24 @@
+import RGB from './RGB';
+
+type ColormapRegistration = {
+  ColorSpace: string;
```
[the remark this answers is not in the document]
sedghi: it is vtk.js based

PR #603, opened 2023-05-09
fix/multiframe wadorui

sedghi:
- Add an extra resetCamera for the renderToCanvas to try to fix the werid thumbnail on some systems
- Fix a long lasting multiframe problem where the frameIndex for metadata was getting used which starts at 1 instead of 0 causing first slice to be the second slice
- Try to adjust windowlevel multiplyer for the range for the non scaled PT


packages/dicomImageLoader/src/imageLoader/wadouri/loadImage.ts
```diff
@@ -23,6 +23,19 @@ function addDecache(imageLoadObject: Types.IImageLoadObject, imageId: string) {
   };
 }
 
+/**
+ *
```
jbocce, line 27: Perhaps add a one line description? "Loads an image from the given data set promise."

jbocce (approved):
Minor comments.

jbocce (approved):

PR #605, opened 2023-05-11
fix(stackViewport): better error handling for disabled viewports

sedghi:
- feat(stackViewport): better error handling for disabled viewports
- some fix in crosshair

## To Test
Run stackManipulationTools example locally

after setStack
re-enable the element 
and setStack again

- You have lost the scroll, but with these changes it will throw since that is what has happened

### DIFF
--- a/packages/core/src/RenderingEngine/StackViewport.ts
+++ b/packages/core/src/RenderingEngine/StackViewport.ts
@@ -1494,6 +1494,8 @@ class StackViewport extends Viewport implements IStackViewport {
     imageIds: Array<string>,
     currentImageIdIndex = 0
   ): Promise<string> {
+    this._throwIfDestroyed();
+
     this.imageIds = imageIds;
     this.currentImageIdIndex = currentImageIdIndex;
     this.targetImageIdIndex = currentImageIdIndex;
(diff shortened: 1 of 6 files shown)


packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -1527,6 +1529,19 @@ class StackViewport extends Viewport implements IStackViewport {
...
   }
 
+  /**
+   * Throws an error if you are using a destroyed instance of the stack viewport
+   */
+  private _throwIfDestroyed() {
+    if (this.isDisabled) {
+      throw new Error(
+        'The stack viewport has been destroyed and is no longer usable. Renderings will not be performed. If you ' +
+          'are using the same viewportId and has re-enabled the viewport, you need to grab the new viewport instance ' +
```
jbocce, line 1539: "...and has re-enabled..." should be "... and have re-enabled..."

packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -1527,6 +1529,19 @@ class StackViewport extends Viewport implements IStackViewport {
...
+          'using renderingEngine.getViewport(viewportId), instead of using your lexical reference to the viewport instance.'
```
jbocce, line 1540: What do you mean by lexcial here? Wouldn't 'previous' or 'existing' be better words to use?

sedghi: I mean lexical scope https://developer.mozilla.org/en-US/docs/Web/JavaScript/Closures
sure i will add those too

jbocce: Right. Sorry. Then say "lexical scoped reference"

PR #607, opened 2023-05-12
fix PlanarFreehandROITool bugs

m00n620:
Fixed these bugs on PlanarFreehandROITool

- Cached stats are not re-calculated on modification.
- ANNOTATION_COMPLETED events are triggered when moving textbox.


m00n620:
@lscoder , can you please review this ASAP?

PR #609, opened 2023-05-15
fix(Circle and VolumeViewport): fixes to ensure measurements are rendered properly

jbocce:
Applied dashing to drawCircle.
Refactored VolumeViewport.getCurrentImageId to use the utilities getClosestImageId.

### DIFF
--- a/packages/core/src/RenderingEngine/VolumeViewport.ts
+++ b/packages/core/src/RenderingEngine/VolumeViewport.ts
@@ -12,7 +12,7 @@ import type {
   Point3,
 } from '../types';
 import type { ViewportInput } from '../types/IViewport';
-import { actorIsA } from '../utilities';
+import { actorIsA, getClosestImageId } from '../utilities';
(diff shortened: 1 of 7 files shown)


packages/core/src/RenderingEngine/VolumeViewport.ts
```diff
@@ -344,7 +344,27 @@ class VolumeViewport extends BaseVolumeViewport {
    * @returns The slice index
    */
   public getCurrentImageIdIndex = (): number | undefined => {
-    return this._getImageIdIndex();
+    if (!this.getCurrentImageId()) {
```
sedghi, line 347: this method should not care if it is acquisition or not, this probably will break imagIndex on the non acquisition viewports.

jbocce: Ok will change.

sedghi (approved):

Issue #390, opened 2023-01-27
add getImage (cornerstone legacy) as a utility


sedghi:
In cornerstone legacy we had `getImage` which would return the `image` object from the enabled element. https://github.com/cornerstonejs/cornerstone/blob/master/src/getImage.js

Since our rendering pipeline has changed and we have removed that function, to accomodate migrations from legacy to cornerstone3D we should add a new utility function `utilities.getCornerstoneLegacyImageForStackViewport` that would return the [image object. ](https://www.cornerstonejs.org/docs/concepts/cornerstone-core/images)

jbocce:
As discussed with @sedghi, we will call it `utilities.getImageLegacy` instead

PR #613, opened 2023-05-17
fix(utilities): added getImageLegacy for migrations of cornerstone legacy's getImage

jbocce:
See issue #390

Testing:

Test A: 
1. Run the basic stack viewport example.
2. Open the browser's dev tools.
3. In the browser console get the viewport element for the image displayed and assign it to a variable - say `viewportElement`.
4. In the browser console issue the command `cornerstone.utilities.getImageLegacy(viewportElement)`
5. The corresponding `IImage` object should be returned.

Test B: 
1. Run the basic volume viewport example.
2. Open the browser's dev tools.
3. In the browser console get the viewport element for the image displayed and assign it to a variable - say `viewportElement`.
4. In the browser console issue the command `cornerstone.utilities.getImageLegacy(viewportElement)`
5. An exception to the following should be output...
```
getImageLegacy.ts:21 Uncaught Error: An image can only be fetched for a stack viewport and not for a viewport of type: orthographic
    at Module.getImageLegacy (getImageLegacy.ts:21:11)
    at <anonymous>:1:23
```

### DIFF
--- a/packages/core/src/RenderingEngine/StackViewport.ts
+++ b/packages/core/src/RenderingEngine/StackViewport.ts
@@ -509,6 +509,13 @@ class StackViewport extends Viewport implements IStackViewport {
   };
 
   /**
+   * Returns the raw/loaded image being shown inside the stack viewport.
+   */
+  public getImage = (): IImage => {
+    return this.csImage;
+  };
+
+  /**
    * Creates imageMapper based on the provided vtkImageData and also creates
    * the imageSliceActor and connects it to the imageMapper.
    * For color stack images, it sets the independent components to be false which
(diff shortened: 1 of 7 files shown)


packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -508,6 +508,13 @@ class StackViewport extends Viewport implements IStackViewport {
     return imagePlaneModule.frameOfReferenceUID;
   };
 
+  /**
+   * Returns the raw/loaded image being shown inside the stack viewport.
+   */
+  public getImage = (): IImage => {
```
sedghi, line 514: since we have getImageData, let's make this one `getCornerstoneImage`

sedghi (commented):
see my only comment please

sedghi (approved):

PR #617, opened 2023-05-18
WIP Fixing problem related to left+right click in ZoomTool

rodrigobasilio2022:
This fixes the problem when ZoomTool is activated by left+right buttons, as not always the two button combination fires preMouseDownCallback, leading to an error in zoom Center implementation


rodrigobasilio2022:
Superssed by PR 616

PR #634, opened 2023-05-29
DRAFT: fix(Cobb Angle): use the two closest line segment points as the tail of each respectful vector

jbocce:
### Context

Depending on the order that the points of the two line segments for the Cobb Angle were drawn, either the angle between 0 and 90 or the angle between 90 and 180 was output. If the exact same pairs of line are drawn, the angle displayed should be the same regardless of the order of the placement of the end points.

Furthermore since only one angle (and not the angle and its compliment) are shown it is best to show the angle consistent with how the tool is typically used.

### Changes & Results
The solution is based on the assumption  that the Cobb angle line segments are drawn
such that the segments intersect nearest the segment endpoints
that are closest to each other AND those closest endpoints are the tails of the
vectors used to calculate the angle between the vectors/line segments.

### Testing

Run the stackannotationtools and draw various Cobb angle measurements.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] OS: Windows 11
- [x] Node version: 19.8.1 
- [x] Browser: Chrome 113.0.5672.127
  


packages/tools/src/tools/annotation/CobbAngleTool.ts
```diff
@@ -48,6 +48,7 @@ import {
 } from '../../types';
 import { AngleAnnotation } from '../../types/ToolSpecificAnnotationTypes';
 import { StyleSpecifier } from '../../types/AnnotationStyle';
+import { vec3 } from 'gl-matrix';
```
sedghi, line 51: move to top please

sedghi:
I think this is wrong right?

[https-deploy-preview-634--cornerstone-3d-docs-netlify-app-live-examples-stackannotationtools.webm](https://github.com/cornerstonejs/cornerstone3D-beta/assets/7490180/9733c4eb-6767-4c66-941c-f41c4bc0b323)

jbocce:
> I think this is wrong right?
> 
>  https-deploy-preview-634--cornerstone-3d-docs-netlify-app-live-examples-stackannotationtools.webm

That looks correct to me.

packages/tools/src/tools/annotation/CobbAngleTool.ts
```diff
@@ -381,6 +382,52 @@ class CobbAngleTool extends AnnotationTool {
     this.isDrawing = false;
   };
 
+  /**
+   * Handles the mouse down for all points after the very first. The first is is handled by addAnnotation.
```
sedghi, line 386: some typos in the doc: 
- very first what?
- is is

packages/tools/src/tools/annotation/CobbAngleTool.ts
```diff
@@ -381,6 +382,52 @@ class CobbAngleTool extends AnnotationTool {
...
+   * This method ensures that the state of the tool is correct for the drawing of the second line segment.
+   * In particular it ensures that the second segment can be created via a mouse down and drag.
+   */
+  _mouseDownCallback = (
+    evt: EventTypes.MouseUpEventType | EventTypes.MouseClickEventType
+  ) => {
+    console.warn('mouse down');
```
sedghi, line 393: remove console warn

packages/tools/src/tools/annotation/CobbAngleTool.ts
```diff
@@ -532,6 +579,10 @@ class CobbAngleTool extends AnnotationTool {
       Events.MOUSE_CLICK,
       this._mouseUpCallback as EventListener
     );
+    element.addEventListener(
```
sedghi, line 582: it is very odd that we are adding this here, since no other tools has it, but at the same time we don't have any other two step tool. So I don't know

sedghi (commented):
see my comments

packages/tools/src/tools/annotation/CobbAngleTool.ts
jbocce: I understand. 

Without it, one cannot mouse down+drag to draw the second line segment which is ALSO odd.  If we were to remove it, the unit tests would require some work to fix. Up to you.

sedghi:
> > I think this is wrong right?
> > https-deploy-preview-634--cornerstone-3d-docs-netlify-app-live-examples-stackannotationtools.webm
> 
> That looks correct to me.

Why it is correct? these two should have the same numbers, the only difference is I rotate the second one 360 degree

image

jbocce:
> > > I think this is wrong right?
> > > https-deploy-preview-634--cornerstone-3d-docs-netlify-app-live-examples-stackannotationtools.webm
> > 
> > 
> > That looks correct to me.
> 
> Why it is correct? these two should have the same numbers, the only difference is I rotate the second one 360 degree
> 
> image

It is correct by the algorithm we are using. The two end points (circled in red below) are the closest end points and are treated as the tails of the vectors to find the angle between. I have also drawn the tip/head of each vector as well.

cobb angle

Typically, the tool is NOT used as you have drawn/used it above. Typically the lines are drawn to either side a the vertebrae tending to a point of intersection as below.

image


One possibility would be to always show the smaller angle, but for cases of severe scoliosis that would NOT be correct. I could see two other possibilities that are likely too time consuming to pursue at present:
1. depicting the actual point of intersection together with placing both angles correctly where they lie between the lines
2. requiring more lines to be drawn like the horizontal lines in the following method https://josr-online.biomedcentral.com/articles/10.1186/s13018-018-0928-5

sedghi:
I don't know much about this area, but those should be the same to me looking at them by eye. Maybe @wayfarer3130 can chime in

wayfarer3130:
> > I think this is wrong right?
> > https-deploy-preview-634--cornerstone-3d-docs-netlify-app-live-examples-stackannotationtools.webm
> 
> That looks correct to me.

That also looks correct to me - the closest endpoint is the one considered "interesting", so I think the angles look right.  Of course that means the end angle may change several times when both of them are just about the same distance.

sedghi (approved):

PR #637, opened 2023-06-01
feat(transfer function editor): various improvements and fixes

jbocce:
Fixed the position of the toolitp to the top-left and added transparency to its background.

Clamped the anchor point such that no portion of it goes beyond the canvas bounds.

Prevent deletion of control points when only two remain.

Control endpoint tooltips added. They are shown on anchor hover and drag.

Added a control point tooltip when hovering and movig a control point.

Update the TF_widget anchor whenever a control point gets deleted.

Updated histogram colours.

Moved context menus to document.body so that they are not clipped by the TF_panel.

Added a div (with css class name 'control-point-tooltip-parent') containing the control point tooltips for each separate TF_widget. This simplifies showing/hiding the tooltips on a per TF_widget basis.

### DIFF
--- a/packages/core/examples/volumeViewport3D/index.ts
+++ b/packages/core/examples/volumeViewport3D/index.ts
@@ -164,7 +164,8 @@ async function run() {
     panel: {
       isCollapsible: true,
       position: 'bottom',
-      background: '#090c29',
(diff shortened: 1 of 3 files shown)


sedghi:
looks good thanks

PR #638, opened 2023-06-02
feat(measurement): Add calibration type labels (ERMF, PROJ, USER)

wayfarer3130:
### Context

The user needs to know how the calibration of size on an image is derived, whether it is ERMF or USER projected or none for volumetric measurements.  Add the ability to specify a label, and instead of updating measurement positions, provide a scaling capability to scale just the measurements and not the separate positioning of everything.

### Changes & Results

Stores the calibration object inside the custom metadata provider
Uses the scale to scale all the measurements appropriately for display, but NOT changing the internal values.
Uses px/mm in the units decision.

### Testing

`yarn example stackAnnotationTools`

Run the various measurements and see how they change with different scaling applied.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  


packages/core/src/enums/CalibrationTypes.ts
```diff
@@ -0,0 +1,11 @@
+export enum CalibrationTypes {
```
sedghi, line 1: please add documentation on top of each enum

packages/tools/src/enums/index.js
```diff
@@ -13,4 +15,5 @@ export {
   Events,
   SegmentationRepresentations,
   Swipe,
+  CalibrationTypes,
```
sedghi, line 18: I don't understand why we are re import/exporting this here, since we already have it from core

packages/tools/src/utilities/roundMeasurement.ts
```diff
@@ -0,0 +1,13 @@
+/** Rounds the measurement value appropriately */
+function roundMeasurement(value, scaling = 1) {
```
sedghi, line 2: needs documentation on the logic, especially on the scaling

packages/core/src/types/IImageCalibration.ts
```diff
@@ -0,0 +1,15 @@
+import CalibrationTypes from '../enums/CalibrationTypes';
+
+/**
+ * IImageCalibration is an object that stores information about the type
+ * of image calibration.
+ */
+export interface IImageCalibration {
```
sedghi, line 7: it is very odd that we have small casing and upper casing in this, maybe we should be consistent or if group the ones that need to be upper case in some other property

sedghi: please add documentation to each entry/property

packages/tools/src/utilities/lengthUnits.ts
```diff
@@ -0,0 +1,19 @@
+const lengthUnits = (handles, image): string => {
```
sedghi, line 1: since these two functions involve looking at the calibration, I prefer other namings suh as `calibratedLenghUnit` (btw why units?), or `getCalibratedLengthUnit` same for area

packages/tools/src/utilities/roundMeasurement.ts
```diff
@@ -0,0 +1,13 @@
...
+  if (value === undefined || value === null || value === '') return 'NaN';
+  const scaleValue = value * scaling;
+  if (scaleValue >= 100) return Math.round(scaleValue);
+  if (scaleValue >= 10) return scaleValue.toFixed(1);
+  if (scaleValue >= 0.1) return scaleValue.toFixed(2);
+  if (scaleValue >= 0.001) return scaleValue.toFixed(4);
+  if (scaleValue >= 0.00001) return scaleValue.toFixed(6);
+  return scaleValue;
+}
+
```
sedghi, line 12: How about something like 

```js
const precision = 
    scaleValue >= 100 ? 0 : 
    scaleValue >= 10 ? 1 : 
    scaleValue >= 0.1 ? 2 : 
    scaleValue >= 0.001 ? 4 : 
    scaleValue >= 0.00001 ? 6 : 
    undefined;

precision !== undefined ? scaleValue.toFixed(precision) : scaleValue;
```

sedghi: also the naming doesn't need to be `measurement`, it is kind of `roundValue`

packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -74,6 +74,7 @@ import {
   ImagePixelModule,
   ImagePlaneModule,
 } from '../types';
+import { CalibrationTypes } from '../enums';
```
sedghi, line 77: can you merge all the above enums imports here too and remove them

sedghi (commented):
Looks good, see my comments please

sedghi:
seems like you need to do the api build again

jbocce (commented):
Just a few things thus far. Still need to look at more, but as you indicated to me in a side conversation, you have some touch-ups for it. So I will wait.

packages/tools/src/utilities/calibratedLengthUnits.ts
```diff
@@ -0,0 +1,37 @@
...
+ *   between various points (eg angled ERMF or US Region)
+ * @param image - to extract the calibration from
+ * @param image.calibration - calibration value to extract units form
+ * @returns String containing the units and type of calibration
+ */
+const calibratedLengthUnits = (handles, image): string => {
+  const { calibration, hasPixelSpacing } = image;
+  const units = hasPixelSpacing ? 'mm' : 'px';
+  if (!calibration || !calibration.type) return units;
+  if (calibration.SequenceOfUltrasoundRegions) return 'US Region';
```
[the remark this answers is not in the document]
wayfarer3130: Yes, it does, but right now I don't have code for that, so just show a warning in that case.

jbocce (commented):
So far it looks fine.

Just a few comments, suggestions and questions I would like the answers to before I do more testing.

packages/core/src/types/IImageCalibration.ts
```diff
@@ -0,0 +1,32 @@
...
+   */
+  type: CalibrationTypes;
+
+  // A tooltip which can be used to explain the calibration information
+  tooltip?: string;
+  /** Indication that the image has some spacing information (the pixelSpacing
+   * when hasPixelSpacing is null can just be 1,1) */
+  hasPixelSpacing?: boolean;
+  /** Indication of projection (eg X-Ray type) spacing and volumetric type */
+  isProjection?: boolean;
```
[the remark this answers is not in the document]
wayfarer3130: The information that getPixelSpacingInformation returns is intended to eventually be changed to use the same type, but
right now the only thing that is used in that return is actually the spacing itself.

packages/core/src/types/IImageCalibration.ts
```diff
@@ -0,0 +1,32 @@
...
+  /** The type of the pixel spacing, distinguishing between various
+   * types projection (CR/DX/MG) spacing and volumetric spacing ('')
+   */
+  type: CalibrationTypes;
+
+  // A tooltip which can be used to explain the calibration information
+  tooltip?: string;
+  /** Indication that the image has some spacing information (the pixelSpacing
+   * when hasPixelSpacing is null can just be 1,1) */
+  hasPixelSpacing?: boolean;
```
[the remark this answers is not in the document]
wayfarer3130: Yes, but they aren't necessarily the same value.  Probably having a consistent type makes sense but that is more scope and this is already large enough.

packages/core/src/types/EventTypes.ts
```diff
@@ -223,8 +223,9 @@ type ImageSpacingCalibratedEventDetail = {
   viewportId: string;
   renderingEngineId: string;
   imageId: string;
-  rowScale: number;
-  columnScale: number;
+  rowScale?: number;
```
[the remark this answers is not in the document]
wayfarer3130: Scale is a single scaling value.  rowScale and columnScale are used to apply a non linear calibration to the image - that isn't supported by the current code, and was broken in the previous version of the code, but is something I'd like to see eventually.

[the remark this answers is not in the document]
wayfarer3130: Removed the old values.

packages/core/src/types/CPUIImageData.ts
```diff
@@ -24,6 +25,8 @@ type CPUIImageData = {
   scaling: Scaling;
   /** whether the image has pixel spacing and it is not undefined */
   hasPixelSpacing?: boolean;
```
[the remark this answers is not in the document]
wayfarer3130: Yes for the has pixel spacing, but not for the scaling parameter, which I think is a display parameter.

packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -2709,6 +2648,8 @@ class StackViewport extends Viewport implements IStackViewport {
       imageId
```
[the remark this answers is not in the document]
wayfarer3130: Deleted the obsolete code.

packages/tools/src/utilities/calibratedUnits.ts
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * TODO: Handle region calibration
+ *
+ * @param handles - used to detect if the spacing information is different
+ *   between various points (eg angled ERMF or US Region)
+ * @param image - to extract the calibration from
+ * @param image.calibration - calibration value to extract units form
+ * @returns String containing the units and type of calibration
+ */
+const calibratedLengthUnits = (handles, image): string => {
```
sedghi, line 16: `calibratredLengthUnits` does not give me impression of this is a function rather it is a variable. How about `getCalibratedLengthUnits` and same for area.

packages/core/src/types/IImageCalibration.ts
```diff
@@ -0,0 +1,38 @@
...
+  /**
+   * Boolean to indicate that the image has spacing information (the pixelSpacing
+   * can just be 1,1 to use image pixels, but this
+   * isn't distinguishable from a real spacing of 1mm square pixels so a real
+   * flag here is needed.)
+   */
+  hasPixelSpacing?: boolean;
+  /** Indication of projection (eg X-Ray type) spacing and volumetric type */
+  isProjection?: boolean;
+  // The DICOM defined ultrasound regions.  Used for non-distance spacing
```
sedghi, line 33: please be consistent in using /** or // , also please remove the empty lines

packages/core/src/types/IImageCalibration.ts
```diff
@@ -0,0 +1,38 @@
...
+  // A tooltip which can be used to explain the calibration information
+  tooltip?: string;
+  /**
+   * Boolean to indicate that the image has spacing information (the pixelSpacing
+   * can just be 1,1 to use image pixels, but this
+   * isn't distinguishable from a real spacing of 1mm square pixels so a real
+   * flag here is needed.)
+   */
+  hasPixelSpacing?: boolean;
+  /** Indication of projection (eg X-Ray type) spacing and volumetric type */
```
sedghi, line 31: can't we remove this and derive it from row/column pixelSpacing? having row/column and hasPixelSpacing sounds redundant to me, no?

packages/core/src/types/IImageCalibration.ts
```diff
@@ -0,0 +1,38 @@
+import CalibrationTypes from '../enums/CalibrationTypes';
+
+/**
+ * IImageCalibration is an object that stores information about the type
+ * of image calibration.
+ */
+export interface IImageCalibration {
+  /** The pixel spacing for the image, in mm between pixel centers */
```
sedghi, line 8: so I'm wondering whether there is a structure to this? If I understand this there are various calibration types, ERF, USER, PROJ, US. So is it true that for each of these we can/should have `row/columnPixelSpacing` and `scale`? or is it like `scale` is only valid when there is `PROJ`? and doesn't makes sense when it is `USER`.

packages/core/src/types/EventTypes.ts
```diff
@@ -223,8 +223,7 @@ type ImageSpacingCalibratedEventDetail = {
   viewportId: string;
   renderingEngineId: string;
   imageId: string;
-  rowScale: number;
-  columnScale: number;
+  scale?: number;
```
sedghi, line 226: why are we getting rid of scale? is it the case that we can't have different scale in rows vs columns?

packages/core/src/utilities/imageToWorldCoords.ts
```diff
@@ -23,12 +23,15 @@ export default function imageToWorldCoords(
...
     rowCosines,
-    rowPixelSpacing,
     imagePositionPatient: origin,
   } = imagePlaneModule;
 
+  let { columnPixelSpacing, rowPixelSpacing } = imagePlaneModule;
+  // Use ||= to convert null and 0 as well as undefined to 1
+  columnPixelSpacing ||= 1;
+  rowPixelSpacing ||= 1;
+
```
sedghi, line 34: it would be good to have a console warn here too

packages/core/src/utilities/worldToImageCoords.ts
```diff
@@ -27,14 +27,15 @@ function worldToImageCoords(
...
     rowCosines,
-    rowPixelSpacing,
     imagePositionPatient: origin,
-    rows,
-    columns,
   } = imagePlaneModule;
 
+  let { columnPixelSpacing, rowPixelSpacing } = imagePlaneModule;
+  // Use ||= to convert null and 0 as well as undefined to 1
+  columnPixelSpacing ||= 1;
```
sedghi, line 36: console warn here too

packages/core/test/stackViewport_gpu_render_test.js
```diff
@@ -778,7 +778,10 @@ describe('renderingCore -- Stack', () => {
       });
     });
 
-    it('Should be able to calibrate an image', function (done) {
+    const skipIt = () => null;
+    // TODO - renable this when affine transforms are supported as part of
+    // the calibration event instead of simple calibration ratios
+    skipIt('Should be able to calibrate an image', function (done) {
```
sedghi, line 784: why skipit

packages/tools/examples/stackAnnotationTools/index.ts
```diff
@@ -185,6 +191,83 @@ addButtonToToolbar({
...
+    const calibration = calibrations.find(
+      (it) => it.value === newCalibrationValue
+    );
+    if (!calibration) return;
+    const f = calibrationFunctions[calibration.selected];
+    if (!f) return;
+    f.apply(calibration);
+  },
+});
+
```
sedghi, line 270: Thanks a lot!

packages/tools/examples/stackAnnotationTools/index.ts
```diff
@@ -185,6 +191,83 @@ addButtonToToolbar({
...
 /**
```
sedghi, line 271: I prefer a dedicated example to make it more searchable. Can you literally copy paste and remove pieces that you don't want?

packages/tools/src/tools/annotation/CircleROITool.ts
```diff
@@ -996,7 +1002,11 @@ class CircleROITool extends AnnotationTool {
           worldPos2
         );
         const isEmptyArea = worldWidth === 0 && worldHeight === 0;
-        const area = Math.abs(Math.PI * (worldWidth / 2) * (worldHeight / 2));
+        const scale = getScale(image);
+        const area =
+          Math.abs(Math.PI * (worldWidth / 2) * (worldHeight / 2)) /
+          scale /
```
sedghi, line 1008: instead can we have `worldWidth/scale/2` and `worldHeight/scale/2` for readability?

packages/tools/src/utilities/calibrateImageSpacing.ts
```diff
@@ -15,19 +15,17 @@ const { calibratedPixelSpacingMetadataProvider } = utilities;
...
 ): void {
-  // 1. Add the calibratedPixelSpacing metadata to the metadata provider
-  // If no column spacing provided, assume square pixels
-  if (!columnPixelSpacing) {
-    columnPixelSpacing = rowPixelSpacing;
+  // Handle simple parameter version
+  if (typeof spacing === 'number') {
+    spacing = {
+      type: Enums.CalibrationTypes.USER,
+      scale: spacing,
```
sedghi, line 24: `scale : spacing` does not makes sense to me at all. Why we are saying scale is equal to spacing? Shouldn't it be the division?

package.json
```diff
@@ -159,5 +159,8 @@
     "not ie < 11",
     "not op_mini all"
   ],
-  "dependencies": {}
+  "dependencies": {},
+  "resolutions": {
+    "canvas": "2.11.2"
```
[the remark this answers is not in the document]
wayfarer3130: It is a sub-package that uses an old canvas, and that doesn't work any longer with new node versions on windows.  So, resolutions is really the only way to go here.

packages/tools/src/utilities/calibratedUnits.ts
```diff
@@ -0,0 +1,41 @@
...
+ */
+const calibratedAreaUnits = (handles, image): string => {
+  const { calibration, hasPixelSpacing } = image;
+  const units = (hasPixelSpacing ? 'mm' : 'px') + SQUARE;
+  if (!calibration || !calibration.type) return units;
+  if (calibration.SequenceOfUltrasoundRegions) return 'US Region';
+  return `${units} ${calibration.type}`;
+};
+
+const getScale = (image) => image.calibration?.scale || 1;
```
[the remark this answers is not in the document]
jbocce: Or perhaps `getCalibratedScale`

packages/core/src/types/IImageCalibration.ts
wayfarer3130: I'm going to try to use /** consistently as that is processed by the tsdoc parser.  In other files, I will update the newly added items but not things in other classes/types.

packages/core/src/types/IImageCalibration.ts
wayfarer3130: No actually, you can't reliably get it from row/column spacing because some images are shown at different aspect ratios and we mix pixel spacing, aspect ratio display and volumetric display.  There are unfortunately more things to fix before this becomes cleaner.

packages/core/src/types/IImageCalibration.ts
wayfarer3130: Scale is used for user projection as well as projected images.  The use of scale actually fixes the storing/loading of user calibrated data, and so is quite critical to the entire fix.  What I'd really like to have is:
1. Aspect ratio - the displayed aspect ratio, irrespective of pixel spacing
2. Imager pixel spacing - the raw imaging spacing
3. Scale - to account for user calibration changes as well as projected units
4. Type - to actually indicate the type, with one of them being "Uncalibrated" or something like that to indicate px units
5. Calibration function, to account for weird cases like US calibration (not fully supported initially)

That would then get rid of hasPixelSpacing entirely, and we would pass around a calibration object instead of pixel spacing etc.  The pixel spacing values would always be defined but would simply be in whatever units were specified by the type, which might be px.

packages/core/src/types/EventTypes.ts
wayfarer3130: Because the rescale for measurements is fundamentally broken in ways that corrupt the underlying data depending on exactly what is and isn't applied at any given time.  I don't see any way to fix that in a way that works.
IF we want to have aspect ratio measurement changes, then we need to apply it to each measurement type calculations.

packages/core/src/utilities/imageToWorldCoords.ts
wayfarer3130: Not sure why you want that - it is a perfectly valid case, just one that we weren't handling correctly previously.

packages/core/test/stackViewport_gpu_render_test.js
wayfarer3130: Because they didn't implement it.skip(....) and I wanted the code to be compiled but not executed right now.  It is a test that we probably want in the future.

packages/tools/src/utilities/calibrateImageSpacing.ts
wayfarer3130: It doesn't work with the old values anyways, so I'm removing that code.

package.json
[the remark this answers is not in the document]
sedghi: what is the package? which one?

packages/core/test/stackViewport_gpu_render_test.js
sedghi: can't you comment it out then?

packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -590,89 +599,20 @@ class StackViewport extends Viewport implements IStackViewport {
...
-    ) {
-      // No calibration is required
-      return imagePlaneModule;
-    }
-
-    // Calibration is required
+    this.calibration = calibration;
     this._publishCalibratedEvent = true;
-
+    console.log('Calibration event is', scale, calibration);
```
sedghi, line 612: remove console log please

packages/core/src/types/IImageCalibration.ts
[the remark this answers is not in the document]
wayfarer3130: Removed hasPixelSpacing from the calibration class, making it into a type instead.

package.json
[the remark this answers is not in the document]
wayfarer3130: resemblejs uses it, updating that now.

packages/core/src/types/IImageCalibration.ts
wayfarer3130: Removed hasPixelSpacing in favour of using a custom type

packages/core/test/stackViewport_gpu_render_test.js
wayfarer3130: I wanted it to continue compiling, and if you comment it out, then it stops being updated/fixed when the API changes.

sedghi (commented):
I'm trying to understand the `calibrateImageSpacing` function. 

```js
export default function calibrateImageSpacing(
  imageId: string,
  renderingEngine: Types.IRenderingEngine,
  calibrationOrScale: Types.IImageCalibration | number
): void {
```

it has calibrationOrScale, while the IImageCalibration itself is below

```js
export interface IImageCalibration {
  /** The pixel spacing for the image, in mm between pixel centers */
  rowPixelSpacing?: number;
  columnPixelSpacing?: number;
  /** The scaling of this image - new spacing = original pixelSpacing/scale */
  scale?: number;
  /** The type of the pixel spacing, distinguishing between various
   * types projection (CR/DX/MG) spacing and volumetric spacing (the type is
   * an empty string as it doesn't get a suffix, but this distinguishes it
   * from other types)
   */
  type: CalibrationTypes;
  /** A tooltip which can be used to explain the calibration information */
  tooltip?: string;
  /** The DICOM defined ultrasound regions.  Used for non-distance spacing units. */
  sequenceOfUltrasoundRegions?: Record<string, unknown>[];
}
```

and it has scale too. It is very confusing to me

packages/tools/src/utilities/calibrateImageSpacing.ts
```diff
@@ -9,25 +9,22 @@ const { calibratedPixelSpacingMetadataProvider } = utilities;
...
- * @param columnPixelSpacing - Spacing in column direction
- * @param renderingEngine - Cornerstone RenderingEngine instance
+ * @param calibrationOrScale - either the calibration object or a scale value
  */
 export default function calibrateImageSpacing(
   imageId: string,
   renderingEngine: Types.IRenderingEngine,
-  rowPixelSpacing: number,
-  columnPixelSpacing: number
+  calibrationOrScale: Types.IImageCalibration | number
```
wayfarer3130, line 17: The intent here is that it is a simple scale value, to agree (almost) with the old API, OR it is a full calibration object.  If it is still confusing, could you suggest a naming for it that explains that it is either a calibration object entirely or a scale value used to construct a calibration object?

sedghi: I mean you are changing the API from row/column pixelSpacing to a single one, so it is changing the API right now. I don't know if it really makes sense to change the API + make it backward compatible

packages/core/src/types/EventTypes.ts
```diff
@@ -225,8 +226,8 @@ type ImageSpacingCalibratedEventDetail = {
...
+  calibration?: IImageCalibration;
```
sedghi, line 230: does `scale` here live because of legacy implementation? since there is scale inside calibration too.

sedghi (commented):
looks good, can you respond to my remaining comment only?

packages/core/src/types/EventTypes.ts
wayfarer3130: I'm removing scale, and adding some more documentation.
Also adding an aspect ratio, but the code currently does not handle it effectively.  I've figured out how to fix that, but I'm not quite sure yet where the changes would be.  The correct way to handle aspect ratio is to:
1. Render the image in the specified aspect ratio
2. Add an SVG transform to apply the aspect ratio as well as pan/zoom settings

wayfarer3130:
> looks good, can you respond to my remaining comment only?

I pushed an update to deal with that.

packages/core/src/types/IImageCalibration.ts
```diff
@@ -5,11 +5,23 @@ import CalibrationTypes from '../enums/CalibrationTypes';
...
   columnPixelSpacing?: number;
   /** The scaling of this image - new spacing = original pixelSpacing/scale */
   scale?: number;
+  /**
+   * The aspect ratio of the screen.
+   * Defaults to 1 if not specified.
+   * Not well handled currently as changing the aspect ratio does not result in
+   * updating measurements in any meaningful way.
+   */
+  aspect?: number;
```
sedghi, line 24: I suggest either aspectRatio or pixelAspectRatio as I searched inside dicom standard

wayfarer3130: I added documentation to the aspectRatio indicating it is neither of the DICOM standard ones - it is the measurement aspect ratio as compared to display screen ratio.

sedghi:
can you run `yarn run build:update-api`? and include them? https://www.cornerstonejs.org/docs/contribute/update-api

sedghi (approved):

PR #641, opened 2023-06-05
fix(build): Allow debugging into the tools package

wayfarer3130:
Fix the tools build package to have a local tsconfig setup, and define the package names/setup in a way that should allow debugging both pre-built and linked packages

### DIFF
--- a/packages/tools/package.json
+++ b/packages/tools/package.json
@@ -4,7 +4,7 @@
   "description": "Cornerstone3D Tools",
   "main": "dist/umd/index.js",
   "types": "dist/esm/index.d.ts",
-  "module": "dist/esm/index.js",
+  "module": "src/index.ts",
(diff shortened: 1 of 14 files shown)


packages/tools/package.json
```diff
@@ -4,7 +4,7 @@
   "description": "Cornerstone3D Tools",
   "main": "dist/umd/index.js",
   "types": "dist/esm/index.d.ts",
-  "module": "dist/esm/index.js",
+  "module": "src/index.ts",
```
sedghi, line 7: this should not be src, there should not be any src in the package json for publish

wayfarer3130: It works fine that way as a minimal change, now that the tsconfig is published, AND the source is published - otherwise, I don't see how to debug in both linked and unlinked versions - if you have another suggestion to enable that, please let me know.

PR #642, opened 2023-06-05
fix(measurements): The image stack sync tool fails to work on non-FOR instances and hangs the browser

wayfarer3130:
### Context

Synchronizers hang the browser if they do asynchronous actions.

The imageStackSync uses two algorithms, of which only one actually works correctly.  
Changed to use a single algorithm, just different starting conditions depending on options.

### Changes & Results

Made the synchronizers thread safe

Made the imageStackSync work with non same FOR images, and same FOR using current positions when configured to do so.

### Testing
Display 4 up (2x2 mode) with the same FOR series in all frames, turn on image stack synchronizer - depending on the series selected, it may hang the browser.

Display 2+up mode with different FOR but coplanar series.  Activate image stack sync (without doing anything special - just activate it, eg in the hanging protocol).  Ensure that image stacks navigate together for coplanar series in stack mode.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  


packages/tools/src/store/SynchronizerManager/Synchronizer.ts
```diff
@@ -211,13 +212,20 @@ class Synchronizer {
...
-        this._eventHandler(this, sourceViewport, targetViewport, sourceEvent);
+        promises.push(
+          this._eventHandler(this, sourceViewport, targetViewport, sourceEvent)
+        );
       }
     } catch (ex) {
       console.warn(`Synchronizer, for: ${this._eventName}`, ex);
     } finally {
-      this._ignoreFiredEvents = false;
+      if (promises.length) {
```
mateusfreira, line 222: @wayfarer3130  and I reviewed this on a call; this is causing issues in the TVM mode. This is also an important fix that was making the browser hang so we want to keep it.
Mentioning here just so we know we need to fix it before merging

packages/core/src/utilities/calculateViewportsSpatialRegistration.ts
```diff
@@ -1,9 +1,11 @@
 import { vec3, mat4 } from 'gl-matrix';
 import { IStackViewport } from '../types';
-import { StackViewport } from '../RenderingEngine';
 import spatialRegistrationMetadataProvider from './spatialRegistrationMetadataProvider';
 import { metaData } from '..';
-import isEqual from './isEqual';
+
+// Allow a fairly broad delta for angle differences, as it is common to have
+// some variance here, but the scrolling still works
```
wayfarer3130, line 7: Add comment here

packages/tools/src/store/SynchronizerManager/Synchronizer.ts
wayfarer3130: Use the viewport status instead of making this async

wayfarer3130: Seems to be ok in the TVM mode now, at least I can't reproduce any hangs.  I believe the issue was caused by multiple initial renders causing a loop, and the existing OHIF changes to prevent extra renders resolved the base issue.

packages/tools/src/store/SynchronizerManager/Synchronizer.ts
```diff
@@ -211,13 +212,20 @@ class Synchronizer {
         if (targetIsSource) {
           continue;
         }
-
-        this._eventHandler(this, sourceViewport, targetViewport, sourceEvent);
+        promises.push(
```
wayfarer3130, line 215: I don't know how to avoid this synchronization - the basic synchronizer design relies on being able to turn off secondary notifications so that only a single change applies, and if this doesn't use async, then you get an event from the target viewport, which changs the source viewport, which sends an event etc.

packages/tools/src/store/SynchronizerManager/Synchronizer.ts
wayfarer3130, line 1: The changes in this file fix the browser hanging issue.

packages/tools/src/synchronizers/callbacks/stackImageSyncCallback.ts
```diff
@@ -70,69 +64,69 @@ export default async function stackImageSyncCallback(
...
-
-      return;
+      registrationMatrixMat4 = mat4.identity(mat4.create());
+    } else {
+      utilities.calculateViewportsSpatialRegistration(sViewport, tViewport);
+      registrationMatrixMat4 =
+        utilities.spatialRegistrationMetadataProvider.get(
+          'spatialRegistrationModule',
+          [targetViewport.viewportId, sourceViewport.viewportId]
+        );
```
sedghi, line 90: can you refactor this since it is used above too, in the same file i mean

packages/tools/src/synchronizers/callbacks/stackImageSyncCallback.ts
```diff
@@ -70,69 +64,69 @@ export default async function stackImageSyncCallback(
...
     if (!registrationMatrixMat4) {
-      throw new Error(
-        `No registration matrix found for sourceViewport: ${sourceViewport.viewportId} and targetViewport: ${targetViewport.viewportId}, viewports with different frameOfReferenceUIDs must have a registration matrix in the registrationMetadataProvider. Use calculateViewportsRegistrationMatrix to calculate the matrix.`
+      console.log(
+        "Viewports still don't have registration:",
+        sourceViewport.viewportId,
+        targetViewport.viewportId
       );
+      return;
     }
```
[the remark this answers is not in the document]
wayfarer3130: No, there are legitimate times this won't return a value, and you can't set it up to simply deny the display in the viewprot - for example, simple CR images won't have that, and may still want to default the viewport to have the stack sync on for when the HP is used for another image set.

sedghi:
can you do a `git pull origin main --no-rebase` ? it is showing 67 commits

packages/tools/src/store/SynchronizerManager/Synchronizer.ts
sedghi: @mateusfreira can you try again, It seems to work with tmtv now

wayfarer3130: @mateusfreira  - you have to try it with the merge update branch of FlexView, as it was changes outside of CS3D that fixed the hang issue.

sedghi: @mateusfreira can you try and let me know?

mateusfreira (approved):
LGTM

sedghi (approved):

PR #651, opened 2023-06-12
link textbox to vertex unless moved by user

IbrahimCSAE:
### Context

In the angle tool the label for degrees is not linked to the vertex of the angle... this can be confusing as the line that links to the label is similar to the lines of the angle and it can appear that you've drawn some other shape rather than an angle.

image

### Changes & Results

Now by default, the textbox line is linked to the vertex, and it will stay that way unless the text box is specifically moved by the user so that they still have that option if they want it.

Before:-

before

After:-

Screenshot_2

### Testing

- Go to the examples directory in the repo
- Use the double click with stack annotations tool example to test the angle tool

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [X] "OS: Windows 10
- [X] "Node version: 18.15.0
- [X] "Browser: Chrome

### DIFF
--- a/packages/tools/src/tools/annotation/AngleTool.ts
+++ b/packages/tools/src/tools/annotation/AngleTool.ts
@@ -738,7 +738,8 @@ class AngleTool extends AnnotationTool {
       const textLines = this._getTextLines(data, targetId);
 
       if (!data.handles.textBox.hasMoved) {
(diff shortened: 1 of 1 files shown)


wayfarer3130 (approved):
@sedghi  - I think this is cleaner/more obvious, but it is a UI change so I wanted to let you see the update and approve/merge it as you will.

sedghi (approved):

PR #670, opened 2023-06-23
[PROTOTYPE] Segmentations in StackViewport

rodrigobasilio2022:
This PR aims to add in Cornerstone the display of segmentations in StackViewports.
Viewports in Cornerstone 3D are basically vtk.js scenes and vtk accepts multiple actors in a scene. 

Cornerstone 3D uses basically two types of actors: vtkImageSlice in StackViewports to display one image slice and vtkVolume to display a volume information in a VolumeViewport or Volumeviewport3D. The methods implemented in both viewports deal with the specifics of each data they are trying to display and they end up preventing other data to be displayed in the same viewport.

In very brief words, The segmentation display in Cornerstone 3D is constructed by placing two vtkVolume actors in the same scene with the same position information and leaving vtk,js handling the display. Exactly the same setup could be used with one vtkImageslice actor and a vtkVolume actor.

The changes I made aims to remove the limitations in a viewport implementation to display different data. Basically it adds to Viewport the ability to handle multiple actors and removes the requirement in some functions in the SegmentationDisplay to only accept VolumeViewports. 

One good benefit is optimize memory usage, because when a segmentation is added to a viewport, all viewports that share the same FrameOfReference will be converted to a VolumeViewport, and the data related will be converted to an interpolated  volume structure, that uses unecessary extra memory. This could be problematic if you have in study series with 30, 40 images derived from a series of 300 images.


sedghi:
closing in favor of #894

PR #679, opened 2023-07-05
[WIP] Add Segmentations to large images (> 2048 in any dimension)

rodrigobasilio2022:
Implements the ability to create segmentations for large images (> 20348 in any dimension)


packages/tools/src/types/LabelmapTypes.ts
```diff
@@ -35,7 +35,18 @@ export type LabelmapRenderingConfig = {
...
 };
 
-export type LabelmapSegmentationData = {
+export type LabelmapSegmentationDataVolume = {
+  type: 'volume';
   volumeId: string;
   referencedVolumeId?: string;
 };
+
+export type LabelmapSegmentationDataImage = {
```
sedghi, line 44: I prefer `LabelmapSegmentationDataStack`

packages/tools/src/types/LabelmapTypes.ts
```diff
@@ -35,7 +35,18 @@ export type LabelmapRenderingConfig = {
   ofun?: vtkPiecewiseFunction;
 };
 
-export type LabelmapSegmentationData = {
+export type LabelmapSegmentationDataVolume = {
+  type: 'volume';
```
sedghi, line 39: I think we can check if there exists `volumeIds` or `imageIds` then infer the type instead of having explicit type here, no?

packages/tools/src/tools/segmentation/SphereScissorsTool.ts
```diff
@@ -118,7 +118,9 @@ class SphereScissorsTool extends BaseTool {
     const { representationData } = getSegmentation(segmentationId);
 
     // Todo: are we going to support contour editing with rectangle scissors?
-    const { volumeId } = representationData[type] as LabelmapSegmentationData;
+    const { volumeId } = representationData[
+      type
+    ] as LabelmapSegmentationDataVolume;
```
sedghi, line 123: I guess this requires a Todo that says `handle stack later`?

packages/core/src/loaders/imageLoader.ts
```diff
@@ -183,6 +197,121 @@ export function loadAndCacheImages(
...
+export function createAndCacheDerivedImage(
+  imageId: string,
+  options: DerivedImageOptions
+): Promise<IImage> {
+  if (imageId === undefined) {
+    throw new Error(
+      'createAndCacheDerivedImage: parameter imageId must not be undefined'
+    );
+  }
+  const imagePlaneModule = metaData.get('imagePlaneModule', imageId);
```
sedghi, line 219: you have now helpers for these extracted out right? `getImagePlaneModule` and ...

packages/tools/examples/StackSegmentationMammography/index.ts
```diff
@@ -0,0 +1,347 @@
+import { vec3 } from 'gl-matrix';
```
sedghi, line 1: can you make this one also use the new API?

packages/tools/examples/StackSegmentation/index.ts
```diff
@@ -0,0 +1,385 @@
...
+
+    if (!colormapObj) {
+      throw new Error(`Colormap ${name} not found`);
+    }
+
+    cfun.applyColorMap(colormapObj);
+    cfun.setMappingRange(0, 255);
+    properties.setRGBTransferFunction(0, cfun);
+  }
+}
```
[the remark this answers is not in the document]
sedghi: We should not have this code here, the labelmap rendering pipeline should apply the colorLUTIndex 0 by default, inside the `addSegmentationRepresentation`

packages/tools/examples/StackSegmentation/index.ts
```diff
@@ -0,0 +1,385 @@
...
+        background: <Types.Point3>[0, 0, 0],
+      },
+    },
+  ];
+  renderingEngine.setViewports(viewportInputArray);
+  toolGroup.addViewport(viewportId, renderingEngineId);
+  const viewport = renderingEngine.getViewport(viewportId);
+  const sortedImageIds = sortImageIds(imageIds);
+  const { imageIds: derivedImageIds } =
+    await imageLoader.createCacheDerivedImages(sortedImageIds);
```
sedghi, line 360: createAndCache?

packages/tools/examples/StackSegmentation/index.ts
```diff
@@ -0,0 +1,385 @@
...
+ * Adds two concentric circles to each axial slice of the demo segmentation.
+ */
+function createMockEllipsoidSegmentation(dimensions, imageIds) {
+  const center = [dimensions[0] / 2, dimensions[1] / 2, dimensions[2] / 2];
+  const outerRadius = 64;
+  const innerRadius = 32;
+
+  for (let z = 0; z < dimensions[2]; z++) {
+    let voxelIndex = 0;
+    const { image } = cache.getCachedImageBasedOnImageURI(imageIds[z]);
```
sedghi, line 153: You know how we have a `cache.getVolume`, we should have a `cache.getImage` too instead of this `getCachedImage.....`

sedghi (changes requested):
Good first draft, major comments:

1. the `render` function inside the `labelmapDisplay` should have logic to check if the requested segImageId has an actor in the viewport, if not creates it (similar to how it does it for the volume via `addLabelmapToViewport`. The actor creation should happen on the fly (on the image_rendered which is hooked to the seg display), and not be done before hand. 

2. setDerivedColorMap is not necessary and should be removed, the colorTF and opacityTF should be handled same way that the volume seg is getting handled (in the `render` via `_setLabelmapColorAndOpacity` ), since it should respect the colorLUTIndex in the seg configs. 

3. getDerivedImage function should get removed, you don't need it. There should me 0 mention of segmentation inside the cornerstonejs/core. Again the render function of the lablemap display can handle this. That function is the one who name the image id (stackSeg:...) and knows how retrieve them too.

5. I understand some of the functions you refactored are needed since they can/should be called elsewhere, but I'm pretty sure you are over doing refactoring. Which makes the code reading harder since the reader should jump file to file to see the logic. Let's only refactor the pieces of the code that is needed, and don't abstract too much

packages/tools/src/types/LabelmapTypes.ts
rodrigobasilio2022: But how that way there isnt no typescript checking. I am ok if you are

packages/tools/src/tools/segmentation/SphereScissorsTool.ts
rodrigobasilio2022: yes

packages/tools/examples/StackSegmentationMammography/index.ts
rodrigobasilio2022: no I will delete in future versions

packages/tools/src/types/LabelmapTypes.ts
rodrigobasilio2022: done

packages/tools/src/types/LabelmapTypes.ts
sedghi: you can check if volumeId is present or imageIds. At least make the type: 'stack'

packages/tools/src/tools/displayTools/Labelmap/validateRepresentationData.ts
```diff
@@ -1,6 +1,10 @@
 import { SegmentationPublicInput } from '../../../types/SegmentationStateTypes';
 import { cache } from '@cornerstonejs/core';
-import { LabelmapSegmentationData } from '../../../types/LabelmapTypes';
+import type {
+  LabelmapSegmentationData,
```
sedghi, line 4: rename this to `LabelmapSegmentationDataStack`

packages/core/src/types/IStackInput.ts
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * VolumeInput that can be used to add a volume to a viewport. It includes
+ * mandatory `volumeId` but other options such as `visibility`, `blendMode`,
+ * `slabThickness` and `callback` can also be provided
+ */
+interface IStackInput {
+  /** imageId of the image in the cache */
+  imageId: string;
+  // actorUID for segmentations
```
sedghi, line 22: actorUID of anything

packages/core/src/types/IActor.ts
```diff
@@ -20,4 +21,6 @@ export type ActorEntry = {
   referenceId?: string;
   /** slab thickness for the actor */
   slabThickness?: number;
+  /** the vtkImage data related to an ImageActor */
+  imageData?: vtkImageData;
```
sedghi, line 25: I don't think this belongs here. We have the actor, you can grab the imageData directly from it

`actor.getMapper().getInputData()``

packages/core/src/loaders/imageLoader.ts
```diff
@@ -183,6 +197,122 @@ export function loadAndCacheImages(
...
+      throw new Error('TargetBuffer should be Float32Array or Uint8Array');
+    }
+  } else {
+    // Use Uint8 if no targetBuffer is provided
+    numBytes = length;
+    TypedArray = Uint8Array;
+  }
+
+  if (targetBuffer?.sharedArrayBuffer) {
+    const buffer = new SharedArrayBuffer(numBytes);
```
sedghi, line 249: we don't need sharedArrayBuffer for images

packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -1973,6 +1817,24 @@ class StackViewport extends Viewport implements IStackViewport {
...
+    stackInputs: Array<IStackInput>,
+    immediateRender: boolean,
+    suppressEvents: boolean
+  ): Promise<void> {
+    const actors = this.getActors();
+    stackInputs.forEach((stackInput) => {
+      const {
+        segmentationActor: imageActor,
+        segmentationImageData: imageData,
+      } = createActorMapperFromImageId(stackInput.imageId);
```
sedghi, line 1830: there should not be any mention of segmentation in core

packages/tools/src/tools/displayTools/Labelmap/labelmapDisplay.ts
```diff
@@ -192,30 +210,55 @@ async function render(
...
+      }
 
+      viewport.element.removeEventListener(
+        Enums.Events.IMAGE_RENDERED,
+        updateSegmentationImage
+      );
+      viewport.element.addEventListener(
+        Enums.Events.IMAGE_RENDERED,
+        updateSegmentationImage
+      );
```
sedghi, line 260: These should go inside tools/src/eventListeners. You need a new StackImageChangeEventListener that gets initiated on the init and triggers `triggerSegmentationRender` or similar to make the flow reach here. You shouldn't subscribe here.

packages/core/src/loaders/imageLoader.ts
rodrigobasilio2022: I was thinking about very large images (like tomosynthesis). Anyway its not a default option

sedghi: SAB doesn't have any benefit regarding size

packages/tools/src/tools/displayTools/Labelmap/addLabelmapToElement.ts
```diff
@@ -31,27 +36,46 @@ async function addLabelmapToElement(
...
+      renderingEngine,
+      volumeInputs,
+      [viewportId],
+      immediateRender,
+      suppressEvents
+    );
+  } else {
+    const stackInputs: Types.IStackInput[] = [
+      {
+        imageId: viewport.getCurrentImageId(),
```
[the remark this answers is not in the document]
sedghi: No, I meant something else. You are assigning the imageId here the imageId of the viewport (which has the image e.g., ct), but I expected this imageId be the imageId of the segmentation which you should grab from the segmentation state manager. No?

packages/core/src/utilities/getDerivedImageId.ts
```diff
@@ -0,0 +1,5 @@
+import imageIdToURI from './imageIdToURI';
+
+export default function getDerivedImageId(imageId: string): string {
+  return 'stackSeg:derived_' + imageIdToURI(imageId);
```
sedghi, line 4: There should not be any mention of the stackSeg anywhere in the core

packages/core/src/loaders/imageLoader.ts
```diff
@@ -183,6 +197,116 @@ export function loadAndCacheImages(
...
+  if (!imageIds || imageIds.length === 0) {
+    throw new Error(
+      'createAndCacheDerivedImages: parameter imageIds must be list of image Ids'
+    );
+  }
+
+  const derivedImageIds = [];
+  const allPromises = imageIds.map((imageId) => {
+    const options: DerivedImageOptions = {
+      imageId: getDerivedImageId(imageId),
```
sedghi, line 301: You don't need to getDerivedImageId here, the `derived` imageId should be part of argument to get passed inside the function, take a look at the createDerivedVolume and see how that has options including volumeId

packages/core/src/RenderingEngine/helpers/getDerivedImage.ts
```diff
@@ -0,0 +1,11 @@
...
+import { IImage, ICachedImage } from '../../types';
+import { imageIdToURI } from '../../utilities';
+
+export default function getSegmentationImage(image: IImage): IImage {
+  return getSegmentationImageFromImageId(image.imageId);
+}
+
+export function getSegmentationImageFromImageId(imageId: string): IImage {
+  return cache.getDerivedImage(imageId);
+}
```
sedghi, line 11: This file should not exist at all in the core

packages/core/src/RenderingEngine/helpers/createVTKImageDataFromImage.ts
```diff
@@ -0,0 +1,24 @@
...
+    dimensions,
+    spacing,
+    numComps: 1,
+    pixelArray: image.getPixelData(),
+  });
+  return imagedata;
+}
+
+export function createVTKImageDataFromImageId(imageId) {
+  const image = getSegmentationImageFromImageId(imageId);
```
sedghi, line 22: Should not exist, the creation of segmentation image from image Id should happen inside the tools

packages/tools/src/tools/segmentation/PaintFillTool.ts
```diff
@@ -77,18 +85,64 @@ class PaintFillTool extends BaseTool {
...
+
+      // Hacking the spacing information.
+      const spacing = imageData.getSpacing();
+      spacing[2] = zSpacing;
+      imageData.setSpacing(spacing);
+      // Hacking the origin
+      imageData.setOrigin(origin);
+      direction = calculatedDirection;
+      auxiliaryData.imageIds = sortedImageIds;
+      const image = cache.getDerivedImage(currentImageId);
```
sedghi, line 140: you don't need this cache.getDerivedImage. just put a enum/constant in the tools as segmentation prefix and just do `${segPrefix}:${imageId}`

packages/tools/src/tools/segmentation/strategies/stack/eraseCircle.ts
```diff
@@ -0,0 +1,18 @@
+import type { Types } from '@cornerstonejs/core';
```
sedghi, line 1: I'm pretty sure you don't need to refactor this either. You can just branch out the logic for the 'stack' vs 'volume' in the current files. Please don't do refactoring if not necessary

packages/tools/src/tools/segmentation/strategies/fillRectangle.ts
```diff
@@ -30,55 +17,11 @@ function fillRectangle(
...
-  });
-
-  const boundsIJK = getBoundingBoxAroundShape(rectangleCornersIJK, dimensions);
-
-  if (boundsIJK.every(([min, max]) => min !== max)) {
-    throw new Error('Oblique segmentation tools are not supported yet');
+  if (operationData.editData.type === 'volume') {
+    volumeStrategy.fillRectangle(enabledElement, operationData);
+  } else {
+    stackStrategy.fillRectangle(enabledElement, operationData);
```
sedghi, line 23: These operations share a lot of logic. Only branch out the logic for the scalarData that is needed, they share almost 80% of the code

packages/tools/src/tools/segmentation/PaintFillTool.ts
```diff
@@ -77,18 +85,64 @@ class PaintFillTool extends BaseTool {
...
+        labelmapData as LabelmapSegmentationDataStack;
+      const {
+        zSpacing,
+        sortedImageIds,
+        origin,
+        direction: calculatedDirection,
+      } = sortImageIds(referencedImageIds);
+      const currentImageId = viewport.getCurrentImageId();
+
+      imageData = createVTKImageDataFromImageId(currentImageId);
```
sedghi, line 124: This is not consistent with what is happening in our volumes, you are here actually creating the imageData though in volumes the segmentation should be pre-created and above we only grab it by `cache.getVolume`. I think it is fine but we need to document this much better, also the following couple of lines are shared between all tools so THIS sure should get refactored

packages/tools/src/tools/segmentation/PaintFillTool.ts
```diff
@@ -77,18 +85,64 @@ class PaintFillTool extends BaseTool {
...
+      imageData = segmentation.imageData;
+    } else {
+      const { referencedImageIds } =
+        labelmapData as LabelmapSegmentationDataStack;
+      const {
+        zSpacing,
+        sortedImageIds,
+        origin,
+        direction: calculatedDirection,
+      } = sortImageIds(referencedImageIds);
```
sedghi, line 121: I don't understand why it needs to sort referencedImageIds. It should just grab the current image id and gets its positionpatient, its direction and that is it. you can grab those from viewport.getImageData

packages/tools/src/tools/segmentation/RectangleScissorsTool.ts
```diff
@@ -145,25 +148,44 @@ class RectangleScissorsTool extends BaseTool {
...
+        segmentsLocked,
+        segmentColor,
+        segmentationId,
+        viewportIdsToRender,
+        handleIndex: 3,
+        movingTextBox: false,
+        newAnnotation: true,
+        hasMoved: false,
+      };
+    }
```
sedghi, line 188: this is alsmot repeated code, this can better organized like 

```js
this.editData = {
  annotation,
  segmentIndex,
  segmentsLocked,
  segmentColor,
  segmentationId,
  viewportIdsToRender,
  handleIndex: 3,
  movingTextBox: false,
  newAnnotation: true,
  hasMoved: false,
};

// Conditionally 
if (labelmapData.type === 'volume' || !labelmapData.type) {
  const { volumeId } = representationData[type] as LabelmapSegmentationDataVolume;
  const segmentation = cache.getVolume(volumeId);
  
  this.editData.segmentation = segmentation;
} else {
  this.editData.currentImageId = viewport.getCurrentImageId();
}

```

packages/core/src/loaders/imageLoader.ts
rodrigobasilio2022: I pass getDerivedImageId as a function param

fedorov:
Per discussion today, this will be reopened. The discussion and work stopped due to lack of resources to continue the development.

PR #683, opened 2023-07-10
fix(color volume viewport): fix incorrect property on volume actor

sedghi:
### Context

Webloader was returning RGBA true which volume viewport could not handle. It is fixed now. Also the volume viewports color was not working either because of it is changed now to reflect `setIndependentComponents(false)` . In addition streaming into the volume for color was not working since we had a short circute

### Changes & Results

- Modify webloader to be volumetric 
- add setIndependentComponents for color volume actor
- !isColorImage(imageFrame.photometricInterpretation) is not necessary in decdeImageFrame

### Testing

yarn run example webloader 

should work now

https://deploy-preview-683--cornerstone-3d-docs.netlify.app/live-examples/webloader

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  

### DIFF
--- a/packages/core/examples/webLoader/registerWebImageLoader.ts
+++ b/packages/core/examples/webLoader/registerWebImageLoader.ts
@@ -16,18 +16,34 @@ function createImage(image, imageId) {
   const rows = image.naturalHeight;
   const columns = image.naturalWidth;
 
-  function getPixelData() {
+  function getPixelData(targetBuffer) {
     const imageData = getImageData();
 
-    // modify original image data and remove alpha channel (RGBA to RGB)
-    const data = new Uint8Array(imageData.width * imageData.height * 3);
-    for (let i = 0, j = 0; i < imageData.data.length; i += 4, j += 3) {
-      data[j] = imageData.data[i];
-      data[j + 1] = imageData.data[i + 1];
-      data[j + 2] = imageData.data[i + 2];
+    let targetArray;
+
+    // Check if targetBuffer is provided for volume viewports
+    if (targetBuffer) {
+      targetArray = new Uint8Array(
+        targetBuffer.arrayBuffer,
+        targetBuffer.offset,
+        targetBuffer.length
+      );
+    } else {
+      targetArray = new Uint8Array(imageData.width * imageData.height * 3);
     }
 
-    return data;
+    // modify original image data and remove alpha channel (RGBA to RGB)
+    convertImageDataToRGB(imageData, targetArray);
+
+    return targetArray;
+  }
+
+  function convertImageDataToRGB(imageData, targetArray) {
+    for (let i = 0, j = 0; i < imageData.data.length; i += 4, j += 3) {
+      targetArray[j] = imageData.data[i];
+      targetArray[j + 1] = imageData.data[i + 1];
+      targetArray[j + 2] = imageData.data[i + 2];
+    }
   }
 
   function getImageData() {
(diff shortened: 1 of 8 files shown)


packages/core/examples/webLoader/registerWebImageLoader.ts
```diff
@@ -70,11 +78,12 @@ function createImage(image, imageId) {
     height: rows,
     width: columns,
     color: true,
-    rgba: true,
+    // since the canvas return rgba we should tell the cornerstone that we have 4 components per pixel
```
wayfarer3130, line 81: I don't get this comment - it implies that we should be setting rgba to true is how I read it.

packages/core/examples/webLoader/registerWebImageLoader.ts
```diff
@@ -202,11 +211,17 @@ function _loadImageIntoBuffer(
           // If we have a target buffer, write to that instead. This helps reduce memory duplication.
           const { arrayBuffer, offset, length } = options.targetBuffer;
 
+          // since the web image was loaded via canvas api it has rgba format
+          // below we convert it to rgb since the vtk image data can work
+          // only with rgb format
+
           // @ts-ignore
-          const pixelDataRGBA = image.getPixelData();
+          const pixelDataRGB = image.getPixelData();
```
wayfarer3130, line 219: Why do you need yet another image data array?  You don't even appear to use targetArray, so can you throw it away entirely?    I don't get this logic here.

packages/streaming-image-volume-loader/src/BaseStreamingImageVolume.ts
```diff
@@ -107,6 +107,8 @@ export default class BaseStreamingImageVolume extends ImageVolume {
       windowCenter,
       windowWidth,
       color,
+      // we don't use rgb for the volume actors, so this is always false
```
wayfarer3130, line 110: But this is saying rgba is false, not rgb is false, and I thought the point of the entire exercise was to be able to display color volumes?  Could you make the comment more descriptive.

packages/core/examples/webLoader/registerWebImageLoader.ts
sedghi: targetArray is used to view the portion of the full volume at the offset and length, we need it to fill the volume

wayfarer3130: Oh, I see what you are doing here - the target array is created from the passed in arrayBuffer instance, so it is a view onto the destination object and you are setting it.  Missed that and thought you were creating a new instance.
You can avoid the copy entirely by passing targetBuffer into getPixelData and creating the Uint8array view inside the getPixelData.  Then, that getPixelData has enough information to do a bounds check to see that nothing is being cut off, and to throw an exception if the provided array is too small.

sedghi: It is not about only volume viewport and target buffer, it should handle both cases (Stack and volume) so I don't want to modify the getPixelData since that correctly return the pixelData for stack viewport, does that makes sense?

wayfarer3130: That is why I would put the logic into getPixelData, setting it up as an optional parameter.  The return value will be identical in both cases, just if you are passed an array buffer, it uses that array buffer (+ offsets info), otherwise create new data.  That prevents the copy operation entirely, while being 100% compatible with either call method.

sedghi: Ok applied it

wayfarer3130 (approved):

PR #686, opened 2023-07-12
fix(PT measurement units): Non-SUV scaled, but pre-scaled PTs should show proper units

jbocce:
### Context
The various ROI measurements (i.e. rectangle, ellipse, circle, etc.) on non-SUV scaled PTs were showing no units at all. Instead, the units should come from DICOM (0054,1001)/units tag.

SUV scaled PTs should continue to show 'SUV' as the units.

### Changes & Results

Non-SUV scaled, but pre-scaled PTs show the units from the DICOM (0054,1001)/units tag.
SUV scaled PTs show SUV units. 
The modality unit displayed as part of the measurement is now also added to the cached stats and fired with the ANNOTATION_ADDED event.

### Testing

Add an ROI measurement to a PT and depending on its scaling, the appropriate unit should display for the statistics units.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] "OS: Windows 11 
- [x] "Node version: 16.14.0
- [x] "Browser: Chrome 114.0.5735.199 
  

### DIFF
--- a/packages/tools/src/tools/annotation/DragProbeTool.ts
+++ b/packages/tools/src/tools/annotation/DragProbeTool.ts
@@ -155,6 +155,14 @@ class DragProbeTool extends ProbeTool {
 
     const color = this.getStyle('color', styleSpecifier, annotation);
 
+    const isPreScaled = isViewportPreScaled(viewport, targetId);
+
+    const isSuvScaled = this.isSuvScaled(
+      viewport,
+      targetId,
+      annotation.metadata.referencedImageId
+    );
+
     if (!data.cachedStats[targetId]) {
       data.cachedStats[targetId] = {
         Modality: null,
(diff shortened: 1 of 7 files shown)


packages/tools/src/tools/annotation/DragProbeTool.ts
```diff
@@ -155,16 +155,36 @@ class DragProbeTool extends ProbeTool {
...
         value: null,
       };
 
-      this._calculateCachedStats(annotation, renderingEngine, enabledElement);
+      this._calculateCachedStats(
+        annotation,
+        renderingEngine,
+        enabledElement,
+        isPreScaled,
+        isSuvScaled
```
sedghi, line 178: can we add them as object at the end as `options`? instead of positional arguments?

packages/tools/src/utilities/getModalityUnit.ts
```diff
@@ -1,16 +1,30 @@
...
+      const units = petSeriesModule?.units;
+
+      if (units) {
+        return units;
+      } else {
+        return 'unitless';
+      }
+    } else {
+      return 'raw';
+    }
```
sedghi, line 27: can we please refactor this into handlePTModality? at the bottom of this function?

sedghi: I believe it can be written simpler like 

```js
function handlePTModality(isPreScaled, isSuvScaled, imageId, metaData) {
  if (!isPreScaled) {
    return 'raw';
  }
  
  if (isSuvScaled) {
    return 'SUV';
  }

  const petSeriesModule = metaData.get('petSeriesModule', imageId);
  return petSeriesModule?.units ?? 'unitless';
}
```

sedghi (commented):
see my comments please

sedghi (approved):

PR #692, opened 2023-07-14
feat(segmentation export): add new cornerstone3D segmentation export adapter

sedghi:
### Context

This PR adds DICOM SEG export for cornerstone3D. It also uses the old importer and optimize and improve upon it for parsing DICOM SEG resulting in 3x faster read time for DICOM SEG.

### Changes & Results

modifies the DICOM SEG export in adapters

### Testing

Export
- Should be able to run the `yarn run example segmentationExport` and export a SEG

Parsing
- Test should be done later in OHIF when I create the PR

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  


packages/adapters/examples/segmentationExport/index.ts
```diff
@@ -0,0 +1,328 @@
...
+
+addButtonToToolbar({
+    title: "Create SEG",
+    onClick: async () => {
+        const volume = cache.getVolume(volumeId);
+
+        const volumeImages = volume.convertToCornerstoneImages();
+        const imagePromises = volumeImages.map(image => image.promise);
+
+        await Promise.all(imagePromises).then(images => {
```
rogerbramon, line 115: Is PixelData needed to generate the segmentation, or metadata would be enough?

sedghi: I believe dcmjs only need metadata to create a derived multiframe SEG from the reference

packages/adapters/src/adapters/Cornerstone3D/Segmentation/generateLabelMaps2DFrom3D.ts
```diff
@@ -0,0 +1,52 @@
...
+
+        if (segmentsOnLabelmap.length === 0) {
+            continue;
+        }
+
+        segmentsOnLabelmap.forEach(segmentIndex => {
+            segmentsOnLabelmap3D.add(segmentIndex);
+        });
+
+        labelmaps2D[dimensions[2] - z] = labelmap2D;
```
rogerbramon, line 41: Should that be `dimensions[2] - 1 - z`?

packages/adapters/src/adapters/Cornerstone/Segmentation_4X.js
```diff
@@ -1235,7 +1375,10 @@ function unpackPixelData(multiframe) {
     }
 
     if (segType === "BINARY") {
-        return BitArray.unpack(data);
+        // For extreme big data, we can't unpack the data at once and we need to
+        // chunk it and unpack each chunk separately.
+        // MAX 2GB is the limit right now to allocate a buffer
+        return getUnpackedChunks(data, 199000000);
```
wayfarer3130, line 1381: This is nice to change so that larger data sizes can be supported

packages/adapters/src/adapters/helpers/codeMeaningEquals.ts
```diff
@@ -0,0 +1,9 @@
+const codeMeaningEquals = codeMeaningName => {
```
wayfarer3130, line 1: Could you add a docs here - this isn't a strict comparison of two values, but generates a function that compares the originally provided value with a content item.

packages/adapters/src/adapters/helpers/graphicTypeEquals.ts
```diff
@@ -0,0 +1,7 @@
+const graphicTypeEquals = graphicType => {
```
wayfarer3130, line 1: Ditto

packages/dicomImageLoader/src/imageLoader/getDatasetModule.ts
```diff
@@ -0,0 +1,44 @@
...
+    } catch (error) {
+      console.error(`Error retrieving ${t} data:`, error);
+    }
+  }
+
+  return result;
+}
+
+function capitalizeTag(tag) {
+  const parts = tag.split(/([0-9]+)/);
```
wayfarer3130, line 38: Could you explain why this is splitting on numbers and removing them?  I understand it capitalizes the first character, but not why it deletes any numbers in it.
Also, this should be const capitalizeTag = tag => etc

packages/dicomImageLoader/src/imageLoader/wadors/metaData/metaDataProvider.ts
```diff
@@ -266,6 +267,28 @@ function metaDataProvider(type, imageId) {
       actualFrameDuration: getNumberValue(metaData['00181242']),
     };
   }
+
+  // This is used for gathering all the metadata for export
+  if (type === 'dataset') {
+    const types = [
```
wayfarer3130, line 273: Can you externalize this list of types, and add a comment on it?  It definitely is NOT all the types, but there doesn't seem to be a safe way to get all the actual types available.  Would be nice to have a way for a module to register a new type if needed.

packages/dicomImageLoader/src/imageLoader/wadors/metaData/metaDataProvider.ts
```diff
@@ -266,6 +267,28 @@ function metaDataProvider(type, imageId) {
...
+      'modalityLutModule',
+      'sopCommonModule',
+      'petIsotopeModule',
+      'overlayPlaneModule',
+      'transferSyntax',
+      'petSeriesModule',
+      'petImageModule',
+    ];
+
+    return getDatasetModule(imageId, metaDataProvider, types);
```
wayfarer3130, line 290: Can you register a new type 'instance' similar to what OHIF uses, so that if a different metadata provider is used, it can safely provide a complete set of tags rather than relying on this rather adhoc collection?  Then, specifically in teh case of OHIF, it becomes possible to get a complete set.

packages/dicomImageLoader/src/imageLoader/wadouri/metaData/metaDataProvider.ts
```diff
@@ -245,6 +246,27 @@ function metaDataProvider(type, imageId) {
...
+      'voiLutModule',
+      'sopCommonModule',
+      'petIsotopeModule',
+      'overlayPlaneModule',
+      'transferSyntax',
+      'petSeriesModule',
+      'petImageModule',
+    ];
+
+    return getDatasetModule(imageId, metaDataProvider, types);
```
wayfarer3130, line 268: Ditto.

packages/adapters/src/adapters/Cornerstone3D/Segmentation/generateSegmentation.ts
```diff
@@ -0,0 +1,57 @@
...
+        metadata,
+        options
+    );
+    return fillSegmentation(segmentation, labelmaps, options);
+}
+
+/**
+ * _createMultiframeSegmentationFromReferencedImages - description
+ *
+ * @param images - An array of the cornerstone image objects.
```
wayfarer3130, line 25: Could you add a bit more information here on what the images are - I think given the context that the images are the segmentation images, and are thus single bit image objects containing the actual data in the segmentation, but it isn't at all obvious to me externally to this function that the images are just the segmentation single bit images.

packages/dicomImageLoader/src/imageLoader/wadouri/metaData/metaDataProvider.ts
```diff
@@ -245,6 +246,27 @@ function metaDataProvider(type, imageId) {
       actualFrameDuration: dicomParser.intString(dataSet.string('x00181242')),
     };
   }
+
+  if (type === 'dataset') {
```
wayfarer3130, line 250: Doesn't OHIF use 'instance'   Lets try to keep consistent with what is already existing if possible.  I think that is an equivalent type setup, and will allow all tags to be accessed for such providers.

packages/adapters/src/adapters/Cornerstone3D/Segmentation/generateSegmentation.ts
wayfarer3130: Looking at the dcmjs side, these are the original images, and in fact the generation code for the derivation throws away all the pixel data itself, using only the length.  I suggest not passing in the actual pixel data, but generating a dummy buffer containing just the byte lengths as required.   That way, even the temporary buffer will have the right lengths, as the generated buffer is simply a copy of the derived dataset buffers, and that is usually twice the required length.

packages/adapters/src/adapters/Cornerstone3D/Segmentation/generateLabelMaps2DFrom3D.ts
wayfarer3130: Is this generating label maps that are frame indexed starting with 1?  If so, that fact should be in a comment here and should be noted up top.

packages/adapters/src/adapters/Cornerstone3D/Segmentation/generateLabelMaps2DFrom3D.ts
```diff
@@ -0,0 +1,52 @@
+function generateLabelMaps2DFrom3D(labelmap3D) {
```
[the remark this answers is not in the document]
wayfarer3130: I think the z dimension isn't necessarily the image slice direction, but rather the dimensions array indexes the actual images to the specified image sets in the correct direction.

packages/adapters/src/adapters/Cornerstone/Segmentation_4X.js
```diff
@@ -1541,3 +1660,109 @@ function getSegmentMetadata(multiframe, seriesInstanceUid) {
         data
     };
 }
+
+function readFromUnpackedChunks(chunks, offset, length) {
```
wayfarer3130, line 1664: Just add a comment indicating that this combines the chunks back into a Uint8Array.  You might consider renaming the file to Segmentation_4X.ts so that you can add type information here.  That would definitely be more clear.

wayfarer3130, line 1664: Consider using the code from StaticDICOMweb that has all this functionality, exposing it as though it were a single Uint8Array object - that is asyncIterableToBuffer.  That allows reading very large buffers and automatically splitting them as required.  I think we will need that type of functionality in the future for better handling of streaming results so it would be nice to include.  I could add it into dcmjs if that is where we think it belongs.

packages/adapters/src/adapters/Cornerstone/Segmentation_4X.js
```diff
@@ -1541,3 +1660,109 @@ function getSegmentMetadata(multiframe, seriesInstanceUid) {
...
+        endChunkIndex++;
+    }
+
+    return {
+        start: { chunkIndex: startChunkIndex, offset: startOffsetInChunk },
+        end: { chunkIndex: endChunkIndex, offset: endOffsetInChunk }
+    };
+}
+
+function calculateCentroid(imageIdIndexBufferIndex, multiframe) {
```
wayfarer3130, line 1731: It feels to me like some of these should be added to a helper/utils export rather than being directly included here, as they may well be useful in other segmentation generations in the future, and also that allows explaining what they are doing cleanly.

wayfarer3130 (changes requested):
Some clean up, but definitely looks good.

packages/adapters/src/adapters/Cornerstone3D/Segmentation/generateSegmentation.ts
sedghi: we already have the pixelData from the volume. So the cost is negligible compared to creating another temporary buffer etc.

packages/adapters/src/adapters/Cornerstone/Segmentation_4X.js
sedghi: it is really not a general utility as it is relying on imageIdIndexBufferIndex

packages/dicomImageLoader/src/imageLoader/wadouri/metaData/metaDataProvider.ts
```diff
@@ -245,6 +249,10 @@ function metaDataProvider(type, imageId) {
       actualFrameDuration: dicomParser.intString(dataSet.string('x00181242')),
     };
   }
+
+  if (type === 'instance') {
+    return getInstanceModule(imageId, metaDataProvider, instanceModuleNames);
```
rogerbramon, line 254: Not sure if it would make sense to use the `naturalizeDataset` or `namifyDataset` methods of `dcmjs` [DicomMetaDictionary](https://github.com/dcmjs-org/dcmjs/blob/f844e025f86055a918c13269f6d762a711f3a4bf/src/DicomMetaDictionary.js#L111) to get the full metadata of the instance that already match the dcmjs naming.

sedghi: the problem is that we don't have the instance stored anywhere to naturalize or namify it.

wayfarer3130 (approved):

PR #696, opened 2023-07-18
feat(nifti): Add nifti volume loader to cornerstone 3D repo

sedghi:
🎉 The work on this PR was made possible thanks to the generous funding from **Mayo Clinic**, underlining their dedication to fostering open-source community development.

👏  A hearty round of applause for James A. Petts (@JamesAPetts) and Davide Punzo (@Punzo), whose diligent work on the nifti loader proved invaluable.

### Context

The nifti format is widely recognized for storing medical images and is the preferred choice for some research labs. This PR introduces a new package, `@cornerstonejs/nifti-volume-loader`, crafted to fetch a nifti file from a URL and load it as a volume within cornerstone3D.

We have included two new demonstrations:

- BasicNifti: Loads and renders a nifti file.
- NiftiWithTools: Loads a nifti file, enables tools, and displays a progress bar during the rendering process.

Given that the metadata for the nifti file is incorporated within the file header, and there is no standard procedure for servers to return the metadata, both metadata and pixelData will be fetched from the entire binary blob from the URL via an XHR request.

### API

1. Prepend `nifti:` to the URL for the volume

```
const niftiURL = 'https://ohif-assets.s3.us-east-2.amazonaws.com/nifti/MRHead.nii.gz';
const volumeId = 'nifti:' + niftiURL;

```

2.1 create rendering engine and enable viewports like every other use case in cs3d

1. Create and Cache the volume, this will also load the volume as opposed to our dicom volume loader which needs a final `.load` (since in dicom we can fetch the metadata, create the volume, and stream data in)

```
  const volume = await volumeLoader.createAndCacheVolume(volumeId);

```

1. assign it to viewports

```
setVolumesForViewports(
    renderingEngine,
    [{ volumeId, callback: setCtTransferFunctionForVolumeActor }],
    viewportIds
  );

```

### Changes & Results

Add nifti loader

### Testing

Test the two examples with these data

1. https://ohif-assets.s3.us-east-2.amazonaws.com/nifti/CTACardio.nii
2. https://ohif-assets.s3.us-east-2.amazonaws.com/nifti/CTACardio.nii.gz
3. https://ohif-assets.s3.us-east-2.amazonaws.com/nifti/MRBrainTumor1.nii.gz
4. https://ohif-assets.s3.us-east-2.amazonaws.com/nifti/MRHead.nii.gz

### PR

### Code

etc.)

### Public Documentation Updates

additions or removals.

### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:


packages/nifti-volume-loader/src/helpers/scaleNiftiArray.ts
```diff
@@ -0,0 +1,23 @@
+/**
+ * Given a pixel array, rescale the pixel values using the rescale slope and
+ * intercept and if modality is PT it uses the suv values to scale the array
```
jbocce, line 3: I do not see in the code where the check for PT is, nor the SUV scaling. What am I missing?

packages/nifti-volume-loader/src/NiftiImageVolume.ts
```diff
@@ -0,0 +1,84 @@
...
+  loading: boolean;
+  callbacks: Array<(...args: unknown[]) => void>;
+};
+
+type NiftiImageProperties = {
+  loadStatus: LoadStatus;
+  controller: AbortController;
+};
+
+// const requestType = Enums.RequestType.Prefetch;
```
jbocce, line 15: Consider removing these comments.

packages/nifti-volume-loader/src/helpers/fetchAndAllocateNiftiVolume.ts
```diff
@@ -0,0 +1,263 @@
...
+        onProgress(e.loaded, e.total);
+      };
+    }
+
+    if (signal && signal.aborted) {
+      xhr.abort();
+      reject(new Error('Request aborted'));
+    }
+
+    signal.addEventListener('abort', () => {
```
jbocce, line 51: Consider removing the event listener for the various situations...
- the load completes (successfully)
- the xhr is aborted

packages/nifti-volume-loader/src/helpers/fetchAndAllocateNiftiVolume.ts
```diff
@@ -0,0 +1,263 @@
...
+      if (onload && typeof onload === 'function') {
+        onload();
+      }
+
+      resolve(xhr.response);
+    };
+
+    if (onProgress && typeof onProgress === 'function') {
+      xhr.onprogress = function (e) {
+        //  if (e.lengthComputable) {
```
jbocce, line 39: Consider uncommenting this code as the `lengthComputable` flag indicates if `loaded` and `total` are "useable".

packages/nifti-volume-loader/src/helpers/fetchAndAllocateNiftiVolume.ts
```diff
@@ -0,0 +1,263 @@
...
+      return new Int16Array(niftiImageBuffer);
+    default:
+      throw new Error(`datatypeCode ${datatypeCode} is not yet supported`);
+  }
+};
+
+export default async function fetchAndAllocateNiftiVolume(
+  volumeId: string
+): Promise<NiftiImageVolume> {
+  // nifti volumeIds start with 'nifti:' so we need to remove that
```
jbocce, line 76: Just a suggestion, but does it make sense to declare a constant somewhere called (for example) `nifitiFilePrefix = 'nifti:'` and then the following line would be `volumeId.substring(niftiFilePrefix.length)`?

packages/nifti-volume-loader/src/helpers/fetchAndAllocateNiftiVolume.ts
```diff
@@ -0,0 +1,263 @@
...
+  const volumeMetadata = makeVolumeMetadata(
+    niftiHeader,
+    orientation,
+    typedNiftiArray
+  );
+
+  // Todo: we should grab this from somewhere else but it doesn't really
+  // prevent us from rendering the volume so we'll just hardcode it for now.
+  Object.assign(volumeMetadata, {
+    Modality: 'MR',
```
jbocce, line 140: This seems odd to me, but I am sure I am missing something.

jbocce (commented):
Looks good so far. See my comments.

packages/core/src/types/AffineMatrix.ts
```diff
@@ -0,0 +1,8 @@
+type AffineMatrix = [
```
wayfarer3130, line 1: Do you really need to define yet another matrix format?  How about affineMatrix as a function producing a gl-matrix mat4 instance instead, taking four arrays like this as the arguments, and producing the standard mat4 representation.

longuto:
Why not continue doing this

longuto:
Why not continue with this PR

sedghi:
@longuto it is still under review (@wayfarer3130)

packages/nifti-volume-loader/package.json
```diff
@@ -0,0 +1,42 @@
...
+  "repository": "https://github.com/cornerstonejs/cornerstone3D",
+  "files": [
+    "dist/"
+  ],
+  "directories": {
+    "test": "test"
+  },
+  "sideEffects": false,
+  "scripts": {
+    "build:cjs": "tsc --project ./tsconfig.cjs.json",
```
wayfarer3130, line 17: I thought we were just doing two of the builds now?  I wouldn't do all three if we are shortly dropping one.

packages/nifti-volume-loader/package.json
```diff
@@ -0,0 +1,42 @@
...
+    "build:esm": "tsc --project ./tsconfig.esm.json",
+    "build:umd": "cross-env NODE_ENV=production webpack --config .webpack/webpack.prod.js",
+    "build:all": "yarn run build:umd && yarn run build:cjs && yarn run build:esm",
+    "build": "yarn run build:all",
```
wayfarer3130, line 21: Can you make sure there is a clean target too?

packages/nifti-volume-loader/src/enums/Events.ts
```diff
@@ -0,0 +1,16 @@
+/**
+ * Cornerstone Nifti volume loader events
+ */
+enum Events {
+  /**
+   * Event fired when a Nifti volume is loaded completely
+   *
+   */
+  NIFTI_VOLUME_LOADED = 'CORNERSTONE_NIFTI_VOLUME_LOADED',
```
wayfarer3130, line 9: Is it necessary to have different events from the dicom volume loader?  Really they should be the same events and just distinguish the type of data being loaded.  It might be necessary to have them be slightly different because of bulk loading, but ideally making them identical would be useful.

packages/nifti-volume-loader/src/helpers/affineUtilities.ts
```diff
@@ -0,0 +1,79 @@
...
+    affine[0][2] / spacing[2],
+    affine[1][0] / spacing[0],
+    affine[1][1] / spacing[1],
+    affine[1][2] / spacing[2],
+    affine[2][0] / spacing[0],
+    affine[2][1] / spacing[1],
+    affine[2][2] / spacing[2],
+  ] as Types.Mat3;
+
+  return { origin, orientation: orientation, spacing };
```
wayfarer3130, line 76: Just include orientation directly - no need to rename it to itself

packages/nifti-volume-loader/src/cornerstoneNiftiImageLoader.ts
```diff
@@ -0,0 +1,22 @@
...
+}
+
+export default function cornerstoneNiftiImageVolumeLoader(
+  volumeId: string
+): IVolumeLoader {
+  const niftiVolumePromise = fetchAndAllocateNiftiVolume(volumeId);
+
+  return {
+    promise: niftiVolumePromise,
+    cancel: () => {
```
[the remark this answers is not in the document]
sedghi: since it is part of the interface

packages/core/src/types/AffineMatrix.ts
sedghi: As discussed in the meeting, I will do it later

packages/nifti-volume-loader/src/enums/Events.ts
sedghi: I think so... , since there is no streaming happening?

packages/nifti-volume-loader/src/helpers/convert.ts
```diff
@@ -0,0 +1,119 @@
+import type { Types } from '@cornerstonejs/core';
+import { parseAffineMatrix } from './affineUtilities';
+
+/**
+ * This converts scalar data: LPS to RAS and RAS to LPS
```
wayfarer3130, line 5: Could you write out LPS/RAS?  At least once

packages/nifti-volume-loader/src/helpers/fetchAndAllocateNiftiVolume.ts
```diff
@@ -0,0 +1,266 @@
...
+  urlsMap.delete(niftiURL);
+
+  let niftiHeader = null;
+  let niftiImage = null;
+
+  if (NiftiReader.isCompressed(niftiBuffer)) {
+    niftiBuffer = NiftiReader.decompress(niftiBuffer);
+  }
+
+  if (NiftiReader.isNIFTI(niftiBuffer)) {
```
wayfarer3130, line 124: This will throw a NPE if it isn't NIFTI data - probably throw an error if it isn't nifti data specifically, and otherwise continue handling it.

packages/nifti-volume-loader/src/helpers/fetchAndAllocateNiftiVolume.ts
```diff
@@ -0,0 +1,266 @@
...
+    ImageOrientationPatient[5]
+  );
+
+  const { dims } = niftiHeader;
+
+  const numFrames = dims[3];
+
+  // Spacing goes [1] then [0], as [1] is column spacing (x) and [0] is row spacing (y)
+  const dimensions = <Types.Point3>[Columns, Rows, numFrames];
+  const direction = new Float32Array([
```
wayfarer3130, line 176: Things like this feel like they could be put into a function rather than making this function so long.

packages/nifti-volume-loader/src/helpers/fetchAndAllocateNiftiVolume.ts
```diff
@@ -0,0 +1,266 @@
...
+    scanAxisNormal[0],
+    scanAxisNormal[1],
+    scanAxisNormal[2],
+  ]) as Types.Mat3;
+  const signed = PixelRepresentation === 1;
+
+  // Check if it fits in the cache before we allocate data
+  // TODO Improve this when we have support for more types
+  // NOTE: We use 4 bytes per voxel as we are using Float32.
+  let bytesPerVoxel = 1;
```
wayfarer3130, line 192: And a fair bit of this could also be put into functions that provide specific bits of data/usage, and can have their own documentation and maybe even their own unit tests.

packages/nifti-volume-loader/src/helpers/scaleNiftiArray.ts
```diff
@@ -0,0 +1,25 @@
+/**
+ * Given a pixel array, rescale the pixel values using the rescale slope and
+ * intercept
+ *
+ * Todo: add the scaling of PT and SUV
+ * @param array - The array to be scaled.
+ * @param scalingParameters - The scaling parameters
+ * @returns The array being scaled
+ */
+export default function scaleNiftiArray(
```
wayfarer3130, line 10: Could we call this modalityScaleNifti or something that explains it isn't a scaling of the image, but of the pixel values.

packages/nifti-volume-loader/src/helpers/scaleNiftiArray.ts
```diff
@@ -0,0 +1,25 @@
...
+  array: Float32Array | Int16Array | Uint8Array,
+  niftiHeader
+): void {
+  const arrayLength = array.length;
+  const { scl_slope, scl_inter } = niftiHeader;
+
+  if (scl_slope === 0 || Number.isNaN(scl_slope)) {
```
wayfarer3130, line 17: Can you check !scl_slope || ...  so that if it ends up null/endefined, you still skip out here?

packages/nifti-volume-loader/src/helpers/scaleNiftiArray.ts
```diff
@@ -0,0 +1,25 @@
...
+    // No scaling encoded in NIFTI header
+    return;
+  }
+
+  for (let i = 0; i < arrayLength; i++) {
```
wayfarer3130, line 22: You are modifying this in place - would it be safer to make a copy?  What if the data is exposed as a Uint8array and the scaling no longer works?  This just feels like it might cause problems, but the memory usage is definitely bigger if you copy the data instead.

packages/nifti-volume-loader/tsconfig.json
```diff
@@ -0,0 +1,4 @@
+{
+  "extends": "../../tsconfig.json",
```
wayfarer3130, line 2: Didn't we find that referencing a global tsconfig file has issues with linking?

wayfarer3130 (approved):
There are a number of small comments/issues, nothing that HAS to be addressed, and I know this really should get in soon, so I'm ok with you reviewing the comments and deciding which ones to address.

jbocce (approved):
Looks good.

packages/nifti-volume-loader/src/helpers/scaleNiftiArray.ts
sedghi: This is the same behaviour in the dicom image loader

packages/nifti-volume-loader/tsconfig.json
sedghi: no that was something else that I fixed in webpack

Issue #673, opened 2023-06-28
[Bug] ScaleOverlayTool is not displaying in viewport with different _FrameOfReferenceUID


kubanm3:
### Describe the Bug

Each series I put in viewport changes it's _FrameOfReferenceUID so that same annotations can be displayed in multiple viewports that share the same series.

With ScaleOverlayTool it remebers whether it got added to some viewport and after series is changed in that same viewport it doesn't add new scaleOverlay annotation, making it not render in that viewport.

Error is because of adding viewport to `const viewportsWithAnnotations = [];`  in line 28 of ScaleOverlayTool.ts after changing series it doesn't add new annotation, because it thinks it already exists. But when rendering it gets annotations by FrameOfReferenceUID which doesnt exists.

### Steps to Reproduce

1. Create viewport with _FrameOfReferenceUID of seriesId.
2. render that series in viewport
3. enable ScaleOverlayTool so it gets added
4. change series in viewport, also changing it's _FrameOfReferenceUID
5. ScaleOverlayTool errors out and is not responsive

### The current behavior

after changing frameofreference and rendering new series in viewport rendering ScaleOverlay tool errors in console and is not responsive.

### The expected behavior

After series is changed and new _FrameOfReferenceUID is assigned to viewport ScaleOverlay annotation renders properly.

### OS

Ubuntu 22.04.2 LTS

### Node version

v20.1.0

### Browser

Brave Version 1.52.122 Chromium: 114.0.5735.110 (Official Build) (64-bit)

PR #700, opened 2023-07-20
fix(scale tool): render on viewports with different frameOfReference

kubanm3:
### Context
Fixes https://github.com/cornerstonejs/cornerstone3D/issues/673. 

### Changes & Results
Viewport may have assigned multiple frameOfReferences (one at a time) wich have caused not adding new scale annotation for new frameOfReference within one viewport.

Previously checking if scale annotation is added was done by comparing whether it had been added to viewport. But rendering is filtering annotations based on both frameOfReference and viewportId. Adding only one scale annotation per viewport caused it to not show on same viewport but with changed frameOfReference.

PR fixes it by comparing added scale annotation with both viewportId and frameOfReference. With normal way application works nothing changes as all viewports share one frameOfReference, however when vieport may be given different frameOfReference based on loaded series for example it will also check if frameOfReference is the same, allowing new scale annotation for same viewport.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [X] "OS:  Windows 10,
- [X] "Node version:  v20.1.0
- [X] "Browser: Microsoft Edge 114.0.1823.82
  

### DIFF
--- a/packages/tools/src/tools/ScaleOverlayTool.ts
+++ b/packages/tools/src/tools/ScaleOverlayTool.ts
@@ -25,7 +25,7 @@ import { StyleSpecifier } from '../types/AnnotationStyle';
 import { getToolGroup } from '../store/ToolGroupManager';
 
 const SCALEOVERLAYTOOL_ID = 'scaleoverlay-viewport';
-const viewportsWithAnnotations = [];
+const viewportsFrameWithAnnotations = [];
 
 /**
  * @public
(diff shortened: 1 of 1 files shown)


wayfarer3130:
Could you add some info on how to test this?

kubanm3:
There needs to be loaded another series in same viewport and frameofreference for viewport be set to something different.

While making it, I edited one of examples to test it. I could include it in this pr for testing purpose but it doesn't really have a purpose being new example so I've not inculed it in this pull request.

packages/tools/src/tools/ScaleOverlayTool.ts
```diff
@@ -114,9 +114,17 @@ class ScaleOverlayTool extends AnnotationDisplayTool {
...
+    // if the viewport and FrameOfReference does not have an annotation, create a new one
+    const viewportId = viewport.id;
+    const isScaleNotInViewportAndFrameOfReference =
+      !viewportsFrameWithAnnotations.find(
+        (viewport) =>
+          viewport.viewportId == viewportId &&
+          viewport.FrameOfReferenceUID == FrameOfReferenceUID
+      );
+
+    if (isScaleNotInViewportAndFrameOfReference) {
```
sedghi, line 127: I'm not sure I follow what is happening here. How is this solving the problem you mentioned? 

I think drag and drop is an app-level concern, so if you just make sure you call setToolEnabled again, it would solve this issue

PR #702, opened 2023-07-24
fix(loader): Load colour images correctly when specified Float32Array

wayfarer3130:
### Context

Sometimes the decoder specifies the wrong type of buffer for decoding.  This change prevents using the wrong
buffer type.

### Changes & Results

Hard code to use Uint8array for colour images even when Float32array is specified.

### Testing

Display a colour image in the stack view in OHIF, such as the test:e2e:serve "Horse" images

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  


packages/dicomImageLoader/src/shared/decodeImageFrame.ts
```diff
@@ -201,7 +201,11 @@ function postProcessDecodedPixels(
     Float32Array,
   };
 
-  if (options.targetBuffer && options.targetBuffer.type) {
+  const type = options.targetBuffer?.type;
+  const invalidTargetType =
+    isColorImage(imageFrame.photometricInterpretation) && type !== 'Uint8Array';
```
sedghi, line 206: why this is the case that we only care about uint8?

wayfarer3130: DICOM only defines 8 bit representations of colour data, and all of the decoded versions produce 8 bit data.  If you can tell me how to figure out the right representation before calling the decode, then I can change this, but the call in StackViewport doesn't have any data to base the required representation on.

sedghi: can you please add that comment then in the code?

wayfarer3130: Agreed to just check length and offset here and go to the next if otherwise.  This will result in not using the type here.

sedghi (approved):

PR #718, opened 2023-08-02
feat(toolEvent): added an event that is triggered when a tool is activated

lscoder:
Added an event that is triggered when a tool is activated to make it possible to the viewer detect any change to the active tool and update the UI showing a proper UI component (config)


packages/tools/src/store/ToolGroupManager/ToolGroup.ts
```diff
@@ -372,6 +376,13 @@ export default class ToolGroup implements IToolGroup {
       toolInstance.onSetToolActive();
     }
     this._renderViewports();
+
+    const eventDetail: ToolActivatedEventDetail = {
```
sedghi, line 380: it should get the toolGroup info too

sedghi (commented):
see my comment, also we can add the other events too or at least put a TODO

sedghi (approved):

PR #721, opened 2023-08-03
fix (angle tool): No text box if angle is incomplete/ value is NaN

IbrahimCSAE:
### Context

In the angle tool, if a user draws an angle where point 1, 2 or point 2,3 have the same coordinates, the calculated angle value will be NaN because it is impossible to calculate the angle. This causes the textbox not to be visible. It would be helpful to at least have the textbox say "Incomplete Angle" letting them know to adjust their points.

A better fix could be to prevent the creation of such invalid measurements and to prevent adjusting an existing angle to end up in that invalid state.

bug

### Changes & Results

Modified the  _calculateCachedStats method to set the angle value to be "Incomplete Angle" instead of NaN if the value is NaN.

In complete angle

### Testing

- visit -> [OHIF](https://v3-demo.ohif.org/viewer?StudyInstanceUIDs=2.16.840.1.114362.1.11972228.22789312658.616067305.306.2)
- the link above will already open a study. Select the angle tool
- try to draw an angle where the third point is on top of the second point
- If the above is too difficult, draw a normal angle, then try to adjust the points to overlap.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [X] "OS: Windows 10
- [X] "Node version: 18.15.0
- [X] "Browser: Chrome Version 115.0.5790.170 (Official Build) (64-bit)

### DIFF
--- a/packages/tools/src/tools/annotation/AngleTool.ts
+++ b/packages/tools/src/tools/annotation/AngleTool.ts
@@ -817,7 +817,7 @@ class AngleTool extends AnnotationTool {
       );
 
       cachedStats[targetId] = {
-        angle,
+        angle: isNaN(angle) ? 'Incomplete Angle' : angle,
       };
     }
 


wayfarer3130 (approved):
I think this is a good way of handling things.  
Please test saving/restoring to see that we still get the right values.  Note that the canonical way of representing NaN in the JSON metadata is (will be?) "NaN" - I have a DICOMweb CP submitted to change that.

IbrahimCSAE:
> I think this is a good way of handling things. Please test saving/restoring to see that we still get the right values. Note that the canonical way of representing NaN in the JSON metadata is (will be?) "NaN" - I have a DICOMweb CP submitted to change that.

@wayfarer3130 I just tested it, looks like it works as expected


Incomplete angle save/restore:

load SR


normal angle save/restore:

load SR normal angle

PR #723, opened 2023-08-03
Added statistics calculator for Annotation Tools

Celian-abd:
### Context

Refactored the stats calculation code for Annotation Tools and added a way to create and initialize new calculator with custom statistics.

### Changes
- Added new class BasicStatsCalculator, which has the default stats such as max, mean , stdDev..  used by the annotation tools instead of using callbacks.
- Made RectangleROITool statistics calculation match the other tools. (Using pointInShapeCallback instead of 3 'for')

### Testing
stackAnnotationTools example.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  


Celian-abd:
@sedghi Hey, I've been working on annotation tools and I added the feature to custom the text shown by the tool. So we can add custom text and also custom statistics calculator by the tool configuration.

So now we can set our custom text and calculator when adding the tool into the toolgroup with it's configuration or also by `setToolConfiguration` method.

I've created a new example aswell (stackAnnotationToolsSpecificConfiguration)

packages/tools/src/tools/annotation/DragProbeTool.ts
```diff
@@ -205,7 +205,7 @@ class DragProbeTool extends ProbeTool {
 
     renderStatus = true;
 
-    const textLines = this._getTextLines(data, targetId);
+    const textLines = this.configuration.getTextLines(data, targetId);
```
sedghi, line 208: shouldn't we add the default to configuration too? similar to others?

packages/tools/src/tools/annotation/LengthTool.ts
```diff
@@ -770,20 +771,20 @@ class LengthTool extends AnnotationTool {
     return renderStatus;
   };
 
-  // text line for the current active length annotation
-  _getTextLines(data, targetId) {
-    const cachedVolumeStats = data.cachedStats[targetId];
-    const { length, unit } = cachedVolumeStats;
+  // // text line for the current active length annotation
```
sedghi, line 774: remove

packages/tools/src/utilities/math/basic/BasicStatsCalculator.ts
```diff
@@ -0,0 +1,40 @@
...
+  let stdDevWithSumSquare = 0;
+
+  points.forEach((point) => {
+    const valueMinusMean = point.value - mean;
+    stdDev += valueMinusMean ** 2;
+  });
+
+  stdDevWithSumSquare = Math.sqrt(sumSquares / points.length - mean ** 2);
+  stdDev = Math.sqrt(stdDev / points.length);
+
```
sedghi, line 33: for these repetitive loop you should stick to simmple for loop so that browser can optimize it. Also one for loop is enough

packages/tools/src/utilities/math/basic/BasicStatsCalculator.ts
```diff
@@ -0,0 +1,40 @@
+//Basic calculator
+import { PointInShape, StatisticValue } from '../../../types';
+
+export default function BasicStatsCalculator(
```
sedghi, line 4: some doc would be appreciated

packages/tools/src/types/CalculatorTypes.ts
```diff
@@ -0,0 +1,16 @@
+import { Types } from '@cornerstonejs/core';
+
+export type PointInShape = {
```
sedghi, line 3: repetitive

packages/tools/src/tools/annotation/EllipticalROITool.ts
```diff
@@ -1116,57 +1092,29 @@ class EllipticalROITool extends AnnotationTool {
...
 
-        stdDev /= count;
-        stdDev = Math.sqrt(stdDev);
-
-        const modalityUnit = getModalityUnit(
-          metadata.Modality,
-          annotation.metadata.referencedImageId,
-          modalityUnitOptions
-        );
+        const stats = this.configuration.statsCalculator(pointsInShape);
```
sedghi, line 1108: so seems like there is some optimization issue happening here. 
Previously we had `pointInShapeCallback` which would take care of looping over the points in the shape and running a callback (e.g., meanMaxCalculator), but now we run the point in shape to grab the points and then again in the statsCalculator to run over them again (two loops instead of one), can you just make the statsCalculator a callbackto the pointInShape so that we don't run twice?

sedghi (commented):
see my comments

sedghi (commented):
see my comments

packages/tools/src/tools/annotation/EllipticalROITool.ts
Celian-abd: So if I want to use a callback to calculate the stats directly into the `pointInShapeCallback `, I would need to make my statsCalculator a function of a class that stores the sum of points and could return the calculated stats (such as `meanMaxCalculator`).

Does that seems correct to you ?

sedghi: I think so

packages/tools/src/store/ToolGroupManager/ToolGroup.ts
```diff
@@ -24,6 +24,7 @@ import {
 
 import { MouseCursor, SVGMouseCursor } from '../../cursors';
 import { initElementCursor } from '../../cursors/elementCursor';
+import Calculator from '../../types/CalculatorTypes';
```
sedghi, line 27: import it from the ../../types and merge it with above imports

packages/tools/src/store/ToolGroupManager/ToolGroup.ts
```diff
@@ -87,9 +88,12 @@ export default class ToolGroup implements IToolGroup {
    * to set the tool to be active or passive or in other states.
    *
    * @param toolName - string
-   * @param configuration - Tool configuration objects
+   * @param configuration - Tool configuration objects and a custom statistics calculator if needed
    */
-  addTool(toolName: string, configuration = {}): void {
+  addTool(
+    toolName: string,
+    configuration: Record<any, any> & { statsCalculator?: Calculator } = {}
```
sedghi, line 95: can you refactor this above to a ToolConfiguration and use it here and below?

packages/tools/src/tools/annotation/CircleROITool.ts
```diff
@@ -1011,57 +970,29 @@ class CircleROITool extends AnnotationTool {
...
           Modality: metadata.Modality,
           area,
-          mean,
-          max,
-          stdDev,
+          mean: stats[1]?.value,
+          max: stats[0]?.value,
+          stdDev: stats[2]?.value,
+          statsArray: stats,
+          pointsInShape: pointsInShape,
```
sedghi, line 995: why are we storing the pointsInShape?

packages/tools/src/utilities/math/basic/BasicStatsCalculator.ts
```diff
@@ -0,0 +1,60 @@
...
+   * Basic function that calculates statictics for a given array of points.
+   * @param points
+   * @returns An object that contains :
+   * max : The maximum value of the array
+   * mean : mean of the array
+   * stdDev : standard deviation of the array
+   * stdDevWithSumSquare : standard deviation of the array using sum²
+   */
+
+  getStatistics = (): StatisticValue[] => {
```
sedghi, line 40: I would call it Statistics instead of StatisticValue

packages/tools/src/utilities/math/basic/BasicStatsCalculator.ts
```diff
@@ -0,0 +1,60 @@
...
+  sumSquares = 0;
+  squaredDiffSum = 0;
+  count = 0;
+
+  /**
+   * This callback is used when we verify if the point is in the annotion drawn so we can get every point
+   * in the shape to calculate the statistics
+   * @param value of the point in the shape of the annotation
+   */
+  statsCallback = ({ value: newValue }): void => {
```
sedghi, line 17: I would call this `run` or something similar. And Also I would make the class an static class so that we don't do `new ...`

```
export default class BasicStatsCalculator extends Calculator {
  static max = -Infinity;
  static currentMax = 0;
  static sum = 0;
  static sumSquares = 0;
  static squaredDiffSum = 0;
  static count = 0;

  static statsCallback({ value: newValue }): void {
```

sedghi (commented):
getting there, I really like this PR.

packages/tools/src/tools/annotation/CircleROITool.ts
salimkanoun: For custom calculations in particular histogram based calculation, we will need it in several nuclear medicine applications, for instance if you want to calculate the nth percentile of the ROI. 
This way a custom calculator or a custom renderer could access to data to make histogram based calculation

packages/tools/src/types/IToolGroup.ts
```diff
@@ -19,8 +20,13 @@ export default interface IToolGroup {
   getViewportsInfo: () => Array<Types.IViewportId>;
   /** Get the toolInstance of the toolName */
   getToolInstance: { (toolName: string): any };
-  /** Add a tool to toolGroup with its configuration */
-  addTool: { (toolName: string, toolConfiguration?: any): void };
+  /** Add a tool to toolGroup with its configuration and custom calculator if wanted */
+  addTool: {
+    (
+      toolName: string,
+      toolConfiguration?: Record<any, any> & { statsCalculator?: Calculator }
```
sedghi, line 27: This should be ToolConfiguration

packages/tools/src/tools/annotation/EllipticalROITool.ts
```diff
@@ -1116,57 +1092,29 @@ class EllipticalROITool extends AnnotationTool {
...
-
-          stdDev += valueMinusMean * valueMinusMean;
-        };
-
-        pointInShapeCallback(
+        const pointsInShape = pointInShapeCallback(
           imageData,
           (pointLPS, pointIJK) => pointInEllipse(ellipseObj, pointLPS),
-          stdCalculator,
+          this.configuration.statsCalculator.run,
```
sedghi, line 1104: ok, I guess my bad, `.run` reads too werid, let's have your callback back. Sorry about that

sedghi (commented):
one last review comments, thanks

packages/tools/src/types/IToolGroup.ts
Celian-abd: Oh true , I forgot about it. I'll update it now

packages/tools/src/tools/annotation/EllipticalROITool.ts
Celian-abd: Sure, no worries

sedghi (approved):

PR #725, opened 2023-08-03
Fix incorrect area calculation in PlanarFreehandRoiTool

kevinshen1101:
This PR fixes a bug with the area calculation in the PlanarFreehandROI tool. Previously, we would input `Point2` canvas coordinates into the `calculateAreaOfPoints` method, which does not yield mm^2 as the output units. By converting first to image coordinates and utilizing the `PixelSpacing` metadata from the viewports image, we can then have `calculateAreaOfPoints` output mm^2 correctly.

Tested via `yarn example stackAnnotationTools` and comparing freehand-drawn ROI area stats to other ROI tools.


packages/tools/src/tools/annotation/PlanarFreehandROITool.ts
```diff
@@ -743,11 +744,16 @@ class PlanarFreehandROITool extends AnnotationTool {
         continue;
       }
 
-      const { imageData, metadata } = image;
-      const canvasCoordinates = points.map((p) => viewport.worldToCanvas(p));
+      const { imageData, metadata, hasPixelSpacing } = image;
+      const imageId = viewport.getCurrentImageId();
```
sedghi, line 748: this would not work for recontructed planes (non-acquisition) since they don't have currentImageId

packages/tools/src/tools/annotation/PlanarFreehandROITool.ts
```diff
@@ -743,11 +744,16 @@ class PlanarFreehandROITool extends AnnotationTool {
...
+      const imageCoordinates = points.map((p) => csUtils.worldToImageCoords(imageId, p));
```
sedghi, line 749: is points in world? why do we need to go to imageCoords?

kevinshen1101: The polyline `calculateAreaOfPoints` method only accepts point2 as input

packages/tools/src/tools/annotation/PlanarFreehandROITool.ts
kevinshen1101: Ah I see, I can change it to use canvas coordinates in that case

packages/tools/src/tools/annotation/PlanarFreehandROITool.ts
```diff
@@ -745,9 +747,25 @@ class PlanarFreehandROITool extends AnnotationTool {
 
       const { imageData, metadata } = image;
       const canvasCoordinates = points.map((p) => viewport.worldToCanvas(p));
+
+      const canvasPoint = canvasCoordinates[0];
+      const deltaX = viewport.canvasToWorld([
+        canvasPoint[0] + 1,
```
sedghi, line 753: why there is a +1?

kevinshen1101: The aim is to convert a step in the x direction to mm. We can pick any arbitrary point as our origin, take one step in the x and one step in the y and use those to calculate the actual mm spacing in the x and y canvas axes.

sedghi: I see, then the code benefits from this comment above, can you add it on top of the code?

packages/tools/src/tools/annotation/PlanarFreehandROITool.ts
```diff
@@ -752,9 +751,27 @@ class PlanarFreehandROITool extends AnnotationTool {
...
       const { imageData, metadata } = image;
       const canvasCoordinates = points.map((p) => viewport.worldToCanvas(p));
+
+      // Using an arbitrary start point (canvasPoint), calculate the
+      // mm spacing for the canvas in the X and Y directions.
+      const canvasPoint = canvasCoordinates[0];
+      const deltaX = viewport.canvasToWorld([
+        canvasPoint[0] + 1,
+        canvasPoint[1],
+      ]);
```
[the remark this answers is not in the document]
kevinshen1101: It's not really a deltaX per se. I can rename it to something like stepX or unitX. We're just looking to calculate the spacing on the canvas.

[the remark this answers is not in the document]
sedghi: I think the correct computation is the following, can you confirm?

```js
const originalWorldPoint = viewport.canvasToWorld(canvasPoint);
const deltaXPoint = viewport.canvasToWorld([
  canvasPoint[0] + 1,
  canvasPoint[1],
]);
const deltaYPoint = viewport.canvasToWorld([
  canvasPoint[0],
  canvasPoint[1] + 1,
]);

const deltaInX = vec3.distance(originalWorldPoint, deltaXPoint);
const deltaInY = vec3.distance(originalWorldPoint, deltaYPoint);

```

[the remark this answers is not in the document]
kevinshen1101: Yes this is correct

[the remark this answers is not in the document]
sedghi: Then can you update and push?

sedghi (approved):
Thanks for this PR

PR #726, opened 2023-08-03
fix: Prefetch that is actually position aware and multi-viewport capable

wayfarer3130:
### Context
The older stackPrefetch tool ended up just prefetching the entire series, marking all prefetched images as actively accessed, so that when the data didnt' fit, it would over-write the cache with images not near the currently active image.  This version is context sensitive to the position, direction of travel, image size and cache size.

The behaviour is: (rest is notes from the code):

 1. On navigating to a new image initially, or one that is at a different position:
  * Fetch the next/previous 2 images
 2. If the user is navigating forward/backward by less than 5 images, then
  * Prefetch additional images in the direction of navigation, up to 100
 3. If all the images in a given prefetch have completed, then:
  * Use the last prefetched image size as an image size for the stack
  * Fetch up to 1/4 of the cache size images near the current image

 This is designed to:
   * Get nearby images immediately so that they are available for navigation
     * Under the assumption that users might click and view an image, then
       navigate to next/previous image to see the exact image they want
   * Not interfere with loading other viewports if they are still loading
     * Load priority is prefetch, and minimal images are requested initially
   * Load an entire series if it will fit in memory
     * Allows navigating to other parts of the series and display images immediately
   * Have images available for CINE/navigation in one direction even when
     there is more image data than will fit in memory.
     * Up to 100 images in the direction of travel will be prefetched

### Changes & Results

Change the computation for which images to prefetch by setting a start/end prefetch range based on:
1. +/- 2 images initially and on navigation to completely different parts of the stack
2. Images in the direction of travel (eg +N or -N for forward or backward, making N larger as the user navigates in that same direction)
3. images around the image navigated to, of size such that it fills 1/4 of the cache, assuming equal size images

### Testing

Use the visible light human image dataset from IDC to test very large image datasets.  This dataset allows for 125 images in the cache roughly.  Play CINE in either direction, at about 10 frames per second for me to allow prefetch (on existing setup this would have gone blank - I think it will now skip images, showing some images)
Note how the prefetch extends further out as the CINE progresses - try at 1 fps to see this easily

Navigate in this series manually, note how it fetches +/- 2, then fetches nearby images in proportion to where the image click happens (so that it still fetches the same number at the ends)

Load a medium size CT study, go to 2x2 layout.  Wait for prefetch to complete (you can check console logs for details, one log per viewport at the end of prefetch).  Click into the middle of the series - should not load any new images.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  


IbrahimCSAE (approved):
I tested this locally in OHIF and it works as expected 👍

wayfarer3130:
@sedghi - any ideas on the build failure?  Nothing obvious when I looked at the logs.

sedghi:
Not working as expected. See video below

Right now:
- It correctly sends couple of requests per viewports on the grid
- It does send one request upon scroll

expected behaviour
- Upon viewport click it should send out the rest of requests for that viewport all together and not one by one

https://github.com/cornerstonejs/cornerstone3D/assets/7490180/128b8aa4-a926-4f5d-a17a-9cacdb59bc7f

packages/tools/src/utilities/stackPrefetch/stackPrefetch.ts
```diff
@@ -257,11 +257,11 @@ const updateToolState = (element) => {
 
   const minIndex = Math.max(
     0,
-    stack.currentImageIdIndex - configuration.minBefore
+    stack.currentImageIdIndex - (configuration.minBefore || 1)
   );
   const maxIndex = Math.min(
     stack.imageIds.length - 1,
-    stack.currentImageIdIndex + configuration.maxAfter
```
sedghi: how is it possible that the configuration is undefined? it is defined above. Why do mean || 5 and || 1 above

sedghi:
Still not fixed after your commit

packages/tools/src/utilities/stackPrefetch/stackPrefetch.ts
wayfarer3130: If the user overrides the configuration but doesn't provide the values, then the || 5 or || 1 provide default values so that the add/subtract aren't undefined.

wayfarer3130:
> Not working as expected. See video below
> 
> Right now:
> 
> * It correctly sends couple of requests per viewports on the grid
> * It does send one request upon scroll
> 
> expected behaviour
> 
> * Upon viewport click it should send out the rest of requests for that viewport all together and not one by one
> 

That is actually working as expected.  The behaviour is:
* When the viewport image has been loaded at the given position (using request priority 'interaction'), THEN start prefetching
* Prefetch only current position - minBefore to current position + maxAfter
* Ignore any viewport selected/not selected (eg any clicks on the viewport).

That behaviour gives exactly what you see 
* on navigating by 1 position, only 1 additional image is eligible for prefetch, since everything else is already in cache.
* on clicking to activate a viewport, nothing additional is fetched
* on opening to 2x2 layout, it fetches all 4 immediate images in view, then fetches about 5 additional images per viewport (20 extra), for a total of 24 images

Now, the code is relatively easy to extend in various ways, so I could do any/all of:
1. Choose how many images to read for the current viewport based on the current image size
2. Choose how much of the cache space to use based on the number of viewports enabled
3. Allocate additional space to the currently active viewport (would need to get a notification on viewport being active)
4. Perform additional fetches to continue filling cache once the first set is complete
5. Automatically choose how many prefetches to run based on how many viewports are open
6. Allocate extra prefetches to the active viewport
7. Update the cache to separate out volume cache size from stack cache max size
8. Only count stack caches towards the number of relevant image sets

All of these would provide more responsiveness to the cache - I'm not sure which ones are optimal.

sedghi:
@wayfarer3130 it is not
If you see the video at 20 seconds, when I'm on the top right (PT viewport). Upon scroll it is sending out one by one requests which doesn't makes sense since it should have already sent out all those before hand. The behaviour we were talking was 
regarding overwriting the images that are in the cache. It was about we should not allow the overwrite to happen IF the current viewports in the grid are on that imageId/Index.

wayfarer3130:
@sedghi - I think this is close to what we agreed on - if it is good enough, then I'd like to proceed with it, and then if there are more changes afterwards, then we can open new issues to deal with them.

sedghi:
@wayfarer3130 I thought the conclusion was to keep this one simple and add the advanced one to OHIF since it requires more metadata

wayfarer3130:
> @wayfarer3130 I thought the conclusion was to keep this one simple and add the advanced one to OHIF since it requires more metadata

Yes, I've kept this one fairly simple - it needed a bit more complexity to fix basic CINE display, but otherwise it is quite simple.  Fetch up to 6 images at a time, the only complexity being which 6 gets fetched.

sedghi:
let me know if it is ready for review

tsconfig.base.json
```diff
@@ -15,7 +15,7 @@
     "forceConsistentCasingInFileNames": true,
     "useDefineForClassFields": false,
     "jsx": "react-jsx",
-    "lib": ["ES6", "dom"], // "ES6", "DOM" | "es2015", "dom"
+    "lib": ["ESNext", "dom"], // "ES6", "DOM" | "es2015", "dom"
```
sedghi, line 18: we should change this in the beta branch not here

packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -2236,6 +2238,8 @@ class StackViewport extends Viewport implements IStackViewport {
 
     const imageAlreadyLoaded = cache.isImageIdCached(targetImageId);
 
+    this.viewportStatus = ViewportStatus.PRE_RENDER;
+
```
sedghi, line 2242: why we are setting PRE_RENDER twice here

packages/tools/src/utilities/cine/playClip.ts
```diff
@@ -330,7 +333,16 @@ function _createStackViewportCinePlayContext(
       // It is always in acquired orientation
       return true;
     },
+    tries: 0,
```
sedghi, line 336: can we extract the CINE changes into a separate PR and dedicate this PR only to stackPrefetch?

packages/core/src/cache/cache.ts
```diff
@@ -15,34 +15,40 @@ import { triggerEvent, imageIdToURI } from '../utilities';
...
 import Events from '../enums/Events';
 
-const MAX_CACHE_SIZE_1GB = 1073741824;
+const ONE_GB = 1073741824;
 
+/**
+ * Stores images, volumes and geometry.
+ * There are two sizes - the max cache size, that controls the overal maximum
+ * size, and the instance size, which controls how big any single object can
+ * be.  Defaults are 3 GB and 2 GB - 8 bytes (just enough to allow allocating it
```
sedghi, line 24: nice doc

packages/tools/src/utilities/stackPrefetch/index.ts
```diff
@@ -5,4 +5,8 @@ import {
   getConfiguration,
 } from './stackPrefetch';
 
-export { enable, disable, setConfiguration, getConfiguration };
+import * as stackContextPrefetch from './stackContextPrefetch';
+
+const stackPrefetch = { enable, disable, setConfiguration, getConfiguration };
```
sedghi, line 10: can't we do the same `import * as ` for the basic stackPrefetch?

packages/tools/src/utilities/stackPrefetch/stackContextPrefetch.ts
```diff
@@ -0,0 +1,386 @@
...
+  const imageIdSet = new Set<string>();
+  imageIds.forEach((imageId) => imageIdSet.add(imageId));
+  return (requestDetails) =>
+    requestDetails.type !== requestType ||
+    !imageIdSet.has(requestDetails.additionalDetails.imageId);
+};
+
+function prefetch(element) {
+  const stack = getStackData(element);
+  if (!stack || !stack.imageIds || stack.imageIds.length === 0) {
```
sedghi, line 83: !stack?.imageIds?.length

packages/tools/src/utilities/stackPrefetch/stackContextPrefetch.ts
```diff
@@ -0,0 +1,386 @@
...
+
+  if (!stackPrefetchData) {
+    return;
+  }
+
+  const stackPrefetch = stackPrefetchData || {};
+
+  // If all the requests are complete, disable the stackPrefetch tool
+  if (
+    !stackPrefetch.indicesToRequest ||
```
sedghi, line 99: ditto

packages/tools/src/utilities/stackPrefetch/stackContextPrefetch.ts
```diff
@@ -0,0 +1,386 @@
...
+  return {
+    currentImageIdIndex: viewport.getCurrentImageIdIndex(),
+    imageIds: viewport.getImageIds(),
+  };
+}
+
+const clearFromImageIds = (stack) => {
+  const { imageIds } = stack;
+  const imageIdSet = new Set<string>();
+  imageIds.forEach((imageId) => imageIdSet.add(imageId));
```
sedghi, line 75: we prob can do `const imageIdSet = new Set(stack.imageIds);`

packages/tools/src/utilities/stackPrefetch/stackContextPrefetch.ts
```diff
@@ -0,0 +1,386 @@
...
+      'stackPrefetch: element must be a StackViewport, VolumeViewport stackPrefetch not yet implemented'
+    );
+  }
+
+  return {
+    currentImageIdIndex: viewport.getCurrentImageIdIndex(),
+    imageIds: viewport.getImageIds(),
+  };
+}
+
```
sedghi, line 71: this is shared between two prefetchers, can we refactor in to stackprefetch/utils?

packages/tools/src/utilities/stackPrefetch/stackContextPrefetch.ts
```diff
@@ -0,0 +1,386 @@
...
+      return;
+    }
+
+    const distance = Math.abs(currentImageIdIndex - imageIdIndex);
+    // Use getter for nearby objects to ensure the recent access is updated
+    // and is cached for distant objects, to let them be decached if necessary
+    const imageCached =
+      distance < 6
+        ? cache.getImageLoadObject(imageId)
+        : cache.isImageIdCached(imageId);
```
sedghi, line 139: I looked at isImageIdCached in the cache and it is basically isLoaded, we should rename it 

```js
public isImageIdCached(imageId: string): boolean {
    const cachedImage = this._imageCache.get(imageId);

    if (!cachedImage) {
      return false;
    }

    return cachedImage.loaded;
  }
```

packages/tools/src/utilities/stackPrefetch/stackContextPrefetch.ts
```diff
@@ -0,0 +1,386 @@
...
+    const imageId = stack.imageIds[imageIdIndex];
+
+    if (!imageId) {
+      return;
+    }
+
+    const distance = Math.abs(currentImageIdIndex - imageIdIndex);
+    // Use getter for nearby objects to ensure the recent access is updated
+    // and is cached for distant objects, to let them be decached if necessary
+    const imageCached =
```
sedghi, line 136: I'm not sure if I follow this, you are saying if distance is less than 6 check if the image is requested to be loaded (since that is basically the getImageLoadObject, it does not gaurantee the pixel load, since it is a promise), but if the distance is more than 6 check if it is pixel loaded. right? why? can you add this clarfication and a bit more to the comment above?

packages/tools/src/utilities/stackPrefetch/stackContextPrefetch.ts
```diff
@@ -0,0 +1,386 @@
...
+
+const updateToolState = (element) => {
+  const stack = getStackData(element);
+  if (!stack || !stack.imageIds || stack.imageIds.length === 0) {
+    console.warn('CornerstoneTools.stackPrefetch: No images in stack.');
+    return;
+  }
+
+  const { currentImageIdIndex } = stack;
+  const { maxAfter = 5, minBefore = 1 } = configuration;
```
sedghi, line 268: there is a default in the top of the page, so I don't think we need to default again

packages/tools/src/utilities/stackPrefetch/stackContextPrefetch.ts
sedghi: can you update the doc based on what we discussed

packages/tools/src/utilities/stackPrefetch/stackContextPrefetch.ts
wayfarer3130: Configuration can be supplied not containing the defaults, and I like the actual default configuration to have it, so both are required.

sedghi:
Still not sure if the behaviour is expected, 

The very first image after scrolling down is only 1 request, then it goes 5 and 5 after that

Another issue is in detecting the direction, I thought it is smart and detects we are moving up so it should fire 5 previous imageIds. no?


https://github.com/cornerstonejs/cornerstone3D/assets/7490180/96aa1dc5-7094-4476-b568-c4a8a88c7e10

wayfarer3130:
@sedghi  - the logic for the stackContextPrefetch is now quite a bit cleaner/simpler, but it actually handles more use cases, as documented in the enable method for stackContextPrefetch.  I think at this point it handles all the cases that the original stackPrefetch handled, plus the directional CINE ones, plus stack size aware initial prefetch, so I would vote to deprecate the original one entirely and replace it with this one.

sedghi:
@wayfarer3130 did you fix this? https://github.com/cornerstonejs/cornerstone3D/pull/726#issuecomment-1714880321

wayfarer3130:
> @wayfarer3130 did you fix this? [#726 (comment)](https://github.com/cornerstonejs/cornerstone3D/pull/726#issuecomment-1714880321)

Yes - so the logic on initial display OR on new display at a completely different position is:
1. Fetch the nearby images - +/- 2 images 
2. Once the +/- 2 images is fetched, computed how much the last image fetched takes space wise, assume every image is like that and cache around the current image according to the position +/-  1/4 of the total cacheable space

wayfarer3130:
I added comparison code to OHIF and this code change to allow comparing the performance
Reports average load and decode times, as well as total times.  Over a number of loads, it appears that the performance for displaying the first image display time is about 5% faster using the new code - probably due to having slightly better caching order, but it might have just been random too, as there was a lot of variability in the responses.

packages/tools/src/utilities/stackPrefetch/stackContextPrefetch.ts
```diff
@@ -0,0 +1,373 @@
...
+        if (!stackPrefetch.cacheFill) {
+          stats.initialTime = Date.now() - stats.start;
+          stats.initialSize = stats.imageIds.size;
+          updateToolState(element, usage);
+          prefetch(element);
+        } else if (stats.imageIds.size) {
+          stats.fillTime = Date.now() - stats.start;
+          const { size } = stats.imageIds;
+          stats.fillSize = size;
+          console.log(
```
wayfarer3130, line 138: @sedghi - I'd like to leave this here for now as we don't have a good way to report the statistics, but eventually I'd like to move it into a unified event on the element that reports stats for both volume and stack prefetch.  It is certainly important to have them for evaluating this issue.

sedghi:
So still confused really here. Based what I tried, the 2x2 would cache all of the images (all) but the explanation in your comment [here](https://github.com/cornerstonejs/cornerstone3D/pull/726#issuecomment-1737292373) was something else. I just noticed that the `stackContextPrefetch` does not have any comment explaining what it is and what it does on top of the file. Can you add that?

https://github.com/cornerstonejs/cornerstone3D/assets/7490180/3b69b58c-e7cf-4ad1-89ce-8cba5f0a08cf

sedghi (approved):
Just putting here some more experiments which shows the value of this PR

Putting the cache size deliberately at 4MB it makes it go crazy and the old prefetcher for the 130 slices would request 260 network requests

https://github.com/cornerstonejs/cornerstone3D/assets/7490180/fafe467e-b67b-499a-8d15-70da04d7c7f6

However, using this PR

https://github.com/cornerstonejs/cornerstone3D/assets/7490180/6549b053-c936-4120-ba99-917fc34fc81e

PR #730, opened 2023-08-08
fix(color): Convert color space for useRGBA false

wayfarer3130:
Fixes #3354

### Context

The convertColorSpace function wasn't being called consistently or correclty for all situations.
According to the DICOM standard, as best I understand it:
1. The Photometric Interpretation (PMI) is used for uncompressed + RLE encodings
2. For colour images encoded in compressed syntaxes (JPEG, JPEG-LS, JPEG-LOSSLESS, JP2000, HTJ2K), the PMI and Planar Configuration are assumed to be whatever the decompressor provides - which is typically RGB, as the colour space is (usually) encoded in the image.  

This does NOT get 100% of images correct because there can be mismatches between what is encoded and decoded, however, it gets the majority of them correct, and it is really obvious when they are wrong.

### Changes & Results

Consistently use the RGB colour space as the output from compressed images, assuming the decoder takes care of conversion to RGB
Apply RGBA to RGB for RGBA decoders not applying colour space (that is, for JPEG)
Add colour space RGB with RGBA to RGB conversion included

### Testing

Run the test dataset from 
https://github.com/OHIF/Viewers/issues/3354#issuecomment-1669470931
Suggest using the dcm4che toolkit to convert some of these to various encodings
Then use static-dicomweb to store locally (including some unconverted instances)
**** Use a local/udpated version of mkdicomweb as a change was added to fix a bug in conversion

Make sure the images look right (see the original bug issue)

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  

### DIFF
--- a/packages/dicomImageLoader/src/imageLoader/createImage.ts
+++ b/packages/dicomImageLoader/src/imageLoader/createImage.ts
@@ -160,7 +160,6 @@ function createImage(
   return new Promise<DICOMLoaderIImage | ImageFrame>((resolve, reject) => {
     // eslint-disable-next-line complexity
     decodePromise.then(function (imageFrame: ImageFrame) {
-      console.log('decodePromise resolve', imageFrame);
       // if it is desired to skip creating image, return the imageFrame
       // after the decode. This might be useful for some applications
       // that only need the decoded pixel data and not the image object
--- a/packages/dicomImageLoader/src/imageLoader/createImage.ts
+++ b/packages/dicomImageLoader/src/imageLoader/createImage.ts
@@ -283,7 +283,18 @@ function createImage(
           imageFrame.pixelDataLength = imageFrame.pixelData.length;
         } else {
           console.log('Calling convert color space', imageFrame);
+          canvas.height = imageFrame.rows;
+          canvas.width = imageFrame.columns;
+          const context = canvas.getContext('2d');
+          const imageData = context.createImageData(
+            imageFrame.columns,
+            imageFrame.rows
+          );
+
           convertColorSpace(imageFrame, imageData.data, useRGBA);
+          imageFrame.imageData = imageData;
+          imageFrame.pixelData = imageData.data;
+          imageFrame.pixelDataLength = imageData.data.length;
         }
 
         /** @todo check as any */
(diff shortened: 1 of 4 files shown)


wayfarer3130:
The image below is against the original bug report:
image

packages/dicomImageLoader/src/imageLoader/createImage.ts
```diff
@@ -18,6 +18,15 @@ import { getOptions } from './internal/options';
 import isColorImageFn from '../shared/isColorImage';
 import isJPEGBaseline8BitColor from './isJPEGBaseline8BitColor';
 
+// Transfer syntaxes that have the photometric color apply
+// Other ones have the colour PMI applied by the decoder
```
sedghi, line 22: This can benefit from more details if possible

packages/dicomImageLoader/src/imageLoader/createImage.ts
```diff
@@ -214,27 +223,24 @@ function createImage(
...
+            imageFrame.columns,
+            imageFrame.rows
+          );
+
+          convertColorSpace(imageFrame, imageData.data, useRGBA);
+          imageFrame.imageData = imageData;
+          imageFrame.pixelData = imageData.data;
+          imageFrame.pixelDataLength = imageData.data.length;
+        } else if (
+          !useRGBA &&
```
sedghi, line 241: i'm surpised that this says `!useRGBA`, but then below in the logic of if statement we have remove A from RGBA

sedghi (commented):
Seems like it works with my color test data that I have. I have a US study that currently does not work in the `main` and this PR doesn't fix it either. Do you mind looking at it to see what is going on? It is YBR_FULL

Download from here: 
[5b5440d6-2e1a0df8-beca22f2-3de904ab-f51b5df7.zip](https://github.com/cornerstonejs/cornerstone3D/files/12302650/5b5440d6-2e1a0df8-beca22f2-3de904ab-f51b5df7.zip)


image

image

packages/dicomImageLoader/src/imageLoader/createImage.ts
wayfarer3130: I updated the comment below, but this case is when we NEED RGB (that is !useRGBA) but we have RGBA, so we need to convert.

sedghi:
interesting, thanks!

PR #736, opened 2023-08-11
Fix an off by 0.5/1 error in window level calculations

wayfarer3130:
### Context

The DICOM standard specifies that the window center is offset by 0.5 and the width to be the count of unique values to map, rather than being (upper+lower)/2 and upper-lower respectively.  This was done to allow for integer values for most
standard window levels for an even number of instances.

### Changes & Results

Change the values for the range.

### Testing

This is hard to test because the change is very subtle - very few images will change at all.  Suggest examining the TG-18 dataset (which is incorporated into the test dataset) for the mapping values being consistent.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  

### DIFF
--- a/packages/core/src/RenderingEngine/helpers/cpuFallback/rendering/getVOILut.ts
+++ b/packages/core/src/RenderingEngine/helpers/cpuFallback/rendering/getVOILut.ts
@@ -15,6 +15,10 @@
  */
 
 /**
+ * Generates the linear VOI LUT function.
+ * From the DICOM standard, that is:
+ * ((x - (c - 0.5)) / (w-1) + 0.5) * (ymax- ymin) + ymin
+ * clipped to the ymin...ymax range
  *
  * @param {Number} windowWidth Window Width
  * @param {Number} windowCenter Window Center
@@ -23,7 +27,10 @@
  */
 function generateLinearVOILUT(windowWidth: number, windowCenter: number) {
   return function (modalityLutValue) {
-    return ((modalityLutValue - windowCenter) / windowWidth + 0.5) * 255.0;
+    const value =
+      ((modalityLutValue - (windowCenter - 0.5)) / (windowWidth - 1) + 0.5) *
+      255.0;
+    return Math.min(Math.max(value, 0), 255);
   };
 }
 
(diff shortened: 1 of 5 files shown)


packages/core/src/RenderingEngine/helpers/cpuFallback/rendering/getVOILut.ts
```diff
@@ -15,6 +15,10 @@
  */
 
 /**
+ * Generates the linear VOI LUT function.
+ * From the DICOM standard, that is:
+ * ((x - (c - 0.5)) / (w-1) + 0.5) * (ymax- ymin) + ymin
+ * clipped to the ymin...ymax range
```
sedghi, line 21: can we have some links to the standard on why this is the formula? and how you are sure this is correct? since this is significant change that we didn't even have in v2 or cornerstone legacy

packages/core/src/utilities/windowLevel.ts
```diff
@@ -1,5 +1,8 @@
 /**
  * Given a low and high window level, return the window width and window center
+ * Formulas from note 4 in
+ * https://dicom.nema.org/medical/dicom/current/output/html/part03.html#sect_C.11.2.1.2.1
```
sedghi, line 4: oh I see you have the link here, thanks. add it there too.

packages/core/src/RenderingEngine/helpers/cpuFallback/rendering/getVOILut.ts
wayfarer3130: It is the definition in the standard - the other one can be made to work, but they tweaked the definition in the standard in order to have 2 things:
1. Make the width be the count of values rather than the absolute length
2. Make the center point of an even width (as in #1) be an integer value

That makes it easy to represent the 2^n bit lengths with an identity window level/width, as the number of points is 2^n and the center being 2^(n-1) as opposed to 2^n-1 and 2^(n-1)-0.5

Added the link.

wayfarer3130: Note that it is the same function, just that width and center are offset by 1 and 0.5 respectively.

sedghi (approved):
Thanks for this

PR #741, opened 2023-08-15
fix sorting priority groups

m00n620:
### Context

There's a bug when negative numbers are used for priorities in RequestPoolManager.

### Changes & Results

Fixed the sort compareFn to resolve the issue.


wayfarer3130 (approved):
This looks good to me.

PR #754, opened 2023-08-24
fix(4D utility): wrong array type returned by getDataInTime

lscoder:
Since `segmentationVolume` is an `Uint8Array` and the type returned by a `map` function is exaclty the same which made  `indexArray` array to be an `Uint8Array` but that type of array may not have values greater than 255 breaking all pixel indexes returned.

### DIFF
--- a/packages/tools/src/utilities/dynamicVolume/getDataInTime.ts
+++ b/packages/tools/src/utilities/dynamicVolume/getDataInTime.ts
@@ -40,7 +40,6 @@ function getDataInTime(
 
   if (options.maskVolumeId) {
     const segmentationVolume = cache.getVolume(options.maskVolumeId);
-    const startTime = performance.now();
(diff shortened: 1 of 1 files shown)


packages/tools/src/utilities/dynamicVolume/getDataInTime.ts
```diff
@@ -40,13 +40,20 @@ function getDataInTime(
...
+
+    // Get the index of every non-zero voxel in mask
+    for (let i = 0, len = segScalarData.length; i < len; i++) {
+      if (segScalarData[i] !== 0) {
+        indexArray.push(i);
+      }
+    }
+
+    // eslint-disable-next-line
+    console.log('>>>>> total time: ', performance.now() - startTime);
```
sedghi, line 55: can you remove performance?

sedghi (approved):

PR #755, opened 2023-08-25
feat(rendering): Correct the positioning for the viewport

wayfarer3130:
### Context

The sizing and positioning of the canvas was being computed incorrectly, causing the rendering to be in the wrong size/position.  

### Changes & Results

Added type/scale for pixel scaling mode to viewport to enable a scaling value to be used
Fixed thumbnail rendering so it renders in the correct aspect ratio when provided a default canvas
Fixed rendering so that it draws exactly what is provided/rendered from the vtk side
Fixed viewport sizing so it is a pixel exact sizing, to prevent half pixel renders

### Testing

In OHIF, use the hpScale hanging protocol to setup 1:1 display rendering
Take a screen capture of the TG-18 resolution 'LN' series 
Open in an image editor with replicate zoom
Zoom in on the image.  The little squares should have sharp edges

Alternatively - view the multi purpose TG-18 1k resolution image in the 1:1 scale mode (type=SCALE, scale=1)
The 5 horizontal/vertical single pixel line pairs should NOT show any moire or grayscale patterns:
image

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  


packages/core/src/RenderingEngine/RenderingEngine.ts
```diff
@@ -632,8 +632,15 @@ class RenderingEngine implements IRenderingEngine {
     const devicePixelRatio = window.devicePixelRatio || 1;
 
     const rect = canvas.getBoundingClientRect();
+    console.log('canvas rectangle', rect.width, rect.height, devicePixelRatio);
```
wayfarer3130, line 635: I will remove all the console logs before moving from WIP

packages/core/src/RenderingEngine/RenderingEngine.ts
```diff
@@ -1023,22 +1044,22 @@ class RenderingEngine implements IRenderingEngine {
...
   ): ViewportDisplayCoords {
     const { canvas } = viewport;
-    const { clientWidth, clientHeight } = canvas;
+    // const { clientWidth, clientHeight } = canvas;
     const devicePixelRatio = window.devicePixelRatio || 1;
-    const height = clientHeight * devicePixelRatio;
-    const width = clientWidth * devicePixelRatio;
+    // const sHeight = clientHeight * devicePixelRatio;
+    // const sWidth = clientWidth * devicePixelRatio;
+    const { width: sWidth, height: sHeight } = canvas;
```
wayfarer3130, line 1051: This is the primary fix for the size differences - using the actual width/height makes things actually work

packages/core/src/RenderingEngine/helpers/getOrCreateCanvas.ts
```diff
@@ -54,5 +55,24 @@ export default function getOrCreateCanvas(
   const internalDiv =
     element.querySelector(viewportElement) || createViewportElement(element);
 
-  return internalDiv.querySelector(canvasSelector) || createCanvas(internalDiv);
+  const canvas = (internalDiv.querySelector(canvasSelector) ||
+    createCanvas(internalDiv)) as HTMLCanvasElement;
+  const rect = internalDiv.getBoundingClientRect();
```
sedghi, line 60: The new additions require a lot of documentation of why it is needed, can you please add that?

packages/core/src/RenderingEngine/helpers/getOrCreateCanvas.ts
```diff
@@ -54,5 +55,24 @@ export default function getOrCreateCanvas(
...
+  const devicePixelRatio = window.devicePixelRatio || 1;
+  // The width and height will be floored by the canvas to produce integers
+  const left = Math.ceil(rect.x * devicePixelRatio) - rect.x * devicePixelRatio;
+  const top = Math.ceil(rect.y * devicePixelRatio) - rect.y * devicePixelRatio;
+  const width = Math.floor(rect.width * devicePixelRatio - left);
+  const height = Math.floor(rect.height * devicePixelRatio - top);
```
sedghi, line 66: why one is ceil and one is floor

packages/core/src/RenderingEngine/Viewport.ts
```diff
@@ -571,12 +571,48 @@ class Viewport implements IViewport {
...
+        viewportId: this.id,
+        displayArea: displayArea,
+        storeAsInitialCamera: storeAsInitialCamera,
+      };
+
+      triggerEvent(this.element, Events.DISPLAY_AREA_MODIFIED, eventDetail);
+    }
+  }
+
+  protected setDisplayAreaScale(displayArea: DisplayArea) {
```
sedghi, line 603: requires doc please

packages/core/src/RenderingEngine/Viewport.ts
```diff
@@ -571,12 +571,48 @@ class Viewport implements IViewport {
...
+    const canvas = this.canvas;
+    const height = canvas.height;
+    const width = canvas.width;
+    if (height < 8 || width < 8) {
+      return;
+    }
+    this.setCamera({ parallelScale: height / (2 * scale) });
+  }
+
+  protected setDisplayAreaFit(displayArea: DisplayArea) {
```
sedghi, line 614: fn doc please

packages/core/src/utilities/loadImageToCanvas.ts
```diff
@@ -57,6 +57,14 @@ export default function loadImageToCanvas(
 
       image.isPreScaled = image.isPreScaled || image.preScale?.scaled;
 
+      // If a dummy canvas, then setup the right size
+      if (canvas.width === 300 && canvas.height === 150) {
+        const width = Math.min(256, image.columns);
+        const height = (width * image.rows) / image.columns;
+        canvas.width = width / devicePixelRatio;
+        canvas.height = height / devicePixelRatio;
+      }
```
sedghi, line 66: can't you refactor the part in getOrCreateCanvas for this puprose?

packages/core/src/utilities/renderToCanvasGPU.ts
```diff
@@ -135,8 +133,10 @@ export default function renderToCanvasGPU(
     element.addEventListener(Events.IMAGE_RENDERED, onImageRendered);
     viewport.renderImageObject(image);
 
-    // force a reset camera to center the image
+    // force a reset camera to center the image and undo the small scaling
     viewport.resetCamera();
+    const parallelScale = viewport.getCamera().parallelScale / 1.04;
```
sedghi, line 138: why 1.04

packages/core/src/utilities/renderToCanvasGPU.ts
```diff
@@ -39,28 +39,29 @@ export default function renderToCanvasGPU(
   const viewportId = `renderGPUViewport-${imageIdToPrint}`;
   const imageId = image.imageId;
   const element = document.createElement('div');
-  element.style.width = `${canvas.width}px`;
-  element.style.height = `${canvas.height}px`;
+  const devicePixelRatio = window.devicePixelRatio || 1;
+  const originalWidth = canvas.width;
+  const originalHeight = canvas.height;
+  // Account for rounding errors by adding a half device pixel
```
sedghi, line 45: can you add more about what is rounding error and why half devide pixel ratio would help (what is the problem that it would even help)

sedghi (commented):
see my comments, I can take care of the tests since most of the comparison ones would fail

packages/core/src/utilities/loadImageToCanvas.ts
wayfarer3130: Because it isn't a canvas that getOrCreateCanvas ever creates - this is the final destination canvas, which might not match the created canvas used for rendering.  I've added a few parameters to control this, which I think makes it more obvious.

packages/core/src/utilities/renderToCanvasGPU.ts
wayfarer3130: Added a comment in the code.

sedghi (commented):
I'm trying to understand what is happening to the canvas, regarding when it is float size. I believe a sketch would help a lot something like this since there is a ceil and floor etc. Just assume an arbitrary misalignment between the div and pixels and show the finalized canvas, I guess it will go outside of the div then we need overflow hidden?

image

can you please draw what would be the drawn canvas if the div is not exactly located on the pixels?

wayfarer3130:
It is easier to show the positioning with the exact positions - both CSS and physical pixels:
Suppose we have a 3 pixel wide image - that is, the image pixels:
0,255,0

Now, when we go to render this, we have a div element to put our canvas into which has:
devicePixelRatio: 2
left CSS: 1.25 px
left physical: 2.5 px
width CSS: 3 px
width physical: 6 px

Now, we want to see how the image pixels get rendered to the viewport.
First, how much space do we have?
  Well, if we start with assuming we can display the entire width, then we get 6 physical pixels - BUT, the left and right edges each have a half physical pixel.
So, if we want to use physical pixels only, then we need to look at the CSS view for the nearest physical pixel on the left - that is:
ceiling(left edge in physical) - left edge in physical= ceiling (2.5) = 3 - 2.5 px = 0.5 px
that is, the left most edge loses 0.5 physical PX
Now, the width is reduced by 0.5 px, and we don't want to use any physical fractional pixels on the right, so we get:
width usable = floor(width physical - fractional left) = floor(6 - 0.5) = 5.5 physical px = 2.75 CSS px

wayfarer3130:
For the previous example, with 3 image pixels, suppose the mapping for the image pixels to canvas pixels is from the center of the image pixel, to the edge of the screen pixels - that is, the center of the canvas pixels are:
-0.5, 0.5, 1.5, 2.5, 3.5
That is, the 3 image pixels are smeared across 5 physical pixels - in this case, each physical pixel has the average value of both image pixels on either side, those being:

(0 * 0.5 + 0*0.5) = 0
(0 * 0.5 + 255*0.5) = 127.5
(255 * 0.5 + 0* 0.5) = 127.5
(0 * 0.5 + 0*0.5) = 0
That is, where we had a nice bright line down the center, we have doubled the size of it, and halved the intensity.
Consider what happens if this were a single pixel line denoting a fracting in a bone.  You suddenly get a line half the intensity, but quite a bit wider.  Maybe that is visible, and maybe not.  Certainly a lot harder to view.

This can happen in either of 2 places - the GPU rendering to the GPU canvas can do it.  This needs to ensure that the offscreen canvas is exact and that the computations for positioning etc are exact.

It can also happen if the onscreen canvas mismatches the offscreen canvas.

packages/core/src/RenderingEngine/helpers/getOrCreateCanvas.ts
```diff
@@ -54,5 +61,32 @@ export default function getOrCreateCanvas(
...
+  const devicePixelRatio = window.devicePixelRatio || 1;
+
+  // The left/top can be fractional physical pixels, round UP to the nearest
+  // physical pixel on the left/top hand edge.
+  const left = Math.ceil(rect.x * devicePixelRatio) - rect.x * devicePixelRatio;
+  const top = Math.ceil(rect.y * devicePixelRatio) - rect.y * devicePixelRatio;
+  // The width/height is the number of physical pixels which fit into the
+  // remaining space without any fractional pixels left over
+  // The floor is because we don't want any fractional pixels left over.
+  const width = Math.floor(rect.width * devicePixelRatio - left);
```
wayfarer3130, line 77: @sedghi  - as you indicated in the review, this calculation can leave a single pixel wide wrong colour.  I'm going to address your comment by having the viewable canvas just larger than the div, and cut it off so that one gets a fractional extra pixel rather than missing a fractional pixel.

wayfarer3130:
@sedghi  
- fixed a rendering issue with even sized viewports caused by rendering an image pixel to a physical pixel offset by a half pixel.
- fixed the issue you noticed during the in person review where the canvas size could be smaller than the div size causing the background to not be rendered for a single pixel.

sedghi:
I started reviewing this. I don't see much improvement for the tg18-cx-2k-01.dcm

Am i doing something wrong?
I'm dragging and dropping in the local example in cornerstone3D
Maybe the new changes you made has affected it?

## upstream (current behaviour)

CleanShot 2023-09-06 at 10 33 35




## You PR

CleanShot 2023-09-06 at 10 34 15

sedghi:
same also for image diff 

https://github.com/cornerstonejs/cornerstone3D/assets/7490180/4a87d65a-a572-45ab-ae8e-fed38e8cc0cb

sedghi:
I tried setting the displayedArea to 'SCALE' and tried your PR versus the main branch (with the setDisplayArea changes, NOT the initial radius changes) but don't see any difference again

packages/core/src/RenderingEngine/RenderingEngine.ts
```diff
@@ -1240,6 +1220,7 @@ class RenderingEngine implements IRenderingEngine {
 
     const onScreenContext = canvas.getContext('2d');
 
+    const rect = canvas.getBoundingClientRect();
```
sedghi, line 1223: is there any value in calling getBoundingClientRect here? I don't see it being used

packages/core/src/RenderingEngine/RenderingEngine.ts
```diff
@@ -1358,6 +1339,7 @@ class RenderingEngine implements IRenderingEngine {
 
       const onScreenContext = canvas.getContext('2d');
 
+      const rect = canvas.getBoundingClientRect();
```
sedghi, line 1342: is there any value in calling getBoundingClientRect here? I don't see it being used

packages/core/src/RenderingEngine/Viewport.ts
```diff
@@ -571,12 +571,114 @@ class Viewport implements IViewport {
...
+    const imageData = this.getDefaultImageData();
+    const spacingWorld = imageData.getSpacing();
+    const spacing = spacingWorld[1];
+    this.setCamera({ parallelScale: (height * spacing) / (2 * scale) });
+    // Need to ensure the focal point is aligned with the canvas size/position
+    // so that we don't get half pixel rendering, which causes additional
+    // moire patterns to be displayed.
+    // This is based on the canvas size having the center pixel be at a fractional
+    // position when the size is even, so matching a fractional position on the
+    // focal point to the center of an image pixel.
```
sedghi, line 631: nice comment

packages/core/src/RenderingEngine/Viewport.ts
```diff
@@ -705,33 +781,25 @@ class Viewport implements IViewport {
...
     const canvasAspectRatio = canvasSize[0] / canvasSize[1];
 
-    let radius;
+    let parallelScale;
 
-    if (boundsAspectRatio < canvasAspectRatio) {
-      // can fit full height, so use it.
-      radius = heightWorld / 2;
+    if (boundsAspectRatio <= canvasAspectRatio) {
+      // can fit full height, so use it minus 10%
```
sedghi, line 787: where is this minus 10% happening?

packages/core/src/RenderingEngine/helpers/getOrCreateCanvas.ts
```diff
@@ -54,5 +63,26 @@ export default function getOrCreateCanvas(
...
+
+  // The width/height is the number of physical pixels which will completely
+  // cover the div so that no pixels, fractional or full are left uncovered.
+  // Thus, it is the ceiling of the CSS size times the physical pixels.
+  // In theory, the physical pixels can be offset from CSS pixels, but in practice
+  // this hasn't been observed.
+  const width = Math.ceil(rect.width * devicePixelRatio);
+  const height = Math.ceil(rect.height * devicePixelRatio);
+  canvas.width = width;
+  canvas.height = height;
```
sedghi, line 80: I expected some left and top manipulation as per our discussion, howcome the width and height ceiling handles making the green canvas below

CleanShot 2023-09-11 at 22 19 24

packages/core/src/utilities/renderToCanvasGPU.ts
```diff
@@ -39,28 +39,33 @@ export default function renderToCanvasGPU(
...
-  element.style.height = `${canvas.height}px`;
+  const devicePixelRatio = window.devicePixelRatio || 1;
+  const originalWidth = canvas.width;
+  const originalHeight = canvas.height;
+  // The canvas width/height are set by flooring the CSS size converted
+  // into physical pixels, but because these are float values, the conversion
+  // isn't exact, and using the exact value sometimes leads to an off by 1
+  // in the actual size, so adding a half physical pixel to the size resolves
+  // the problem.
+  element.style.width = `${originalWidth + devicePixelRatio / 2}px`;
```
sedghi, line 50: why this is different than the one in `getOrCreateCanvas`, I know one is element and the other is canvas but this one is doign `devicePixelRatio/2` 

```js
canvas.style.width = (width + 0.0001) / devicePixelRatio + 'px';
canvas.style.height = (height + 0.0001) / devicePixelRatio + 'px';
```

wayfarer3130: Should I change it to + 0.0001 as well?  Anything smaller than the device pixel ratio and larger than the next fractional size should be ok.

sedghi: why one of them has devided by two and other does not? Shouldn't it be consistent?

packages/core/src/RenderingEngine/helpers/getOrCreateCanvas.ts
wayfarer3130: That is because of your comment about leaving a tiny bit of unfilled background.  What I did was set it up so that there is just more than enough canvas pixels to cover the physical pixels.  That ensures the canvas is completely filled, with potentially a fractional pixel not displayed, but all physical pixels covered.

sedghi:
I was updating visual tests and capturing new ones to be compared to and saw an unusual behaviour 

Original

image

Flipped Horizontally

image

Flipped Vertically

image

sedghi:
Also, it seems like the volume rendering tests show this incorrect pattern of a half bar. (I'm quite familiar with this bug, and it is related to the focal point. It has caused me a lot of trouble during the development of cornerstone3D.)

In main

image

In new PR:

CleanShot 2023-10-23 at 10 03 34

sedghi:
@Max-Kharitonov There are some issues regarding when the image is flipped and I didn't have time to look at it

packages/core/src/RenderingEngine/CanvasActor/index.ts
```diff
@@ -209,7 +209,8 @@ export default class CanvasActor implements ICanvasActor {
     const imageData = this.viewport.getImageData();
     Object.assign(this.image, {
       worldToIndex: (worldPos) => imageData.imageData.worldToIndex(worldPos),
-      indexToWorld: (index) => imageData.imageData.indexToWorld(index),
+      indexToWorld: (index, destPoint) =>
```
wayfarer3130, line 212: There are calls to indexToWorld assuming a second destination parameter, so make it consistent.

packages/core/src/RenderingEngine/VideoViewport.ts
```diff
@@ -870,13 +873,13 @@ class VideoViewport extends Viewport implements IVideoViewport {
       canvasPos[1] - panOffsetCanvas[1],
     ];
 
-    const worldPos: Point3 = [
-      subCanvasPos[0] / worldToCanvasRatio,
-      subCanvasPos[1] / worldToCanvasRatio,
+    destPos.splice(
```
sedghi, line 876: i'm not sure what is happening here

packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -1201,8 +1206,10 @@ class StackViewport extends Viewport implements IStackViewport, IImagesLoader {
   }
 
   private setRotationGPU(rotation: number): void {
+    const panFit = this.getPan(this.fitToCanvasCamera);
```
sedghi, line 1209: can you please add a comment on why all this pan magic is required for rotation?

packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -2881,6 +2893,27 @@ class StackViewport extends Viewport implements IStackViewport, IImagesLoader {
     };
   }
 
+  /**
+   * Applies the view reference, which may navigate the slice index and apply
+   * other camera modifications
+   */
+  public setView(viewRef?: ViewReference, viewPres?: ViewPresentation): void {
```
sedghi, line 2900: can we rename to setViewPresentation, since we have getViewPresentation

packages/core/src/types/IViewport.ts
```diff
@@ -105,6 +106,61 @@ export type ViewReference = {
...
+  slabThickness?: boolean;
+  // Camera relative parameters
+  rotation?: boolean;
+  displayArea?: boolean;
+  zoom?: boolean;
+  pan?: boolean;
+  // Transfer function relative parameters
+  windowLevel?: boolean;
+  paletteLut?: boolean;
+};
```
sedghi, line 162: can you please document this, what is a viewPresentation selector and how is it different from viewPresentation, then each item requires doc please

packages/core/src/RenderingEngine/Viewport.ts
```diff
@@ -51,6 +55,18 @@ import IImageCalibration from '../types/IImageCalibration';
...
  */
 class Viewport implements IViewport {
+  static readonly CameraViewPresentation: ViewPresentationSelector = {
+    rotation: true,
+    pan: true,
+    zoom: true,
+    displayArea: true,
+  };
+
+  static readonly TransferViewPresentation: ViewPresentationSelector = {
```
sedghi, line 65: requires doc on what is TransferViewPresentation, and how is it different from CameraViewPresentation

packages/core/src/RenderingEngine/Viewport.ts
```diff
@@ -1437,6 +1577,58 @@ class Viewport implements IViewport {
     return true;
   }
 
+  /**
+   * Gets a basic view reference.
+   */
+  public getViewPresentation(
```
sedghi, line 1583: more docs please, specially on the selector

packages/core/src/RenderingEngine/Viewport.ts
```diff
@@ -1437,6 +1577,58 @@ class Viewport implements IViewport {
...
+    const initZoom = this.getZoom();
+
+    if (zoom) {
+      target.zoom = initZoom;
+    }
+    if (pan) {
+      target.pan = this.getPan();
+      vec2.scale(target.pan, target.pan, 1 / initZoom);
+    }
+    return target;
```
sedghi, line 1609: i'm not sure why we don't return all of them? and at the destination we use the selector too choose speicifc one? aren't these pure functoins? they are all get... (getRotation, getDisplayArea) etc

sedghi (commented):
almost there, please see my comments

packages/core/src/RenderingEngine/Viewport.ts
sedghi: I don't see getting used anywhere

packages/core/src/RenderingEngine/VideoViewport.ts
wayfarer3130: The first two values are getting replaced to agree with the contract coming from vtk - which we use in 1 place.

packages/core/src/RenderingEngine/StackViewport.ts
wayfarer3130: It is a single combined method that can navigate as well as apply a view presentation.  It probably doesn't QUITE work perfectly yet without extra event notifications, but the idea is to only get one set of notifications.

sedghi (approved):

PR #761, opened 2023-08-30
Fix/cine black images

wayfarer3130:
### Context

If the image load time is sufficiently slow, then the frame will never be displayed before the next frame is requested, and the CINE viewport goes black.  This change ensures that every frame is displayed, not continuing to the next frame until a frame is displayed or the retries limit is hit.

### Changes & Results

Ensures that CINE plays every frame, waiting till it is available to be rendered.

### Testing

Hard to test without a very slow back end.  Play CINE against a slow/large dataset like the visible light human dataset.  Previously, the images would go black.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  


packages/tools/src/types/CINETypes.ts
```diff
@@ -1,12 +1,12 @@
-import { Types } from '@cornerstonejs/core';
-
 type PlayClipOptions = {
   framesPerSecond?: number;
   frameTimeVector?: number[];
   reverse?: boolean;
   loop?: boolean;
   dynamicCineEnabled?: boolean;
   frameTimeVectorSpeedMultiplier?: number;
+  // How many items to wait for.  Set to 0 to play immediately
```
sedghi, line 8: what is an `item` here? I don't follow what it does

sedghi:
is this for the case where framerate is slow or fast? I thought it is for fast

packages/tools/src/utilities/cine/playClip.ts
```diff
@@ -330,7 +333,16 @@ function _createStackViewportCinePlayContext(
       // It is always in acquired orientation
       return true;
     },
+    tries: 0,
     scroll(delta: number): void {
+      if (
+        this.tries <= waitForRendered &&
```
sedghi, line 339: waitForRenderedCount

packages/tools/src/types/CINETypes.ts
sedghi: for a 60 frames per seconds this is roughly 16 or 17 ms

sedghi: a typical value for this would be 30 or 40 (~half a second =:> 30 * 16 ms )

sedghi: the unit for this is frameNumber

sedghi (approved):

sedghi (approved):

PR #784, opened 2023-09-15
feat(referenceLines): showFullDimension option to ReferenceLines tool

m00n620:
### Context

At the moment reference lines are moving on other tool interactions(Zoom, Pan). I think we need a way to make the reference lines not moving and fit to the image.

### Changes & Results

Added `showFullDimension` option to `ReferenceLines` tool to change annotation renders.


sedghi (approved):
Great work thanks

PR #787, opened 2023-09-18
fix(Stack prefetch): should not remove other requests 

sedghi:
### Context

If the viewport Grid has a stack and a volume of the same displaySet then the volume would not load complete for the bug this PR fixes. I don't know why we set the clearRequests for the stack prefetch

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  

### DIFF
--- a/packages/tools/src/utilities/stackPrefetch/stackPrefetch.ts
+++ b/packages/tools/src/utilities/stackPrefetch/stackPrefetch.ts
@@ -16,7 +16,13 @@ const addToBeginning = true;
 
 let configuration = {
   maxImagesToPrefetch: Infinity,
-  preserveExistingPool: false,
+  // Preserving the existing pool should be the default behaviour, as there might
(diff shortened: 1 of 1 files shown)


packages/tools/src/utilities/stackPrefetch/stackPrefetch.ts
```diff
@@ -16,7 +16,13 @@ const addToBeginning = true;
 
 let configuration = {
   maxImagesToPrefetch: Infinity,
-  preserveExistingPool: false,
+  // Preserving the existing pool should be the default behaviour, as there might
```
wayfarer3130, line 19: The new position aware cache deals with this a lot better by being aware of how the prefetch gets allocated/setup.  This will cause some slow down when prefetching a large display set which isn't in view any longer, but for now, I think this is a reasonable change.

wayfarer3130 (approved):

PR #806, opened 2023-10-03
feat(Tools): add Eraser Tool

ramonemiliani93:
### Context

Creates an eraser tool that allows the deletion of annotations.

### Changes & Results

- Add new tool

What are the effects of this change?
- New tool available to delete annotations

### Testing
- There is an `EraserTool_test` file to run tests on the tool.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  


jbocce:
Thank you so much for this PR. It is a nice feature to get in.

A few overall things for this PR...

- since we are changing the public api here, please run `yarn build:update-api` and commit and push the changes; this will get the API_CHECK passing
- I think it would be best to add the eraser tool to the stack annotation tools example (i.e. https://www.cornerstonejs.org/live-examples/stackannotationtools); the code for it is https://github.com/cornerstonejs/cornerstone3D/blob/main/packages/tools/examples/stackAnnotationTools/index.ts

packages/tools/src/tools/EraserTool.ts
```diff
@@ -0,0 +1,78 @@
...
+    const annotationsToRemove = [];
+
+    for (const toolName in tools) {
+      const toolInstance = tools[toolName];
+
+      if (typeof toolInstance.isPointNearTool !== 'function') {
+        continue;
+      }
+
+      const annotations = state.getAnnotations(toolName, viewportId);
```
jbocce, line 50: I think here we need `element` instead of `viewportId`

packages/tools/test/EraserTool_test.js
```diff
@@ -0,0 +1,208 @@
...
+        128
+      );
+
+      this.DOMElements.push(element);
+
+      const imageId1 = 'fakeImageLoader:imageURI_64_64_10_5_1_1_0';
+      const vp = this.renderingEngine.getViewport(viewportId);
+
+      const addEventListenerForAnnotationRemoved = () => {
+        element.addEventListener(csToolsEvents.ANNOTATION_REMOVED, () => {
```
jbocce, line 120: I think here instead of adding the listener to `element`, `eventTarget` should be used.

jbocce:
@ramonemiliani93 it looks good so far. Perhaps make the suggested changes, push and then we can take this out of draft and go from there. Thanks.

ramonemiliani93:
Changes done @jbocce 👌 Let me know If I need to add/update something 💪

jbocce:
> Changes done @jbocce 👌 Let me know If I need to add/update something 💪

@ramonemiliani93 thank you for the updates. However the unit test you added is not passing. Does it pass for you?

Also I was hoping you would have added the eraser tool to the stack annotation tools example [as I had requested](https://github.com/cornerstonejs/cornerstone3D/pull/806#issuecomment-1768630705). 

Without these two items it is difficult to test your code and complete the PR. If you could please update accordingly then that would be very much appreciated.

ramonemiliani93:
Sorry, @jbocce! It seems like I did not push the changes for the stack annotation tools example. Regarding the test, I have yet to figure out why it is not capturing the event (partly the reason why I opened the PR in draft). If you could point me in the right direction on how to figure out why the `ANNOTATION_REMOVED` event is not being caught, I will fix it as soon as possible.

packages/tools/test/EraserTool_test.js
```diff
@@ -0,0 +1,208 @@
...
+          target: element,
+          buttons: 1,
+          clientX: clientX1,
+          clientY: clientY1,
+          pageX: pageX1,
+          pageY: pageY1,
+        });
+
+        addEventListenerForAnnotationRemoved();
+        document.dispatchEvent(evt);
```
jbocce, line 195: Dispatch the event on the viewport element itself. (i.e. `element.dispatchEvent(evt)`)

packages/tools/test/EraserTool_test.js
jbocce, line 1: The only two changes I needed to have the test pass are suggested below. Please give them a try.

packages/tools/examples/stackAnnotationTools/index.ts
```diff
@@ -118,6 +119,7 @@ const toolsNames = [
   CobbAngleTool.toolName,
   ArrowAnnotateTool.toolName,
   PlanarFreehandROITool.toolName,
+  EraserTool.toolName,
```
jbocce, line 122: Thanks for adding the eraser tool to the example. I tried it and was able to delete every annotation type except the probe. Could you please have a look at why this might be? Thanks.

jbocce (changes requested):
Please update the unit test and if it passes push your changes and remove the draft status of the PR so that we can move forward. Also be sure to determine why the probe tool cannot be erased. Thanks.

sedghi (approved):

PR #808, opened 2023-10-03
feat(surface rendering): Add surface rendering as segmentation representation

rodrigobasilio2022:
### Context

Adds Surface geometry support to stack, volume and volume3D viewports. 

See these examples 

- https://deploy-preview-808--cornerstone-3d-docs.netlify.app/live-examples/surfacerendering
- https://deploy-preview-808--cornerstone-3d-docs.netlify.app/live-examples/surfacerenderingstack

### Changes & Results

New Geometry is added called Surface for segmentations. We also added a new tool called SurfaceIntersection to show overlay for each surface as well. 

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  


packages/core/src/RenderingEngine/Viewport.ts
```diff
@@ -1159,6 +1163,37 @@ class Viewport implements IViewport {
         viewPlaneNormal,
         focalPoint
       );
+      if (actorEntry?.vtkPlanes && actorEntry?.outline) {
```
sedghi, line 1166: This is getting called on EVERY reset camera while it needs to only run for when it is needed (scroll), so in the setCamera in the viewport.ts you should have some logic similar to following that identify if it is an in-plane (pan, zoom) or out of plane and only when out of plane it modifies the planes

```
 if (oldFocalPoint && focalPoint) {
        const currentViewPlaneNormal = <Point3>vtkCamera.getViewPlaneNormal();
        const deltaCamera = <Point3>[
          focalPoint[0] - oldFocalPoint[0],
          focalPoint[1] - oldFocalPoint[1],
          focalPoint[2] - oldFocalPoint[2],
        ];

        const cameraModifiedInPlane =
          Math.abs(vtkMath.dot(deltaCamera, currentViewPlaneNormal)) < 1e-2;

        if (!cameraModifiedInPlane) {
          this.updateClippingPlanesForActors(updatedCamera);
          this.hasPerformedInitialClipping = true;
        }
      }
```

packages/core/src/RenderingEngine/Viewport.ts
```diff
@@ -1159,6 +1163,37 @@ class Viewport implements IViewport {
...
+            passPointData: false,
+          });
+          filter.setInputData(normals.getOutputData());
+          filter.setGenerateOutline(true);
+          filter.setGenerateFaces(false);
+          filter.update();
+          const filterData = filter.getOutputData();
+          mapper.setInputData(filterData);
+        } else {
+          const filter = vtkClipClosedSurface.newInstance({
```
sedghi, line 1184: Also this is getting called a lot, you don't need to `newInstance` it again, i think we can just assign it to the actor and create for the first time and read from it afterwards

sedghi: This also feels a bit weird that we have this logic here, ideally the viewport should trigger an event that say CLIPPING_PLANEES_CHANGED and we listen to that in the `displayTools/surface` and react to that by changing the surface planes

sedghi:
seems like we need to call reset camera clipping range AFTER the surface is added I beleive so we need to figure out when is that 

https://share.cleanshot.com/rYwKdRHq

sedghi (commented):
Much better now

I feel like we can set the planes if they were already computed. Since that closedSurfaceFilter is too slow

What I mean is if I'm slice 10 and I go to slice 11 it will be slow and that is because of the filter which I understand. Then if someone goes to 12 again it will be slow since it is the first time that we are in 12, BUT if we go back to 11, I think we should read from cache somewhere so that it will be fast.

rodrigobasilio2022:
> Much better now
> 
> I feel like we can set the planes if they were already computed. Since that closedSurfaceFilter is too slow
> 
> What I mean is if I'm slice 10 and I go to slice 11 it will be slow and that is because of the filter which I understand. Then if someone goes to 12 again it will be slow since it is the first time that we are in 12, BUT if we go back to 11, I think we should read from cache somewhere so that it will be fast.

I did that. Not sure if the cache is in the right place

packages/core/src/loaders/utils/contourSet/createContourSet.ts
```diff
@@ -14,6 +14,7 @@ export function createContourSet(
     id: contourSetData.id,
     data: contourSetData.data,
     color: contourSetData.color,
+    segmentIndex: 1,
```
sedghi, line 17: why?

packages/core/src/RenderingEngine/Viewport.ts
```diff
@@ -1133,9 +1150,11 @@ class Viewport implements IViewport {
    * Updates the actors clipping planes orientation from the camera properties
    * @param updatedCamera - ICamera
    */
-  protected updateClippingPlanesForActors(updatedCamera: ICamera): void {
+  protected async updateClippingPlanesForActors(
+    updatedCamera: ICamera
+  ): Promise<void> {
     const actorEntries = this.getActors();
-    actorEntries.forEach((actorEntry) => {
+    await actorEntries.forEach(async (actorEntry) => {
```
sedghi, line 1157: what is async here that you need an await?

packages/tools/src/tools/displayTools/utils/pointFunctions.ts
```diff
@@ -0,0 +1,42 @@
+/**
+ * Fast distance calculation between two points
+ * @param pointA
+ * @param pointB
+ * @returns
+ */
+export function fastPointDistance(pointA, pointB) {
```
sedghi, line 7: can we augment the function packages/tools/src/utilities/math/point/distanceToPoint.ts to handle Point3 and use that instead of this?

packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -2035,9 +2035,14 @@ class StackViewport extends Viewport implements IStackViewport {
...
-    const actors = [];
-    actors.push({ uid: this.id, actor });
-    this.setActors(actors);
+    const oldActors = this.getActors();
+    if (oldActors.length && oldActors[0].uid === this.id) {
+      oldActors[0].actor = actor;
+    } else {
+      oldActors.unshift({ uid: this.id, actor });
+    }
+    this.setActors(oldActors);
```
sedghi, line 2044: can you explain what is happening here?

packages/tools/src/tools/displayTools/utils/pointFunctions.ts
rodrigobasilio2022: yes

packages/core/src/RenderingEngine/Viewport.ts
rodrigobasilio2022: Because if we remove the await, the 3d volume render will not correctly call the renderer.resetCameraClippingRange();

packages/core/src/RenderingEngine/StackViewport.ts
rodrigobasilio2022: If one adds an actor in the stackViewport this part of code will remove it from the actors list, because old code creates a list with only the ImageSlice actor and sets it as the actors list. So what I am doing here is checking if the imageslice actor is in the actor list. If not adding as the first item of the list

packages/core/src/RenderingEngine/Viewport.ts
sedghi: But which part of the the inner functions is asynchronous? Do you know?

packages/core/src/RenderingEngine/StackViewport.ts
sedghi: I'm not following unfortunately. 

So this says that if the uid of the actor is the same as the viewportId (what does that even mean), then replace the actor at the first index? otherwise throw it out?

Why?

packages/core/src/RenderingEngine/Viewport.ts
rodrigobasilio2022: Yes I understand what you are saying... I really dont know.... Any of these functionas appears to be the one. I remember if I removed this await if will not render correctly the volumeViewport3D. Maybe you can do this test again...

packages/tools/src/tools/SegmentationIntersectionTool.ts
```diff
@@ -0,0 +1,283 @@
...
+      let actorWorldPointsMap = actorsWorldPointsMap.get(actorEntry.uid);
+      if (!actorWorldPointsMap) {
+        actorWorldPointsMap = new Map();
+        actorsWorldPointsMap.set(actorEntry.uid, actorWorldPointsMap);
+      }
+      if (!actorWorldPointsMap.get(focalPointString)) {
+        const polyData = actorEntry.clippingFilter.getOutputData();
+        let worldPointsSet = getPolyDataPoints(polyData);
+        if (worldPointsSet) {
+          worldPointsSet = removeExtraPoints(viewport, worldPointsSet);
```
sedghi, line 189: I commented this line out (to not remove extra points) and everything worked fine. Do you have a reason to include this?

sedghi (approved):
This is great work, thanks

salimkanoun:
Wow guys this is so impressive !

PR #815, opened 2023-10-06
fix(BaseStreamingImageVolume) Issue with typescript type

### LINKED ISSUE #807: [Bug] Cannot import cornerstoneStreamingImageVolumeLoader
### Describe the Bug

I'm trying to import **cornerstoneStreamingImageVolumeLoader** in my project and I get this error:
`
Error: node_modules/@cornerstonejs/streaming-image-volume-loader/dist/esm/StreamingImageVolume.d.ts:24:29 - error TS2307: Cannot find module 'packages/core/dist/esm/enums/RequestType' or its corresponding type declarations.

24         requestType: import("packages/core/dist/esm/enums/RequestType").default;
`

Versions:
    "@cornerstonejs/core": "^1.16.6",
    "@cornerstonejs/dicom-image-loader": "^1.16.6",
    "@cornerstonejs/streaming-image-volume-loader": "^1.16.6",
    "@cornerstonejs/tools": "^1.16.6",

### Steps to Reproduce

Import cornerstoneStreamingImageVolumeLoader with:
`import { cornerstoneStreamingImageVolumeLoader } from '@cornerstonejs/streaming-image-volume-loader';`

### The current behavior

Get an error
Seems related to the way it's declared (streaming-image-volume-loader.api.md):
`
import { default as default_2 } from 'packages/core/dist/esm/enums/RequestType';
`

### The expected behavior

Should work without error

### OS

Windows 11

### Node version

18.16.0

### Browser

NA

frolic06:
### Context

Fixes #807 : not able to import from streaming-image-volume-loader

### Changes & Results

Change the type of requestType from enum to string

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  


sedghi (commented):
I read about the bug your reported. and I'm surprised on the solution. Do you know what is happening? what are you using? vite? webpack?

frolic06:
I'm using angular, so webpack. It seems to me that the enum value 'Enums.RequestType.Prefetch' is not see as the enum type Enums.RequestType, although the enum is correctly exported. That's why I change the type to string. It's probably not the cleaner solution, nor really elegant, but it seems to work.
If you know a better way to fix this, I will change my PR.

frolic06:
requestType is already defined as a string here for example:
`interface ImageLoaderOptions {
    additionalDetails?: Record<string, unknown>;
    priority: number;
    requestType: string;
}`

frolic06:
@sedghi I can also change with, if you prefer:
`const requestType: any = Enums.RequestType.Prefetch;`

There must be another solution, but I haven't found it. This solution has the advantage of being simple and removing unnecessary enum exports.

sedghi:
what is the type declaration for other Enums that we have, are they strings?

frolic06:
I searched the code history and found that the bug was introduced by PR #438 . The difference is that we have a class that inherits from another class. This seems to be the source of the problem, because the signature of  one of the method is created from an inherited private method.

frolic06:
My solution is therefore to explicitly add an interface for the return of one of the methods

frolic06:
@sedghi is this ok for you ?

packages/streaming-image-volume-loader/src/ImageLoadRequests.ts
```diff
@@ -0,0 +1,30 @@
+import { Types, Enums } from '@cornerstonejs/core';
+
+export default interface ImageLoadRequests {
+  callLoadImage: (
+    imageId: any,
```
sedghi, line 5: string?

packages/streaming-image-volume-loader/src/ImageLoadRequests.ts
```diff
@@ -0,0 +1,30 @@
...
+    imageIdIndex: any,
```
sedghi, line 6: number?

packages/streaming-image-volume-loader/src/ImageLoadRequests.ts
```diff
@@ -0,0 +1,30 @@
...
+    options: any
+  ) => Promise<void>;
+  imageId: string;
+  imageIdIndex: number;
+  options: {
+    targetBuffer: {
+      arrayBuffer: SharedArrayBuffer;
+      offset: number;
+      length: number;
+      type: any;
```
sedghi, line 16: string?

packages/streaming-image-volume-loader/src/ImageLoadRequests.ts
```diff
@@ -0,0 +1,30 @@
...
+  callLoadImage: (
+    imageId: any,
+    imageIdIndex: any,
+    options: any
+  ) => Promise<void>;
+  imageId: string;
+  imageIdIndex: number;
+  options: {
+    targetBuffer: {
+      arrayBuffer: SharedArrayBuffer;
```
sedghi, line 13: I think if we don't use the sharedArrayBuffer this is undefine so it is SharedArrayBuffer | undefined

frolic06:
@sedghi fixed, anything else ?
Thanks for the review

sedghi (approved):

PR #828, opened 2023-10-13
feat(vide): Add new Video Viewport with zoom and pan

sedghi:
🎉 This work in progress (WIP) is the outcome of initial exceptional efforts by @JamesAPetts 

This is the prototype for a new Video Viewport, currently with zoom and pan using Cornerstone3D rendering Engine. Future plans is to add annotation and segmentation tools to this viewport and treat it exactly the same as other types of viewports.


jlopes90:
The example does not have the video viewport...

sedghi:
@jlopes90 I know and didn't want to spend time on making the loader for webpack. You can run `yarn run example videoTools`

packages/core/src/types/index.ts
```diff
@@ -83,13 +83,18 @@ import type { PixelDataTypedArray } from './PixelDataTypedArray';
 import type { ImagePixelModule } from './ImagePixelModule';
 import type { ImagePlaneModule } from './ImagePlaneModule';
 import type { AffineMatrix } from './AffineMatrix';
+import type VideoViewportProperties from './VideoViewportProperties';
+import type IVideoViewport from './IVideoViewport';
+
+type IVtkViewport = IStackViewport | IVolumeViewport;
```
sedghi, line 89: hmmm, our StackViewport CPU is not vtk. so not sure what to call this

utils/ExampleRunner/example-info.json
```diff
@@ -203,6 +203,16 @@
         "description": "Demonstrates how to render a 4D data into multiple viewports and fuse them"
       }
     },
+    "video" : {
```
sedghi, line 206: this right now doesn't appear, you should add it to the top as well

utils/ExampleRunner/template-config.js
```diff
@@ -18,7 +18,7 @@ module.exports = function buildConfig(
   root,
   exampleBasePath
 ) {
-  console.log('root=', root);
+  console.log('root=', root, 'destPath=', destPath);
```
sedghi, line 21: remove

utils/ExampleRunner/template-config.js
```diff
@@ -51,6 +51,11 @@ module.exports = {
             '../../../node_modules/@cornerstonejs/dicom-image-loader/dist/dynamic-import',
           to: '${destPath.replace(/\\/g, '/')}',
         },
+        {
+          from:
+            '../../../rendered.mp4',
+          to: '${destPath.replace(/\\/g, '/')}/video/',
```
sedghi, line 57: let's remove the local rendered.mp4 if the https one works

packages/tools/src/utilities/scroll.ts
```diff
@@ -19,7 +20,7 @@ import { ScrollOptions, EventTypes } from '../types';
  * @returns
  */
 export default function scroll(
-  viewport: Types.IStackViewport | Types.IVolumeViewport,
+  viewport: Types.IStackViewport | Types.IVolumeViewport | Types.IVideoViewport,
```
sedghi, line 23: IViewport?

packages/tools/examples/videoTools/index.ts
```diff
@@ -0,0 +1,228 @@
...
+rangeDiv.innerHTML =
+  '<div id="time" style="float:left;width:2.5em;">0 s</div><input id="range" style="width:400px;height:8px;float: left" value="0" type="range" /><div id="remaining">unknown</div>';
+content.appendChild(rangeDiv);
+const rangeElement = document.getElementById('range');
+rangeElement.onchange = () => {
+  viewport.setTime(Number(rangeElement.value));
+};
+rangeElement.oninput = () => {
+  viewport.setTime(Number(rangeElement.value));
+};
```
sedghi, line 57: we have a utiltity for this addSliderToToolbar

packages/tools/src/tools/annotation/VideoRedactionTool.ts
```diff
@@ -0,0 +1,935 @@
...
+import * as rectangle from '../../utilities/math/rectangle';
+import {
+  resetElementCursor,
+  hideElementCursor,
+} from '../../cursors/elementCursor';
+import triggerAnnotationRenderForViewportIds from '../../utilities/triggerAnnotationRenderForViewportIds';
+
+import { SVGDrawingHelper } from '../../types';
+import { StyleSpecifier } from '../../types/AnnotationStyle';
+import { vec3, vec2 } from 'gl-matrix';
```
sedghi, line 34: move to top

packages/tools/src/tools/annotation/VideoRedactionTool.ts
```diff
@@ -0,0 +1,935 @@
...
+import getWorldWidthAndHeightFromTwoPoints from '../../utilities/planar/getWorldWidthAndHeightFromTwoPoints';
+
+// interface VideoRedactionSpecificToolData extends ToolSpecificToolData {
```
sedghi, line 37: please fix this type

packages/tools/src/tools/annotation/VideoRedactionTool.ts
```diff
@@ -0,0 +1,935 @@
...
+    });
+
+    this._throttledCalculateCachedStats = throttle(
+      this._calculateCachedStats,
+      100,
+      { trailing: true }
+    );
+  }
+
+  // TODO_JAMES -> We are just adding this as a "global redaction for now".
```
sedghi, line 86: remove TODO_JAMES

packages/tools/src/tools/annotation/VideoRedactionTool.ts
```diff
@@ -0,0 +1,935 @@
...
+      data: {
+        invalidated: true,
+        handles: {
+          points: [
+            <Types.Point3>[...worldPos],
+            <Types.Point3>[...worldPos],
+            <Types.Point3>[...worldPos],
+            <Types.Point3>[...worldPos],
+          ],
+          textBox: {
```
sedghi, line 118: is there a textBox associated with this tool? I didn't think so

packages/tools/src/tools/annotation/VideoRedactionTool.ts
```diff
@@ -0,0 +1,935 @@
...
+      },
+    };
+
+    addAnnotation(annotation, element);
+
+    const viewportUIDsToRender = getViewportIdsWithToolToRender(
+      element,
+      this.getToolName(),
+      false // TODO_JAMES: Just trying to bypass the vtk specific stuff as fast as possible.
+    );
```
sedghi, line 141: I believe this should be at the ned of the render before setting the this.editData

packages/tools/src/tools/annotation/VideoRedactionTool.ts
```diff
@@ -0,0 +1,935 @@
...
+
+  getHandleNearImagePoint = (element, annotation, canvasCoords, proximity) => {
+    const enabledElement = getEnabledElement(element);
+    const { viewport } = enabledElement;
+
+    const { data } = annotation;
+    const { points, textBox } = data.handles;
+    const { worldBoundingBox } = textBox;
+
+    if (worldBoundingBox) {
```
sedghi, line 173: There is no textBox for this tool as I see

packages/tools/src/tools/annotation/VideoRedactionTool.ts
```diff
@@ -0,0 +1,935 @@
...
+    ) {
+      removeAnnotation(annotation.annotationUID);
+    }
+
+    triggerAnnotationRenderForViewportIds(
+      renderingEngine,
+      viewportUIDsToRender
+    );
+  };
+
```
sedghi, line 359: I feel like we should extend the RectangleROI Tool and just override the renderAnnotation. Similar to how we do. There are a lot of repetitive code here

packages/tools/src/tools/segmentation/RectangleROIThresholdTool.ts

packages/tools/src/tools/annotation/VideoRedactionTool.ts
```diff
@@ -0,0 +1,935 @@
...
+        drawHandlesSvg(
+          svgDrawingHelper,
+          annotationUID,
+          handleGroupUID,
+          activeHandleCanvasCoords,
+          {
+            color,
+          }
+        );
```
sedghi, line 630: Let make the edge color black too. It is odd looking right now

packages/tools/src/tools/annotation/VideoRedactionTool.ts
```diff
@@ -0,0 +1,935 @@
...
+  };
+
+  /**
+   * _getTextLines - Returns the Area, mean and std deviation of the area of the
+   * target volume enclosed by the rectangle.
+   *
+   * @param {object} data - The toolDatas tool-specific data.
+   * @param {string} targetUID - The volumeUID of the volume to display the stats for.
+   */
+  _getTextLines = (data, targetUID: string) => {
```
sedghi, line 710: Repetitive and unnessary code

packages/tools/src/tools/base/AnnotationDisplayTool.ts
```diff
@@ -123,7 +124,10 @@ abstract class AnnotationDisplayTool extends BaseTool {
   };
 
   protected getReferencedImageId(
-    viewport: Types.IStackViewport | Types.IVolumeViewport,
+    viewport:
+      | Types.IStackViewport
+      | Types.IVolumeViewport
+      | Types.IVideoViewport,
```
sedghi, line 130: can we say IViewport?

packages/core/src/types/VideoViewportProperties.ts
```diff
@@ -0,0 +1,17 @@
+import { ViewportProperties } from './ViewportProperties';
+import Point2 from './Point2';
+
+/**
+ * Stack Viewport Properties
+ */
+type VideoViewportProperties = ViewportProperties & {
+  loop?: boolean;
+  muted?: boolean;
```
sedghi, line 9: hmmmm, since we are using offscreen video tag, do we really support audio? I'm confused

packages/core/src/types/VideoViewportProperties.ts
```diff
@@ -0,0 +1,17 @@
...
+  pan?: Point2;
```
sedghi, line 10: having pan here is odd to me what does this mean?

packages/core/src/types/VideoViewportProperties.ts
```diff
@@ -0,0 +1,17 @@
+import { ViewportProperties } from './ViewportProperties';
+import Point2 from './Point2';
+
+/**
+ * Stack Viewport Properties
+ */
+type VideoViewportProperties = ViewportProperties & {
```
sedghi, line 7: please add doc to each

packages/core/src/types/VideoViewportProperties.ts
```diff
@@ -0,0 +1,17 @@
...
+  loop?: boolean;
+  muted?: boolean;
+  pan?: Point2;
+  playbackRate?: number;
+  // The zoom factor, naming consistent with vtk cameras for now,
+  // but this isn't necessarily necessary.
+  parallelScale?: number;
```
sedghi, line 14: again odd, these should be set via `setProperties`

packages/core/src/types/VideoViewportProperties.ts
sedghi: for pan and parallelScale we should treat it similar to the viewport which we have setPan and setZoom programmetically

packages/core/src/types/IVideoViewport.ts
```diff
@@ -0,0 +1,72 @@
...
+   */
+  setCamera(cameraInterface: ICamera): void;
+  /**
+   * Centers Pan and resets the zoom for stack viewport.
+   */
+  resetCamera(resetPan?: boolean, resetZoom?: boolean): boolean;
+  /**
+   * If the renderer is CPU based, throw an error. Otherwise, returns the `vtkRenderer` responsible for rendering the `Viewport`.
+   */
+  getRenderer(): any;
```
sedghi, line 71: IViewport has a lot of these, like setCamera etc. why we are extending those again here

packages/core/src/RenderingEngine/VideoViewport.ts
```diff
@@ -0,0 +1,509 @@
...
+  type: ViewportType;
+  element: HTMLDivElement;
+  sx: number;
+  sy: number;
+  sWidth: number;
+  sHeight: number;
+  defaultOptions: any;
+  canvas: HTMLCanvasElement;
+};
+
```
sedghi, line 36: Would be great to refactor out inside src/utilities/video.ts

packages/core/src/RenderingEngine/VideoViewport.ts
```diff
@@ -0,0 +1,509 @@
...
+  readonly canvasContext: CanvasRenderingContext2D;
+  private videoElement?: HTMLVideoElement;
+  private videoWidth = 0;
+  private videoHeight = 0;
+
+  private loop = false;
+  private mute = true;
+  private isPlaying = false;
+  private scrollSpeed = 1;
+  private fps = 30; // TODO We need to find a good solution for this.
```
sedghi, line 56: what is the TODO say here? I think we resolved it

packages/core/src/RenderingEngine/VideoViewport.ts
```diff
@@ -0,0 +1,509 @@
...
+  private videoCamera: VideoCamera = {
```
sedghi, line 57: This is not compatible with other viewports we currently have. Therefore, it is unclear how to handle this situation. Ideally, we should use focalPoint and position, just like we do with other elements. I am confident that if we don't address this now, it will cause issues again when working with annotation and segmentation tools.

packages/core/src/RenderingEngine/VideoViewport.ts
```diff
@@ -0,0 +1,509 @@
...
+      this.elementDisabledHandler
+    );
+  }
+
+  private elementDisabledHandler() {
+    this.removeEventListeners();
+    this.videoElement.remove();
+  }
+
+  public async setVideo(videoURL: string) {
```
sedghi, line 104: I feel like this is misleading, can we say setVideoURL?

packages/core/src/RenderingEngine/VideoViewport.ts
```diff
@@ -0,0 +1,509 @@
...
+      (worldPos[0] + pan[0]) * worldToCanvasRatio,
+      (worldPos[1] + pan[1]) * worldToCanvasRatio,
+    ];
+
+    const canvasPos: Point2 = [subCanvasPos[0], subCanvasPos[1]];
+
+    return canvasPos;
+  };
+
+  private refreshRenderValues() {
```
sedghi, line 405: sounds like `updateCameraValues` to me

packages/core/src/RenderingEngine/VideoViewport.ts
```diff
@@ -0,0 +1,509 @@
...
+      this.renderFrame();
+    }
+  };
+
+  /**
+   * Returns the rendering engine driving the `Scene`.
+   *
+   * @returns The RenderingEngine instance.
+   */
+  getRenderingEngine() {
```
sedghi, line 351: we don't need this already implemented in viewport

packages/core/src/RenderingEngine/VideoViewport.ts
sedghi: or resestCameraPanAndZoom?

packages/core/src/RenderingEngine/VideoViewport.ts
```diff
@@ -0,0 +1,509 @@
...
+      this.renderFrame();
+    }
+  }
+
+  public getCamera(): VideoViewportProperties {
+    return {
+      pan: this.videoCamera.pan,
+      parallelScale: this.videoCamera.parallelScale,
+    };
+  }
```
sedghi, line 305: I have significant concerns regarding annotation and segmentation tools. It would be ideal to calculate the ICamera and export it here, as this would ensure the proper functioning of our annotation/segmentation tools once we implement getImageData. Otherwise, we would need to extensively add custom logic for the Video in the tools, which would be quite challenging. Therefore, I strongly prefer to handle it in this manner.

The reason why our GPU and CPU StackViewport work seamlessly with the tools is because we simulate the computation of all ICameras for the CPU route (even though vtk Camera is not used). By enforcing this computation, we eliminate any explicit mention of the CPU vs GPU within the tools. See `getCameraCPU`

packages/core/src/RenderingEngine/VideoViewport.ts
```diff
@@ -0,0 +1,509 @@
...
+  };
+
+  public resetProperties() {
+    this.setProperties({
+      loop: false,
+      muted: true,
+    });
+  }
+
+  public setCamera(
```
sedghi, line 281: same for this, I'm already seeing some custom logic for the pan and zoom in the tools. And it will happen for other tools. 

As mentioned there is no mention of CPU vs GPU in the Pan/Zoom tool since we take care of them at the viewport level and their camera are two different cameras. But we force them to implement the same interface ICamera. 

Then when setCameraCPU we just convert them back to internal CPU camera

packages/core/src/RenderingEngine/VideoViewport.ts
```diff
@@ -0,0 +1,509 @@
...
+      return;
+    }
+    if (!this.videoElement) {
+      return;
+    }
+    this.videoElement.playbackRate = rate;
+    this.play();
+  }
+
+  public setScrollSpeed(scrollSpeed = 1, unit = 'f') {
```
sedghi, line 263: unit should be Enums

packages/tools/src/tools/PanTool.ts
```diff
@@ -29,28 +29,38 @@ class PanTool extends BaseTool {
...
-    enabledElement.viewport.render();
+    if (viewport instanceof VideoViewport) {
+      const videoViewport: VideoViewport = <VideoViewport>viewport;
+
+      const { pan } = videoViewport.getCamera();
+
+      videoViewport.setCamera({
+        pan: [pan[0] + deltaPoints.world[0], pan[1] + deltaPoints.world[1]],
+      });
+    } else {
```
sedghi, line 42: Not following the ICamera leads to this side effect, which will occur during Phase 2 for annotation and segmentations.

packages/tools/src/tools/ZoomTool.ts
```diff
@@ -48,6 +51,14 @@ class ZoomTool extends BaseTool {
     const worldPos = currentPoints.world;
     const enabledElement = getEnabledElement(element);
 
+    if (enabledElement.viewport instanceof VideoViewport) {
```
sedghi, line 54: side effect of ICamera

packages/tools/src/tools/ZoomTool.ts
```diff
@@ -131,6 +147,55 @@ class ZoomTool extends BaseTool {
     viewport.render();
   }
 
+  _dragCallbackVideoViewport(evt) {
```
sedghi, line 150: Ok this is exactly what i was looking for a map between ICamera and IVideoCamera, if you put this in the setCamera you have half of it done

sedghi (commented):
Nice work, major comment is that we should implement ICamera for the VideoViewport so that there be not custom logic for VideoViewport in the tools. Not implementing the same interface would bite us very bad in phase 2 with annotations and segmentation. You have preliminary implemented that mapper between Interfaces as I see in `_dragCallbackVideoViewport`

packages/core/src/RenderingEngine/VideoViewport.ts
```diff
@@ -0,0 +1,522 @@
...
+
+    // All other viewports have the focal point in canvas coordinates in the center
+    // of the canvas, so to make tools work the same, we need to do the same here
+    // and convert to the world coordinate system since focal point is in world coordinates.
+    const canvasCenterWorld = this.canvasToWorld(canvasCenter);
+
+    return {
+      parallelProjection: true,
+      focalPoint: canvasCenterWorld,
+      position: [0, 0, 0],
```
wayfarer3130, line 322: Probably we should generate more values here sometime, but I don't think it matters for now.

packages/tools/src/tools/annotation/VideoRedactionTool.ts
wayfarer3130: Agreed, but that belongs in the tool section for the video.  Right now it is fine as an example that is intended to still evolve.

packages/tools/src/eventListeners/touch/touchStartListener.ts
```diff
@@ -134,9 +134,9 @@ let tapState: ITouchTapListnenerState = JSON.parse(
 function triggerEventCallback(ele, name, eventDetail) {
   if (runtimeSettings.get('debug')) {
     if (name === 'CORNERSTONE_TOOLS_TOUCH_DRAG') {
-      console.debug(name);
```
wayfarer3130, line 137: I'm not keen on having a log message here at all unless we can turn it on/off consistently - if it is ok with you, I will delete hte entire line.

wayfarer3130 (approved):
It looks pretty good now.  I think this is a reasonable first pass, good enough to include until the next phase.

PR #829, opened 2023-10-14
fix(color): Fix RGB inconsistencies

rodrigobasilio2022:
### Context

This PR aims to fix the problem with RGB inconsistencies in loading specific color dicom images

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  

### DIFF
--- a/packages/dicomImageLoader/src/imageLoader/createImage.ts
+++ b/packages/dicomImageLoader/src/imageLoader/createImage.ts
@@ -228,8 +228,8 @@ function createImage(
         cornerstone.metaData.get('modalityLutModule', imageId) || {};
       const sopCommonModule: MetadataSopCommonModule =
         cornerstone.metaData.get('sopCommonModule', imageId) || {};
-      if (isColorImage) {
-        const { rows, columns } = imageFrame;
+      const { rows, columns } = imageFrame;
+      if (isColorImage && imageFrame.pixelDataLength !== 3 * rows * columns) {
         if (TRANSFER_SYNTAX_USING_PHOTOMETRIC_COLOR[transferSyntax]) {
           canvas.height = imageFrame.rows;
           canvas.width = imageFrame.columns;


rodrigobasilio2022:
> This part of the code is very tricky to touch. Please explain the logic behind your addition

As stated in the lines 21-28 of the same file, the purpose of the code is to transform from RGBA -> RGB, as the main job was already been done by the decompressors. So if the data is already in RGB (3 bytes per voxel), then there is no need to go further in the code. Also the code in here https://github.com/cornerstonejs/cornerstone3D/blob/b33eac99097417d700cdf4ba27c70c47f2eac48a/packages/dicomImageLoader/src/imageLoader/colorSpaceConverters/convertYBRFull422ByPixel.ts#L8-L11 didnt make much sense

sedghi:
I'm pretty certain this breaks a lot of other color images. It is not just RGBA -> RGB, it also handles color space conversions

```js
if (TRANSFER_SYNTAX_USING_PHOTOMETRIC_COLOR[transferSyntax]) {
          canvas.height = imageFrame.rows;
          canvas.width = imageFrame.columns;
          const context = canvas.getContext('2d');
          const imageData = context.createImageData(
            imageFrame.columns,
            imageFrame.rows
          );
          convertColorSpace(imageFrame, imageData.data, useRGBA);
          imageFrame.imageData = imageData;
          imageFrame.pixelData = imageData.data;
          imageFrame.pixelDataLength = imageData.data.length;
        }
```

rodrigobasilio2022:
> Yeah i know exactly what you are talking about. I tried to find the correct condition, but everytime I tried something it failed something else. @wayfarer3130 maybe he has an idea of what is the best way to identify if something comes out of the decompresser as RGB

What if for the moment we get the requirements for each conversion and only allow if the input is correct: for example ybr_full_422 expects two bytes data. What if we check that and if the requirements does match we allow proceed and if not we return the data as is? The only extepcting two byte data is the YBR_FULL_422, all others expects three byte data

rodrigobasilio2022:
Adde a new function that checks the requirements for color space convertions. It convers only one color space conversion for now, the YBR_FULL_422, as it tests if the voxel byte size is equal to 2.

packages/dicomImageLoader/src/imageLoader/colorSpaceConverters/convertYBRFullByPixel.ts
```diff
@@ -6,10 +6,12 @@ export default function (
   useRGBA: boolean
 ): void {
   if (imageFrame === undefined) {
-    throw new Error('decodeRGB: ybrBuffer must not be undefined');
+    throw new Error('convertYBRFullByPixel: ybrBuffer must not be undefined');
```
wayfarer3130, line 9: If you are going to change the text, how about must be defined

packages/dicomImageLoader/src/imageLoader/colorSpaceConverters/convertYBRFullByPixel.ts
```diff
@@ -6,10 +6,12 @@ export default function (
...
   }
   if (imageFrame.length % 3 !== 0) {
-    throw new Error('decodeRGB: ybrBuffer length must be divisble by 3');
+    throw new Error(
+      'convertYBRFullByPixel: ybrBuffer length must be divisible by 3'
```
wayfarer3130, line 13: How about include the length here?

packages/dicomImageLoader/src/imageLoader/convertColorSpace.ts
```diff
@@ -23,6 +23,25 @@ function convertYBRFull(imageFrame, colorBuffer, useRGBA) {
...
 }
 
+/**
+ * This function checks color space conversion data requirements before
+ * apply them
+ * @param imageFrame
+ * @param RGBA
+ * @returns
+ */
+export function isColorConversionRequirementsFulfilled(imageFrame, RGBA) {
```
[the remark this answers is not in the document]
wayfarer3130: According to the DICOM standard:
If images with Photometric Interpretation (0028,0004) YBR_FULL_422 or YBR_PARTIAL_422, are encoded with JPEG coding Process 1 (non hierarchical with Huffman coding), identified by DICOM Transfer Syntax UID "1.2.840.10008.1.2.4.50" the minimum compressible unit is YYCBCR, where Y, CB, and CR are 8 by 8 blocks of pixel values. The data stream encodes two Y blocks followed by the corresponding CB and CR blocks.

Thus, for JPEG encoded values, the columns is a multiple of 16, while the height is a multiple of 8, with the row length being a multiple of 2, as specified.
It is permissible to encode this as LEI YBR_FULL_422, so this would be just a straight multiple of 2 for the number of columns.

For YBR_FULL_420, the equivalent is:
Y Y
Y Y CB CR
So, the number of columns and number of rows must be a multiple of 2, AND the number of elements a multiple of 6/4  (eg 4 Y element, 1 CB and 1 CR for every 4 pixel elements)

packages/dicomImageLoader/src/imageLoader/convertColorSpace.ts
```diff
@@ -23,6 +23,25 @@ function convertYBRFull(imageFrame, colorBuffer, useRGBA) {
   }
 }
 
+/**
```
wayfarer3130, line 26: Suggest extracting this function into it's own file.

packages/dicomImageLoader/src/imageLoader/convertColorSpace.ts
```diff
@@ -23,6 +23,25 @@ function convertYBRFull(imageFrame, colorBuffer, useRGBA) {
...
+ * apply them
+ * @param imageFrame
+ * @param RGBA
+ * @returns
+ */
+export function isColorConversionRequirementsFulfilled(imageFrame, RGBA) {
+  const { rows, columns } = imageFrame;
+  if (
+    imageFrame.photometricInterpretation === 'YBR_FULL_422' &&
+    imageFrame.pixelDataLength === 2 * rows * columns
```
wayfarer3130, line 37: Specifically, the rows element must be even length, and the total length be 2 * rows * columns

rodrigobasilio2022:
Changed as suggested

sedghi:
Testing on https://www.cornerstonejs.org/live-examples/dicomimageloaderwadouri

These two are broken which is not in main right now

CleanShot 2023-10-17 at 11 37 33

rodrigobasilio2022:
Fixed the issue!!!

packages/dicomImageLoader/src/imageLoader/isColorConversionRequirementsFulfilled.ts
```diff
@@ -0,0 +1,28 @@
...
+ * @returns
+ */
+export default function isColorConversionRequirementsFulfilled(
+  imageFrame,
+  RGBA
+) {
+  const { rows, columns } = imageFrame;
+  if (imageFrame.photometricInterpretation === 'RGB') {
+    return imageFrame.length % 3 === 0;
+  } else if (imageFrame.photometricInterpretation === 'YBR_RCT') {
```
sedghi, line 15: Is there a specific reference available that can provide an explanation for these concepts? If the photometric is YBR_RCT and the frame is a multiple of 3, can we confidently say that it was converted to RGB in the decoder?

Note: I'm not very familiar with color spaces, so this might be a commonly known fact.

packages/dicomImageLoader/src/imageLoader/isColorConversionRequirementsFulfilled.ts
```diff
@@ -0,0 +1,28 @@
+/**
+ * This function checks color space conversion data requirements before
+ * apply them
+ * @param imageFrame
+ * @param RGBA
+ * @returns
+ */
+export default function isColorConversionRequirementsFulfilled(
```
sedghi, line 8: @wayfarer3130 Can you please look at this function? Thanks

packages/dicomImageLoader/src/imageLoader/isColorConversionRequirementsFulfilled.ts
rodrigobasilio2022: I just got all checks in the correspondent functions and put them here. In future we can create better checks here.

packages/dicomImageLoader/src/imageLoader/isColorConversionRequirementsFulfilled.ts
rodrigobasilio2022: I will limit the checks to only YBR_FULL_422

packages/dicomImageLoader/src/imageLoader/isColorConversionRequirementsFulfilled.ts
rodrigobasilio2022: I will limit the check to only YBR_FULL_422

rodrigobasilio2022:
Changed as suggested

sedghi:
I don't think we need this after this PR can someone confirm? https://github.com/OHIF/Viewers/pull/3883

packages/dicomImageLoader/src/imageLoader/colorSpaceConverters/convertYBRFull422ByPixel.ts
```diff
@@ -6,10 +6,12 @@ export default function (
   useRGBA: boolean
 ): void {
   if (imageFrame === undefined) {
-    throw new Error('decodeRGB: ybrBuffer must not be undefined');
+    throw new Error('convertYBRFull422ByPixel: ybrBuffer must be defined');
   }
   if (imageFrame.length % 2 !== 0) {
```
wayfarer3130, line 11: Please add a comment here that the length is divisible by 2 because for 4 pixels (a 2x2 array), there are 4 y and 2 b and 2 r values, so 4+2+2 / 4 = 2 bytes per pixel

packages/dicomImageLoader/src/imageLoader/isColorConversionRequirementsFulfilled.ts
```diff
@@ -0,0 +1,24 @@
+/**
+ * This function checks color space conversion data requirements before
+ * apply them. This function was created to solve problems like the one
```
wayfarer3130, line 3: applying

packages/dicomImageLoader/src/imageLoader/isColorConversionRequirementsFulfilled.ts
```diff
@@ -0,0 +1,24 @@
...
+ */
+export default function isColorConversionRequirementsFulfilled(
+  imageFrame,
+  RGBA
+) {
+  if (imageFrame === undefined) {
+    return false;
+  }
+  const { rows, columns } = imageFrame;
+  if (imageFrame.photometricInterpretation === 'YBR_FULL_422') {
```
wayfarer3130, line 19: You can check endsWith 422 or endsWIth 420 to support all the test values we can check.

packages/dicomImageLoader/src/imageLoader/isColorConversionRequirementsFulfilled.ts
```diff
@@ -0,0 +1,24 @@
...
+    return imageFrame.pixelDataLength === 2 * rows * columns && rows % 2 === 0;
+  } else {
```
wayfarer3130, line 21: Add a settable flag allowing coercion to RGB if the response is uncompressed, and the original is YBR of some flavour.  That flag could go in cornerstone settings, and maybe be triggered based on the URL prefix (eg url.startsWith), or a regular expression match.

sedghi: Can you please add more to this? I don't understand

packages/dicomImageLoader/src/imageLoader/isColorConversionRequirementsFulfilled.ts
sedghi: as we talked to Bill
the formula should be edited to 

number of ybr y ybr y ybr y elements -> 3 * ceil(cols/2) + florr(cols/2) * ceil(rows/2)
+ cols * floor(rows/2)

wayfarer3130: > as we talked to Bill the formula should be edited to
> 
> for ybr_Full_422
> 
> ```
> ybr y ybr y ybr .... (ybr or y)
> y   y  y  y  y ........ y
> ..
> ..
> ..
> (last row)
> ybr y ybr y ybr y ... (ybr or y)
> OR 
> y   y  y   y  y   y 
> ```
> 

> -> 3 * ceil(cols/2) + florr(cols/2) * ceil(rows/2) + cols * floor(rows/2)

That is:
(3 * ceil(cols/2) + floor(cols/2)) * ceil(rows/2) + cols * floor(rows/2)

Note the parenthesis around the first expression.  That breaks down into:
There are ceil(rows/2) YBR containing rows (think about 1,2,3 rows, having 1 1 and 2 YBR containing rows)
There are floor(rows/2) NON YBR containing rows.  Those have length cols as they are just y containing.
For each YBR row, there are ceil(cols/2) YBR elements - think about hte same thing as the rows, but for columns - the first, third, fifth etc column has YBR.  Each YBR has 3 elements (Y b and r)
Then, for each YBR row, there are floor(cols/2) Y element only.

packages/dicomImageLoader/src/imageLoader/isColorConversionRequirementsFulfilled.ts
wayfarer3130: If the photometric interpretation is YBR_FULL_422, BUT the length is 3*rows*columns, then assume it is the configurable value of either RGB or YBR_FULL.  That is invalid data, but is common enough that we see it regularly as a "bug" fix request.
To decide whether RGB or YBR_FULL, check the datasource configuration - something like:
UncompressedColorEncoding=YBR_FULL | RGB  (default is RGB)
and perhaps
ForceColorEncoding=AllUncompressed | All | WrongLength  (default is WrongLength)

Then, have the above function return the PMI to use for the given image.   

const {ForceColorEncoding = WrongLength, UncompressedColorEncoding = RGB} = dataSourceConfiguration;

if( pixel length === 3 * rows * columns && pmi ==='YBR_FULL_422 && forceColorEncoding===WrongLength ) {
   return UncompressedColorEncoding

indicating that this image needs decompression handling, in the specified color encoding.

PR #832, opened 2023-10-15
fix(StreamingImageLoader): allow image volume loading to work when SharedArrayBuffer is being used.

sgielen:
### Context

I was investigating why my 3D stacks load normally, but all volumes are completely
grayed out.

For me, scalarData.buffer was a SharedArrayBuffer and not an ArrayBuffer here.
This was causing handleArrayBufferLoad to ignore the successful loading of
pixel data, keeping the volume scalarData all-zero, and eventually causing the
volume to display completely empty even though all images were loaded.

I suppose, at this point, the idea was to check whether there was any
scalarData to load into -- not that the type of buffer was explicitly
ArrayBuffer. By changing the check accordingly, the slice pixel data is being
loaded into the volume scalarData correctly, and the volume is rendered
normally.

### Changes & Results

Before:

Scherm­afbeelding 2023-10-15 om 19 29 12

After:

Scherm­afbeelding 2023-10-15 om 19 28 24

(The black rectangle is another issue on my end, not related to this one.)

### Testing

This bug can be triggered simply by calling `volume.load()` on the result of a `streamingImageVolumeLoader.loadVolume()` while in a scope that allows cross-origin isolation (i.e. the Cross-Origin-Opener-Policy & Cross-Origin-Embedder-Policy headers are set to 'same-origin' and 'require-corp' respectively). This will cause Cornerstone to use SharedArrayBuffer instead of ArrayBuffer.

If you'd like I can supply some code examples, but it's not zero-effort so let me know if this is wanted.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] "OS: macOS 14.0"
- [x] "Node version: v20.8.0"
- [x] "Browser: Chrome 117.0"

### DIFF
--- a/packages/streaming-image-volume-loader/src/BaseStreamingImageVolume.ts
+++ b/packages/streaming-image-volume-loader/src/BaseStreamingImageVolume.ts
@@ -469,7 +469,7 @@ export default class BaseStreamingImageVolume extends ImageVolume {
     }
 
     function handleArrayBufferLoad(scalarData, image, options) {
-      if (!(scalarData.buffer instanceof ArrayBuffer)) {
+      if (!scalarData.buffer) {
         return;
       }
 


sgielen:
@sedghi thanks for your quick reply! I'll make a test case and will report back. Thanks!

sgielen:
@sedghi Alright, I went ahead and wrote a test case reproducing the issue. :-)

So the test case is doing a few weird things. I'm sure there are some things that could be done simpler. Please let me know if you have any suggestions on that front! I really appreciate you looking into this.

The test case touches multiple issues I am having with Cornerstone at the moment, but for this PR, let's just focus on one: when SharedArrayBuffer is used, the volume is rendered fully gray. Once we looked into this, and the other issues still remain, I can file issues for those, too.

Here's the zip file: https://sjorsgielen.nl/cs3d.zip. It's pretty large since it contains a 90MB volume, so Github doesn't allow me to upload it directly. For future record, here is a zip file containing everything except the volume: 
[cs3d.zip](https://github.com/cornerstonejs/cornerstone3D/files/12921184/cs3d.zip)

Run `yarn install` and `yarn dev` to start the webserver, then browse to the indicated port. Note that `vite.config.js` sets the necessary headers for cross-origin isolation to work.

In short, the test case:
- Uses a custom loader inside the `cornerstoneDICOMImageLoader` to download the images. This may be relevant to the issue. I use this approach because I want to turn my DICOM bytes (coming from a custom protocol) into cornerstone `IImage`s.
- Renders them as a volume in acquisition order.
- Scrolls through the volume continuously.

By default, it does this with a very high window center/width, because of another issue I have while GPU rendering (similar to #225).

For me, this renders like this:

Scherm­afbeelding 2023-10-16 om 21 38 47

If we disable SharedArrayBuffer (see the constants at the top of `main.ts`), the window level is off, but at least the volume renders somewhat:

Scherm­afbeelding 2023-10-16 om 21 40 03

Likewise, if we render as a stack instead of as a volume (regardless of SharedArrayBuffer config):

Scherm­afbeelding 2023-10-16 om 21 40 41

Also, if we use the same configuration as the third image, but we use my patched version of the streaming loader from this PR:

```
$ cd node_modules/@cornerstonejs
$ mv streaming-image-volume-loader{,.old}
$ ln -s /Users/sjors/...../cornerstone3D/packages/streaming-image-volume-loader .
$ cd ../..
$ rm -rf node_modules/.vite/deps # clear out dependency cache
$ vite dev
```

Now, the volume renders the same as the second image, even with `useVolume` true and `SharedArrayBuffer` enabled:

Scherm­afbeelding 2023-10-16 om 21 44 14

Looking forward to hearing what you think about this. Perhaps it's just my code being silly. :-)

sedghi:
So the wadors (using orthanc) works fine, there might be a bug in wadouri (I'm looking now)

CleanShot 2023-10-16 at 17 16 33

sedghi:
so seems like SAB is not supported in wadoURI. For the windowlevel,  since you have a custom loader you have not passed the prescale true and that is the issue. 

So you should do 

```js
registerImageLoader('custom', (imageId: string) => {
  const filename = imageId.substring('custom:'.length);
  return cornerstoneDICOMImageLoader.wadouri.loadImage(imageId, {
    loader: () => {
      console.log(filename);
      return fetch('/' + filename).then((r) => r.arrayBuffer());
    },
    preScale: {
      enabled: true,
    },
  });
});
```

which makes me think maybe we should force it to true... maybe in packages/dicomImageLoader/src/imageLoader/wadouri/loadImage.ts


CleanShot 2023-10-16 at 17 45 53


Do you want to create that fix maybe instead of your initial proposal?

sgielen:
> so seems like SAB is not supported in wadoURI. For the windowlevel, since you have a custom loader you have not passed the prescale true and that is the issue.

If I understand you correctly these are two separate issues, right? We have (1) when loading images through wadoURI with SAB enabled we can't load those images into a volume, and (2) the windowlevel is wrong and we should enable prescale. I can apply your fix for (2) in a separate PR (even though I must admit I don't really understand the relation between having a custom loader and needing prescaling). I have some more thoughts about the window leveling (e.g. `setProperties({ voiRange })` doesn't seem to have any effect on a VolumeViewport) but I'll keep this PR focused on #1.

Even with `preScale: { enabled: true }}` the original issue this PR intends to fix still exists - do you already have a gut feeling why this is? Is there anything I could do to help on that front to come up with a better PR than this one?

sgielen:
Closing this since the fix is incorrect according to @sedghi -- but have filed #886 to investigate the issue further.

PR #835, opened 2023-10-18
fix(getDataInTime): to consider different image spacing for reference

sedghi:
### Context

when we use the mask to grab the information from a reference volume, if they are not the same properties (spacing, direction, etc.) we should consider the voxel overlaps and average (maybe we re-configure this later or maybe do weighted average based on the overlap percentage)

image

### Changes & Results

calculates the voxel overlaps

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  

### DIFF
--- a/packages/tools/src/utilities/dynamicVolume/getDataInTime.ts
+++ b/packages/tools/src/utilities/dynamicVolume/getDataInTime.ts
@@ -158,6 +158,8 @@ function _getTimePointDataMask(frames, dynamicVolume, segmentationVolume) {
       segPointLPS
     );
 
+    // count represents the number of voxels of the dynamic volume that represents
+    // one voxel of the segmentation mask
     let count = 0;
     const perFrameSum = new Map();
 
@@ -165,7 +167,7 @@ function _getTimePointDataMask(frames, dynamicVolume, segmentationVolume) {
     frames.forEach((frame) => perFrameSum.set(frame, 0));
 
     const averageCallback = ({ index }) => {
-      for (let i = 0, len = frames.length; i < len; i++) {
+      for (let i = 0; i < frames.length; i++) {
         const value = dynamicVolumeScalarDataArray[i][index];
         const frame = frames[i];
         perFrameSum.set(frame, perFrameSum.get(frame) + value);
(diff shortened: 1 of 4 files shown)


packages/tools/src/utilities/dynamicVolume/getDataInTime.ts
```diff
@@ -95,19 +94,104 @@ function _getTimePointDataCoordinate(frames, coordinate, volume) {
...
+    // Pre-initialize the Map
+    frames.forEach((frame) => perFrameSum.set(frame, 0));
+
+    const averageCallback = ({ index }) => {
+      for (let i = 0, len = frames.length; i < len; i++) {
+        const value = dynamicVolumeScalarDataArray[i][index];
+        const frame = frames[i];
+        perFrameSum.set(frame, perFrameSum.get(frame) + value);
+      }
+      count++;
```
rodrigobasilio2022, line 173: Should leave a comment here. count is representing the number of voxels of the dynamicVolume that represents one voxel in the mask, but the fact that a callback function modifying a global variable is dangerous

packages/tools/src/utilities/dynamicVolume/getDataInTime.ts
```diff
@@ -95,19 +94,104 @@ function _getTimePointDataCoordinate(frames, coordinate, volume) {
...
+    );
+
+    let count = 0;
+    const perFrameSum = new Map();
+
+    // Pre-initialize the Map
+    frames.forEach((frame) => perFrameSum.set(frame, 0));
+
+    const averageCallback = ({ index }) => {
+      for (let i = 0, len = frames.length; i < len; i++) {
```
rodrigobasilio2022, line 168: Why creating a variable len here?

packages/tools/src/utilities/dynamicVolume/getDataInTime.ts
```diff
@@ -95,19 +94,104 @@ function _getTimePointDataCoordinate(frames, coordinate, volume) {
...
+    // average the values
+    const averageValues = [];
+    perFrameSum.forEach((sum) => {
+      averageValues.push(sum / count);
+    });
+
+    values.push(averageValues);
+  };
+
+  pointInShapeCallback(maskImageData, () => true, callback);
```
rodrigobasilio2022, line 192: As you have the nonZeroVoxelIndices array why you are not using it here, instead of using the pointInShapeCallback? In case of a CT mask in a PET you are calculationg for each CT voxel, that could be a very large number, lots of calculations that doens need to be done

packages/tools/src/utilities/dynamicVolume/getDataInTime.ts
sedghi: added

packages/tools/src/utilities/dynamicVolume/getDataInTime.ts
sedghi: Good point. because I need the segPointLPS of each of the masks voxels, and also I have an early return in the first line of callback. My benchmark don't show any performance issue, but I added a todo

PR #837, opened 2023-10-18
fix: Add degree to rad conversion for rotate mouse wheel

salimkanoun:
### Context

RotateMouseWheel as a configuration unit in degree (which is fine)
The probleme is that the value in degree was passed to glmatrix which is using radian rotation unit

### Changes & Results
Convert degree to radian before application

### Testing

N/A

No modification needed

#### Tested Environment


sedghi:
good catch, though it is a breaking change, I will merge into beta for now

sedghi (approved):

PR #844, opened 2023-10-20
feat(segmentation): Add polymorph segmentation and representation conversion capabilities

sedghi:
- 👏 Kudos to Mo Alsad (@moalasad) from [XNAT](https://www.xnat.org/) team and [ICR](https://www.icr.ac.uk/) for their amazing new repository [`polySEG-wasm`](https://bitbucket.org/icrimaginginformatics/polyseg-wasm/src/master/) to put the ground work for the WASM converters, and also kudos to @rodrigobasilio2022 who did an excellent job by adding the remaining converters. 🎉

### Changes & Results
polymorph segmentation is the idea of "Segmentation object storing each structure of an entity with each representation in one data object" [formulated and developed initially by](https://github.com/PerkLab/PolySeg) @cpinter and 3D Slicer. 

image

image

Through the development of Cornerstone3D, we [aimed to address the separation between representation and segmentation](https://www.cornerstonejs.org/docs/concepts/cornerstone-tools/segmentation/). We built the segmentation framework around it, initially providing the Labelmap representation. 

This PR will add the capability to create/compute different representations from each other without a breaking change.

This PR incorporates the necessary logic and connectors into the internals of cornerstone3D. It also enables the execution of each conversion within a web worker to prevent the main thread from being blocked. Enjoy!

[](https://youtu.be/7UCmzz_HGfw)

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  


packages/core/src/RenderingEngine/Viewport.ts
```diff
@@ -765,6 +763,10 @@ class Viewport implements IViewport {
     // compute the radius of the enclosing sphere
     radius = Math.sqrt(radius) * 0.5;
 
+    // For 3D viewport, we should increase the radius to make sure the whole
+    // volume is visible and we don't get clipping artifacts.
+    radius = this.type === ViewportType.VOLUME_3D ? radius * 10 : radius;
```
wayfarer3130, line 768: Is there a performance impact of this?  Is there a smaller value that would still fix things, but have less impact?

packages/core/src/RenderingEngine/helpers/setDefaultVolumeVOI.ts
```diff
@@ -175,7 +175,13 @@ async function getVOIFromMinMax(
   // instead. For the first scenario, we use the arrayBuffer of the volume to get the correct
   // slice for the imageScalarData, and for the second scenario we use the getPixelData
   // on the Cornerstone IImage object to get the pixel data.
-  const image = await loadAndCacheImage(imageId, options);
+  let image;
+  // Note: we don't want to use the derived or generated images for setting the
```
wayfarer3130, line 179:
```suggestion
  // Note: we don't want to use the derived or generated images for setting the
```
Couple of thoughts on this - one idea is to create a utility function to do the test on the imageId portion, which at least abstracts it out of this class.  Another idea is to not check on the imageId, but to check on the image object itself.  That means that loadAndCacheImage would have to response with a resolved promise, but that should probably already be true, so doing it that way means you can add a field on the image response object.

packages/core/src/RenderingEngine/RenderingEngine.ts
```diff
@@ -305,6 +305,14 @@ class RenderingEngine implements IRenderingEngine {
 
     this.setVtkjsDrivenViewports(vtkDrivenViewportInputEntries);
     this.setCustomViewports(customRenderingViewportInputEntries);
+
+    // Making sure the setViewports api also can fill the canvas
```
wayfarer3130, line 309: I suspect there are a couple of spots that the background colors are filled.  Any reason to not make a method and call it consistently for filling?

packages/core/src/cache/classes/Surface.ts
```diff
@@ -43,6 +43,18 @@ export class Surface {
     return this.polys;
   }
 
+  public setColor(color: Point3): void {
```
wayfarer3130, line 46: Don't we have a color type?  If it has the same shape, using a different name allows changing things int he future and finding everywhere using it.

packages/core/src/RenderingEngine/helpers/setDefaultVolumeVOI.ts
wayfarer3130: Actually, the cache itself has a referenced image/volume, so you can use that to determine if it is derived/generated?

packages/core/src/types/IImageVolume.ts
```diff
@@ -81,10 +81,13 @@ interface IImageVolume {
   destroy(): void;
 
   /** decache */
-  decache?: () => void;
+  decache?: (completelyRemove?: boolean) => void;
 
   /** */
   get imageCacheOffsetMap(): Map<string, any>;
+
+  /** updates volume internals  */
```
wayfarer3130, line 89: The comment should be more generic - something like "mark the volume as having had the pixel data changed externally".  That explains what the call does, not how it does it or what it does internally to performt he operation.

packages/core/src/types/ISurface.ts
```diff
@@ -0,0 +1,14 @@
...
+  readonly id: string;
+  readonly sizeInBytes: number;
+  readonly frameOfReferenceUID: string;
+  getColor(): Point3;
+  setColor(color: Point3): void;
+  getPoints(): number[];
+  getPolys(): number[];
+  getSizeInBytes(): number;
+  setPoints(points: number[]): void;
+  setPolys(polys: number[]): void;
```
wayfarer3130, line 13: Since we spell everything else out, could we use setPolygons?

packages/tools/examples/PolySegWasmContourToStackLabelmap/index.ts
```diff
@@ -0,0 +1,302 @@
...
+  eventTarget.addEventListener(
+    cornerstoneTools.Enums.Events.POLYSEG_CONVERSION_COMPLETED,
+    (evt) => {
+      const p = document.getElementById('progressDetailP');
+      p.innerHTML = `Conversion completed`;
+    }
+  );
+
+  // Add tools to Cornerstone3D
+  cornerstoneTools.addTool(PanTool);
```
wayfarer3130, line 153: Can you use the addManipulationBindings to start removing heavily duplicated code?

utils/ExampleRunner/template-config.js
```diff
@@ -57,6 +57,9 @@ module.exports = {
   module: {
     rules,
   },
+  experiments: {
```
wayfarer3130, line 60: Is this a good idea for production code?  Or does it only affect running the examples?

packages/tools/src/utilities/segmentation/getUniqueSegmentIndices.ts
```diff
@@ -0,0 +1,74 @@
...
+
+    const keySet = {};
+
+    if (isVolumeSegmentation(labelmapData)) {
+      const volume = cache.getVolume(segmentationId);
+      const scalarData = volume.getScalarData();
+
+      for (let i = 0; i < scalarData.length; i++) {
+        const segmentIndex = scalarData[i];
+        if (segmentIndex !== 0 && !keySet[segmentIndex]) {
```
wayfarer3130, line 28: If you use a set [], then you can just use the add function, then remove just 0 at the end.  Makes this part all 1 line.  If you prefer to leave as an object, then keySet[segmentIndex] ||= true works, and you can still delete the 0 at the end.

packages/tools/src/utilities/segmentation/getUniqueSegmentIndices.ts
```diff
@@ -0,0 +1,74 @@
...
+        for (let i = 0; i < scalarData.length; i++) {
+          const segmentIndex = scalarData[i];
+          if (segmentIndex !== 0 && !keySet[segmentIndex]) {
+            keySet[segmentIndex] = true;
+          }
+        }
+      });
+    }
+
+    return Object.keys(keySet).map((it) => parseInt(it, 10));
```
wayfarer3130, line 46: The ,10 is implicit in the parseInt

wayfarer3130: Also, the other response is sorted, but this one isn't - please sort this one or don't sort the others.

packages/tools/src/utilities/math/polyline/isPointInsidePolyline.ts
```diff
@@ -0,0 +1,98 @@
+import type { Types } from '@cornerstonejs/core';
+
+/**
+ * Determines whether a 2D point is inside a polyline.
```
wayfarer3130, line 4: Please indicate the boundary conditions - what happens if a point is on the boundary?  I believe that should be considered inside, but with the naming/description, I just don't know.

packages/tools/src/utilities/segmentation/getSegmentAtWorldPoint.ts
```diff
@@ -0,0 +1,146 @@
...
+import { getAnnotation } from '../../stateManagement';
+import { isPointInsidePolyline3D } from '../math/polyline';
+
+/**
+ * Get the segment at the specified world point in the viewport.
+ * @param viewport - The viewport where the point resides.
+ * @param worldPoint - The world point to get the segment for.
+ * @param segmentationId - The ID of the segmentation to lookup.
+ * @param representationType - The type of segmentation representation (labelmap, contour etc.)
+ * @returns The index of the segment at the world point, or undefined if not found.
```
wayfarer3130, line 27: What happens if there are multiple contours at the same location?  I know for our current labelmaps, we don't allow that, but do you need to support it for the general case?

packages/tools/src/utilities/segmentation/getSegmentAtWorldPoint.ts
```diff
@@ -0,0 +1,146 @@
...
+    return;
+  }
+
+  const segmentationActor = viewport.getActor(segmentationRepresentationUID);
+
+  const imageData = segmentationActor?.actor.getMapper().getInputData();
+  const indexIJK = utilities.transformWorldToIndex(imageData, worldPoint);
+
+  // Since it is a stack, we don't need to check the z
+  const flattenedIndex = indexIJK[0] + indexIJK[1] * image.columns;
```
wayfarer3130, line 101: Please try to use VoxelManager so we can start isolating the behaviour of how to lookup the index values into the scalar data.

packages/tools/src/utilities/math/polyline/isPointInsidePolyline.ts
```diff
@@ -0,0 +1,98 @@
...
+  polyline: Types.Point2[]
+) {
+  const n = polyline.length;
+  let windingNumber = 0;
+  const x = point[0],
+    y = point[1];
+
+  for (let i = 0; i < n; i++) {
+    const j = (i + 1) % n;
+    const xi = polyline[i][0],
```
wayfarer3130, line 24: const [xi,yi] = polyline[i]

packages/tools/src/utilities/math/polyline/isPointInsidePolyline.ts
```diff
@@ -0,0 +1,98 @@
...
+  point: Types.Point2,
+  polyline: Types.Point2[]
+) {
+  const n = polyline.length;
+  let windingNumber = 0;
+  const x = point[0],
+    y = point[1];
+
+  for (let i = 0; i < n; i++) {
+    const j = (i + 1) % n;
```
wayfarer3130, line 23: How about const iNext or something desriptive - I think of j as another iteration index.

packages/tools/src/utilities/math/polyline/isPointInsidePolyline.ts
```diff
@@ -0,0 +1,98 @@
...
+    const xi = polyline[i][0],
+      yi = polyline[i][1];
+    const xj = polyline[j][0],
```
wayfarer3130, line 26: Ditto

packages/tools/src/utilities/math/polyline/isPointInsidePolyline.ts
```diff
@@ -0,0 +1,98 @@
...
+      yj = polyline[j][1];
+
+    if (yi === yj && y !== yi) {
+      continue;
+    }
+
+    if (
+      ((yi < y && yj >= y) || (yj < y && yi >= y)) &&
```
wayfarer3130, line 34: This is a little opaque here - I can figure out what you are doing, but would like a reference to the algorithm in the docs above, or an explanation here - how about:
// If y is between yi and yj, and x is to the right of the line between (xi,yi) and (xj,yj), 
// then the line between (xi,yi) and (xj,yj) is a relevant boundary for being inside/outside the point.

packages/tools/src/utilities/math/polyline/isPointInsidePolyline.ts
```diff
@@ -0,0 +1,98 @@
...
+    throw new Error(
+      'Cannot find a shared dimension index for the polyline points'
+    );
+  }
+
+  // convert polyline list and point to 2D
+  const points2D = [] as Types.Point2[];
+
+  for (let i = 0; i < polyline.length; i++) {
+    points2D.push([
```
wayfarer3130, line 86: Can you extract the min/max here, and do a boundary check?  That is quite a bit less expensive than winding number, and will typically exclude more than half the points in the image area.

packages/tools/src/utilities/contours/getDeduplicatedVTKPolyDataPoints.ts
```diff
@@ -0,0 +1,61 @@
...
+    return { a: cell[0], b: cell[1] };
+  });
+
+  if (bypass) {
+    return { points: pointsArray, lines: linesArray };
+  }
+
+  const newPoints = [];
+  for (const [i, pt] of pointsArray.entries()) {
+    const index = newPoints.findIndex(
```
wayfarer3130, line 27: This is an n^2 algorithm - consider using a Map<string,Point3>.  Generates a reasonable amount of garbage, but I think the performance of that is better than doing repeated compares across the entire array.

packages/tools/src/utilities/contours/getDeduplicatedVTKPolyDataPoints.ts
```diff
@@ -0,0 +1,61 @@
+/**
+ * Iterate through polyData from vtkjs and merge any points that are the same
+ * then update merged point references within lines array
+ * @param polyData - vtkPolyData
+ * @param bypass - bypass the duplicate point removal
+ * @returns the updated polyData
+ */
+export function getDeduplicatedVTKPolyDataPoints(polyData, bypass = false) {
```
wayfarer3130, line 8: Could you add a unit test for this function?

packages/tools/src/utilities/math/polyline/isPointInsidePolyline.ts
```diff
@@ -0,0 +1,98 @@
+import type { Types } from '@cornerstonejs/core';
+
```
wayfarer3130, line 2: Please add unit tests for this function - particularly with directly on the line tests.

packages/tools/src/utilities/contours/detectContourHoles.ts
```diff
@@ -18,6 +18,7 @@
  * @returns
  */
 const getIsPointInsidePolygon = (point, vertices) => {
+  // Todo: this is a duplicated logic we should merge it with the other one
```
wayfarer3130, line 21: Reference to where?

packages/tools/src/utilities/segmentation/getSegmentAtWorldPoint.ts
```diff
@@ -0,0 +1,146 @@
...
+  LabelmapSegmentationDataVolume,
+} from '../../types/LabelmapTypes';
+import { isVolumeSegmentation } from '../../tools/segmentation/strategies/utils/stackVolumeCheck';
+import {
+  ContourSegmentationAnnotation,
+  ContourSegmentationData,
+} from '../../types';
+import { getAnnotation } from '../../stateManagement';
+import { isPointInsidePolyline3D } from '../math/polyline';
+
```
wayfarer3130, line 20: Could we get some unit tests on this?  I'd particularly like to see tests that check that if the slice is different/same as the expected data, we get the right results.

wayfarer3130: Oh, and data inside/outside the range of available indices should also be tested.

packages/tools/src/types/SegmentationStateTypes.ts
```diff
@@ -266,9 +270,15 @@ export type SegmentationPublicInput = {
...
+};
+
+export type RepresentationPublicInputOptions = {
+  segmentationRepresentationUID?: string;
+  // color lut to use for this representation (optional), it can
+  // be either a colorLUT array or the index of the colorLUT in the state
+  colorLUTOrIndex?: Types.ColorLUT | number;
+  // whether to use polymorphic segmentation utilities to convert
+  // from other representations to this representation
+  polySeg?: boolean;
```
wayfarer3130, line 283: hasPolymorphicSeg?  Just from the name, I was thinking it was a polygon segmentation?

packages/tools/src/tools/segmentation/SegmentSelectTool.ts
```diff
@@ -106,75 +116,28 @@ class SegmentSelectTool extends BaseTool {
...
-      const indexIJK = csUtils.transformWorldToIndex(imageData, worldPoint);
-
-      // since it is a stack we don't need to check the z
-      const flattenedIndex = indexIJK[0] + indexIJK[1] * image.columns;
-
-      const scalars = imageData.getPointData().getScalars().getData();
-
-      hoveredSegmentIndex = scalars[flattenedIndex];
-    }
+    const hoveredSegmentIndex = getSegmentAtWorldPoint(
```
wayfarer3130, line 122: What is the impact of this hover test call when there are a lot of segments?  Although it is nice to do on hover, I do wonder about the total cost of this and impact on interaction.

packages/tools/examples/PolySegWasmContourToStackLabelmap/index.ts
```diff
@@ -0,0 +1,302 @@
+import { RenderingEngine, Enums, eventTarget } from '@cornerstonejs/core';
```
wayfarer3130, line 1: Couldn't get this example working, not sure what is going wrong - it just never converts.

packages/tools/examples/PolySegWasmContourToStackLabelmap/index.ts
```diff
@@ -0,0 +1,302 @@
...
+  toolGroup1.addTool(ZoomTool.toolName);
+  toolGroup1.addTool(StackScrollMouseWheelTool.toolName);
+  toolGroup1.addTool(PlanarFreehandContourSegmentationTool.toolName);
+  toolGroup1.addTool(SegmentationDisplayTool.toolName);
+  toolGroup1.addTool(SegmentSelectTool.toolName);
+
+  // Segmentation Tools
+  toolGroup2.addTool(PanTool.toolName);
+  toolGroup2.addTool(StackScrollMouseWheelTool.toolName);
+  toolGroup2.addTool(TrackballRotateTool.toolName);
```
wayfarer3130, line 175: Not added up above.  Suggest adding trackball to the standard test tools, and removing them from here - should be ok as long as trackball doesn't fail on stack viewports.

packages/tools/examples/PolySegWasmContourToSurface/index.ts
```diff
@@ -0,0 +1,310 @@
...
+
+/**
+ * Runs the demo
+ */
+async function run() {
+  // Init Cornerstone and related libraries
+  await initDemo();
+
+  // Add tools to Cornerstone3D
+  cornerstoneTools.addTool(PanTool);
```
wayfarer3130, line 147: Please use standard tools, and then also add a interpolation tool as default so that it is easy to add a volume here.

packages/tools/examples/PolySegWasmContourToVolumeLabelmap/index.ts
```diff
@@ -0,0 +1,284 @@
...
+  cornerstoneTools.addTool(SegmentationDisplayTool);
+  cornerstoneTools.addTool(PlanarFreehandContourSegmentationTool);
+  cornerstoneTools.addTool(SegmentSelectTool);
+
+  // Define tool groups to add the segmentation display tool to
+  toolGroup1 = ToolGroupManager.createToolGroup(toolGroupId1);
+  toolGroup2 = ToolGroupManager.createToolGroup(toolGroupId2);
+
+  // Manipulation Tools
+  toolGroup1.addTool(PanTool.toolName);
```
wayfarer3130, line 143: Same comment as for surface - add the interpolation tools so it is easy to create a volume.

wayfarer3130 (commented):
Could you merge the conversion examples to select between stack/volume on the source, and stack/volume/volume3d on the destination, and only differentiate between conversion type?  That would make it easier to test/look at the options.

packages/tools/src/enums/Events.ts
```diff
@@ -164,6 +164,10 @@ enum Events {
    */
   SEGMENTATION_DATA_MODIFIED = 'CORNERSTONE_TOOLS_SEGMENTATION_DATA_MODIFIED',
 
+  POLYSEG_CONVERSION_STARTED = 'CORNERSTONE_TOOLS_POLYSEG_CONVERSION_STARTED',
```
wayfarer3130, line 167: Could you move to one type for polyseg, and just have a flag as to the progress (0 means start, and 100 means done, anything else is in between)?  Having too many types of events for a fairly fine grained change like that is harder to manage.

packages/tools/src/eventListeners/annotations/contourSegmentation/contourSegmentationCompleted.ts
```diff
@@ -129,25 +130,18 @@ function getValidContourSegmentationAnnotations(
   sourceAnnotation: ContourSegmentationAnnotation
 ): ContourSegmentationAnnotation[] {
   const { annotationUID: sourceAnnotationUID } = sourceAnnotation;
```
wayfarer3130, line 132: Definitely cleaner to use a getAll like this.

packages/tools/src/stateManagement/annotation/FrameOfReferenceSpecificAnnotationManager.ts
```diff
@@ -365,6 +365,24 @@ class FrameOfReferenceSpecificAnnotationManager implements IAnnotationManager {
     }
   };
 
+  getAllAnnotations = (): Annotations => {
+    const annotations = this.annotations;
+    const allAnnotations = [];
+
+    for (const groupKey in annotations) {
```
wayfarer3130, line 372: What about:
return Object.values(this.annotations).flat();

packages/tools/src/stateManagement/annotation/annotationState.ts
```diff
@@ -170,6 +175,7 @@ function removeAllAnnotations(): void {
 }
 
 export {
+  getAllAnnotations,
```
wayfarer3130, line 178: Thank you.

packages/tools/src/stateManagement/segmentation/polySeg/Labelmap/convertContourToLabelmap.ts
```diff
@@ -0,0 +1,285 @@
...
+      dimensions,
+      scalarData,
+      origin,
+      direction,
+      spacing,
+      annotationUIDsInSegmentMap,
+    },
+    {
+      callbacks: [
+        (progress) => {
```
wayfarer3130, line 90: Consider adding a progress trigger event - maybe with a callback to cancel it?  That would then call events with increasing progress count, and would all use a single event type.

packages/tools/src/stateManagement/segmentation/polySeg/Surface/convertContourToSurface.ts
```diff
@@ -0,0 +1,43 @@
...
+  const { annotationUIDsMap } = contourRepresentationData;
+
+  // loop over all annotations in the segment and flatten their polylines
+  const polylines = [];
+  const numPointsArray = [];
+  const annotationUIDs = annotationUIDsMap.get(segmentIndex);
+
+  for (const annotationUID of annotationUIDs) {
+    const annotation = getAnnotation(annotationUID);
+    const polyline = annotation.data.contour.polyline;
```
wayfarer3130, line 21: const { polyline } = annotation.data.contour

packages/tools/src/stateManagement/segmentation/polySeg/Surface/computeAndAddSurfaceRepresentation.ts
```diff
@@ -0,0 +1,28 @@
...
+ * @param segmentationId - The ID of the segmentation.
+ * @param options - Additional options for computing the surface representation.
+ * @param options.segmentIndices - The indices of the segments to compute the surface for.
+ * @param options.segmentationRepresentationUID - The UID of the segmentation representation to compute the surface for.
+ *
+ * @returns A promise that resolves when the surface representation is computed and added.
+ */
+export function computeAndAddSurfaceRepresentation(
+  segmentationId: string,
+  options: {
```
wayfarer3130, line 17: These options are common enough that you should extract the type and use it consistently for all calls where it is relevant.

packages/tools/src/stateManagement/segmentation/polySeg/computeAndAddRepresentation.ts
```diff
@@ -0,0 +1,98 @@
...
+function subscribeToSegmentationChanges(updateFunction) {
+  // Named function for handling the event
+  const handleSegmentationChange = (event) => {
+    _debouncedSegmentationModified(event, updateFunction);
+  };
+
+  // Remove the event listener for cleanup
+  eventTarget.removeEventListener(
+    Events.SEGMENTATION_DATA_MODIFIED,
+    handleSegmentationChange
```
wayfarer3130, line 74: This will never unsubscribe - the subscription removal does an === comparison on the function provided, and you end up creating a new one every time you call this.  You need to have the update function be the one that gets mapped with a ===, and then remove it based on identity on the updateFunction.

packages/tools/src/tools/displayTools/Contour/contourDisplay.ts
```diff
@@ -66,27 +72,91 @@ async function render(
...
   }
 
-  if (!geometryIds?.length) {
-    // This is a normal occurrence when contours activated, but not yet drawn
+  const contourData = segmentation.representationData[Representations.Contour];
+
+  const { geometryIds } = contourData;
+
+  // this means we would like to use vtk actors for contour data
+  // Note: We really should get out of
```
wayfarer3130, line 85: Dangling senten

packages/tools/src/tools/displayTools/Contour/contourDisplay.ts
```diff
@@ -66,27 +72,91 @@ async function render(
...
   // add the contour sets to the viewport
-  addOrUpdateContourSets(
+  addOrUpdateVTKContourSets(
     viewport,
     geometryIds,
     representationConfig,
     toolGroupConfig
   );
+
+  // if (contourData?.points?.length) {
```
wayfarer3130, line 101: Do you want to leave this commented out - if so, could you add a comment as to why this is still left in the code?

packages/tools/src/tools/base/ContourSegmentationBaseTool.ts
```diff
@@ -59,7 +59,7 @@ abstract class ContourSegmentationBaseTool extends ContourBaseTool {
       activeSegmentation.getActiveSegmentationRepresentation(toolGroupId);
 
     if (!activeSegmentationRepresentation) {
-      throw new Error(
+      console.warn(
```
wayfarer3130, line 62: You need to actively skip the segmentation specific changes here, otherwise the next line will throw.

wayfarer3130: You could create a non-segmentation contour, but I think that is a bad idea.  Could you revert to the throw?

packages/tools/src/stateManagement/segmentation/segmentationState.ts
```diff
@@ -97,6 +97,42 @@ function getAllSegmentationRepresentations(): Record<
...
+  segmentationRepresentationUID: string
+): {
+  toolGroupId: string;
+  segmentationRepresentation: ToolGroupSpecificRepresentation;
+} {
+  const allToolGroupRepresentations = getAllSegmentationRepresentations() || [];
+
+  const toolGroupIds = Object.keys(allToolGroupRepresentations);
+
+  for (let i = 0; i < toolGroupIds.length; i++) {
```
wayfarer3130, line 116: for (const [toolGroupId,toolGroupRepresentations] of Object.entries(getAllSegmentationRepresentations())) {

is in general a better idiom to use than looping through the indices on the key values.

Even just const toolGroupId of toolGroupIds is probably better.

packages/core/src/RenderingEngine/Viewport.ts
sedghi: There should not be any performance impact

packages/core/src/types/ISurface.ts
sedghi: polys ans setPolys is vtk terminology https://kitware.github.io/vtk-js/api/Common_DataModel_PolyData.html#setPolys

utils/ExampleRunner/template-config.js
sedghi: I think this is the new standard in webpack https://webpack.js.org/configuration/experiments/

> asyncWebAssembly: Support the new WebAssembly according to the [updated specification](https://github.com/WebAssembly/esm-integration), it makes a WebAssembly module an async module. And it is enabled by default when [experiments.futureDefaults](https://webpack.js.org/configuration/experiments/#experimentsfuturedefaults) is set to true.

packages/core/src/RenderingEngine/RenderingEngine.ts
sedghi: There is already fillCanvasWithBackgroundColor I think this is good enough abstraction

packages/tools/src/utilities/segmentation/getSegmentAtWorldPoint.ts
```diff
@@ -0,0 +1,146 @@
...
+      const { polyline } = annotation.data.contour;
+
+      if (
+        !utilities.isEqual(viewPlaneNormal, annotation.metadata.viewPlaneNormal)
+      ) {
+        continue;
+      }
+
+      if (isPointInsidePolyline3D(worldPoint, polyline)) {
```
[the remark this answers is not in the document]
sedghi: Being perfectly on the contour means being on the 0.1 precision in the canvas, which is highly unlikely. I added a comment to explain this.

packages/tools/src/utilities/segmentation/getSegmentAtWorldPoint.ts
sedghi: It has the segmentationRepresentationUID so that would not be an issue

packages/tools/src/utilities/math/polyline/isPointInsidePolyline.ts
sedghi: I switched the what Leo implemented so all of this can be resolved

packages/tools/src/utilities/math/polyline/isPointInsidePolyline.ts
sedghi: will add this to the containsPoint in Leo's PR

packages/tools/src/utilities/contours/getDeduplicatedVTKPolyDataPoints.ts
sedghi: I just renamed that function, I will add todo

packages/tools/src/utilities/segmentation/getSegmentAtWorldPoint.ts
sedghi: I believe it would be great to include that, but unfortunately, the budget for this ticket has been used up. I have added it to the to-do list.

packages/tools/src/types/SegmentationStateTypes.ts
sedghi: I'd like to use the naming convention of the 3D Slicer, which is polySEG. You can find it at this link: [PolySeg](https://github.com/PerkLab/PolySeg). Is that alright with you?

packages/tools/src/tools/segmentation/SegmentSelectTool.ts
sedghi: This will test against only one representation UID (active one), so it will not have performance degradation.

packages/tools/examples/PolySegWasmContourToStackLabelmap/index.ts
sedghi: I think i have seen this I will investigate more today

packages/tools/src/utilities/segmentation/getSegmentAtWorldPoint.ts
```diff
@@ -0,0 +1,146 @@
...
+): number | undefined {
+  if (isVolumeSegmentation(labelmapData, viewport)) {
+    const { volumeId } = labelmapData as LabelmapSegmentationDataVolume;
+    const segmentationVolume = cache.getVolume(volumeId);
+
+    if (!segmentationVolume) {
+      return;
+    }
+
+    const segmentIndex =
```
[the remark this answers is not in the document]
sedghi: I don't understand, can you rephrase?

packages/tools/examples/PolySegWasmContourToStackLabelmap/index.ts
sedghi: I switched to Leo's code for containsPoint and seem to work based on my tests

packages/tools/src/stateManagement/segmentation/polySeg/polySegConverters.js
```diff
@@ -0,0 +1,387 @@
...
+      imageData.modified();
+
+      segmentationVoxelManagers.set(referencedImageId, { manager, imageData });
+    });
+
+    for (const index of segmentIndices) {
+      const annotations = annotationUIDsInSegmentMap.get(index);
+
+      for (const annotation of annotations) {
+        if (!annotation.data) {
```
wayfarer3130, line 187: I get an undefined occasionally for annotation - not that annotation.data isn't defined, but that annotation itself is undefined.

packages/tools/src/eventListeners/annotations/contourSegmentation/contourSegmentationCompleted.ts
```diff
@@ -129,25 +129,18 @@ function getValidContourSegmentationAnnotations(
...
-        targetAnnotation.annotationUID !== sourceAnnotationUID &&
-        contourSegUtils.isContourSegmentationAnnotation(targetAnnotation) &&
-        contourSegUtils.areSameSegment(targetAnnotation, sourceAnnotation) &&
-        contourUtils.areCoplanarContours(targetAnnotation, sourceAnnotation)
-    );
-
-    return validAnnotations.concat(toolAnnotations);
-  }, []);
+  const allAnnotations = getAllAnnotations();
+  debugger;
```
wayfarer3130, line 135: Still seeing some debuggers around.

packages/tools/examples/PolySegWasmContourToStackLabelmap/index.ts
```diff
@@ -0,0 +1,245 @@
...
+  setTitleAndDescription,
+  addButtonToToolbar,
+  addDropdownToToolbar,
+  createInfoSection,
+  addManipulationBindings,
+} from '../../../../utils/demo/helpers';
+import * as cornerstoneTools from '@cornerstonejs/tools';
+
+// This is for debugging purposes
+console.warn(
```
wayfarer3130, line 14: This example works the first time now, but does not convert a second time.

packages/tools/examples/segmentSelect/index.ts
```diff
@@ -38,41 +39,48 @@ let renderingEngine;
...
 
-const stackViewportId = 'STACK_VIEWPORT';
-const volumeViewportId = 'VOLUME_VIEWPORT';
-
-const stackToolGroupId = 'TOOL_GROUP_STACK';
-const volumeToolGroupId = 'TOOL_GROUP_VOLUME';
-
 // ======== Set up page ======== //
 setTitleAndDescription(
   'Segment Select Tool in Both Stack and Volume Viewport',
```
wayfarer3130, line 44: There is some odd behaviour on this - I can select sometimes, but not consistently.  Sometimes it goes to select and sometimes to brush tools.

wayfarer3130: OK, read the code on the select tool.  It is a hover select on delay.  Please update the example to explain to hover over the existing annotations to select one or the other. 
Also, could you add pre-existing annotations for the bottom row so that it is testable?  I think they are supposed to be there, just that I don't see anything.

packages/tools/examples/surfaceRenderingStack/index.ts
```diff
@@ -46,6 +41,7 @@ const volumeLoaderScheme = 'cornerstoneStreamingImageVolume'; // Loader id which
 const volumeId = `${volumeLoaderScheme}:${volumeName}`; // VolumeId with loader id + volume id
 const toolGroupIdStack = 'MY_TOOLGROUP_ID_STACK';
 const toolGroupIdVolume = 'MY_TOOLGROUP_ID_VOLUME';
+const segmentationId = 'MY_SEGMENTATION_ID';
 
 // ======== Set up page ======== //
 setTitleAndDescription(
```
wayfarer3130, line 47: Could you initially navigate the stack to something that shows the data?  It wasn't obvious to me what to do to get it to show up.  Once I started scrolling, then I started seeing the data, so I think it is either a missing render on complete or the wrong position initially.

packages/tools/src/stateManagement/segmentation/polySeg/Labelmap/convertContourToLabelmap.ts
```diff
@@ -0,0 +1,275 @@
...
+const workerManager = getWebWorkerManager();
+
+export async function convertContourToVolumeLabelmap(
+  contourRepresentationData: ContourSegmentationData,
+  options: PolySegConversionOptions = {}
+) {
+  const { viewport } = options;
+
+  let imageIds;
+  if (viewport instanceof VolumeViewport) {
```
wayfarer3130, line 33: Could you move this into a utility function to abstract the idea of getting a set of imageIds?

packages/tools/src/stateManagement/segmentation/polySeg/Labelmap/convertContourToLabelmap.ts
```diff
@@ -0,0 +1,275 @@
...
+    const defaultActor = viewport.getDefaultActor();
+    const volumeId = defaultActor.uid;
+    const volume = cache.getVolume(volumeId);
+    imageIds = volume.imageIds;
+  } else if (viewport instanceof StackViewport) {
```
wayfarer3130, line 38: How about  if( (viewport as IStackViewport).getImageIds ) ...
That way it works for anything that can get a list of imageIds.  Even better, for VolumeViewport, you could implement getImageIds as the code you just have above, then you don't need to care about whether it is volume or stack or video or ....

packages/tools/src/stateManagement/segmentation/polySeg/Labelmap/convertContourToLabelmap.ts
```diff
@@ -0,0 +1,275 @@
...
+      scalarData,
+      origin,
+      direction,
+      spacing,
+      annotationUIDsInSegmentMap,
+    },
+    {
+      callbacks: [
+        (progress) => {
+          triggerEvent(eventTarget, Events.POLYSEG_CONVERSION, { progress });
```
wayfarer3130, line 91: Thank you.

packages/tools/src/stateManagement/segmentation/polySeg/Surface/convertContourToSurface.ts
```diff
@@ -0,0 +1,48 @@
+import { Types, eventTarget, triggerEvent } from '@cornerstonejs/core';
+import { getWebWorkerManager } from '@cornerstonejs/core';
+import { ContourSegmentationData } from '../../../../types';
+import { getAnnotation } from '../../../annotation/annotationState';
+import { Events } from '../../../../enums';
+
+const workerManager = getWebWorkerManager();
+
+export async function convertContourToSurface(
```
wayfarer3130, line 9: docs please - useful to know that this calls the wasm conversion as a webworker manager and returns the promise when it is done converting, with the provided data.

packages/tools/src/stateManagement/segmentation/polySeg/Surface/convertLabelmapToSurface.ts
```diff
@@ -0,0 +1,60 @@
...
+  LabelmapSegmentationData,
+  LabelmapSegmentationDataStack,
+  LabelmapSegmentationDataVolume,
+} from '../../../../types/LabelmapTypes';
+import { computeVolumeSegmentationFromStack } from '../../convertStackToVolumeSegmentation';
+import { Events } from '../../../../enums';
+
+const workerManager = getWebWorkerManager();
+
+export async function convertLabelmapToSurface(
```
wayfarer3130, line 13: At least a short docs here please

wayfarer3130: Need to know that this will fail if the volume segmentation from stack fails - least that is what it appear to happen to me.  Certainly that is an important bit of information to know when calling this against stack data.

packages/tools/src/stateManagement/segmentation/polySeg/Surface/createAndCacheSurfacesFromRaw.ts
```diff
@@ -0,0 +1,76 @@
...
+    const segmentIndex = rawSurfaceData.segmentIndex;
+
+    // Get the color either from the segmentation representation or randomly generated
+    const color = segmentationRepresentation
+      ? getColorForSegmentIndex(
+          toolGroupId,
+          segmentationRepresentation.segmentationRepresentationUID,
+          segmentIndex
+        ).slice(0, 3)
+      : [Math.random() * 255, Math.random() * 255, Math.random() * 255];
```
wayfarer3130, line 50: Does this backup happen often?  I think this should be moved into a function if it happens regularly.

packages/tools/src/stateManagement/segmentation/polySeg/Surface/surfaceComputationStrategies.ts
```diff
@@ -0,0 +1,157 @@
+import { Types } from '@cornerstonejs/core';
+import {
```
wayfarer3130, line 2: Can you import with type - just that it avoids circular dependencies, and I see you export a type down below, so I think it should have the types.

packages/tools/src/stateManagement/segmentation/polySeg/Surface/surfaceComputationStrategies.ts
```diff
@@ -0,0 +1,157 @@
...
+        | LabelmapSegmentationDataVolume
+        | LabelmapSegmentationDataStack,
+      index,
+      isVolume
+    );
+
+    return surface;
+  });
+
+  const surfaces = await Promise.all(promises);
```
wayfarer3130, line 114: Should this allow reporting individual failures when they occur, using alLSettled perhaps, and then mapping to an error issue below?

packages/tools/src/stateManagement/segmentation/polySeg/Surface/updateSurfaceData.ts
```diff
@@ -0,0 +1,99 @@
...
+            [{ segmentIndex, data }],
+            {
+              segmentationRepresentationUID:
+                segmentationRepresentation.segmentationRepresentationUID,
+            }
+          );
+        });
+      });
+    } else {
+      if (indices.includes(segmentIndex)) {
```
wayfarer3130, line 80: You don't need to nest this if condition, it can be directly included in the else if on the previous line.

packages/tools/src/stateManagement/segmentation/polySeg/Surface/updateSurfaceData.ts
```diff
@@ -0,0 +1,99 @@
...
+        // if the geometry already exists and the segmentIndex is
+        // still present, update the geometry data
+        const surface = geometry.data as Types.ISurface;
```
wayfarer3130, line 83: Don't need a return here or the next one.

packages/tools/src/stateManagement/segmentation/polySeg/Surface/updateSurfaceData.ts
```diff
@@ -0,0 +1,99 @@
...
+        surface.setPoints(data.points);
+        surface.setPolys(data.polys);
+        return;
```
wayfarer3130, line 86: This return will result in undefined in the promises list.  Checked on the Promise.all method and that just results in an undefined result for that entry, which is ok.  Might add a comment here saying something along those lines, as I certainly didn't know what happened in Promise.all on undefined values.

packages/tools/src/stateManagement/segmentation/polySeg/canComputeRequestedRepresentation.ts
```diff
@@ -0,0 +1,122 @@
...
+    } else {
+      supportedTypes.push(representationType);
+    }
+  });
+
+  return supportedTypes;
+}
+
+async function canConvertFromTo(fromRepresentationType, toRepresentationType) {
+  const availablePaths = conversionPaths.get(fromRepresentationType);
```
wayfarer3130, line 113: return conversionPaths.get(fromRepresentationType)?.has(toRepresentationType) || false;
is a one line version of this if you want.

packages/tools/src/stateManagement/segmentation/polySeg/polySegConverters.js
```diff
@@ -0,0 +1,387 @@
+import { expose } from 'comlink';
+import vtkImageData from '@kitware/vtk.js/Common/DataModel/ImageData';
+import vtkDataArray from '@kitware/vtk.js/Common/Core/DataArray';
+import ICRPolySeg from '@icr/polyseg-wasm';
+import { utilities } from '@cornerstonejs/core';
+import { getBoundingBoxAroundShapeWorld } from '../../../utilities/boundingBox';
+import { pointInShapeCallback } from '../../../utilities';
+import { isPointInsidePolyline3D } from '../../../utilities/math/polyline';
+
+const obj = {
```
wayfarer3130, line 10: Can we call this something else - like polySegConverters

wayfarer3130: Also, would like a bit of docs here saying this is an exposed set of objects shared between webworker and primary contexts,  `@see ....` so that there is a bit of help to someone new getting into this framework.  If I hadn't seen your earlier reviews, I would be COMPLETELY clueless as to what you were doing here.

packages/tools/src/stateManagement/segmentation/polySeg/polySegConverters.js
```diff
@@ -0,0 +1,387 @@
...
+  polySeg: null,
+  polySegInitializing: false,
+  polySegInitializingPromise: null,
+  async initializePolySeg(progressCallback) {
```
wayfarer3130, line 14: Could these be static members of a class instead of object members in a typescript class?  Don't know if that works, but if it does it would be nice.

packages/tools/src/stateManagement/segmentation/polySeg/polySegConverters.js
```diff
@@ -0,0 +1,387 @@
...
+        })
+        .then(() => {
+          this.polySegInitializing = false;
+          resolve();
+        });
+    });
+
+    await this.polySegInitializingPromise;
+  },
+  async convertContourToSurface(args, ...callbacks) {
```
wayfarer3130, line 39: Can you at least add spacing between functions and maybe a short comment?

packages/tools/src/types/SegmentationStateTypes.ts
```diff
@@ -107,6 +107,13 @@ export type ToolGroupSpecificRepresentationState = {
    * using to render
    */
   colorLUTIndex: number;
+  /**
+   * Poly Seg generated
```
wayfarer3130, line 111: Can you spell out in comment that this is a polymorphic segmentation?  At least if you get some information in the type definition it means it isn't a "polygon" or other "poly" type.

packages/tools/src/utilities/segmentation/getSegmentAtWorldPoint.ts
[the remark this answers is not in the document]
wayfarer3130: I think this is sufficiently isolated for now, and we can look into adding utilities later.

packages/tools/src/utilities/segmentation/getSegmentAtWorldPoint.ts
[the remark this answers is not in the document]
wayfarer3130: If you are displaying 1:1 then it is almost certain to occur.  However, lets leave it for now.

utils/demo/helpers/addManipulationBindings.ts
```diff
@@ -22,22 +24,35 @@ const { MouseBindings, KeyboardBindings } = csToolsEnums;
  * * Stack Scroll on Mouse Wheel, Primary+Alt
  * * Length Tool on fourth button
  */
-export default function addManipulationBindings(toolGroup, register = true) {
-  if (register) {
-    cornerstoneTools.addTool(LengthTool);
-    cornerstoneTools.addTool(StackScrollMouseWheelTool);
+export default function addManipulationBindings(
+  toolGroup,
+  options: {
```
wayfarer3130, line 29: I think this is a good option as it allows us to add standard bindings for other types of views, like volume.

packages/tools/src/stateManagement/segmentation/polySeg/polySegConverters.js
sedghi: I think I saw that sometime (that is why i have a debugger) but tried it again and can't reproduce it at all

packages/tools/examples/PolySegWasmContourToStackLabelmap/index.ts
sedghi: I will enhance this later to update on any representation edits, when I do the contours

packages/tools/examples/segmentSelect/index.ts
sedghi: Sure I will add some in the bottom

packages/tools/src/stateManagement/segmentation/polySeg/Surface/createAndCacheSurfacesFromRaw.ts
sedghi: Not really, I will remove it it was for debugging

packages/tools/src/stateManagement/segmentation/polySeg/polySegConverters.js
sedghi: It does not work last time i tried the worker stuff

packages/tools/examples/surfaceRenderingStack/index.ts
sedghi: I will re-work the stack segmentation with surface intersections pretty soon, but rather have the PR reviewed

packages/tools/src/tools/segmentation/SegmentSelectTool.ts
```diff
@@ -42,7 +42,7 @@ class SegmentSelectTool extends BaseTool {
       configuration: {
         hoverTimeout: 100,
         mode: SegmentSelectTool.SelectMode.Border,
-        searchRadius: 20, // search for border in a 10px radius
+        searchRadius: 6, // search for border in a 10px radius
```
wayfarer3130, line 45: Comment doesn't match - please fix that to just specify the type of radius.

wayfarer3130 (approved):
Looks good now.

PR #863, opened 2023-11-01
feat(registration): example (poc)

lscoder:
Registration (POC)

### DIFF
--- a/packages/core/examples/registration/index.ts
+++ b/packages/core/examples/registration/index.ts
@@ -1,0 +1,852 @@
+import {
+  defaultParameterMap,
+  elastix,
+  DefaultParameterMapOptions,
+  DefaultParameterMapResult,
+  ElastixOptions,
+} from '@itk-wasm/elastix';
+import { Image, ImageType, FloatTypes, PixelTypes, Metadata } from 'itk-wasm';
+import {
+  RenderingEngine,
+  Types,
+  Enums,
+  cache,
+  volumeLoader,
+  getRenderingEngine,
+} from '@cornerstonejs/core';
+import {
+  initDemo,
+  createImageIdsAndCacheMetaData,
+  setCtTransferFunctionForVolumeActor,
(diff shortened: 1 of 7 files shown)


packages/core/package.json
```diff
@@ -31,6 +31,7 @@
   },
   "dependencies": {
     "@kitware/vtk.js": "27.3.1",
+    "@itk-wasm/elastix": "0.2.2",
```
sedghi, line 34: this should not be in the core, just the examples

packages/core/examples/registration/index.ts
```diff
@@ -0,0 +1,852 @@
...
+  'bspline',
+  'spline',
+  // 'groupwise', // 2D+time or 3D+time
+];
+
+let activeTransformName = transformNames[1];
+let currentParameterMap = {};
+
+const defaultParameterMaps = {};
+const parametersSettings = {
```
sedghi, line 131: maybe extract these out into a separeate file so that example becomes smaller

packages/core/examples/registration/index.ts
```diff
@@ -0,0 +1,852 @@
...
+  const month = `0${now.getMonth() + 1}`.slice(-2);
+  const year = now.getFullYear();
+  const hours = `0${now.getHours()}`.slice(-2);
+  const minutes = `0${now.getMinutes()}`.slice(-2);
+  const seconds = `0${now.getSeconds()}`.slice(-2);
+  const ms = `00${now.getMilliseconds()}`.slice(-3);
+
+  return `${year}-${month}-${day} ${hours}:${minutes}:${seconds}.${ms}`;
+}
+
```
sedghi, line 600: this can be in a utils file that sits beside this example, just to save space on this example

sedghi: along with other utilities

sedghi (commented):
see my comments

PR #868, opened 2023-11-06
Use abstract methods for shared information on viewports

wayfarer3130:
### Context

With more viewport types being defined, having custom logic in the individual viewports that isn't abstract in the base
interface becomes difficult.  This change adds more behaviour into the base Viewport class, and makes it abstract.

### Changes & Results

Make Viewport abstract
Add getNumberOfFrames to IViewport and implement it in the viewport types.
Use getNumberOfFrames instead of having custom per-viewport logic to determine number of frames.

### Testing

Run the example page for the videoTools viewport and scroll images with left click up/down
Run another example with the stack scroll tool enabled.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  


packages/core/src/RenderingEngine/VideoViewport.ts
```diff
@@ -339,6 +352,10 @@ class VideoViewport extends Viewport implements IVideoViewport {
     return true;
   };
 
+  public getNumberOfSlices(): number {
```
wayfarer3130, line 355: Here is the implementation of getNumberOfSlices for videos.

packages/core/src/RenderingEngine/Viewport.ts
```diff
@@ -44,7 +44,7 @@ import IImageCalibration from '../types/IImageCalibration';
  * which is camera properties/methods, vtk.js actors, and other common
  * logic.
  */
-class Viewport implements IViewport {
+abstract class Viewport implements IViewport {
```
wayfarer3130, line 47: Note that viewport is abstract now

packages/core/src/RenderingEngine/VolumeViewport.ts
```diff
@@ -82,6 +83,13 @@ class VolumeViewport extends BaseVolumeViewport {
     return super.setVolumes(volumeInputArray, immediate, suppressEvents);
   }
 
+  /** Gets the number of slices the volume is broken up into in the camera direction */
+  public getNumberOfSlices(): number {
```
wayfarer3130, line 87: This is the implementation used elsewhere, seems like it is reasonable to just call it from here.

packages/tools/examples/videoTools/index.ts
```diff
@@ -156,18 +159,35 @@ addDropdownToToolbar({
...
+
 /**
  * Runs the demo
  */
 async function run() {
   // Init Cornerstone and related libraries
   await initDemo();
 
+  // Get Cornerstone imageIds and fetch metadata into RAM
+  const imageIds = await createImageIdsAndCacheMetaData({
```
wayfarer3130, line 172: See how similar this looks to other retrieval types.

packages/tools/examples/videoTools/index.ts
```diff
@@ -191,16 +211,27 @@ async function run() {
       {
         mouseButton: MouseBindings.Auxiliary, // Middle Click
       },
+      {
```
wayfarer3130, line 214: Tweaked the bindings so that it would work without a mouse or with a mouse effectively.

packages/tools/src/tools/StackScrollTool.ts
```diff
@@ -76,21 +75,11 @@ class StackScrollTool extends BaseTool {
...
   _getPixelPerImage(viewport) {
     const { element } = viewport;
-    const numberOfSlices = this._getNumberOfSlices(viewport);
+    const numberOfSlices = viewport.getNumberOfSlices();
 
     // The Math.max here makes it easier to mouseDrag-scroll small or really large image stacks
     return Math.max(2, element.offsetHeight / Math.max(numberOfSlices, 8));
   }
-
-  _getNumberOfSlices(viewport) {
```
wayfarer3130, line 85: Will start replacing some of the viewport specific code with generalized code.

packages/core/examples/video/index.ts
```diff
@@ -100,13 +101,30 @@ addButtonToToolbar({
   },
 });
 
+const useLocal = true;
```
sedghi, line 104: can you remove this pleaes?

packages/core/src/RenderingEngine/VideoViewport.ts
```diff
@@ -82,6 +88,26 @@ class VideoViewport extends Viewport implements IVideoViewport {
     this.videoElement.remove();
   }
 
+  public setVideoImageId(
+    imageIds: string | string[],
+    currentImageIdIndex = 0
+  ): Promise<unknown> {
```
sedghi, line 94: As this is a video specific method, I don't see any harm in not calling the currentImageIdIndex a `timePoint` ? or `frameNumber` ?

packages/core/src/RenderingEngine/VolumeViewport.ts
```diff
@@ -25,6 +25,7 @@ import BaseVolumeViewport from './BaseVolumeViewport';
 import setDefaultVolumeVOI from './helpers/setDefaultVolumeVOI';
 import { setTransferFunctionNodes } from '../utilities/transferFunctionUtils';
 import { ImageActor } from '../types/IActor';
+import { utilities } from '..';
```
sedghi, line 28: this will usually cause a circular dependency, better to import the actual utility you need yourself

packages/tools/examples/videoTools/index.ts
```diff
@@ -156,18 +159,35 @@ addDropdownToToolbar({
   },
 });
 
+const useLocal = true;
```
sedghi, line 162: please remove useLocal

packages/core/examples/video/index.ts
```diff
@@ -100,13 +101,30 @@ addButtonToToolbar({
...
+
 /**
  * Runs the demo
  */
 async function run() {
+  ///dicomweb/studies//series//metadata
```
sedghi, line 110: please remove

packages/core/src/enums/MetadataModules.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Cornerstone Core events
```
sedghi, line 2: wrong doc, maybe mention we should eventually move everything to this

packages/core/src/enums/MetadataModules.ts
```diff
@@ -0,0 +1,9 @@
...
+ */
+enum MetadataModules {
+  CINE_MODULE = 'cineModule',
```
sedghi, line 5: can we follow the naming convention similar to other enums? CINE_MODULE

packages/dicomImageLoader/src/imageLoader/wadors/metaData/metaDataProvider.ts
```diff
@@ -1,3 +1,4 @@
+import { utilities, Enums } from '@cornerstonejs/core';
```
sedghi, line 1: we don't get cornerstone like this in the dicomImageLoader, use `external`

packages/dicomImageLoader/src/imageLoader/wadors/metaData/metaDataProvider.ts
```diff
@@ -251,9 +262,11 @@ function metaDataProvider(type, imageId) {
...
   if (type === 'transferSyntax') {
-    return {
-      transferSyntaxUID: getValue<string>(metaData['00020010']),
-    };
+    return getTransferSyntax(imageId, metaData);
+  }
+
+  if (type === MetadataModules.CINE_MODULE) {
+    return getCineModule(imageId, metaData);
   }
```
sedghi, line 270: why duplicated cine module?

packages/core/src/enums/MetadataModules.ts
```diff
@@ -0,0 +1,9 @@
...
+  IMAGE_URL_MODULE = 'imageUrlModule',
```
sedghi, line 6: I saw the implemntation of this, so our naming convention is `imageId = schema + imageURI``, can you name this `IMAGE_URI_MODULE` ?

sedghi: Hmmm looking again at this, I'm wrong. This seems to be a lot specific to video, can we call this then `VIDEO_URL_MODULE`?

sedghi (commented):
great stuff, see my comments, thanks

packages/core/examples/video/index.ts
wayfarer3130: No, it is important to be able to test against local easily.  I will set it to false by default  - wasn't supposed to get committed true, but the basic ability is important for consistent testing.

packages/core/src/RenderingEngine/VideoViewport.ts
wayfarer3130: Used frameNumber - it isn't a time point, although if preferred I could set it up to take both arguments to specify what type of position it is?

packages/tools/examples/videoTools/index.ts
wayfarer3130: Needed for testing purposes, and it is much cleaner to change a variable.  Ideally this should come from the environment variables so one can build either way.  However, it does need to default to false.

sedghi: You are hardcoding the port 5000 which is another problem

packages/core/src/enums/MetadataModules.ts
wayfarer3130: There is currently some specific to video, but it is a general image url module, neither URI nor video specific.
The intent is to use it for things like thumbnail URLs, rendered endpoints etc, all of which are URL based, not just URI, and those are not necessarily video related.

wayfarer3130: Some of the non-video stuff is things like PDF viewing - which happens to use the same rendered endpoint, I just haven't implemented it yet.

Issue #713, opened 2023-07-31
[Bug] Error in the interaction of ROI and Rotation tools 


mauricio-cdr:
### Describe the Bug

Hello, First and foremost, I want to extend my heartfelt appreciation and admiration for the work you are doing.

The error I found is the following.
When establishing a measurement using the Elliptical ROI Tool, RectangleROITool and then rotating the image the drawing of the tool is deformed

### Steps to Reproduce

1. Establish a measurement using the Elliptical ROI Tool or Rectangle Tool
2. Rotate the image by an angle other than 90° or 80°

### The current behavior

When rotating the image the drawing of the tool is deformed

https://github.com/cornerstonejs/cornerstone3D/assets/17281855/b29d76ea-5e33-4f5b-81d3-a92d73a6ab13

https://github.com/cornerstonejs/cornerstone3D/assets/17281855/3d12c698-2cd5-429b-af84-01c2f0e3a5c4

### The expected behavior

That when rotating the image the drawing of the tool remains

### OS

macOs12.3 , Windows 10, Linux Mint

### Node version

16.14.0

### Browser

Chrome 90+

sedghi:
Yeah seems like it is incorrectly rotating those points (other than 90 or 180)

jbocce:
After some investigation, I discovered that the problem is caused by using the svg `<rect>` and `<ellipse>` elements which draw rectangles and ellipses respectively that are "square" with the canvas element that is rendering the image. So for example, for rotating a rectangle by a non 90 degree multiple, the top-left and bottom-right points are each transformed (correctly) into canvas space (which is a space containing the rotation), but then the svg rectangle is rendered square to the canvas element itself and "drawn" from the transformed top-left point to the transformed bottom-right point. Two problems here:

- the `<rect>` element should NOT be rendered square to the canvas
- the transformed top-left and bottom-right points are NOT necessarily the points to create the `<rect>` element with

If instead an svg `<path>` were used to render the rectangle then each rectangle vertex could be transformed into the rotated canvas space and "joined" in their natural/rectangle ordering. Similarly an ellipse can be rendered with the `<path>` element using its elliptical arc segments.

A prototype solution was quickly tried for correcting the ellipse ROI problem and worked well.

More investigation needed to address (potentially some and maybe all) the following...
- implement rectangle ROI fix (similar to ellipse)
- address any issues with adding and editing ellipse and rectangle ROI in a rotated image space; it might be awkward, but investigation needed
[...]

PR #875, opened 2023-11-08
fix(elliptical): ROI and Rotation Tool Interaction (EllipticalROITool Only)

kom482:
### Context

This fix addresses the[issue #713](https://github.com/cornerstonejs/cornerstone3D/issues/713) where an error occurred in the interaction between the EllipticalROITool and Rotation tools

### Changes & Results

 - Adjusted the `drawEllipse` function by passing `canvasCoordinates` and executing precise calculations for `cx`, `cy`, `rx`, `ry`, and rotation transform angle.
- Replaced `corner1` and `corner2` parameters with `canvasCoordinates` for the `drawEllipse` function, enhancing accuracy.
- Refined `center`, `rotation`, and `radius` calculations for the ellipse's SVG attributes.
- Introduced the `transform: rotate(angle cx cy)` ellipse attribute for appropriate transformations.

Before: 
[issue #713](https://github.com/cornerstonejs/cornerstone3D/issues/713) videos

After:
https://github.com/cornerstonejs/cornerstone3D/assets/2040590/0902358b-1cb3-4009-a3aa-a054b71f6ef3

### Testing

1. run `yarn example stackAnnotationTools`
2. draw multiply EllipticalROITool annotations
3. change viewport rotation

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] OS: Windows 11 
- [x] Node version: v18.17.1 
- [x] "Browser: Chrome  119.0.6045.105
  


packages/tools/examples/stackAnnotationTools/index.ts
```diff
@@ -180,9 +181,14 @@ addButtonToToolbar({
   },
 });
 
-addButtonToToolbar({
-  title: 'Rotate Delta 90',
-  onClick: () => {
+addSliderToToolbar({
```
sedghi, line 184: I know you used this example for debugging, but I rather keep this one simple so can you take this out please?

kom482: Thanks for the suggestion! I've completed the requested tasks. Should I create new `/tools/examples/.../index.ts` example where it is possible to add ROI tools and enable rotation or it's ok for now?

packages/tools/src/drawingSvg/drawEllipse.ts
```diff
@@ -9,8 +9,7 @@ function drawEllipse(
   svgDrawingHelper: SVGDrawingHelper,
   annotationUID: string,
   ellipseUID: string,
-  corner1: Types.Point2,
-  corner2: Types.Point2,
+  canvasCoordinates: [Types.Point2, Types.Point2, Types.Point2, Types.Point2],
```
sedghi, line 12: Is there any way that we don't change the signature of this function to avoid causing breaking changes?

kom482: Thank you for your feedback. 
It's necessity of having more than two corner points to accurately determine the ellipse rotation angle. 
In my opinion, passing all four coordinates is a logical solution.
But I appreciate your flexibility and willingness to maintain the current state of the `drawEllipse` function. 
For me, there are two possible ways to address this issue:

1. Implement `drawEllipseByCoordinates`: This option provides a specific and precise solution for users who require accurate ellipse rendering. Keeping `drawEllipse` unchanged ensures backward compatibility, allowing users to choose the method that best fits their needs.

2. Pass Rotation Angle with `options` object with a default value of 0. This maintains simplicity for users who don't require rotation while offering enhanced functionality for those who do.

I think I will stick to the first option, as it seems to be more developer-friendly.

kom482: @sedghi changes was pushed, ping me if you prefer to stick to second option with `options` object

sedghi (approved):
Thanks for your contribution

kom482:
thank you for approve, next step: fix for RectangleROITool 🙂

sedghi:
That would be super great if you can do that!

PR #876, opened 2023-11-08
feat(video): Apply window level and color balance

wayfarer3130:
### Context

For the video viewport, apply the window level and colour balance using an SVG/CSS filter transform.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  


packages/core/examples/video/index.ts
```diff
@@ -100,13 +101,29 @@ addButtonToToolbar({
...
 async function run() {
   // Init Cornerstone and related libraries
   await initDemo();
 
+  // Get Cornerstone imageIds and fetch metadata into RAM
+  const imageIds = await createImageIdsAndCacheMetaData({
+    StudyInstanceUID: '2.25.96975534054447904995905761963464388233',
+    SeriesInstanceUID: '2.25.15054212212536476297201250326674987992',
+    wadoRsRoot: useLocal
+      ? 'http://localhost:5000/dicomweb'
```
sedghi, line 118: use the other approach

packages/core/src/RenderingEngine/VideoViewport.ts
```diff
@@ -36,6 +42,14 @@ class VideoViewport extends Viewport implements IVideoViewport {
     parallelScale: 1,
   };
 
+  windowLevelTransform: mat4;
+  colorBalanceTransform: mat4;
+  colorTransform: mat4;
+
+  feFilter: string;
+  averageWhite: [number, number, number];
+  windowLevel: { windowWidth: number; windowCenter: number };
```
sedghi, line 51: some documentation above each is appreaciated

packages/core/src/RenderingEngine/VideoViewport.ts
```diff
@@ -82,6 +96,30 @@ class VideoViewport extends Viewport implements IVideoViewport {
...
 
+  /**
+   * Sets the video image id to show and hte frame number.
+   * Requirements are to have the imageUrlModule in the metadata
+   * with the rendered endpoint being the raw video in video/mp4 format.
+   */
+  public setVideoImageId(
+    imageIds: string | string[],
+    frameNumber?: number
+  ): Promise<unknown> {
```
sedghi, line 107: This feels a bit unusuall, since you say `setVideoImageId` but at the same time it accepts imageIds and also frameNumber. 

How about setVideo?

packages/core/src/RenderingEngine/VideoViewport.ts
```diff
@@ -82,6 +96,30 @@ class VideoViewport extends Viewport implements IVideoViewport {
...
+    const { rendered } = metaData.get(MetadataModules.IMAGE_URL, imageId);
+    return this.setVideoURL(rendered).then(() => {
+      const { cineRate = 30 } = metaData.get(MetadataModules.CINE, imageId);
+      this.fps = cineRate;
+      if (frameNumber !== undefined) {
+        this.pause();
+        this.setFrame(frameNumber);
+      }
+      this.setAverageWhite([180, 255, 255]);
+      this.setWindowLevel(128, 128);
```
sedghi, line 119: I think these should get deleted if we want to merge

packages/core/src/RenderingEngine/VideoViewport.ts
```diff
@@ -269,6 +305,46 @@ class VideoViewport extends Viewport implements IVideoViewport {
...
+
+  public setAverageWhite(averageWhite: [number, number, number]) {
+    this.averageWhite = averageWhite;
+    this.setColorTransform();
+  }
+
+  public setColorTransform() {
+    if (!this.windowLevel && !this.averageWhite) {
+      this.feFilter = null;
+    } else {
```
sedghi, line 321: I usually prefer/recommend less indentation, so i would go with

```js
if (!this.windowLevel && !this.averageWhite) {
    this.feFilter = null;
    // probably you need to add the following I guess
    this.canvas.style.filter = '';
    return;
}

// rest of unindented code
```

packages/core/src/RenderingEngine/VideoViewport.ts
```diff
@@ -269,6 +305,46 @@ class VideoViewport extends Viewport implements IVideoViewport {
...
+        <feColorMatrix type="matrix" \
+        values="\
+          ${scaleWhite[0] * wlScale} 0 0 0 ${wlDelta} \
+          0 ${scaleWhite[1] * wlScale} 0 0 ${wlDelta} \
+          0 0 ${scaleWhite[2] * wlScale} 0 ${wlDelta} \
+          0 0 0 1 0" />\
+        </filter>\
+      </svg>#colour')`;
+    }
+    console.log('Setting canvas filter', this.feFilter);
```
sedghi, line 343: we need to remove these logs for merge

packages/core/src/RenderingEngine/VideoViewport.ts
```diff
@@ -269,6 +305,46 @@ class VideoViewport extends Viewport implements IVideoViewport {
...
+      <svg xmlns="http://www.w3.org/2000/svg">\
+        <filter id="colour" color-interpolation-filters="linearRGB">\
+        <feColorMatrix type="matrix" \
+        values="\
+          ${scaleWhite[0] * wlScale} 0 0 0 ${wlDelta} \
+          0 ${scaleWhite[1] * wlScale} 0 0 ${wlDelta} \
+          0 0 ${scaleWhite[2] * wlScale} 0 ${wlDelta} \
+          0 0 0 1 0" />\
+        </filter>\
+      </svg>#colour')`;
```
sedghi, line 341: since you are doing this we need some sanitization happen for XSS

How about you always check those (width and center and average white etc.) are numbers and not texts

sedghi (commented):
great stuff, see my comments

packages/core/examples/video/index.ts
wayfarer3130: Will do as soon as it is available to be merged/used.

packages/core/src/RenderingEngine/VideoViewport.ts
wayfarer3130: They are all calculated values, you can't effectively XSS except if you call the entire set method on the filter/assign to the internal value, in whih case you are dead anyways.

packages/core/src/RenderingEngine/VideoViewport.ts
```diff
@@ -7,18 +11,23 @@ import {
...
   InternalVideoCamera,
   VideoViewportInput,
+  VOIRange,
 } from '../types';
+import * as metaData from '../metaData';
 import { Transform } from './helpers/cpuFallback/rendering/transform';
 import { triggerEvent } from '../utilities';
 import Viewport from './Viewport';
 import { getOrCreateCanvas } from './helpers';
+import { vec3 } from 'gl-matrix';
```
sedghi, line 21: please move to top

packages/core/src/RenderingEngine/VideoViewport.ts
```diff
@@ -30,12 +39,42 @@ class VideoViewport extends Viewport implements IVideoViewport {
...
+   * the CSS filter is actually an link link to a SVG filter.
+   */
+  feFilter: string;
+
+  /**
+   * An average white point value, used to color balance the image so that
+   * the given white is mapped to [255,255,255] via multiplication per channel.
+   */
+  averageWhite: [number, number, number];
+
```
sedghi, line 69: should be private prob?

packages/core/src/RenderingEngine/VideoViewport.ts
```diff
@@ -82,6 +121,88 @@ class VideoViewport extends Viewport implements IVideoViewport {
...
+  private _getImageDataMetadata() {
+    const imagePlaneModule = metaData.get(
+      MetadataModules.IMAGE_PLANE,
+      this.imageId
+    );
+
+    let rowCosines, columnCosines;
+
+    rowCosines = <Point3>imagePlaneModule.rowCosines;
+    columnCosines = <Point3>imagePlaneModule.columnCosines;
```
sedghi, line 133: can be merged let rowCosines = ...

packages/core/src/RenderingEngine/VideoViewport.ts
```diff
@@ -255,6 +376,7 @@ class VideoViewport extends Viewport implements IVideoViewport {
     return {
       loop: this.videoElement.loop,
       muted: this.videoElement.muted,
+      voiRange: { ...this.voiRange },
```
sedghi, line 379: what about playback rate?

packages/core/src/RenderingEngine/VideoViewport.ts
```diff
@@ -265,8 +387,101 @@ class VideoViewport extends Viewport implements IVideoViewport {
...
+    const canvasData = context.getImageData(
+      0,
+      0,
+      this.videoWidth,
+      this.videoHeight
+    );
+    const { data: scalarData } = canvasData;
+    (scalarData as any).getRange = () => [0, 255];
+    return scalarData;
+  }
```
sedghi, line 405: we don't have getScalarData for the rest of the viewports, is this really necessary to have it as public? getImageData is enough imo

packages/core/src/RenderingEngine/VideoViewport.ts
```diff
@@ -265,8 +387,101 @@ class VideoViewport extends Viewport implements IVideoViewport {
...
+      spacing,
+      origin: metadata.origin,
+      direction: metadata.direction,
+      metadata: { Modality: this.modality },
+      imageData: {
+        getDirection: () => metadata.direction,
+        getDimensions: () => metadata.dimensions,
+        getRange: () => [0, 255],
+        getScalarData: () => this.getScalarData(),
+        getSpacing: () => null,
```
sedghi, line 423: you have x, y spacings, I'm surprised this is null

packages/core/src/RenderingEngine/VideoViewport.ts
```diff
@@ -411,6 +630,20 @@ class VideoViewport extends Viewport implements IVideoViewport {
...
+    const pan = this.getPan();
+    return [(x + pan[0]) * ratio, (y + pan[1]) * ratio];
+  };
+
+  public indexToCanvas = (indexPos: Point2): Point2 => {
+    const [x, y] = indexPos;
+    const ratio = this.canvas.width / this.videoWidth;
+    const pan = this.getPan();
+    return [x * ratio - pan[0], y * ratio - pan[1]];
+  };
```
sedghi, line 645: private or protected most likely since other viewports don't implement it

packages/tools/examples/videoColor/index.ts
```diff
@@ -0,0 +1,295 @@
...
+ */
+async function run() {
+  // Init Cornerstone and related libraries
+  await initDemo();
+
+  // Get Cornerstone imageIds and fetch metadata into RAM
+  const imageIds = await createImageIdsAndCacheMetaData({
+    StudyInstanceUID: '2.25.96975534054447904995905761963464388233',
+    SeriesInstanceUID: '2.25.15054212212536476297201250326674987992',
+    wadoRsRoot: useLocal
```
sedghi, line 191: pull main so that you can get the new local approach

sedghi (commented):
Looks great. 
Other than my comments on the code 

Tried out the demo. The interaction seemed too fast. I wonder if we should add a multiplier, similar to what we do for PT. By "fast," I mean a small interaction with the window level results in a significant change in intensities.

We should add a reset button to the demo and a resetProperties similar to other viewports.

packages/tools/examples/videoColor/index.ts
wayfarer3130: It isn't in main yet since it is part of HTJ2K, which hasn't been merged.  Prefer not to have it added twice and it is easy to fix once htj2k is merged.

sedghi (approved):

PR #891, opened 2023-11-15
feat(webworker): Simplify the API for running a compute task off the main thread in a worker

sedghi:
### Context

This PR reworks the web worker API and proposes a new set of methods to work with workers to make developers life so much easier to be able to run a function of the main thread

### Changes & Results

adds webWorkerManager to the Core

**Note:** Currently, we will not transfer the decoders from the old web worker manager to this new one until the next version of cornerstone3D. The only thing you need to be careful is how many compute workers you create (on top of the decode workers which you can configure in the dicom image loader) so that your system can handle it. 

### Testing

Follow the steps here 

https://deploy-preview-891--cornerstone-3d-docs.netlify.app/docs/concepts/cornerstone-core/webWorker

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  


packages/tools/examples/stackManipulationTools/index.ts
```diff
@@ -99,7 +99,7 @@ async function run() {
   toolGroup.addTool(WindowLevelTool.toolName);
   toolGroup.addTool(PanTool.toolName);
   toolGroup.addTool(ZoomTool.toolName);
-  toolGroup.addTool(StackScrollMouseWheelTool.toolName, { loop: true });
+  toolGroup.addTool(StackScrollMouseWheelTool.toolName, { loop: false });
```
wayfarer3130, line 102: This seems unrelated - that is probably ok, but wasn't sure you intended to have this.

sedghi: Yeah the demo was feeling weird that is why i have it reverted

karma.conf.js
```diff
@@ -97,15 +97,15 @@ module.exports = function (config) {
           },
           // NOTE: For better debugging you can comment out the
           // istanbul-instrumenter-loader below
-          {
```
wayfarer3130, line 100: Should this be commented out - seems like this is a change to the coverage information we are seeing.

packages/core/examples/webWorker/heavyTask.js
```diff
@@ -0,0 +1,20 @@
+import { expose } from 'comlink';
+
```
wayfarer3130, line 2: Add a comment about this being a simulated heavy task, not really being heavy.

karma.conf.js
sedghi: Karma hates the comlink package. `coverageIstanbulReporter` is responsible for coverage based on my understanding not this one

packages/core/src/enums/RequestType.ts
```diff
@@ -6,8 +6,10 @@ enum RequestType {
   Interaction = 'interaction',
   /** Second highest priority for loading*/
   Thumbnail = 'thumbnail',
-  /** Lowest priority for loading*/
+  /** Third highest priority for loading, usually used for image loading in the background*/
   Prefetch = 'prefetch',
+  /** Lower priority, often used for background computations in the worker */
```
wayfarer3130, line 11: I think I would say instead that this request type does not consume bandwidth as it is local compute.  Thus, it has a different queue to use.  It is not whether the priority is lower or higher, it is different.

wayfarer3130 (approved):
I think this looks good enough to start working with it.  It might be worth adding a preliminary warning somewhere so that we acknowledge that the API might change a bit still to deal with any issues found, but generally I think it is reasonable.

packages/core/src/requestPool/requestPoolManager.ts
```diff
@@ -110,12 +113,14 @@ class RequestPoolManager {
...
       thumbnail: 0,
       prefetch: 0,
+      compute: 0,
     };
 
     this.maxNumRequests = {
       interaction: 6,
       thumbnail: 6,
       prefetch: 5,
+      compute: 15,
```
lscoder, line 123: Maybe we should change `maxNumRequests`, `requestType`, etc. to a different name because before this change those three first options were related to image requests (download) but now `compute` was added and it is not related to that.

lscoder (approved):

packages/core/src/enums/RequestType.ts
sedghi: No, it is, in fact, lower in priority when queued. This should be comparing a decoding task versus a computation task.

packages/core/src/webWorkerManager/webWorkerManager.js
```diff
@@ -0,0 +1,193 @@
...
+
+    // Check and recreate the worker if it was terminated.
+    if (workerProperties.instances[minLoadIndex] === null) {
+      const worker = workerProperties.workerFn();
+      workerProperties.instances[minLoadIndex] = Comlink.wrap(worker);
+      workerProperties.nativeWorkers[minLoadIndex] = worker;
+    }
+
+    // Update the load counter.
+    workerProperties.loadCounters[minLoadIndex] += 1;
```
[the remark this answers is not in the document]
sedghi: Add a comment about number of workers and load balancing

packages/core/src/webWorkerManager/webWorkerManager.js
```diff
@@ -0,0 +1,193 @@
...
+    let minLoadIndex = 0;
+    let minLoadValue = workerProperties.loadCounters[0] || 0;
+
+    for (let i = 1; i < workerInstances.length; i++) {
+      const currentLoadValue = workerProperties.loadCounters[i] || 0;
+      if (currentLoadValue < minLoadValue) {
+        minLoadIndex = i;
+        minLoadValue = currentLoadValue;
+      }
+    }
```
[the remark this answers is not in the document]
sedghi: Makes a lot of sense, a good candidate for v2.x

PR #893, opened 2023-11-15
feat(video Tools): add video annotation tools

wayfarer3130:
### Context

Add support for video annotation tools

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  


packages/core/src/RenderingEngine/VideoViewport.ts
```diff
@@ -35,11 +35,15 @@ class VideoViewport extends Viewport implements IVideoViewport {
...
-  private loop = false;
+  private loop = true;
   private mute = true;
   private isPlaying = false;
   private scrollSpeed = 1;
   private playbackRate = 1;
+  /**
+   * The range is the set of frames to play
+   */
+  private range: [number, number] = [0, 0];
```
sedghi, line 46: range is too generic in my opinion, how about `frameRange`?

packages/core/src/RenderingEngine/VideoViewport.ts
```diff
@@ -330,6 +353,21 @@ class VideoViewport extends Viewport implements IVideoViewport {
...
+    }
+    if (range.length !== 2 || range[0] === range[1]) {
+      return;
+    }
+    this.range = [range[0], range[1]];
+  }
+
+  public getRange(): [number, number] {
+    return this.range;
+  }
```
sedghi, line 369: setFrameRange and getFrameRange

Range sounds like scalar range

packages/core/src/RenderingEngine/VideoViewport.ts
```diff
@@ -445,6 +483,33 @@ class VideoViewport extends Viewport implements IVideoViewport {
...
+
+      if (minRegion === maxRegion && Math.abs(frame - minRegion) < 10) {
+        return true;
+      }
+      if (frame >= minRegion && frame <= maxRegion) {
+        return true;
+      }
+    }
+    return false;
+  }
```
sedghi, line 511: can you please add some documentation on the logic here? what is regions, what is minRegion and maxRegion etc.

packages/core/src/RenderingEngine/VideoViewport.ts
```diff
@@ -526,6 +589,21 @@ class VideoViewport extends Viewport implements IVideoViewport {
...
 
+  public getCurrentImageId() {
+    const current = this.imageId.replace(
+      '/frames/1',
+      this.isPlaying
+        ? `/frames/1-${this.numberOfFrames}`
+        : `/frames/${this.getFrame()}`
+    );
+    return current;
+  }
```
sedghi, line 600: Does the imageId always have /frames/1? what about a direct link to an mp4 that we initially started with

packages/core/src/RenderingEngine/VideoViewport.ts
```diff
@@ -526,6 +589,21 @@ class VideoViewport extends Viewport implements IVideoViewport {
...
+
+  public getFrame() {
+    // Need to round this as the fps/time isn't exact
+    return 1 + Math.round(this.videoElement.currentTime * this.fps);
+  }
```
sedghi, line 605: getFrameNumber? or getCurrentFrameNumber?

packages/core/src/RenderingEngine/VideoViewport.ts
sedghi: What is this looking to? is the imageURI in the current range being played? in any of the frames?

packages/tools/examples/videoTools/index.ts
```diff
@@ -1,4 +1,10 @@
-import { RenderingEngine, Types, Enums } from '@cornerstonejs/core';
+import {
+  RenderingEngine,
+  Types,
+  Enums,
+  eventTarget,
+  triggerEvent,
+} from '@cornerstonejs/core';
```
sedghi, line 7: in this example, next, previous does not work

sedghi: clear Range unpause the video which is not expected, not even sure what is clear range

There is also Add, Set, Start, End which either is confusing or don't work

sedghi: I prefer scroll to do scrolling through the times but right click is scroll which is unusuall

sedghi: Selection: 1.6 is confusing, You mean the annotation time or something?

packages/tools/examples/videoTools/index.ts
```diff
@@ -92,74 +117,357 @@ addButtonToToolbar({
...
 addButtonToToolbar({
-  id: 'previous',
-  title: 'previous',
+  id: 'Next',
+  title: 'Next',
   onClick() {
-    viewport.scroll(-1);
+    navigateRelative(1);
   },
 });
```
sedghi, line 163: Please rename accordingly to avoid confusion. Something like "Select Next Annotation" and make sure the function name is related to selection.

packages/tools/examples/videoTools/index.ts
```diff
@@ -92,74 +117,357 @@ addButtonToToolbar({
...
+    }
+    const newSelection = setSelection(
+      selectedAnnotation.annotationUID,
+      null,
+      viewport.getFrame()
+    );
+    updateAnnotationDiv(selectedAnnotation.annotationUID, newSelection);
+    fireUpdateEvent();
   },
 });
```
sedghi, line 199: I'm not sure why we need Add Selection, since they should be able to just click and select right?

packages/tools/examples/videoTools/index.ts
```diff
@@ -92,74 +117,357 @@ addButtonToToolbar({
...
+
+function toTime(selection) {
+  const range = toRange(selection).map(
+    (it) => Math.round(((it - 1) * 10) / viewport.fps) / 10
+  );
+  if (range[0] === range[1]) {
+    return String(range[0]);
+  }
+  return toSelection(range);
+}
```
sedghi, line 277: ideally rename for clarity

function convertSelectionToTimeRange(selection) { ... }
function convertSelectionToNumericRange(selection) { ... }
function navigateAnnotationsRelative(direction) { ... }

sedghi (commented):
I'm not sure if this is going in the right direction. 

## Metadata 

I thought the cachedStats data would store the frames it should display, and in the filterAnnotationForDisplay function, it would decide whether to show them or not. But there's a lot of extra logic in the example (videoTools) that doesn't seem to belong there. 

Right now what i see in the cached stats is only the videoId and referencedImageId. 

Something you can do is instead of manipulating the imageId and coming up with frames/range1-range2, just have the imageId of that frame in the cachedStats, and the filter just checks if the current frame is among those then display if not it ignores. 

Or we can have another way to say have one imageId (the frame that it initially was drawn) as the key in the cachedStats and for the values we have `length, unit, frames` where frames is array of frame numbers

---

## Demo too Complicated

Ideally, the example should be as simple as possible, with just a few things:
-  Play/pause button
-  Tool dropdown
-  Button to remove annotation
-  Button to move annotation to another frame (let's say Button: Move to frame 50-60)

The user should be able to draw, and by default, the annotation should appear on 10 frames (which can be changed in the settings).

The user should be able to adjust the ends (handles) of the annotation.

The user should be able to delete the annotation.

And when they click that button, they should be able to move the annotation from the current frame range to another frame range.

---

## Frames of Interest

For frames of interest selections, let's put it in another example (e.g., frames of interests example) . You can implement it similar to this 

image

packages/core/src/RenderingEngine/VideoViewport.ts
```diff
@@ -193,8 +202,22 @@ class VideoViewport extends Viewport implements IVideoViewport {
     this.metadata = this._getImageDataMetadata();
 
     return this.setVideoURL(rendered).then(() => {
-      const { cineRate = 30 } = metaData.get(MetadataModules.CINE, imageId);
+      let { cineRate, numberOfFrames } = metaData.get(
+        MetadataModules.CINE,
+        imageId
```
sedghi, line 207: imageIds must be imageId

packages/core/src/RenderingEngine/VideoViewport.ts
wayfarer3130: get/set FrameNumber

packages/core/src/RenderingEngine/VideoViewport.ts
```diff
@@ -526,6 +582,21 @@ class VideoViewport extends Viewport implements IVideoViewport {
...
 
+  public getCurrentImageId() {
+    const current = this.imageId.replace(
+      '/frames/1',
+      this.isPlaying
+        ? `/frames/1-${this.numberOfFrames}`
+        : `/frames/${this.getFrameNumber()}`
+    );
+    return current;
+  }
```
sedghi, line 593: please add doc on why we do the 1-numberOfFrames, you mentioned it is DICOM way of doing this as well. Function lack documentation

packages/core/src/RenderingEngine/VideoViewport.ts
```diff
@@ -445,6 +485,24 @@ class VideoViewport extends Viewport implements IVideoViewport {
...
+    }
+    const frames = imageURI.match(/\/frames\/([^/?&]*)/);
+    if (!frames?.[1]) {
+      return true;
+    }
+    const range = frames?.[1].split('-').map((it) => Number(it));
+    const frame = this.getFrameNumber();
+    if (range.length === 1) {
+      return Math.abs(range[0] - frame) < 0.25 * this.fps;
+    }
```
sedghi, line 502: I find this logic unusual. what is 0.25?

packages/core/src/RenderingEngine/VideoViewport.ts
```diff
@@ -445,6 +485,24 @@ class VideoViewport extends Viewport implements IVideoViewport {
     };
   }
 
+  public hasImageURI(imageURI: string) {
+    // TODO - move annotationFrameRange into core so it can be used here.
+    const testURI = imageURI.replace(/\/frames\/.*/, '');
+    if (this.imageId.indexOf(testURI) === -1) {
+      return false;
+    }
+    const frames = imageURI.match(/\/frames\/([^/?&]*)/);
```
sedghi, line 494: can we rename this to framesMatch or something since it is a match?

packages/core/src/RenderingEngine/VideoViewport.ts
```diff
@@ -445,6 +485,24 @@ class VideoViewport extends Viewport implements IVideoViewport {
...
+    if (!frames?.[1]) {
+      return true;
+    }
+    const range = frames?.[1].split('-').map((it) => Number(it));
```
sedghi, line 498: You can do `.split('-').map(Number)`

packages/core/src/RenderingEngine/VideoViewport.ts
```diff
@@ -445,6 +485,24 @@ class VideoViewport extends Viewport implements IVideoViewport {
     };
   }
 
+  public hasImageURI(imageURI: string) {
```
sedghi, line 488: also please add documentation, on what is matched and what is not (I guess frame range is matched)

packages/docs/docs/concepts/cornerstone-tools/annotation/annotationGroups.md
```diff
@@ -0,0 +1,33 @@
...
+In order to indicate that annotations are related to each other, there is
+an `AnnotationGroup` class that can be used to group annotations. Currrently,
+the grouping is very basic and isn't saved/restored automatically within the
+adapters. The requirements for enhanced grouping are still being gathered, but the basic
+capability is there to use. Annotations can be added to a group, and navigated
+between them by finding the next/previous annotation.
+
+## Creating a new group
+
+To create a new annotation group, just create an instance of AnnotationGroup
```
sedghi, line 17: please add the code snippets for this file

packages/tools/src/utilities/annotationFrameRange.ts
```diff
@@ -0,0 +1,84 @@
...
+import Events from '../enums/Events';
+import { Annotation } from '../types';
+
+/**
+ * This class handles the annotation frame range values for multiframes.
+ * Mostly used for the Video viewport, it allows references to
+ * a range of frame values.
+ *
+ */
+
```
sedghi, line 11: This should be exactly on top of the class name so that it gets picked up by api extractor for api docs

packages/tools/examples/videoNavigation/index.ts
```diff
@@ -0,0 +1,277 @@
...
+const { ViewportType } = Enums;
+const { MouseBindings, KeyboardBindings } = csToolsEnums;
+
+const toolGroupId = 'VIDEO_TOOL_GROUP_ID';
+
+// ======== Set up page ======== //
+setTitleAndDescription(
+  'Basic Video Tools',
+  'Show a video viewport with controls to allow it to be navigated and zoom/panned'
+);
```
sedghi, line 36: This doesn't match

packages/tools/examples/videoNavigation/index.ts
```diff
@@ -0,0 +1,277 @@
...
+    StudyInstanceUID: '2.25.96975534054447904995905761963464388233',
+    SeriesInstanceUID: '2.25.15054212212536476297201250326674987992',
+    wadoRsRoot:
+      getLocalUrl() || 'https://d33do7qe4w26qo.cloudfront.net/dicomweb',
+  });
+
+  // Only one SOP instances is DICOM, so find it
+  const videoId = imageIds.find(
+    (it) => it.indexOf('2.25.179478223177027022014772769075050874231') !== -1
+  );
```
sedghi, line 181: can you add some comment explaining this as we discuss over zoom?

packages/tools/src/utilities/annotationFrameRange.ts
```diff
@@ -0,0 +1,84 @@
...
+export type FramesRange = [number, number] | number;
+
+export default class AnnotationFrameRange {
+  public static frameRangeExtractor = /(\/frames\/|[&?]frameNumber=)([^/&?]*)/i;
```
sedghi, line 15: I prefer functional programming over the class here since it does not add any value. But it's fine, just for the future.

packages/tools/examples/videoTools/index.ts
```diff
@@ -60,106 +95,285 @@ rangeElement.oninput = () => {
...
+
+    // Set the old tool passive
+    toolGroup.setToolPassive(selectedToolName);
+
+    selectedToolName = <string>newSelectedToolName;
+  },
+});
+
+const nextButton = addButtonToToolbar({
+  id: 'Next',
```
sedghi, line 174: Next Annotation, and Previous Annotation

packages/tools/src/stateManagement/annotation/AnnotationGroup.ts
```diff
@@ -0,0 +1,163 @@
...
+
+  /**
+   * Removes listeners that add to this group on creating a new instance.
+   */
+  public removeListeners() {
+    eventTarget.removeEventListener(
+      Events.ANNOTATION_ADDED,
+      this.addAnnotationListener
+    );
+  }
```
sedghi, line 162: These will cause so many side effects. I imagine there are use cases where you want a group, but you don't want adding or removing to be automated. Can we stick to explicit control over adding and removing to each group?

sedghi: And you get rid of the setActive too by moving to explicit. I think it is more simple and reduces potential unintended side effects

packages/tools/src/stateManagement/annotation/AnnotationGroup.ts
```diff
@@ -0,0 +1,163 @@
...
+
+  public get isActive() {
+    return this._isActive;
+  }
+
+  /** Finds the nearby/next annotation in the given direction */
+  public findNearby(uid: string, direction: 1) {
+    if (!this._isVisible) {
+      return null;
+    }
```
sedghi, line 97: I don't think there is any need for this here. One should be able to get nearby annotations even if they are invisible

packages/tools/src/stateManagement/annotation/AnnotationGroup.ts
```diff
@@ -0,0 +1,163 @@
...
+    if (!uid) {
+      return uids[direction === 1 ? 0 : uids.length - 1];
+    }
+    const index = uids.indexOf(uid);
+    if (
+      index === -1 ||
+      index + direction < 0 ||
+      index + direction > uids.length
```
sedghi, line 109: shouldn't this be >=?

packages/tools/src/stateManagement/annotation/AnnotationGroup.ts
```diff
@@ -0,0 +1,163 @@
...
+        return;
+      }
+      annotation.isVisible = isVisible;
+      const eventDetail = {
+        ...baseEvent,
+        annotation,
+      };
+      triggerEvent(eventTarget, Events.ANNOTATION_MODIFIED, eventDetail);
+    });
+  }
```
sedghi, line 83: I think we also need setSelected, what do you think

packages/tools/src/stateManagement/annotation/AnnotationGroup.ts
```diff
@@ -0,0 +1,163 @@
...
+  public get isVisible() {
+    return this._isVisible;
+  }
+
+  public get isActive() {
+    return this._isActive;
+  }
+
+  /** Finds the nearby/next annotation in the given direction */
+  public findNearby(uid: string, direction: 1) {
```
sedghi, line 94: findNearbyAnnotation or getNearbyAnnotation maybe

packages/tools/src/tools/annotation/KeyImageTool.ts
```diff
@@ -0,0 +1,423 @@
...
+  ToolProps,
+  SVGDrawingHelper,
+} from '../../types';
+import { StyleSpecifier } from '../../types/AnnotationStyle';
+import { Annotation } from '../../types';
+
+type Point2 = Types.Point2;
+
+class KeyImageTool extends AnnotationTool {
+  static toolName = 'KeyImage';
```
sedghi, line 38: This will cause build error in some systems. Please put it outside like others

packages/tools/src/tools/annotation/KeyImageTool.ts
```diff
@@ -0,0 +1,423 @@
...
+    annotation: Annotation,
+    canvasCoords: Types.Point2,
+    proximity: number
+  ): boolean => {
+    const enabledElement = getEnabledElement(element);
+    const { viewport } = enabledElement;
+    const { data } = annotation;
+
+    if (
+      Math.abs(canvasCoords[0] - 15) <= 5 &&
```
sedghi, line 190: Not sure what "are 15, 5" means, but whatever they are, they should be configurable in the `this.configuration`.

packages/tools/examples/videoTools/index.ts
```diff
@@ -226,7 +475,11 @@ async function run() {
   toolGroup.setToolActive(StackScrollTool.toolName, {
     bindings: [
       {
-        mouseButton: MouseBindings.Primary, // Left Click
+        mouseButton: MouseBindings.Secondary,
```
sedghi, line 478: Still it is unusually that scroll does not scroll in the video

sedghi (commented):
Please see my comments. 

The example still needs some pruning

Please remove the [ Single Image ] as it is confusing

Right now 

CleanShot 2023-11-20 at 09 17 02

Ideally should be much simplere

image

And extract out the key Image functionality to another example since this is already too big

packages/tools/src/utilities/annotationFrameRange.ts
```diff
@@ -0,0 +1,84 @@
...
+    const eventDetail = {
+      ...eventBase,
+      annotation,
+    };
+    triggerEvent(eventTarget, Events.ANNOTATION_MODIFIED, eventDetail);
+  }
+
+  public static annotationToFrameRange(annotation: Annotation) {
+    return this.imageIdToFrames(annotation.metadata.referencedImageId);
+  }
```
sedghi, line 72: maybe rename to `getFrameRange`, and the imageIdToFrames to be private

packages/tools/examples/videoTools/index.ts
```diff
@@ -60,106 +95,285 @@ rangeElement.oninput = () => {
...
+      viewport.render();
+    }
   },
 });
 
-const scrollSpeeds = ['1 f', '2 f', '4 f', '0.5 s', '1 s', '2 s', '4 s'];
+addButtonToToolbar({
+  id: 'Set Current',
+  title: 'Single Image',
+  onClick() {
```
sedghi, line 245: can you explain in the description that this means for navigation back and forth it is single and not a range

packages/tools/src/utilities/annotationFrameRange.ts
```diff
@@ -0,0 +1,84 @@
...
+   * Returns an imageId without:
+   * * Any prefixes such as imageId, videoId etc.
+   * * The /frames/... part (or anything after it).
+   */
+  public static baseUrl(imageId: string): string {
+    const match = imageId.match(this.frameRangeExtractor);
+    const beforeFrames = imageId.substring(0, match?.index);
+    return beforeFrames.replace(this.baseUrlExtractor, '');
+  }
+}
```
sedghi, line 84: let's move to hasImageURI

packages/tools/examples/videoTools/index.ts
```diff
@@ -60,106 +95,285 @@ rangeElement.oninput = () => {
...
-    const button = document.getElementById('play');
-    if (button.innerText === 'pause') {
-      button.innerText = 'play';
-    } else {
-      button.innerText = 'pause';
-    }
+    viewport.setFrameRange(null);
+    viewport.play();
   },
 });
```
sedghi, line 134: let's move this along with key images into a separate VideoRange example

packages/tools/src/stateManagement/annotation/AnnotationGroup.ts
wayfarer3130: You can just use add as it takes multi-arguments (now)

packages/tools/examples/videoTools/index.ts
wayfarer3130: On what button?  The primary is already used.

sedghi (approved):
Looks great, thanks

PR #894, opened 2023-11-15
feat(segmentation): add stack viewport segmentations rendering and tools

sedghi:
Thanks @rodrigobasilio2022 for starting this work initially, you rock 🎸

### Context

Add segmentation rendering and tools to stack viewport

### Changes & Results

- Adds `createAndCacheDerivedImages` to imageLoader to be able to create empty segmentation for one slice(s)

```js
const newSegImageId = await imageLoader.createAndCacheDerivedImage(
    viewport.getCurrentImageId(),
);

segmentation.addSegmentations([
    {
      segmentationId,
      representation: {
        type: csToolsEnums.SegmentationRepresentations.Labelmap,
        data: {
            imageIdReferenceMap: new Map([[currentImageId, newSegImageId]]),
        },
      },
    },
  ]);
  
// // Add the segmentation representation to the toolgroup
  await segmentation.addSegmentationRepresentations(toolGroupId, [
    {
      segmentationId,
      type: csToolsEnums.SegmentationRepresentations.Labelmap,
    },
  ]);


```

### Testing

`yarn run example stackSegmentation`

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  


sedghi:
@Ouwen

packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -1606,13 +1601,39 @@ class StackViewport extends Viewport implements IStackViewport, IImagesLoader {
...
+    imageData.setSpacing(spacing);
+    imageData.setDirection(direction);
+    imageData.setOrigin(origin);
+    imageData.getPointData().setScalars(scalarArray);
+
+    return imageData;
+  }
+  /**
+   * Creates vtkImagedata based on the image object, it creates
+   * and empty scalar data for the image based on the metadata
```
wayfarer3130, line 1616: it creates empty scalar data....

packages/core/src/types/IImage.ts
```diff
@@ -10,6 +10,8 @@ import { ImageQualityStatus } from '../enums';
 interface IImage {
   /** Image Id */
   imageId: string;
+  /** referenced imageId if this image is derived from */
+  referencedImageId?: string;
```
wayfarer3130, line 14: This should be an array and should reference all the overlays that go into the image.

packages/core/src/utilities/derivedImageMetadataProvider.ts
```diff
@@ -0,0 +1,41 @@
...
+    };
+  },
+
+  /**
+   * Returns the metadata for an imageId if it exists.
+   * @param type - the type of metadata to enquire about
+   * @param imageId - the imageId to enquire about
+   */
+  get: (type: string, imageId: string): any => {
+    if (!state[imageId] || !state[imageId][type]) {
```
wayfarer3130, line 31: !state[imageId]?.[type]
Really, you care about the final value, so checking it in one go is easier to read.

packages/core/src/utilities/derivedImageMetadataProvider.ts
```diff
@@ -0,0 +1,41 @@
...
+   */
+  add: (imageId: string, payload: any): void => {
+    const type = payload.type;
+
+    if (!state[imageId]) {
+      state[imageId] = {};
+    }
+
+    state[imageId][type] = {
+      ...payload.metadata,
```
wayfarer3130, line 21: What about an object clone?  The problem with this is that you don't clone arrays and nested objects, so it still shares some data with the original item.

packages/core/src/utilities/derivedImageMetadataProvider.ts
```diff
@@ -0,0 +1,41 @@
+import { metaData } from '..';
+
+const state: Record<string, any> = {}; // Calibrated pixel spacing per imageId
+/**
+ * Simple metadata provider that stored the derived metadata in memory.
+ */
+const metadataProvider = {
```
wayfarer3130, line 7: Could we call this something like registerTypeMetadataProvider?  It doesn't actually matter that it is for derived metadata, unless you plan on using that separately to allow clearing the state etc.  Having a single one that allows registering anything seems like it would be useful.

packages/core/src/utilities/derivedImageMetadataProvider.ts
```diff
@@ -0,0 +1,41 @@
...
+  get: (type: string, imageId: string): any => {
+    if (!state[imageId] || !state[imageId][type]) {
+      return;
+    }
+
+    return state[imageId][type];
+  },
+};
+
+metaData.addProvider(metadataProvider.get);
```
wayfarer3130, line 39: Could we start adding a clear method so that it becomes possible to clear the metadata?  Otherwise it just starts building up, and with a lot of overlays, that actually ends up being quite a bit of data.

packages/core/src/utilities/isValidVolume.ts
```diff
@@ -0,0 +1,51 @@
+import { metaData } from '..';
+import isEqual from './isEqual';
+
+function isValidVolume(imageIds: string[]): boolean {
```
wayfarer3130, line 4: Please document exactly what is considered a valid volume here - it seems like it isn't quite the consistent spacing that is used in some other places.

packages/core/src/utilities/updateVTKImageDataWithCornerstoneImage.ts
```diff
@@ -0,0 +1,39 @@
+import vtkImageData from '@kitware/vtk.js/Common/DataModel/ImageData';
+import { IImage } from '../types';
+
+function updateVTKImageDataWithCornerstoneImage(
+  sourceImageData: vtkImageData,
+  image: IImage
+) {
+  const pixelData = image.getPixelData();
+  const scalarData = sourceImageData.getPointData().getScalars().getData() as
+    | Uint8Array
```
wayfarer3130, line 10: Could you declare this somewhere - there are so many times it gets used that I'd rather have a specific name for it.

packages/tools/examples/labelmapSegmentationTools/index.ts
```diff
@@ -275,7 +275,7 @@ async function run() {
   );
   toolGroup.setToolEnabled(SegmentationDisplayTool.toolName);
 
-  toolGroup.setToolActive(brushInstanceNames.CircularBrush, {
+  toolGroup.setToolActive(brushInstanceNames.SphereBrush, {
```
wayfarer3130, line 278: Should this change be included?

packages/tools/src/tools/segmentation/RectangleScissorsTool.ts
```diff
@@ -150,20 +159,47 @@ class RectangleScissorsTool extends BaseTool {
...
       viewportIdsToRender,
       handleIndex: 3,
       movingTextBox: false,
       newAnnotation: true,
       hasMoved: false,
-    };
+      segmentationRepresentationUID,
+    } as any;
+
+    if (isVolumeSegmentation(labelmapData as LabelmapSegmentationData)) {
```
wayfarer3130, line 176: Seems like this code is getting repeated several times in each tool.  Could it be moved into a common top level class?

packages/tools/src/tools/segmentation/strategies/fillCircle.ts
```diff
@@ -88,9 +61,9 @@ function fillCircle(
       }
 
       if (
-        isWithinThreshold(index, imageVolume, strategySpecificConfiguration)
+        isWithinThreshold(index, imageScalarData, strategySpecificConfiguration)
```
wayfarer3130, line 64: This code would all be shared if we merged the dynamic brush strategies first.  Would make it a bit easier to manage.

packages/core/src/utilities/derivedImageMetadataProvider.ts
lscoder: It could be only  `return state[imageId]?.[type]`

packages/tools/src/tools/segmentation/strategies/utils/getStrategyData.ts
```diff
@@ -0,0 +1,65 @@
+import { cache } from '@cornerstonejs/core';
+import { isVolumeSegmentation } from './stackVolumeCheck';
+import { LabelmapToolOperationDataStack } from '../../../../types';
+
+function getStrategyData({ operationData, viewport }) {
+  let segmentationImageData, segmentationScalarData, imageScalarData;
+
+  if (isVolumeSegmentation(operationData)) {
```
wayfarer3130, line 8: Could you drop both halfs of the strategy into individual functions?  It would make the overall function less complex, and that makes it easier to understand.

packages/tools/src/tools/displayTools/Labelmap/addLabelmapToElement.ts
```diff
@@ -31,27 +33,45 @@ async function addLabelmapToElement(
...
+      renderingEngine,
+      volumeInputs,
+      [viewportId],
+      immediateRender,
+      suppressEvents
+    );
+  } else {
+    const stackInputs: Types.IStackInput[] = [
+      {
+        imageId: viewport.getCurrentImageId(),
```
wayfarer3130, line 61: This should mirror the volume one and use the stackseg:.... image id instead, otherwise you don't get the right metadata in the create image data.  If you need to include the current image id as well, that is certainly possible.

packages/core/src/loaders/imageLoader.ts
```diff
@@ -183,6 +196,126 @@ export function loadAndCacheImages(
...
+      'createAndCacheDerivedImage: parameter imageId must not be undefined'
+    );
+  }
+  const imagePlaneModule = metaData.get('imagePlaneModule', imageId);
+
+  const length = imagePlaneModule.rows * imagePlaneModule.columns;
+  let numBytes, TypedArray;
+
+  const { targetBuffer } = options;
+  if (targetBuffer) {
```
wayfarer3130, line 224: Can you put the targetBuffer creation code somewhere shared please?  There are several places that execute this.

packages/core/src/loaders/imageLoader.ts
```diff
@@ -183,6 +196,126 @@ export function loadAndCacheImages(
...
+export function createAndCacheDerivedImage(
+  imageId: string,
+  options: DerivedImageOptions
+): Promise<IImage> {
+  if (imageId === undefined) {
+    throw new Error(
+      'createAndCacheDerivedImage: parameter imageId must not be undefined'
+    );
+  }
+  const imagePlaneModule = metaData.get('imagePlaneModule', imageId);
```
wayfarer3130, line 218: This should use the options.imageId first in case the imagePlaneModule has already been created, and then fallback to the base imageId module otherwise.

packages/core/src/loaders/imageLoader.ts
```diff
@@ -183,6 +196,126 @@ export function loadAndCacheImages(
...
+    rowPixelSpacing: imagePlaneModule.rowPixelSpacing,
+    invert: false,
+  };
+
+  const imageLoadObject = {
+    promise: Promise.resolve(image),
+  };
+  cache.putImageLoadObject(image.imageId, imageLoadObject);
+
+  ['imagePixelModule', 'imagePlaneModule', 'generalSeriesModule'].forEach(
```
wayfarer3130, line 278: Please put the list of copyable objects somewhere, and only create them if they do not already exist - that allows customizing them externally before calling into this.

packages/core/src/loaders/imageLoader.ts
```diff
@@ -183,6 +196,126 @@ export function loadAndCacheImages(
...
+ * which will resolve to the loaded image object or fail if an error occurred.
+ * The image is stored in the cache.
+ *
+ * @param imageId -  A Cornerstone Image Object's imageId
+ * @param options - Options to be passed to the Image Loader
+ *
+ * @returns Image Loader Object
+ */
+export function createAndCacheDerivedImage(
+  imageId: string,
```
wayfarer3130, line 210: For consistency with the other calls, this should be the destination imageId, and the options should contain the referencedImageId - otherwise it is really easy to confuse things.

packages/core/src/loaders/imageLoader.ts
```diff
@@ -183,6 +196,126 @@ export function loadAndCacheImages(
...
+
+/**
+ * Load and cache a list of imageIds
+ *
+ * @param imageIds - list of imageIds
+ * @param options - options for loader
+ *
+ */
+export function createAndCacheDerivedImages(
+  imageIds: Array<string>,
```
wayfarer3130, line 298: Can you call the list referencedImageIds so it is really obvious these are not the derived ids?

packages/tools/src/eventListeners/segmentation/index.ts
```diff
@@ -2,10 +2,12 @@ import segmentationRepresentationModifiedEventListener from './segmentationRepre
 import segmentationDataModifiedEventListener from './segmentationDataModifiedEventListener';
 import segmentationRepresentationRemovedEventListener from './segmentationRepresentationRemovedEventListener';
 import segmentationModifiedListener from './segmentationModifiedEventListener';
+import stackImageChangeEventListener from './stackImageChangeEventListener';
```
wayfarer3130, line 5: Can we not include stack in the naming?  We are starting to get other viewport types and making this a generic iamgeChangeEventListener means that when we have segmentations on videos and so on, we should be able to re-use the events.

packages/tools/src/eventListeners/segmentation/stackImageChangeEventListener.ts
```diff
@@ -0,0 +1,213 @@
...
+      const derivedImage = cache.getImage(derivedImageId);
+
+      const { origin, dimensions, spacing, direction } =
+        viewport.getImageDataMetadata(derivedImage);
+
+      segmentationImageData.setOrigin(origin);
+      segmentationImageData.modified();
+
+      if (segmentationImageData.getDimensions()[0] !== dimensions[0]) {
+        // IMPORTANT: Not sure why we can't just update the dimensions
```
wayfarer3130, line 142: The defect you were seeing is because the other version of it uses the primary image ID for the data, and this one, correctly uses the derivedImageId instead.  If you fix the bug in the other spot, you should be able to update the dimensions and call modified.

packages/tools/src/tools/annotation/CircleROITool.ts
```diff
@@ -984,9 +984,21 @@ class CircleROITool extends AnnotationTool {
           modalityUnitOptions
         );
 
+        const invXRadiusSq =
+          ellipseObj.xRadius !== 0 ? 1 / ellipseObj.xRadius ** 2 : 0;
```
wayfarer3130, line 988: Why bother with the test here - just get the square of the radius, which is 0 if the radius is 0.  Seems like extra work to handle that case.

packages/tools/src/tools/annotation/EllipticalROITool.ts
```diff
@@ -1112,9 +1112,21 @@ class EllipticalROITool extends AnnotationTool {
           modalityUnitOptions
         );
 
+        const invXRadiusSq =
+          ellipseObj.xRadius !== 0 ? 1 / ellipseObj.xRadius ** 2 : 0;
```
wayfarer3130, line 1116: Ditto

packages/tools/src/utilities/math/ellipse/pointInEllipse.ts
```diff
@@ -1,38 +1,38 @@
...
  * @returns A boolean value.
  */
-export default function pointInEllipse(
-  ellipse: Ellipse,
-  pointLPS: Types.Point3
-): boolean {
-  const { center: circleCenterWorld, xRadius, yRadius, zRadius } = ellipse;
-  const [x, y, z] = pointLPS;
-  const [x0, y0, z0] = circleCenterWorld;
+export default function pointInEllipse(ellipse, pointLPS, inverts) {
```
wayfarer3130, line 7: Just set inverts to {}, and then compute the three values if they are missing - that way you can just get the default values by passing in a fixed {}, and it will automatically get the right initial values.

packages/tools/src/utilities/math/ellipse/pointInEllipse.ts
```diff
@@ -1,38 +1,38 @@
...
+  const { center, xRadius, yRadius, zRadius } = ellipse;
 
-  let inside = 0;
-  if (xRadius !== 0) {
-    inside += ((x - x0) * (x - x0)) / (xRadius * xRadius);
+  let invXRadiusSqToUse, invYRadiusSqToUse, invZRadiusSqToUse;
```
wayfarer3130, line 10: Make these constants directly extracted from the inverts object, but have updated the inverts object to have the default values first if not provided.

packages/tools/src/tools/annotation/EllipticalROITool.ts
```diff
@@ -1112,9 +1112,21 @@ class EllipticalROITool extends AnnotationTool {
           modalityUnitOptions
         );
 
+        const invXRadiusSq =
```
wayfarer3130, line 1115: Please remove this bit of code, and instead, just pass an empty object down below - let the internals of pointInEllipse assign the right value

packages/tools/src/tools/annotation/EllipticalROITool.ts
```diff
@@ -1112,9 +1112,21 @@ class EllipticalROITool extends AnnotationTool {
...
+        const invYRadiusSq =
+          ellipseObj.yRadius !== 0 ? 1 / ellipseObj.yRadius ** 2 : 0;
+        const invZRadiusSq =
+          ellipseObj.zRadius !== 0 ? 1 / ellipseObj.zRadius ** 2 : 0;
+
         const pointsInShape = pointInShapeCallback(
           imageData,
-          (pointLPS, pointIJK) => pointInEllipse(ellipseObj, pointLPS),
+          (pointLPS) =>
+            pointInEllipse(ellipseObj, pointLPS, {
```
wayfarer3130, line 1125: Just pass in {}, and let pointInEllipse assign the values on the first call.

packages/tools/src/tools/annotation/CircleROITool.ts
```diff
@@ -984,9 +984,21 @@ class CircleROITool extends AnnotationTool {
           modalityUnitOptions
         );
 
+        const invXRadiusSq =
```
wayfarer3130, line 987: See EllipticalROITool for comments on this.

packages/tools/src/types/LabelmapTypes.ts
```diff
@@ -35,7 +35,16 @@ export type LabelmapRenderingConfig = {
...
 };
 
-export type LabelmapSegmentationData = {
+export type LabelmapSegmentationDataVolume = {
   volumeId: string;
   referencedVolumeId?: string;
 };
+
+export type LabelmapSegmentationDataStack = {
+  imageIds: Array<string>;
```
wayfarer3130, line 44: Make the referendImageIds a Set, or if you want to get updated imageId, then use a map of referenced to updated imageIds - either way, using a Map will be quite a bit faster for large stacks - you should expect to see stacks in the range of 2000+ images working effectively.

packages/tools/src/tools/displayTools/Labelmap/labelmapDisplay.ts
```diff
@@ -192,29 +194,50 @@ async function render(
...
+
+    actorEntry = viewport.getActor(segmentationRepresentationUID);
+  } else {
+    // stack segmentation
+    const imageId = viewport.getCurrentImageId();
+    const { referencedImageIds } = labelmapData;
+
+    // if the stack labelmap is not built for the current imageId that is
+    // rendered at the viewport then return
+    if (!referencedImageIds.includes(imageId)) {
```
wayfarer3130, line 229: Should be .has - which will then be quite quick.

packages/tools/examples/stackSegmentation/index.ts
wayfarer3130, line 1: Could you add a button to convert to/from a volume?  The code should work transparently either way, with the derived data being used to generate the volumetric data and vice versa.  Otherwise I have suspicions that you can load for one but not both.

packages/core/src/loaders/imageLoader.ts
```diff
@@ -183,6 +196,126 @@ export function loadAndCacheImages(
...
+    throw new Error(
+      'createAndCacheDerivedImage: parameter imageId must not be undefined'
+    );
+  }
+  const imagePlaneModule = metaData.get('imagePlaneModule', imageId);
+
+  const length = imagePlaneModule.rows * imagePlaneModule.columns;
+  let numBytes, TypedArray;
+
+  const { targetBuffer } = options;
```
wayfarer3130, line 223: Could you create the target buffer based on the imagePixelModule for the new derived image id?

packages/core/src/loaders/imageLoader.ts
wayfarer3130, line 278: Could you move this creator code up into the createAndCacheDerivedImages - that way the createAndCacheDerivedImage just uses the already registered data, and if you are creating them for some other purpose/setup, then you would need to manage creating the data for them externally - but it allows creating such data externally.

packages/tools/src/tools/segmentation/BrushTool.ts
```diff
@@ -122,30 +125,88 @@ class BrushTool extends BaseTool {
...
+    } else {
+      const { imageIds: segmentationImageIds } =
+        labelmapData as LabelmapSegmentationDataStack;
+
+      // const { zSpacing, sortedImageIds, origin } =
+      //   sortImageIds(referencedImageIds);
+
+      // const orderedSegmentationImageIds = sortedImageIds.map((imageId) =>
+      //   getDerivedImageId(imageId, referencedImageIds, segmentationImageIds)
+      // );
```
lscoder, line 170: Remove?

packages/tools/src/tools/segmentation/BrushTool.ts
```diff
@@ -122,30 +125,88 @@ class BrushTool extends BaseTool {
...
+        throw new Error('No current segmentation image id');
+      }
 
-    const viewportIdsToRender = [viewport.id];
+      // here we should identify if we can perform sphere manipulation
+      // for these stack of images, if the metadata is not present
+      // to create a volume or if there are inconsistencies between
+      // the image metadata we should not allow the sphere manipulation
+      // and should throw an error or maybe simply just allow circle manipulation
+      // and not sphere manipulation
```
lscoder, line 185: IMO it should not allow the sphere manipulation and throw an error because the tool shouldn't be even enabled on the toolbar when the active viewport is a stack viewport

packages/tools/src/tools/segmentation/RectangleScissorsTool.ts
sedghi: I think we are fixing this in your strategy class

packages/core/src/loaders/imageLoader.ts
```diff
@@ -183,6 +196,126 @@ export function loadAndCacheImages(
...
+    windowWidth: 0,
+    color: false,
+    numComps: 1,
+    slope: 1,
+    minPixelValue: 0,
+    maxPixelValue: 255,
+    voiLUTFunction: undefined,
+    rows: imagePlaneModule.rows,
+    columns: imagePlaneModule.columns,
+    sizeInBytes: imageScalarData.byteLength | numBytes,
```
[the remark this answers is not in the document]
sedghi: In no case

packages/core/src/loaders/imageLoader.ts
sedghi: I mean it is derived from that so I'm not sure what is the benefit there, the options.imageId is basically just for naming

packages/core/src/loaders/imageLoader.ts
sedghi: They should be able to overwrite their previous state in the provider, why do we want to see if they exist first?

packages/core/src/loaders/imageLoader.ts
sedghi: the options is only for name, I changed the first argument to referencedImageId

packages/tools/src/eventListeners/segmentation/stackImageChangeEventListener.ts
sedghi: No it is not about that, it is about we have a stack of CT and PT and when we change the slice (from CT to PT) we can't use the CT actor to fill the PT segmentation, we should re-create based on its new metadata

packages/tools/src/tools/annotation/EllipticalROITool.ts
sedghi: I don't want those to get recalculated, the whole point is to not recalculate on every check

packages/core/src/loaders/imageLoader.ts
sedghi: It doesn't need to, since a segmentation can have a different type than the base (reference) imageid

packages/core/src/types/IImage.ts
sedghi: One image will only have always one reference, this is not about segmentation

packages/core/src/utilities/derivedImageMetadataProvider.ts
sedghi: GenericMetadataProvider

packages/core/src/loaders/imageLoader.ts
sedghi: add a random generator

sedghi: createLocalImage

sedghi: refactor this outside


  ['imagePixelModule', 'imagePlaneModule', 'generalSeriesModule'].forEach(
    (type) => {
      derivedImageMetadataProvider.add(image.imageId, {
        type,
        metadata: metaData.get(type, referencedImageId),
      });
    }
  );

  derivedImageMetadataProvider.add(image.imageId, {
    type: 'referencedImageId',
    metadata: {
      referencedImageId,
    },
  });

packages/core/src/loaders/imageLoader.ts
sedghi: createAndCacheLocalImage and derivedImage will run that

packages/core/src/loaders/imageLoader.ts
```diff
@@ -183,6 +219,170 @@ export function loadAndCacheImages(
...
+  preventCache = false
+): Promise<IImage> {
+  if (referencedImageId === undefined) {
+    throw new Error(
+      'createAndCacheDerivedImage: parameter imageId must not be undefined'
+    );
+  }
+
+  if (options.imageId === undefined) {
+    options.imageId = `derived:${uuidv4()}`;
```
wayfarer3130, line 244: Can you switch to crypto.randomUUID() - that generates a v4 uuid without needing a library (might not have the exact name there)

packages/core/src/utilities/findMapKeyByValue.ts
```diff
@@ -0,0 +1,10 @@
+function findMapKeyByValue(map, searchValue) {
```
wayfarer3130, line 1: Can you add comments on this that it will be O(n) for the number of entries?  It needs some sort of warning that it is expensive to use.

packages/core/src/utilities/findMapKeyByValue.ts
wayfarer3130, line 1: You aren't using this - please remove this file.  I don't think including it is a good idea right now due to performance issues.  As suggesed elsewhere, implementing BiMap  would be the right choice

packages/tools/src/tools/annotation/CircleROITool.ts
```diff
@@ -984,9 +981,19 @@ class CircleROITool extends AnnotationTool {
...
+          ellipseObj.xRadius !== 0 ? 1 / ellipseObj.xRadius ** 2 : 0;
```
wayfarer3130, line 985: You don't need to continue computing this here since it is done internally on the first call.

wayfarer3130 (commented):
Just a couple of changes, then I will approve it

packages/core/src/loaders/imageLoader.ts
sedghi: uuid4 is our implementation and is not reading a library, and uses crypto

```js
// prettier-ignore
// @ts-nocheck
/**
 * Generates a unique id that has limited chance of collision
 *
 * @see {@link https://stackoverflow.com/a/2117523/1867984|StackOverflow: Source}
 * @returns a v4 compliant GUID
 */
export default function uuidv4(): string {
  return ([1e7]+-1e3+-4e3+-8e3+-1e11).replace(/[018]/g, c =>
    (c ^ crypto.getRandomValues(new Uint8Array(1))[0] & 15 >> c / 4).toString(16)
  );
}

```

packages/core/src/utilities/updateVTKImageDataWithCornerstoneImage.ts
```diff
@@ -0,0 +1,38 @@
...
+      newPixelData[i * 3 + 1] = pixelData[i * 4 + 1];
+      newPixelData[i * 3 + 2] = pixelData[i * 4 + 2];
+    }
+    // modify the image object to have the correct pixel data for later
+    // use.
+    image.rgba = false;
+    image.getPixelData = () => newPixelData;
+    scalarData.set(newPixelData);
+  } else {
+    scalarData.set(pixelData);
```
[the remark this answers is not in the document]
sedghi: I'm not sure about the impact, but would be similar before

wayfarer3130 (approved):

Sergey1888888:
Hello. Function `_imageChangeEventListener` have lines:

```
const toolGroup = getToolGroupForViewport(viewportId, renderingEngineId);
let toolGroupSegmentationRepresentations =
    SegmentationState.getSegmentationRepresentations(toolGroup.id) || [];
```

But function `getToolGroupForViewport` might return undefined. And I get error 'Uncaught TypeError: Cannot read properties of undefined (reading 'id')'.

How can I fix this error?

Sergey1888888:
Error appears even in your example: https://www.cornerstonejs.org/live-examples/stackbasic

sedghi:
@Sergey1888888 Thanks for the report. I'll fix it

PR #898, opened 2023-11-17
feat(splineROI): Add Cardinal, Linear, Catmull-Rom and B-Spline Spline Tools

lscoder:
Added four spline ROIs to "Stack Annotation Tools" example page:
- Cardinal
- Linear: Cardinal with scale set to 0 but in the future it should inherit directly from Spline instead to make the math simpler using a 2x2 matrix instead 4x4
- Catmull-Rom: Cardinal with scale set to 0.5
- B-Spline

The first three splines the line pass through all control points and the last one does not.
We should be able to add new spline curves in the future with minor changes.

For a better understanding you can watch the following Youtube videos:
- [The Beauty of Bézier Curves](https://www.youtube.com/watch?v=aVwxzDHniEw)
- [The Continuity of Splines](https://www.youtube.com/watch?v=jvPPXbo87ds)


packages/tools/src/utilities/math/point/mirror.ts
```diff
@@ -0,0 +1,14 @@
+import { Types } from '@cornerstonejs/core';
+
+export default function mirror(
```
sedghi, line 3: docs please

packages/tools/src/utilities/math/point/distanceToPoint.ts
```diff
@@ -10,14 +11,5 @@ type Point = Types.Point2 | Types.Point3;
...
-    throw Error('Both points should have the same dimensionality');
-  }
-
-  const [x1, y1, z1 = 0] = p1;
-  const [x2, y2, z2 = 0] = p2;
-
-  return Math.sqrt(
-    Math.pow(x1 - x2, 2) + Math.pow(y1 - y2, 2) + Math.pow(z1 - z2, 2)
-  );
+  return Math.sqrt(distanceToPointSquared(p1, p2));
```
sedghi, line 14: maybe modify the docs that it is sqaured

packages/tools/src/utilities/math/aabb/distanceToPoint.ts
```diff
@@ -0,0 +1,9 @@
+import { Types } from '@cornerstonejs/core';
+import distanceToPointSquared from './distanceToPointSquared';
+
+export default function distanceToPoint(
```
sedghi, line 4: docs please

packages/tools/src/tools/annotation/splines/types/CardinalSplineProps.ts
```diff
@@ -0,0 +1,6 @@
+import { SplineProps } from './SplineProps';
+
+export type CardinalSplineProps = SplineProps & {
```
sedghi, line 3: Types should move to tools/types and maybe we create name space of SplineTypes then we can do SlineTypes.CardinalProps etc.

packages/tools/src/tools/annotation/splines/types/SplineProps.ts
```diff
@@ -0,0 +1,4 @@
+export type SplineProps = {
```
sedghi, line 1: docs please

packages/tools/src/tools/annotation/splines/types/CardinalSplineProps.ts
sedghi: All missing docs

packages/tools/src/tools/annotation/SplineROITool.ts
```diff
@@ -0,0 +1,895 @@
...
+            DEFAULT: {
+              resolution: 20,
+              addControlPointLineDistance: 6,
+              delControlPointDistance: 6,
+              addControlPointEnabled: true,
+              deleteControlPointEnabled: true,
+              showControlPointsConnectors: false,
+            },
+          },
+          type: 'CATMULLROM',
```
sedghi, line 100: different types can be static prop on the tool so that in the example or app SplineROITool.CATMULLROM

packages/tools/src/tools/annotation/SplineROITool.ts
```diff
@@ -0,0 +1,895 @@
...
+            LINEAR: {
+              Class: LinearSpline,
+            },
+            BSPLINE: {
+              Class: BSpline,
+              addControlPointEnabled: false,
+              deleteControlPointEnabled: false,
+              showControlPointsConnectors: true,
+            },
+            DEFAULT: {
```
sedghi, line 91: what is default?

packages/tools/src/tools/annotation/SplineROITool.ts
```diff
@@ -0,0 +1,895 @@
...
+              resolution: 20,
+              addControlPointLineDistance: 6,
+              delControlPointDistance: 6,
+              addControlPointEnabled: true,
+              deleteControlPointEnabled: true,
+              showControlPointsConnectors: false,
```
sedghi, line 97: How about 

controlPointAdditionDistance: 6,
controlPointDeletionDistance: 6,
showControlPointsConnectors: false,
controlPointAdditionEnabled: true,
controlPointDeletionEnabled: true,

packages/tools/src/tools/annotation/SplineROITool.ts
```diff
@@ -0,0 +1,895 @@
...
+            },
+          },
+          type: 'CATMULLROM',
+        },
+        actions: [
+          {
+            method: 'addControlPoint',
```
sedghi, line 104: ideally these should be enums as well that can be used from outside with the tool.actiosn.addControlPoint or somethingg

packages/tools/src/tools/annotation/SplineROITool.ts
```diff
@@ -0,0 +1,895 @@
...
+    const worldPos = currentPoints.world;
+
+    const enabledElement = getEnabledElement(element);
+    const { viewport, renderingEngine } = enabledElement;
+
+    this.isDrawing = true;
+
+    const camera = viewport.getCamera();
+    const { viewPlaneNormal, viewUp } = camera;
+    const splineConfig = this._getActiveSplineConfig();
```
sedghi, line 169: _getActiveSplineConfig is uses only here, how about we just put the logic here instead of refactoring the function, it is one line

packages/tools/src/tools/annotation/SplineROITool.ts
```diff
@@ -0,0 +1,895 @@
...
+    annotation: SplineROIAnnotation;
+    viewportIdsToRender: Array<string>;
+    handleIndex?: number;
+    movingTextBox?: boolean;
+    newAnnotation?: boolean;
+    hasMoved?: boolean;
+  } | null;
+  isDrawing: boolean;
+  isHandleOutsideImage = false;
+  splines: Map<string, ISpline> = new Map();
```
sedghi, line 65: what is this supposed to hold? it is too generic name. Is it all the segments?

packages/tools/src/tools/annotation/SplineROITool.ts
```diff
@@ -0,0 +1,895 @@
...
+    let spline = this.splines.get(annotation.annotationUID);
+
+    if (!spline) {
+      const splineType = annotation.data.spline.type;
+      const splineConfig = this._getSplineConfig(splineType);
+
+      console.log('>>>> create :: splineConfig :: ', splineConfig);
+
+      spline = new splineConfig.Class();
+      spline.resolution = splineConfig.resolution;
```
sedghi, line 845: can resolution be passed via constructor instead of this?

packages/tools/src/tools/annotation/SplineROITool.ts
```diff
@@ -0,0 +1,895 @@
...
+    spline.closed = !!data.spline?.closed;
+
+    return spline;
+  };
+
+  private _getAnnotationSpline = (
+    element: HTMLDivElement,
+    annotation: SplineROIAnnotation
+  ) => {
+    let spline = this.splines.get(annotation.annotationUID);
```
sedghi, line 871: again not sure what is spline here, is this the one spline (with all segments and control points)? then I"m not sure why it is acting as state manager and has knowledge of other annotationUIDs

packages/tools/src/tools/annotation/SplineROITool.ts
```diff
@@ -0,0 +1,895 @@
...
+
+    const spline = this._getAnnotationSpline(element, annotation);
+    const canvasPos = evt.detail.currentPoints.canvas;
+    const closestPointInfo = spline.getClosestPoint(canvasPos);
+
+    if (closestPointInfo.distance <= maxDist) {
+      spline.addControlPointAt(closestPointInfo.uValue);
+      annotation.data.handles.points = spline
+        .getControlPoints()
+        .map((p) => canvasToWorld(p));
```
sedghi, line 765: not sure why you are redoing the canvasToWorld here? shouldn't you only do it for the last added point?

packages/tools/src/tools/annotation/SplineROITool.ts
```diff
@@ -0,0 +1,895 @@
...
+      const splineType = annotation.data.spline.type;
+      const splineConfig = config.spline.configuration[splineType] ?? {};
+      const spline = this._getAnnotationSpline(element, annotation);
+      const splinePolyline = spline.getPolylinePoints();
+
+      // If rendering engine has been destroyed while rendering
+      if (!viewport.getRenderingEngine()) {
+        console.warn('Rendering Engine has been destroyed');
+        return renderStatus;
+      }
```
sedghi, line 661: maybe this should be sooner

packages/tools/src/tools/annotation/SplineROITool.ts
```diff
@@ -0,0 +1,895 @@
...
+    const key = (eventDetail.key ?? '').toLowerCase();
+
+    console.log('>>>>> key:', key);
+    console.log('>>>>> keyCode:', keyCode);
+    const deleteLastPoint =
+      keyCode === 8 ||
+      keyCode === 46 ||
+      key === 'backspace' ||
+      key === 'delete';
+
```
sedghi, line 383: should be configurable

packages/tools/src/tools/annotation/splines/CubicSpline.ts
```diff
@@ -0,0 +1,217 @@
...
+
+        minX = minX <= lineSegAABB.minX ? minX : lineSegAABB.minX;
+        minY = minY <= lineSegAABB.minY ? minY : lineSegAABB.minY;
+        maxX = maxX >= lineSegAABB.maxX ? maxX : lineSegAABB.maxX;
+        maxY = maxY >= lineSegAABB.maxY ? maxY : lineSegAABB.maxY;
+        curveSegmentLength += lineSegment.length;
+      });
+
+      curveSegments[i] = {
+        controlPoints: { p0, p1, p2, p3 },
```
sedghi, line 39: please add a comment which is which, one before, one after etc

packages/tools/src/tools/annotation/splines/CubicSpline.ts
```diff
@@ -0,0 +1,217 @@
...
+        aabb: { minX, minY, maxX, maxY },
+        length: curveSegmentLength,
+        lengthStart,
```
sedghi, line 42: I'm not sure what is lengthStart, maybe rename?

packages/tools/src/tools/annotation/splines/CubicSpline.ts
```diff
@@ -0,0 +1,217 @@
...
+    const numLineSegments = this.resolution + 1;
+    const inc = 1 / numLineSegments;
+    const minU = curveSegmentIndex;
+    let maxU = minU + 1;
+
+    // 'u' must be greater than or equal to 0 and smaller than N where N is the number of segments
+    // otherwise it does not find the spline segment when it is not a closed curve because it is
+    // 0-based indexed. In this case `u` needs to get very close to the end point but never touch it
+    if (!this.closed && curveSegmentIndex === numCurveSegments - 1) {
+      maxU -= 1e-8;
```
sedghi, line 168: maybe enum on top

packages/tools/src/tools/annotation/splines/CubicSpline.ts
```diff
@@ -0,0 +1,217 @@
...
+
+      lineSegments.forEach((lineSegment) => {
+        const { aabb: lineSegAABB } = lineSegment;
+
+        minX = minX <= lineSegAABB.minX ? minX : lineSegAABB.minX;
+        minY = minY <= lineSegAABB.minY ? minY : lineSegAABB.minY;
+        maxX = maxX >= lineSegAABB.maxX ? maxX : lineSegAABB.maxX;
+        maxY = maxY >= lineSegAABB.maxY ? maxY : lineSegAABB.maxY;
+        curveSegmentLength += lineSegment.length;
+      });
```
sedghi, line 36: How about this 

```js
let minX = Infinity;
let minY = Infinity;
let maxX = -Infinity;
let maxY = -Infinity;

lineSegments.forEach(({ aabb: lineSegAABB }) => {
  minX = Math.min(minX, lineSegAABB.minX);
  minY = Math.min(minY, lineSegAABB.minY);
  maxX = Math.max(maxX, lineSegAABB.maxX);
  maxY = Math.max(maxY, lineSegAABB.maxY);
});
}
```

packages/tools/src/tools/annotation/splines/CubicSpline.ts
```diff
@@ -0,0 +1,217 @@
...
+    // each point of the curve
+    const q1 = m00 * 1 + m10 * t + m20 * tt + m30 * ttt;
+    const q2 = m01 * 1 + m11 * t + m21 * tt + m31 * ttt;
+    const q3 = m02 * 1 + m12 * t + m22 * tt + m32 * ttt;
+    const q4 = m03 * 1 + m13 * t + m23 * tt + m33 * ttt;
+
+    const tx = p0[0] * q1 + p1[0] * q2 + p2[0] * q3 + p3[0] * q4;
+    const ty = p0[1] * q1 + p1[1] * q2 + p2[1] * q3 + p3[1] * q4;
+
+    return [tx, ty];
```
sedghi, line 118: most probably we can use gl-matrix here

```js
import { mat4, vec4 } from 'gl-matrix';

const tt = t * t;
const ttt = tt * t;
const T = vec4.fromValues(1, t, tt, ttt);


// might need to do mat4 from transformMatrix i'm not sure
const Q = vec4.create();
vec4.transformMat4(Q, T, transformMatrix);

const P0 = vec4.fromValues(p0[0], p0[1], p0[2], p0[3]
const P1 = //
const P2 = //
const P3 = //

finally the last part

```

packages/tools/src/utilities/math/aabb/distanceToPointSquared.ts
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * @param aabb - Axis-aligned bound box
+ * @param point - 2D point
+ * @returns The closest distance between the 2D point and the AABB
+ */
+export default function distanceToPointSquared(
+  aabb: Types.AABB2,
+  point: Types.Point2
+): number {
+  const aabbSize = [aabb.maxX - aabb.minX, aabb.maxY - aabb.minY];
```
sedghi, line 17: refactor aabb.maxX - aabb.minX it is calculated twice

packages/tools/src/utilities/math/aabb/distanceToPointSquared.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  const dx = point[0] - aabbSize[0] * 0.5;
+  const dy = point[1] - aabbSize[1] * 0.5;
+
+  // dx >  0 && dy >  0: diagonal line connecting the point to AABB's corner
+  // dx >  0 && dy <= 0: a line parallel to x-axis connecting the point to AABB's right side
+  // dx <= 0 && dy >  0: a line parallel to y-axis connecting the point to AABB's top side
+  // dx <= 0 && dy <= 0: the point is inside the AABB
+  if (dx > 0 && dy > 0) {
+    return dx ** 2 + dy ** 2;
```
sedghi, line 45: dx ** dx + dy ** dy is faster

packages/tools/src/utilities/math/line/distanceToPointSquaredInfo.ts
```diff
@@ -0,0 +1,41 @@
...
+): {
+  point: Types.Point2;
+  distanceSquared: number;
+} {
+  let closestPoint: Types.Point2;
+  const d2 = distanceToPointSquared(lineStart, lineEnd);
+
+  if (d2 === 0) {
+    closestPoint = lineStart;
+  }
```
sedghi, line 17: not sure what is happening here, you are using the distance of line to point, but passing both line start and end?

Is this because you want to check if they are the same?

How about simple 

```js
const areEqual = lineStart[0] === lineEnd[0] && // etc.
```

packages/tools/src/tools/annotation/splines/renderer/canvas/CatmullRomRenderer.ts
```diff
@@ -0,0 +1,6 @@
+import { SplineCanvasRenderer } from './SplineCanvasRenderer';
```
sedghi, line 1: Not sure what are these files in the renderer/canvas and renderer/svg

sedghi (commented):
great work, see my comments please. I want to try and see the annotation state, so can you add a dedicated spline example and add to example info json too? 

Thanks! you rock

packages/tools/src/tools/annotation/splines/CubicSpline.ts
lscoder: That works as well but if you have noticed I also don't use to use `.map(...)`, `.forEach(...)`  when I'm working on loops that can run hundreds of times to reduce function calls because switching between contexts when invoking subroutines requires more clock cycles in assembly. I started working on that way after investigating why a cine player was not able to play at 60fps and the annotations was one of the problems.

packages/tools/src/tools/annotation/splines/renderer/canvas/CatmullRomRenderer.ts
lscoder: They shouldn't have been pushed

packages/tools/examples/stackAnnotationTools/index.ts
```diff
@@ -227,6 +236,25 @@ async function run() {
...
+      type: 'LINEAR',
+    },
+  });
+
+  toolGroup.addToolInstance('BSplineROI', SplineROITool.toolName, {
+    spline: {
+      type: 'BSPLINE',
+    },
+  });
+
```
sedghi, line 257: I prefer this example to not have any spline

packages/tools/examples/splineTools/index.ts
```diff
@@ -0,0 +1,278 @@
...
+
+const { ViewportType } = Enums;
+const { MouseBindings } = csToolsEnums;
+const renderingEngineId = 'myRenderingEngine';
+const viewportId = 'CT_STACK';
+
+// ======== Set up page ======== //
+setTitleAndDescription(
+  'Spline Tools Stack',
+  'Spline tools for a stack viewport'
```
sedghi, line 35: please add more information here that you can change the spline type or use sliders etc. etc.

packages/tools/examples/splineTools/index.ts
```diff
@@ -0,0 +1,278 @@
...
+    toolGroup.setToolActive(newSelectedToolName, {
+      bindings: [
+        {
+          mouseButton: MouseBindings.Primary, // Left Click
+        },
+      ],
+    });
+
+    // Set the old tool passive
+    toolGroup.setToolPassive(selectedToolName);
```
sedghi, line 107: maybe do the passive before the active?

packages/tools/examples/splineTools/index.ts
```diff
@@ -0,0 +1,278 @@
...
+
+const Splines = {
+  CardinalSplineROI: {
+    toolConfigKey: 'CARDINAL',
+  },
+  CatmullRomSplineROI: {
+    toolConfigKey: 'CATMULLROM',
+  },
+  LinearSplineROI: {
+    toolConfigKey: 'LINEAR',
```
sedghi, line 80: splineType instead of toolConfigKey?

packages/tools/src/utilities/math/line/distanceToPointSquaredInfo.ts
```diff
@@ -0,0 +1,58 @@
...
+  point: Types.Point2;
+  distanceSquared: number;
+} {
+  let closestPoint: Types.Point2;
+  let distanceSquared;
+
+  // Check if lineStart is the same as lineEnd which means
+  // the line has length equal to 0.
+  if (lineStart[0] === lineEnd[0] && lineStart[1] === lineEnd[1]) {
+    distanceSquared = 0;
```
sedghi, line 30: if the start and end are the same doesn't necessarly mean the distance is 0

This should be 

```js
let distanceSquared = math.point.distanceToPointSquared(lineStart, lineEnd);

if (lineStart[0] === lineEnd[0] && lineStart[1] === lineEnd[1]) {
  
} else {
  
}

```

packages/tools/src/utilities/math/line/distanceToPointSquaredInfo.ts
```diff
@@ -0,0 +1,58 @@
...
+    closestPoint = lineStart;
+  } else {
+    distanceSquared = math.point.distanceToPointSquared(lineStart, lineEnd);
+  }
+
+  if (!closestPoint) {
+    const t =
```
sedghi, line 37: what does `t` mean? maybe some variable name related to `projection` ?

packages/tools/src/utilities/math/aabb/distanceToPointSquared.ts
```diff
@@ -0,0 +1,53 @@
...
+    maxX: aabb.maxX - aabbCenter[0],
+    maxY: aabb.maxY - aabbCenter[1],
+  };
+
+  // The point is not only translated but also mirroed to the first quadrant to simplify the math
+  point = [
+    Math.abs(point[0] - aabbCenter[0]),
+    Math.abs(point[1] - aabbCenter[1]),
+  ];
+
```
sedghi, line 38: you are modifying the function argument, which might have side effect

packages/tools/src/tools/annotation/splines/types/CardinalSplineProps.ts
```diff
@@ -0,0 +1,11 @@
...
+
+/**
+ * Cardinal spline properties
+ */
+export type CardinalSplineProps = SplineProps & {
+  /** Scale that must be in 0-1 range */
+  scale?: number;
+  /** Fixed scale used by children classes (Catmull-Rom and Linear splines) */
+  fixedScale?: boolean;
+};
```
sedghi, line 11: after the review is done, before merging, can you move these to types folder?

packages/tools/src/tools/annotation/SplineROITool.ts
```diff
@@ -0,0 +1,1107 @@
...
+    evt: EventTypes.InteractionEventType
+  ): SplineROIAnnotation => {
+    const eventDetail = evt.detail;
+    const { currentPoints, element } = eventDetail;
+    const worldPos = currentPoints.world;
+
+    const enabledElement = getEnabledElement(element);
+    const { viewport, renderingEngine } = enabledElement;
+
+    this.isDrawing = true;
```
sedghi, line 167: It is odd that we have `isDrawing = true` at the top. Usually, other tools place it at the bottom.

packages/tools/src/tools/annotation/SplineROITool.ts
```diff
@@ -0,0 +1,1107 @@
...
+    const enabledElement = getEnabledElement(element);
+    const { viewport } = enabledElement;
+    const { worldToCanvas } = viewport;
+    const splineType = annotation.data.spline.type;
+    const splineConfig = this._getSplineConfig(splineType);
+    let spline = this.splineInstancesCache.get(annotation);
+
+    if (!spline) {
+      spline = new splineConfig.Class();
+      this.splineInstancesCache.set(annotation, spline);
```
sedghi, line 974: I believe it would be better to have the spline instance caches inside the annotation itself for a couple of reasons. Firstly, it would be more manageable for memory. Secondly, it would align better with our architecture.

packages/tools/src/tools/annotation/SplineROITool.ts
```diff
@@ -0,0 +1,1107 @@
...
+      spline = this._createOrUpdateAnnotationSpline(element, annotation);
+    }
+
+    return spline;
+  };
+
+  private _deleteAnnotationSpline = (annotation: SplineROIAnnotation) => {
+    this.splineInstancesCache.delete(annotation);
+  };
+
```
sedghi, line 1020: This is a side effect of having it outside the annotation manager, ideally remove annotation should remove its spline instances

packages/tools/src/tools/annotation/SplineROITool.ts
```diff
@@ -0,0 +1,1107 @@
...
+    element.removeEventListener(Events.TOUCH_END, this._endCallback);
+    element.removeEventListener(Events.TOUCH_DRAG, this._dragCallback);
+    element.removeEventListener(Events.TOUCH_TAP, this._endCallback);
+  };
+
+  private _activateDraw = (element) => {
+    state.isInteractingWithTool = true;
+
+    element.addEventListener(Events.KEY_DOWN, this._keyDownCallback);
+    element.addEventListener(Events.MOUSE_DOWN, this._mouseDownCallback);
```
sedghi, line 574: why twice MOUSE_DOWN?

packages/tools/src/tools/annotation/SplineROITool.ts
```diff
@@ -0,0 +1,1107 @@
...
+      toolGroupId: this.toolGroupId,
+      toolName: this.getToolName(),
+      viewportId: enabledElement.viewport.id,
+    };
+
+    for (let i = 0; i < annotations.length; i++) {
+      const annotation = annotations[i] as SplineROIAnnotation;
+      const { annotationUID, data, highlighted } = annotation;
+      const { handles } = data;
+      const { points, activeHandleIndex } = handles;
```
sedghi, line 647: is this controlPoints? then can we do const { points: controlPoints} for clarity?

packages/tools/src/tools/annotation/SplineROITool.ts
```diff
@@ -0,0 +1,1107 @@
...
+        annotation
+      ) as string;
+
+      const canvasCoordinates = points.map((p) =>
+        worldToCanvas(p)
+      ) as Types.Point2[];
+
+      const splineType = annotation.data.spline.type;
+      const splineConfig = this._getSplineConfig(splineType);
+      const spline = this._createOrUpdateAnnotationSpline(element, annotation);
```
sedghi, line 673: Is there any scenario where you don't have the spline and you really need to create it? I guess in this line it is more on the update and these lines

```js
 // Update spline resolution in case it has changed
    if (spline.resolution !== splineConfig.resolution) {
      spline.resolution = parseInt(splineConfig.resolution);
      annotation.invalidated = true;
    }

    // Update Cardinal spline scale in case it has changed
    if (
      spline instanceof CardinalSpline &&
      !spline.fixedScale &&
      splineConfig.scale !== undefined &&
      spline.scale !== splineConfig.scale
    ) {
      spline.scale = splineConfig.scale;
      annotation.invalidated = true;
    }

```

maybe we just refactor those and call those here?

packages/tools/src/tools/annotation/SplineROITool.ts
```diff
@@ -0,0 +1,1107 @@
...
+      const splinePolyline = spline.getPolylinePoints();
+
+      data.spline.polyline = splinePolyline.map((point) =>
+        viewport.canvasToWorld(point)
+      );
```
sedghi, line 678: let's use for loop

packages/tools/src/tools/annotation/SplineROITool.ts
```diff
@@ -0,0 +1,1107 @@
...
+    }
+
+    const eventDetail = evt.detail;
+    const { element } = eventDetail;
+
+    const viewportIdsToRender = getViewportIdsWithToolToRender(
+      element,
+      this.getToolName()
+    );
+
```
sedghi, line 850: misplaced, too soon

packages/tools/src/tools/annotation/SplineROITool.ts
```diff
@@ -0,0 +1,1107 @@
...
+    }
+
+    // Add a point at the `u` position from Parameter Space
+    const { index, point: canvasPoint } = spline.addControlPointAt(
+      closestPointInfo.uValue
+    );
+
+    data.handles.points.splice(index, 0, canvasToWorld(canvasPoint));
+    annotation.invalidated = true;
+
```
sedghi, line 870: very nice

packages/tools/src/tools/annotation/splines/Spline.ts
```diff
@@ -0,0 +1,683 @@
...
+    points.forEach((point) => this.addControlPoint(point));
+  }
+
+  /**
+   * Add a control point specifying its `u` value in Parameter Space which is a number from 0 to N
+   * where N is the number of curve segments. The integer part is the curve segment index and the
+   * decimal part is the `t` value on that curve segment.
+   * @param u - `u` value in Parameter Space
+   */
+  public addControlPointAt(u: number): ControlPointInfo {
```
sedghi, line 146: I even prefer addControlPointAtU

packages/tools/src/tools/annotation/splines/Spline.ts
```diff
@@ -0,0 +1,683 @@
...
+    this.addControlPoints(points);
+  }
+
+  /**
+   * Update the coordinate of a control point given its index
+   * @param index - Control point index
+   * @param x - Control point X coordinate
+   * @param y - Control point Y coordinate
+   */
+  public updateControlPoint(index: number, x: number, y: number): void {
```
sedghi, line 214: ByIndex to be similar to method above?

packages/tools/src/tools/annotation/splines/Spline.ts
```diff
@@ -0,0 +1,683 @@
...
+  /**
+   * Finds the closest point on the spline curve given 2D point
+   * @param point - Reference 2D point
+   * @returns Closest point on the spline curve
+   */
+  public getClosestPoint(point: Types.Point2): ClosestSplinePoint {
+    this._update();
+
+    const curveSegmentsDistInfo =
+      this._getCurveSegmmentsDistanceSquaredInfo(point);
```
sedghi, line 293: Segment not Segmments

packages/tools/src/tools/annotation/splines/Spline.ts
```diff
@@ -0,0 +1,683 @@
...
+  /**
+   * Checks if a point is near to the spline curve
+   * @param point - Reference point
+   * @param maxDist - Maximum allowed distance
+   * @returns True if the point is close to the spline curve or false otherwise
+   */
+  public isPointNearCurve(point: Types.Point2, maxDist: number): boolean {
+    this._update();
+
+    const curveSegments = this._getCurveSegmmentsWithinRange(point, maxDist);
```
sedghi, line 442: simmilar typo

packages/tools/src/tools/annotation/splines/Spline.ts
```diff
@@ -0,0 +1,683 @@
...
+    };
+  }
+
+  /**
+   * Finds the closest control point given a 2D point and a maximum distance
+   * @param point - Reference 2D point
+   * @param maxDist - Maximum distance
+   * @returns Closest control point that is within the given range or undefined otherwise
+   */
+  public getClosestControlPointWithinRange(
```
sedghi, line 273: withinDistance?

packages/tools/src/tools/annotation/splines/Spline.ts
```diff
@@ -0,0 +1,683 @@
...
+      : undefined;
+  }
+
+  /**
+   * Finds the closest point on the spline curve given 2D point
+   * @param point - Reference 2D point
+   * @returns Closest point on the spline curve
+   */
+  public getClosestPoint(point: Types.Point2): ClosestSplinePoint {
+    this._update();
```
sedghi, line 290: why we are calling update? shouldn't we check first if it is invalidate?

packages/tools/src/tools/annotation/splines/Spline.ts
```diff
@@ -0,0 +1,683 @@
...
+        }
+      }
+    }
+
+    return numIntersections % 2 === 1;
+  }
+
+  protected abstract getTransformMatrix(): number[];
+
+  protected abstract getSplineCurves(): SplineCurveSegment[];
```
sedghi, line 535: maybe on top?

packages/tools/src/tools/annotation/splines/Spline.ts
```diff
@@ -0,0 +1,683 @@
...
+    curveSegments.forEach((curveSegment) => {
+      const { aabb: curveSegAABB } = curveSegment;
+
+      minX = minX <= curveSegAABB.minX ? minX : curveSegAABB.minX;
+      minY = minY <= curveSegAABB.minY ? minY : curveSegAABB.minY;
+      maxX = maxX >= curveSegAABB.maxX ? maxX : curveSegAABB.maxX;
+      maxY = maxY >= curveSegAABB.maxY ? maxY : curveSegAABB.maxY;
+      length += curveSegment.length;
+    });
+
```
sedghi, line 558: for loop

packages/tools/src/tools/annotation/splines/CubicSpline.ts
```diff
@@ -0,0 +1,234 @@
...
+import type { SplineCurveSegment } from './types/SplineCurveSegment';
+import type { SplineLineSegment } from './types/SplineLineSegment';
+
+// The `u` in Parameter Space used when spliting a curve segment into line segments must
+// be greater than or equal to `curveSegmentIndex` and smaller than `curveSegmentIndex + 1`.
+// In this case we are using `curveSegmentIndex + 1 - MAX_U_ERROR`
+const MAX_U_ERROR = 1e-8;
+
+const times = [];
+(window as any).times = times;
```
sedghi, line 13: remove

sedghi (commented):
Looks great, minor comments + the major one regarding moving the spline instances into the annotation

packages/tools/src/utilities/math/aabb/distanceToPointSquared.ts
lscoder: - Renamed the point to `translatedPoint`.
- Removed the lines that updated AABB's origin because we don't need it anymore because `aabbSize * 0.5` gives what is needed.

packages/tools/src/tools/annotation/splines/types/CardinalSplineProps.ts
lscoder: Moved

packages/tools/src/tools/annotation/SplineROITool.ts
sedghi: The this.isDrawing is usually the last thing (or close to last) in the addNewAnnotation in any tools

packages/tools/src/tools/annotation/SplineROITool.ts
sedghi: Later, we can extract the serializable pieces and reinstantiate the class.

packages/tools/src/tools/annotation/SplineROITool.ts
sedghi: performance

packages/tools/src/tools/annotation/splines/Spline.ts
sedghi: I'm talking about it is abstract

packages/tools/src/tools/annotation/splines/Spline.ts
sedghi: Sorry my bad, should be for loop for performance

packages/tools/src/tools/annotation/splines/CubicSpline.ts
lscoder: Updated

packages/tools/src/tools/annotation/SplineROITool.ts
lscoder: Updated

packages/tools/src/tools/annotation/SplineROITool.ts
lscoder: Updated

sedghi (approved):

sedghi:
One of the most amazing PRs has been merged into cs3d! It's truly a remarkable achievement. Great job!

PR #902, opened 2023-11-20
fix(metadata): exceptions trying to use the metadata providers

wayfarer3130:
### Context

A recent change started making use of non-string parameters for metadata query, causing some providers to fail.  Instead, use varargs for the providers needing more than one query argument.

### Changes & Results

Use varargs.

### Testing

Run OHIF linked to CS3d, plus the examples htj2kStackBasic and htj2kVolumeBasic.  They should still run progressively.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  

### DIFF
--- a/common/reviews/api/core.api.md
+++ b/common/reviews/api/core.api.md
@@ -44,7 +44,7 @@ type ActorSliceRange = {
 };
 
 // @public (undocumented)
-function addProvider(provider: (type: string, query: any) => any, priority?: number): void;
+function addProvider(provider: (type: string, ...query: string[]) => any, priority?: number): void;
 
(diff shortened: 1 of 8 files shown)


sedghi (changes requested):
I think this is once again a side effect of the HTJ2K metadata provider changes. We should modify it to match the signature of others. You can extract the third argument 'stack' from the provider in a different way. I believe we have inadvertently affected several other custom providers in people's apps with this change.

sedghi (approved):
Thanks makes a lot more sense

PR #909, opened 2023-11-22
feat(dynamic brush): Add a dynamic threshold brush tool with preview

wayfarer3130:
### Context

The existing threshold tool is  bit too coarse grained as it doesn't have a good way to find the center values.  This PR adds a dynamic threshold brush strategy to allow dynamically choosing the threshold.

### Changes & Results

Created a new example demo labelmapsegmentationdynamicthreshold to demonstrate the preview and dynamic threshold tools.
Redid how the brush tool works to have a brush strategy that is built on composition to allow combining various tools in different ways.
Wrote a full set of composition objects for the regular and dynamic threshold.
Created a new model for accessing data to hide the index/ijk/scalar data information.

### Testing

Run the examples:
labelmapSegmentationDynamicThreshold
stackSegmentation

Note the dynamic tools work on both the stack and volume segmentations.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  


packages/dicomImageLoader/.webpack/webpack-dev.js
```diff
@@ -13,7 +13,7 @@ const devConfig = {
     // Bundles; takes precedence over contentBase
     // Static content
     static: path.resolve(path.join(__dirname, './../../examples')),
-    port: 3000,
+    port: 3010,
```
sedghi, line 16: please revert

packages/tools/examples/labelmapSegmentationDynamicThreshold/index.ts
```diff
@@ -0,0 +1,504 @@
...
+// ============================= //
+addDropdownToToolbar({
+  options: { values: ['1', '2', '3'], defaultValue: '1' },
+  onSelectedValueChange: (segmentIndex) => {
+    segmentation.segmentIndex.setActiveSegmentIndex(
+      segmentationId,
+      Number(segmentIndex)
+    );
+  },
+});
```
sedghi, line 223: requires a label:

packages/tools/examples/labelmapSegmentationTools/index.ts
```diff
@@ -152,21 +160,31 @@ addDropdownToToolbar({
   },
 });
 
-const thresholdOptions = ['CT Fat: (-150, -70)', 'CT Bone: (200, 1000)'];
+const thresholdOptions = new Map<string, any>();
+thresholdOptions.set('CT Fat: (-150, -70)', {
+  threshold: [-150, -70],
+  isDynamic: false,
```
sedghi, line 166: isDynamic false should be default, I don't know why we have it here

packages/tools/src/drawingSvg/drawEllipseByCoordinates.ts
```diff
@@ -1,4 +1,4 @@
-2
+2;
```
sedghi, line 1: remove 2

utils/ExampleRunner/template-config.js
```diff
@@ -88,7 +88,7 @@ module.exports = {
   devServer: {
     hot: true,
     open: false,
-    port: 3000,
+    port: 3010,
```
sedghi, line 91: revert please

utils/ExampleRunner/template-multiexample-config.js
```diff
@@ -126,7 +126,7 @@ module.exports = {
   devServer: {
     hot: true,
     open: false,
-    port: 3000,
+    port: 3010,
```
sedghi, line 129: revert

packages/tools/test/segmentationSphereScissor_test.js
```diff
@@ -232,18 +231,21 @@ describe('Segmentation Tools --', () => {
         const image3 = canvas3.toDataURL('image/png');
 
         try {
+          console.warn('Testing axial compare');
```
sedghi, line 234: remove i guess

packages/tools/src/utilities/segmentation/brushThresholdForToolGroup.ts
```diff
@@ -16,9 +17,19 @@ export function setBrushThresholdForToolGroup(
 
   const brushBasedToolInstances = getBrushToolInstances(toolGroupId);
 
+  const configuration = {
+    ...otherArgs,
+    threshold,
+  };
+  if (threshold === undefined) {
+    delete configuration.threshold;
+  }
```
sedghi, line 26: const configuration = {
  ...otherArgs,
  ...(threshold !== undefined && { threshold }),
};

packages/tools/src/types/SegmentationStateTypes.ts
```diff
@@ -66,6 +66,11 @@ export type Segmentation = {
    * inside the segmentation tools
    */
   activeSegmentIndex: number;
+  /**
+   * Preview segment index, used for drawing previews.  Will be replaced by
+   * final segment indices when completed.
+   */
+  previewSegmentIndex?: number;
```
sedghi, line 73: I'm not entirely sure about this, to be honest. I believe we should only provide a preview color, and the index should not be influenced externally. If the imported SEG already has that index, it could lead to unexpected side effects that the user may not be aware of before importing. In my opinion, we should not allow this to be controlled.

packages/tools/src/types/LabelmapToolOperationData.ts
```diff
@@ -6,12 +8,29 @@ import {
...
+  previewColors?: Record<number, [number, number, number, number]>;
   segmentsLocked: number[];
   viewPlaneNormal: number[];
   viewUp: number[];
   strategySpecificConfiguration: any;
-  constraintFn: (pointIJK: number) => boolean;
+  // constraintFn: (pointIJK: number) => boolean;
   segmentationRepresentationUID: string;
+  points: Types.Point3[];
+  representationUID: string;
```
sedghi, line 26: How is representationUID different than segmentationRepresentationUID?

packages/tools/src/types/LabelmapToolOperationData.ts
```diff
@@ -6,12 +8,29 @@ import {
 type LabelmapToolOperationData = {
   segmentationId: string;
   segmentIndex: number;
+  /**
+   * The segment index to use for a preview
+   */
+  previewSegmentIndex?: number;
+  /**
+   * The colours to use for previewing
```
sedghi, line 16: please explain if not provided what would happen

packages/tools/src/tools/segmentation/SphereScissorsTool.ts
```diff
@@ -71,6 +72,7 @@ class SphereScissorsTool extends BaseTool {
       configuration: {
         strategies: {
           FILL_INSIDE: fillInsideSphere,
+          ERASE_INSIDE_SPHERE: eraseInsideSphere,
```
sedghi, line 75: ERASES_INSIDE? to match the FILL_INSIDE?

packages/tools/src/tools/segmentation/CircleScissorsTool.ts
```diff
@@ -69,10 +70,10 @@ class CircleScissorsTool extends BaseTool {
       configuration: {
         strategies: {
           FILL_INSIDE: fillInsideCircle,
-          // ERASE_INSIDE: eraseInsideCircle,
+          ERASE_INSIDE: eraseInsideCircle,
         },
-        defaultStrategy: 'FILL_INSIDE',
-        activeStrategy: 'FILL_INSIDE',
+        defaultStrategy: 'ERASE_INSIDE',
+        activeStrategy: 'ERASE_INSIDE',
```
sedghi, line 76: change back the default

packages/tools/src/tools/segmentation/strategies/utils/isWithinThreshold.ts
```diff
@@ -6,11 +6,12 @@ function isWithinThreshold(
   imageScalarData: TypedArray,
```
sedghi, line 6: I think the TypedArray is wrong we have something similar in core

packages/tools/src/tools/segmentation/strategies/utils/initializeTracking.ts
```diff
@@ -0,0 +1,21 @@
+import type { OperationData, InitializedOperationData } from '../BrushStrategy';
+
+/**
+ * Sets up tracking for use by preview and other services.
+ * This sets up a  TRACKING value in the strategy specific configuration which
+ * stores the original value for updated pixels, and allows the changes to be
+ * applied (eg for a preview), reverted, or acted on in other ways.
+ */
+export default {
+  createInitialized: (enabled, operationData: InitializedOperationData) => {
```
sedghi, line 10: is `enabled` `enabledElement`? please rename it to enabledElement, I was struggling with what is this

packages/tools/src/tools/base/BaseTool.ts
```diff
@@ -83,7 +83,35 @@ abstract class BaseTool implements IBaseTool {
...
+      | 'acceptPreview'
+      | 'preview'
+  ): any {
+    const { strategies, activeStrategy } = this.configuration;
+    return strategies[activeStrategy][eventType]?.call(
+      this,
+      enabledElement,
+      operationData
+    );
   }
```
sedghi, line 115: I'm not sure what is this, the name does not suggest what it does nor the documentaiton, what do you mean by `It applies the active strategy down initialization, if any.`

packages/tools/src/tools/segmentation/strategies/fillSphere.ts
```diff
@@ -100,9 +119,23 @@ export function fillInsideSphere(
   enabledElement: Types.IEnabledElement,
   operationData: OperationData
 ): void {
-  fillSphere(enabledElement, operationData, true);
+  SPHERE_STRATEGY.fill(enabledElement, operationData);
 }
 
+SPHERE_STRATEGY.assignMethods(fillInsideSphere);
```
sedghi, line 125: I kind of understood your api for new BrushStrategy, but this bit `SPHERE_STRATEGY.assignMethods(fillInsideSphere);` is quite unusually, can you work around this so that the strategy after creation doesn't need this assignMethods? It is a confusing api

```js
const SPHERE_STRATEGY = new BrushStrategy(
  'Sphere',
  initializeRegionFill,
  initializeSetValue,
  initializeSphere,
  initializeTracking
);

export function fillInsideSphere(
  enabledElement: Types.IEnabledElement,
  operationData: OperationData
): void {
  SPHERE_STRATEGY.fill(enabledElement, operationData);
}

SPHERE_STRATEGY.assignMethods(fillInsideSphere);
```

I thought initializeSetValue is acutally the fill

packages/tools/src/types/SegmentationStateTypes.ts
wayfarer3130: Can do, but be aware it will be quite a lot of work, and will not preview on anything except the primary view.  Probably 3-4 days of additional effort to get this working.  Currently the code depends on being able to set the colour into a segmentation, but if that is no longer possible, then it will become significantly harder.  Probably I will implement a direct preview map to segmentation representation.  That is cleaner of course, because it means the preview can contain multiple colours and in fact only contains the new values, not the old ones, but there is a lot of code depending on the old value.

packages/tools/src/tools/segmentation/SphereScissorsTool.ts
wayfarer3130: No, really bad idea as it means you can't use the strategy name independently.  Rather fix FILL_INSIDE to be FILL_INSIDE_SPHERE

packages/tools/src/tools/segmentation/strategies/utils/initializeTracking.ts
wayfarer3130: Can we use a short name for things that aren't used?  Like na for not applicable.

sedghi: please no!
I had to dig into the code to understand oh this is enabledElement.

packages/tools/src/types/SegmentationStateTypes.ts
sedghi: what if the previewSegmentIndex is actually already claimed by another segment when imported? this will have side effects. Why do you think it will be 3-4 days?

packages/tools/examples/labelmapSegmentationTools/index.ts
```diff
@@ -213,10 +229,16 @@ async function run() {
   // Init Cornerstone and related libraries
   await initDemo();
 
+  utilities.imageRetrieveMetadataProvider.add(
+    'volume',
+    ProgressiveRetrieveImages.interleavedRetrieveStages
+  );
+
```
sedghi, line 236: please add a comment that this is not necessary, just appealing

packages/tools/examples/labelmapSegmentationTools/index.ts
```diff
@@ -370,12 +426,15 @@ async function run() {
   );
 
   // // Add the segmentation representation to the toolgroup
-  await segmentation.addSegmentationRepresentations(toolGroupId, [
-    {
-      segmentationId,
-      type: csToolsEnums.SegmentationRepresentations.Labelmap,
-    },
-  ]);
+  const [representationUID] = await segmentation.addSegmentationRepresentations(
```
sedghi, line 429: no need for representationUID

packages/tools/examples/labelmapSegmentationDynamicThreshold/index.ts
```diff
@@ -0,0 +1,525 @@
...
+
+      toolGroup.setToolActive(toolName, {
+        bindings: [{ mouseButton: MouseBindings.Primary }],
+      });
+    }
+  },
+});
+
+const thresholdOptions = new Map<string, any>();
+thresholdOptions.set('Dynamic r=0', { isDynamic: true, delta: 0 });
```
sedghi, line 171: what does r=0 mean? I see it is setting the delta to 0, is this the radius or the delta for threshold? in either case we better be consistent

packages/tools/examples/labelmapSegmentationDynamicThreshold/index.ts
```diff
@@ -0,0 +1,525 @@
...
+thresholdOptions.set('Dynamic r=1', { isDynamic: true, delta: 1 });
+thresholdOptions.set('Dynamic r=3', { isDynamic: true, delta: 3 });
+thresholdOptions.set('Dynamic r=5', { isDynamic: true, delta: 5 });
+thresholdOptions.set('Same Threshold', { isDynamic: false, delta: 5 });
```
sedghi, line 175: same threshold is confusing

packages/tools/src/stateManagement/segmentation/config/segmentationColor.ts
```diff
@@ -96,7 +96,18 @@ function getColorForSegmentIndex(
 
   // get colorLUT
   const colorLUT = SegmentationState.getColorLUT(colorLUTIndex);
-  return colorLUT[segmentIndex];
+  let colorValue = colorLUT[segmentIndex];
+  if (!colorValue) {
+    if (
+      segmentIndex < 0 ||
+      segmentIndex > 65535 ||
+      Math.floor(segmentIndex) !== segmentIndex
```
sedghi, line 104: what is Math.floor of an integer (segment index)? can you please add a comment what is happening in the code?

packages/tools/src/types/LabelmapToolOperationData.ts
```diff
@@ -6,12 +8,24 @@ import {
...
-  constraintFn: (pointIJK: number) => boolean;
+  // constraintFn: (pointIJK: number) => boolean;
   segmentationRepresentationUID: string;
+  points: Types.Point3[];
+  /**
+   * preview is used for sharing preview data between views/interactions with
+   * a tool, and needs to be maintained by the tool side in order to be able
+   * to accept/reject/update the preview information.
+   */
+  preview: any;
```
sedghi, line 27: is this `previewData` then?

packages/tools/src/utilities/segmentation/brushPreviewForToolGroup.ts
```diff
@@ -0,0 +1,67 @@
...
+  const toolGroup = getToolGroup(toolGroupId);
+
+  if (toolGroup === undefined) {
+    return;
+  }
+
+  const brushBasedToolInstances = getBrushToolInstances(toolGroupId);
+  const configuration = {
+    ...otherArgs,
+    getPreviewSegmentIndex,
```
sedghi, line 21: I don't see anywhere getPreviewSegmentIndex is used

packages/tools/src/utilities/segmentation/brushPreviewForToolGroup.ts
```diff
@@ -0,0 +1,67 @@
...
+  // Use helper to get array of viewportIds, or we just end up doing this mapping
+  // ourselves here.
+  const viewportIds = toolGroup.getViewportIds();
+
+  const renderingEngine = getRenderingEngine(renderingEngineId);
+
+  triggerAnnotationRenderForViewportIds(renderingEngine, viewportIds);
+}
+
+export function getBrushThresholdForToolGroup(toolGroupId: string) {
```
sedghi, line 46: incorrect name

packages/tools/examples/labelmapSegmentationDynamicThreshold/index.ts
```diff
@@ -0,0 +1,525 @@
...
+        modifierKey: KeyboardBindings.Alt,
+      },
+      {
+        numTouchPoints: 1,
+        modifierKey: KeyboardBindings.Meta,
+      },
+    ],
+  });
+
+  const cancelFn = {
```
sedghi, line 357: rename cancelFn

packages/tools/examples/labelmapSegmentationDynamicThreshold/index.ts
```diff
@@ -0,0 +1,525 @@
...
+    fn: (evt) => {
+      const { element, key } = evt.detail;
+      const toolGroup = ToolGroupManager.getToolGroup(toolGroupId);
+      const activeName = toolGroup.getActivePrimaryMouseButtonTool();
+      const brush = toolGroup.getToolInstance(activeName);
+      if (key === 'Escape') {
+        brush.rejectPreview?.(element);
```
sedghi, line 364: This is not the cornerstone way of doing things. We should not allow calling methods directly on tool instances. One approach you can take is similar to how the AdvancedMagnify or SplineTool works. They create their own bindings in their constructor for the `action` methods, and then our listeners would invoke the corresponding method. This way, you won't need any of these.

See AdvancedMagnifty

```js
actions: {
  showZoomFactorsList: {
    method: 'showZoomFactorsList',
    bindings: [
      {
        mouseButton: MouseBindings.Secondary,
        modifierKey: KeyboardBindings.Shift,
      },
    ],
  },
},
```

which we handle internally inside `getToolsWithActionsForMouseEvent`

packages/tools/src/tools/segmentation/strategies/eraseSphere.ts
```diff
@@ -1,21 +1,16 @@
 import type { Types } from '@cornerstonejs/core';
 
-import { fillInsideSphere } from './fillSphere';
-import { LabelmapToolOperationData } from '../../../types';
+import type { OperationData } from './BrushStrategy';
```
sedghi, line 3: remove the two unused imports

packages/tools/src/tools/segmentation/strategies/eraseCircle.ts
```diff
@@ -1,22 +1,13 @@
...
-  // Take the arguments and set the segmentIndex to 0,
-  // Then use existing fillInsideCircle functionality.
-  const eraseOperationData = {
-    ...operationData,
-    segmentIndex: 0,
-  };
-
-  fillInsideCircle(enabledElement, eraseOperationData);
-}
+export { ERASE_CIRCLE_STRATEGY, eraseInsideCircle };
```
sedghi, line 13: In this file and also in `eraseSphere`, I noticed that you export the strategy instances (`ERASE_CIRCLE_STRATEGY` and the other one for sphere), but they are not used anywhere in the code. I was initially wondering about the difference between these two, and I discovered that `ERASE_CIRCLE_STRATEGY` is not used anywhere. Can we simply export the actual one?

packages/tools/src/tools/segmentation/strategies/BrushStrategy.ts
```diff
@@ -0,0 +1,289 @@
+import type { Types } from '@cornerstonejs/core';
+import { utilities, cache, utilities as csUtils } from '@cornerstonejs/core';
```
sedghi, line 2: two utilities import

packages/tools/src/tools/segmentation/strategies/BrushStrategy.ts
```diff
@@ -0,0 +1,289 @@
...
+    enabledElement: Types.IEnabledElement,
+    operationData: OperationData
+  ) => unknown;
+
+  public setValue: (data, operationData: InitializedOperationData) => void;
+
+  public createIsInThreshold: (
+    enabled,
+    operationData: InitializedOperationData
+  ) => any;
```
sedghi, line 257: please add a comment above these that these will get overwritten from the outside by each strategy separately

packages/tools/src/tools/segmentation/strategies/BrushStrategy.ts
```diff
@@ -0,0 +1,289 @@
...
+    enabledElement: Types.IEnabledElement,
+    operationData: OperationData
+  ) => void;
+
+  public acceptPreview: (
+    enabledElement: Types.IEnabledElement,
+    operationData: OperationData
+  ) => void;
+
+  public preview: (
```
sedghi, line 247: a method called preview (noun) does not ring a bell to me, accept/reject preview is quite clear, but `preview` seems unusuall, maybe rename it

packages/tools/src/tools/segmentation/strategies/utils/initializeTracking.ts
```diff
@@ -0,0 +1,27 @@
+import type { InitializedOperationData } from '../BrushStrategy';
+
+/**
+ * Sets up tracking for use by preview and other services.
+ * The tracking is either the preview object, or the previewVoxelValue.
+ * Re-using this existing preview means that the same segment colour and already
+ * tracked values can be extended.
+ */
+export default {
```
sedghi, line 9: let's rename this file to something else since we have tracking in OHIF for measurements and it would be confusing. Another question is that why you can't merge this one into the `createInitialized` of the `initializePreview` ?

packages/tools/src/tools/segmentation/strategies/utils/initializeRegionFill.ts
```diff
@@ -0,0 +1,52 @@
...
+      segmentationVoxelValue,
+      previewVoxelValue,
+      imageVoxelValue,
+      brushStrategy,
+      centerIJK,
+    } = operationData;
+    const isWithinThreshold = brushStrategy.createIsInThreshold?.(
+      enabledElement,
+      operationData
+    );
```
sedghi, line 27: I think there is a bug here, no? If not, when `createIsInThreshold` is defined, the `isWithinThreshold` becomes undefined. Can you live with that?

packages/tools/src/tools/segmentation/strategies/utils/initializeDetermineSegmentIndex.ts
```diff
@@ -0,0 +1,60 @@
...
+      hasPreviewIndex ||= value === previewSegmentIndex;
+    };
+
+    pointInShapeCallback(
+      segmentationImageData as unknown,
+      imageVoxelValue.isInObject,
+      callback,
+      segmentationVoxelValue.boundsIJK
+    );
+
```
sedghi, line 49: I'm surprised that this function is calling pointInShapeCallback. I feel that it is unnecessary or we should not add another loop over the whole bounds just to determine the segment index. Why are we looping here?

packages/tools/src/tools/segmentation/strategies/utils/initializeRegionFill.ts
```diff
@@ -0,0 +1,52 @@
...
+          if (segmentsLocked.includes(value)) {
+            return;
+          }
+          if (!isWithinThreshold(index)) {
+            return;
+          }
+          setValue(data, operationData);
+        }
+      : (data) => setValue(data, operationData);
+
```
sedghi, line 42: I think the same lock logic applys to when isWithinThreshold is undefined, so it is more like


```js
const callback = (data) => {
  const { value, index } = data;
  if (!segmentsLocked.incluede(value) && (!isWithinThreshold || isWithinThreshold(index))) {
    ....


  }
  ```

File: packages/tools/src/tools/segmentation/strategies/BrushStrategy.ts [hunk 37]
```diff
@@ -0,0 +1,289 @@
...
+  /**
+   * Provide some default initializers for various situations, mostly for
+   * external use to allow defining new brushes
+   */
+  public static initializers = {
+    initializePreview,
+    initializeSetValue,
+    initializeThreshold,
+    initializeRegionFill,
+  };
```
sedghi, line 82: Do you really need this? I don't see you default to these anywhere

packages/tools/src/tools/segmentation/strategies/fillSphere.ts
```diff
@@ -1,136 +1,142 @@
...
-    sliceArray = Array.from(
-      { length: maxSlice - minSlice + 1 },
-      (v, k) => k + minSlice
-    );
-    triggerSegmentationDataModified(operationData.volumeId, sliceArray);
-  }
-}
+
+    imageVoxelValue.isInObject = (pointLPS /*, pointIJK */) =>
+      pointInEllipse(ellipseObj, pointLPS);
```
sedghi, line 100: I'm pretty sure you are reducing the performance here. Previously, we had the `pointInSurroundingSphereCallback`, which was quite efficient in taking into account the camera, etc. Why don't I see that anymore in your changes?

I tried the local demo versus the deployed demo and noticed the performance issue in the sphere brush as well.

packages/tools/src/tools/base/BaseTool.ts
```diff
@@ -83,7 +84,38 @@ abstract class BaseTool implements IBaseTool {
...
+   */
+  public applyActiveStrategyCallback(
+    enabledElement: Types.IEnabledElement,
+    operationData: unknown,
+    callbackType:
+      | 'initDown'
+      | 'completeUp'
+      | 'rejectPreview'
+      | 'acceptPreview'
+      | 'preview'
```
sedghi, line 111: Having an accept/reject preview here in the base tool is quite unusual. Maybe these are only specific to the Brush Tool?

packages/tools/src/tools/segmentation/BrushTool.ts
```diff
@@ -181,49 +204,121 @@ class BrushTool extends BaseTool {
...
       this.updateCursor(evt);
+
+      const { currentPoints, element } = evt.detail;
+      const { canvas } = currentPoints;
+
+      const { preview, startPoint, timer, timerStart, isDrag } =
+        this._previewData;
+      const delta = vec2.distance(canvas, startPoint);
+      const time = Date.now() - timerStart;
+      if (delta > 8 || (time > 250 && delta > 1)) {
```
sedghi, line 286: should be configurable time

packages/tools/src/tools/segmentation/BrushTool.ts
```diff
@@ -529,6 +696,17 @@ class BrushTool extends BaseTool {
...
+    drawCircleSvg(
+      svgDrawingHelper,
+      annotationUID,
+      circleUID1,
+      center as Types.Point2,
+      2,
+      {
+        color,
+      }
+    );
```
sedghi, line 709: I think this whole thing should be configurable. Someone might say, "I liked the previous one."

packages/core/src/utilities/VoxelValue.ts
```diff
@@ -0,0 +1,235 @@
...
+  public map: Map<number, T>;
+  public sourceVoxelValue: VoxelValue<T>;
+  public isInObject: (pointIPS, pointIJK) => boolean;
+
+  points: Set<number>;
+  dimensions: Point3;
+  width: number;
+  frameSize: number;
+  _get: (index: number) => T;
+  _set: (index: number, v: T) => boolean | void;
```
sedghi, line 25: please add docs on top of each one

packages/core/src/utilities/VoxelValue.ts
```diff
@@ -0,0 +1,235 @@
+import type { BoundsIJK, Point3, VolumeScalarData } from '../types';
+
+/**
+ * This is a simple, standard interface to values associated with a voxel.
+ */
```
sedghi, line 5: please add documentation about how the `new` it, what is get and what is set etc.

packages/core/src/utilities/VoxelValue.ts
```diff
@@ -0,0 +1,235 @@
...
+export default class VoxelValue<T> {
```
sedghi, line 6: I feel like this is a VoxelManager rather than VoxelValue, since if you take a look at the way we use it (e.g., getStrategyData) it is like

```js
return {
    segmentationImageData,
    segmentationScalarData,
    segmentationVoxelValue: VoxelValue.volumeVoxelValue(
      dimensions,
      segmentationScalarData
    ),
    imageScalarData,
    imageVoxelValue: VoxelValue.volumeVoxelValue(dimensions, imageScalarData),
  };
  ```

So here `segmentationVoxelValue` sounds like it is just a value (one scalar), but it is not, it is set of utilities 

what do you think

File: packages/core/src/utilities/VoxelValue.ts [hunk 45]
```diff
@@ -0,0 +1,235 @@
...
+    _set?: (index: number, v: T) => boolean | void
+  ) {
+    this.dimensions = dimensions;
+    this.width = dimensions[0];
+    this.frameSize = this.width * dimensions[1];
+    this._get = _get;
+    this._set = _set;
+  }
+
+  public getIJK = (i, j, k) => {
```
sedghi, line 39: Lots of these methods should be renamed for clarity. Looking at the name it might mean `getIJKValueFromFlatIndex` which is fair give the name `getIJK`, so how about `getValueAtIJK`

packages/core/src/utilities/VoxelValue.ts
```diff
@@ -0,0 +1,235 @@
...
+    const index = i + j * this.width + k * this.frameSize;
+    return this._get(index);
+  };
+
+  public setIJK = (i: number, j: number, k: number, v) => {
```
sedghi, line 44: how about `setValueAtIJK([i,j,k], value)`

packages/core/src/utilities/VoxelValue.ts
```diff
@@ -0,0 +1,235 @@
...
+
+  public addPoint(point: Point3 | number) {
+    const index = Array.isArray(point)
+      ? point[0] + this.width * point[1] + this.frameSize * point[2]
+      : point;
+    if (!this.points) {
+      this.points = new Set<number>();
+    }
+    this.points.add(index);
+  }
```
sedghi, line 60: it is not quite evident what is this, since I expected this class to handle images, but seems like it can handle points too, add docs please

sedghi: Maybe all `this.points` stuff should be a separate manager?

packages/core/src/utilities/VoxelValue.ts
```diff
@@ -0,0 +1,235 @@
...
+  public static mapVoxelValue<T>(dimension: Point3): VoxelValue<T> {
+    const map = new Map<number, T>();
+    const voxelValue = new VoxelValue(
+      dimension,
+      map.get.bind(map),
+      (index, v) => map.set(index, v) && true
+    );
+    voxelValue.map = map;
+    return voxelValue;
+  }
```
sedghi, line 181: I'm not sure about these two names either but can't think of good ones either. For instance when i was reading the getStrategyData code 

`segmentationVoxelValue: VoxelValue.volumeVoxelValue(
      dimensions,
      segmentationScalarData
    ),`

I absolutely did not expect this `new` logic

sedghi (commented):
Great stuff, regarding renamings

The `createSomething` pattern is not consistently followed. For example, `fill` should be `createFill` or `createAcceptPreview`, etc. However, it might be less confusing to remove the "create" altogether. How about using `initialize` instead of `createInitialized`, and `isWithinThreshold` instead of `createIsInThreshold`?

Additionally, I suggest using `onBrushStart` and `onBrushEnd`, or `onBrushComplete` instead of `initDown` and `completeUp`.

Many files in the segmentation/strategies/utils directory are named `initializeSomething`, even though they don't actually act as initializers. For example, `initializeRegionFill` only has the "fill" functionality. Perhaps it would be better to name it `regionFill` or `setValue` instead. I believe that if we reserve the "initialize" (or "configure") keyword for those functions that truly initialize or configure, it would be easier to understand.

packages/tools/examples/labelmapSegmentationDynamicThreshold/index.ts
wayfarer3130: That doesn't help - those are mouse bindings you are looking at, with a keyboard modifier, not direct actions which are invokeable externally.  Ideally I'd like the preview accept/reject functions to be external, but I don't know how to set that up, and at this point I'm not willing to create a full framework for keyboard handling to invoke tool functions.
Can we leave this for now, and improve on it during the undo phase?  We will need a way to undo/redo, and also to handle preview capabilities, so if we setup listeners that can be fired externally and the tool responds to when active, then that would be optimal.

sedghi: There is no framework, You need these 60 lines of code for keyboard, right now it is `getToolsWithActionsForMouseEvent` you need an exact copy paste to `getToolsWithActionsForKeyboardEvents` which the `keyDown` should call. 

Calling methods on tool instances is just a no no IMO

Alternative would be a utility that deals with the state

packages/tools/src/tools/segmentation/strategies/BrushStrategy.ts
```diff
@@ -0,0 +1,289 @@
...
+import initializeThreshold from './utils/initializeThreshold';
+import { getStrategyData } from './utils/getStrategyData';
+import { isVolumeSegmentation } from './utils/stackVolumeCheck';
+import type {
+  LabelmapToolOperationDataStack,
+  LabelmapToolOperationDataVolume,
+} from '../../../types/LabelmapToolOperationData';
+
+const { VoxelValue } = utilities;
+export type OperationData =
```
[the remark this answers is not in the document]
sedghi: So put it there please

packages/tools/src/tools/segmentation/strategies/utils/initializeRegionFill.ts
wayfarer3130: The isWithinThreshold is created by the createIsInThreshold call - the problem is you can't create it too early because it binds some of the values in the strategy function which aren't defined initially, and I don't want to have to read them dynamically every time because the call gets made tens of thousands of times, so optimizing it is quite important.

wayfarer3130: I'm wondering about optimizing setValue the same way actually, since that one also gets called a gazillion times.

packages/tools/src/tools/segmentation/strategies/BrushStrategy.ts
wayfarer3130: Yes, you really need it if you are ever going to implement a strategy externally.  Otherwise, no you don't need it.

packages/tools/src/tools/segmentation/strategies/fillSphere.ts
wayfarer3130: That is a generic optimization I plan on making for both sphere and circle annotations as part of the 3.x implementation.  What I plan on is a generic set of calls to determine bounds along a given dimension, so that if you provide bounds for zero, one or two of i,j,k, you can get the bounds for the remaining element.  Then, allow one to set it for everything outside, and everything inside.
Probably will end up with 3 versions of the iteration in VoxelValue, each optimized for different situations.

packages/tools/src/tools/base/BaseTool.ts
wayfarer3130: Previews aren't specific to the brush tool - we happen to only use it there, but in general it is a useful concept.

packages/core/src/utilities/VoxelValue.ts
wayfarer3130: What about getAt and setAt for naming?  Then:
getAtIndex
getAtIJK
getAt(ijk: point3)
or the last once could even just be get, but getAt as the prefix is ok.

packages/tools/src/tools/segmentation/strategies/BrushStrategy.ts
sedghi: if there is a strategy externally, then can't we just export those and anyone explicitly use them?

packages/tools/src/tools/base/BaseTool.ts
sedghi: what would be the preview for annotation?

packages/core/src/utilities/VoxelValue.ts
sedghi: I like it

packages/tools/src/stateManagement/segmentation/config/segmentationColor.ts
sedghi: remove entrierly

packages/core/src/utilities/VoxelValue.ts
sedghi: getAtIJK
getAtIJKPoint or getAtIJKArray

packages/tools/src/enums/StrategyCallbacks.ts
```diff
@@ -0,0 +1,52 @@
...
+ * Defines the names of the strategy callbacks used for performing enhanced
+ * strategy operations.
+ */
+enum StrategyCallbacks {
+  /**
+   * startStrategy is called at the start of a strategy, typically on mouse down
+   * Note this is separate from preview and the endings for preview, which could
+   * be called alternatively, but this may be nested within a preview.
+   */
+  onInteractionStart = 'onInteractionStart',
```
sedghi, line 11: All other enums are like OnInteractionStart, start with capital letter

packages/tools/src/eventDispatchers/shared/getActiveToolForKeyboardEvent.ts
```diff
@@ -1,5 +1,5 @@
 import { ToolGroupManager } from '../../store';
-import { ToolModes, MouseBindings } from '../../enums';
+import { ToolModes, MouseBindings, KeyboardBindings } from '../../enums';
```
sedghi, line 2: remove unused

packages/tools/src/tools/segmentation/BrushTool.ts
```diff
@@ -76,41 +106,70 @@ class BrushTool extends BaseTool {
...
+          enabled: false,
+          previewColors: {},
+          // The time before showing a preview
+          previewTimeMs: 250,
+          // The distance to move to show a preview before preview time expired
+          previewMoveDistance: 8,
+          // The distance to drag before being considered a drag rather than click
+          dragMoveDistance: 4,
+          // The time to consider a mouse click a drag when moved less than dragMoveDistance
+          dragTimeMs: 500,
```
sedghi, line 127: seems like you still need this?

packages/tools/src/tools/segmentation/strategies/fillSphere.ts
```diff
@@ -1,146 +1,141 @@
...
+import BrushStrategy from './BrushStrategy';
+import type { InitializedOperationData, Composition } from './BrushStrategy';
+import type { CanvasCoordinates } from '../../../types';
+import compositions from './compositions';
+import StrategyCallbacks from '../../../enums/StrategyCallbacks';
+
+const { transformWorldToIndex } = csUtils;
+
+const sphereComposition = {
+  [StrategyCallbacks.initialize]: function intializeSphere(
```
sedghi, line 19: typo in intialize

sedghi (commented):
1. So seems like the sphere Brush is worse than I thought, since we changed it to use pointInShapeCallback. The fact that we do the last phase sphere dynamic threshold way later means we are adding a regression for sphere brush, i think we should fix that before merge


2. For some reason i can't remove (erase) from the preview, i was able previously

packages/tools/src/tools/segmentation/BrushTool.ts
wayfarer3130: Yes, but it is only used for preview.  Otherwise you can end up with a different selection than you intended.  I can demonstrate if we set the preview time/drag times to fairly large values so it is easy to show how you get the wrong final selection.

packages/tools/src/utilities/math/ellipse/pointInEllipse.ts
```diff
@@ -36,7 +53,16 @@ export default function pointInEllipse(
...
   }
+  return inverts;
+};
 
+/**
+ * If you call the pointInEllipse.precalculateInverts first, then you
+ * can call precalculated directionly instead of having the extra time for
+ * the if conditions.
+ */
+pointInEllipse.precalculated = (ellipse, pointLPS, inverts) => {
```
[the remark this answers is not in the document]
wayfarer3130: Nothing is wrong with that, but I wanted to optimize the time for that version since you were finding it slow.  This is probably the wrong optimization - the real one is to find the boundaries and iterate over the others in a simple loop.

sedghi (approved):
Great job on the PR, really enjoyed reviewing it

PR #922, opened 2023-12-01
feat(vtk.js): Upgrade version and add Segment Select Tool

sedghi:
## Bump Vtk.js version

Bumping vtk.js to incorporate additions 
- https://github.com/Kitware/vtk-js/pull/2960
- https://github.com/Kitware/vtk-js/pull/2960
- https://github.com/Kitware/vtk-js/pull/2914

## New Segment Select Tool on Hover

Adding new Segment Select Tool to be able to change active segment index only by hovering over segments

https://github.com/cornerstonejs/cornerstone3D/assets/7490180/eebae093-9e48-4fad-af65-913fba05986f

## New Utilities 
New createSegmentation utilities in imageLoader and volume loader

`createAndCacheDerivedSegmentationImage`

`createAndCacheDerivedSegmentationVolume`

I understand that this may not be ideal, but currently, we are limited to using Uint8Array for segmentations. Unfortunately, this is not widely known in the community yet. Therefore, until we can address this issue, I strongly advocate for using this method to ensure the use of Uint8Array.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  


utils/test/testUtils.js
```diff
@@ -107,6 +107,43 @@ function createMockEllipsoidStackSegmentation({
   }
 }
 
+/**
+ * Adds two concentric circles to each axial slice of the demo segmentation.
+ */
+function createMockEllipsoidVolumeSegmentation({
```
wayfarer3130, line 113: Could we import this from a file please?  I like the fact that you have added it here, but I think I'd like to see the imports come from individual files.

packages/tools/src/tools/displayTools/Labelmap/labelmapDisplay.ts
```diff
@@ -314,15 +314,32 @@ function _setLabelmapColorAndOpacity(
...
+
+  const { activeSegmentIndex } = SegmentationState.getSegmentation(
+    segmentationRepresentation.segmentationId
+  );
+
+  // create an array that contains all the segment indices and for the active
+  // segment index, use the activeSegmentOutlineWidth, otherwise use the
+  // outlineWidth
+  // Pre-allocate the array with the required size to avoid dynamic resizing.
+  const outlineWidths = new Array(numColors - 1);
```
wayfarer3130, line 330: Would really prefer to just have this map 1-1 with the colours - that means you get an empty spot at the beginning, but I think that is preferable to having to map an off by one, and all the consequent issues you get with that.

packages/tools/src/tools/displayTools/Labelmap/labelmapConfig.ts
```diff
@@ -4,6 +4,7 @@ const defaultLabelmapConfig: LabelmapConfig = {
   renderOutline: true,
   outlineWidthActive: 3,
   outlineWidthInactive: 2,
+  activeSegmentOutlineWidth: 4,
```
wayfarer3130, line 7: Should these be percentages, such that this is the displayed pixel width in actual pixels on screen?

packages/tools/examples/segmentSelect/index.ts
```diff
@@ -0,0 +1,303 @@
...
+
+  await stackViewport.setStack(imageIdsArray, 0);
+
+  createMockEllipsoidStackSegmentation({
+    imageIds: imageIdsArray,
+    segmentationImageIds,
+    cornerstone,
+  });
+
+  segmentation.addSegmentations([
```
wayfarer3130, line 281: We are starting to get an awful lot of this boilerplate code that really is probably shared.  What about having a few more helper methods to setup some of these sorts of things for the examples?

packages/core/src/loaders/volumeLoader.ts
```diff
@@ -561,3 +561,24 @@ export function registerUnknownVolumeLoader(
...
+/**
+ * Creates and caches a derived segmentation volume based on a referenced volume.
+ * This is basically a utility method since for the segmentations we have to specify
+ * Uint8Array as the targetBuffer type for now until we support other types.
+ *
+ * @param referencedVolumeId - The ID of the referenced volume.
+ * @param options - The options for creating the derived volume.
+ * @returns A promise that resolves to the created derived segmentation volume.
+ */
+export async function createAndCacheDerivedSegmentationVolume(
```
wayfarer3130, line 574: Really think this is a better name than just the createAndCacheDerivedVolume, as we can well get other types of derivations.

wayfarer3130:
Just a couple of small changes, plus the test fixes, but generally I like the changes here.
One thing that might be worth considering is exactly where the border is - you will get a cleaner border if the center of the border is that the center of edge position - so that a width of 2 give you:
1 full pixel at 100% right on the border
2 pixels with 50% border colour beside that.

That will give you some anti-aliasing on the curves and lines, but I think that should be specified - or perhaps you can't specify it inside VTK?

packages/tools/src/tools/displayTools/Labelmap/labelmapConfig.ts
sedghi: the unit is pixels, I guess you are right, but It needs to change everything maybe for 2.x

packages/tools/examples/segmentSelect/index.ts
sedghi: I think it is fine for examples, honestly. It is a lot of work. Examples don't hurt to be explicit about the APIs. The mock was a utility rather than showing how our API works.

sedghi: I refactored more mocking data instead

wayfarer3130 (approved):

robustNan:
image
image

This is a great upgrade, the previous 27.3.1 was killing me and sometimes reported errors when getting contours via ImageMarchingSquares! Thanks a lot!

PR #927, opened 2023-12-04
feat(ultrasound regions): Add new US specific tool and augment length and probe tool to better support US

sedghi:
### Context

We will add Ultrasound enhanced regions support for annotations with this PR. Initially two most common usage are added. 

`Enhanced Length Tool`

https://github.com/cornerstonejs/cornerstone3D/assets/7490180/b449420c-cd03-4c00-a121-ed034fd5c2c7

`Enhanced Probe Tool`

https://github.com/cornerstonejs/cornerstone3D/assets/7490180/8a9cd466-872a-4b31-beb5-d042398cbf9b

`New Ultrasound Directional Tool`

https://github.com/cornerstonejs/cornerstone3D/assets/7490180/6d5c3d84-881f-46ba-b509-e3c43f3ac1f2

### Testing

yarn run example ultrasoundenhancedregion

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  


packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -442,7 +443,11 @@ class StackViewport extends Viewport implements IStackViewport, IImagesLoader {
       metadata: { Modality: this.modality },
       scaling: this.scaling,
       hasPixelSpacing: this.hasPixelSpacing,
-      calibration: this.calibration,
+      calibration: Object.assign(
```
wayfarer3130, line 446: Can you use the ... notation for consistency - or change the other examples to use Object.assign, whichever you prefer.

packages/core/src/enums/MetadataModules.ts
```diff
@@ -9,12 +9,17 @@
 enum MetadataModules {
   CINE = 'cineModule',
   IMAGE_URL = 'imageUrlModule',
+  SOP_COMMON = 'sopCommonModule',
```
wayfarer3130, line 12: It seems that some enums are UpperCamelCase and some are UPPER_UNDERSCORE naming.  Which one do we want to use consistently?   Could you fix this one to UpperCamelCase, since I think that is what we are trying to use?

packages/dicomImageLoader/src/imageLoader/wadors/metaData/USHelpers.ts
```diff
@@ -0,0 +1,67 @@
+import getNumberValues from './getNumberValues';
+import getSequenceItems from './getSequenceItems';
+
+function getUSEnhancedRegions(metadata) {
```
wayfarer3130, line 4: Docs please

wayfarer3130: Also, this would be a great place to have a jest unit test, as there is just enough complexity to warrant a unit test, but not quite enough for have a full set of integration tests on this.

packages/dicomImageLoader/src/imageLoader/wadors/metaData/USHelpers.ts
```diff
@@ -0,0 +1,67 @@
...
+      physicalDeltaY,
+      physicalUnitXDirection,
+      physicalUnitYDirection,
+      referencePhysicalPixelValueY,
+      referencePhysicalPixelValueX,
+      regionSpatialFormat,
+      regionDataType,
+      regionFlags,
+      transducerFrequency,
+      pixelSpacing: [physicalDeltaX * 10, physicalDeltaY * 10],
```
wayfarer3130, line 60: This is not a good assumption here - the multiplier should be based on if this is centimeters, and you should probably not return anything here if it isn't pixels.

packages/dicomImageLoader/src/imageLoader/wadors/metaData/getSequenceItems.ts
```diff
@@ -0,0 +1,19 @@
+import { WADORSMetaDataElement } from 'dicomImageLoader/src/types';
+
+function getSequenceItems(element: any): WADORSMetaDataElement[] {
+  if (!element) {
+    return;
+  }
+  // Value is not present if the attribute has a zero length value
+  if (!element.Value) {
```
wayfarer3130, line 8: Use !element.Value?.length since some DICOMweb systems are not quite compliant, and that is a more general test.

packages/core/src/enums/MetadataModules.ts
sedghi: As I see all the enums in the cores/src/enums are actually UPPER_UNDERSCORE, so I don't see any inconsistencies

packages/tools/src/utilities/getCalibratedUnits.ts
```diff
@@ -56,7 +75,154 @@ const getCalibratedAreaUnits = (handles, image): string => {
...
+    // Todo: add support for other scenarios
+  }
+
+  if (calibration.sequenceOfUltrasoundRegions) {
+    // for Probe tool
+    const supportedRegionsMetadata =
+      calibration.sequenceOfUltrasoundRegions.filter(
+        (region) =>
+          SUPPORTED_REGION_DATA_TYPES.includes(region.regionDataType) &&
+          SUPPORTED_PROBE_VARIANT.includes(
```
wayfarer3130, line 178: There just aren't that many units to test - is there any reason to just not include all of them, separately in the x and y?  It isn't as though the behaviour changes other than labelling.

packages/dicomImageLoader/src/imageLoader/wadors/metaData/USHelpers.ts
sedghi: removed it

packages/tools/src/tools/annotation/UltrasoundDirectionalTool.ts
```diff
@@ -0,0 +1,894 @@
...
+    this._throttledCalculateCachedStats = throttle(
+      this._calculateCachedStats,
+      100,
+      { trailing: true }
+    );
+  }
+
+  /**
+   * Based on the current position of the mouse and the current imageId to create
+   * a Length Annotation and stores it in the annotationManager
```
wayfarer3130, line 94: Update the docs here please.

packages/tools/src/tools/annotation/UltrasoundDirectionalTool.ts
```diff
@@ -0,0 +1,894 @@
...
+        // so we just report the px distance
+        value = distanceToPoint(worldPos1, worldPos2);
+      } else {
+        const canvasPoint1 = enabledElement.viewport.worldToCanvas(worldPos1);
+        const canvasPoint2 = enabledElement.viewport.worldToCanvas(worldPos2);
+
+        const canvasDeltaY = canvasPoint2[1] - canvasPoint1[1];
+        const canvasDeltaX = canvasPoint2[0] - canvasPoint1[0];
+
+        const orientation =
```
wayfarer3130, line 838: Can you make this a boolean so that you only have the comparison once, eg
const orientationHorizontal = ....
It is a bit ugly to have the values all over the place for the comparison.

packages/tools/src/tools/annotation/UltrasoundDirectionalTool.ts
```diff
@@ -0,0 +1,894 @@
...
+        const yValues = [values1[1], values2[1]];
+
+        if (orientation === 'HORIZONTAL') {
+          value = Math.abs(xValues[1] - xValues[0]);
+          unit = units1[0];
+        } else if (orientation === 'VERTICAL') {
+          value = Math.abs(yValues[1] - yValues[0]);
+          unit = units1[1];
+        } else {
+          throw new Error('Invalid orientation');
```
wayfarer3130, line 857: If it's a boolean, there is no need for an error condition here.

packages/tools/src/tools/annotation/UltrasoundDirectionalTool.ts
```diff
@@ -0,0 +1,894 @@
...
+            : 'VERTICAL';
+
+        // then for the third point we need to go from first point towards
+        // the second point (it can be left or right in the horizontal orientation)
+        // or up or down in the vertical orientation, and only add
+        // the delta y to the x or y coordinate of the first point
+        const xValues = [values1[0], values2[0]];
+        const yValues = [values1[1], values2[1]];
+
+        if (orientation === 'HORIZONTAL') {
```
wayfarer3130, line 850: Could you just add both values and both units here to a list, and then choose which one(s) to display later?

packages/tools/src/tools/annotation/UltrasoundDirectionalTool.ts
```diff
@@ -0,0 +1,894 @@
...
+          unit = units1[0];
+        } else if (orientation === 'VERTICAL') {
+          value = Math.abs(yValues[1] - yValues[0]);
+          unit = units1[1];
+        } else {
+          throw new Error('Invalid orientation');
+        }
+      }
+
+      cachedStats[targetId] = {
```
wayfarer3130, line 861: If the cached stats has a set of indices for the values to display, then you can easily add the ability to display both - so you have two orientations and three sets of values to display:
horizontal/vertical orientation
both values, eg [0,1]
horizontal value [1]
vertical value [0]
That can be in the configuration for the tool as a whole, so that you can configure the same tool to always display both, regardless of orientation, or only display one or the other regardless of orientation, but put the label in the same spot.

wayfarer3130 (changes requested):
Would like to see the changes so that more units are supported, and also to avoid showing the underlying/base units when the calibration type is unknown.

packages/tools/src/tools/annotation/UltrasoundDirectionalTool.ts
sedghi: Ok done

bun.lockb
wayfarer3130, line 1: What is this?

packages/tools/src/tools/annotation/UltrasoundDirectionalTool.ts
```diff
@@ -0,0 +1,916 @@
...
+
+function defaultGetTextLines(data, targetId, configuration): string[] {
+  const cachedStats = data.cachedStats[targetId];
+  const { xValues, yValues, units, isUnitless, isHorizontal } = cachedStats;
+
+  if (isUnitless) {
+    return [`${roundNumber(xValues[0])} px`];
+  }
+
+  if (configuration.displayBothAxesDistances) {
```
wayfarer3130, line 897: Thank you - I know this is a bit of a pain, but sometimes it really helps to have both.

packages/tools/src/utilities/getCalibratedUnits.ts
```diff
@@ -56,7 +75,154 @@ const getCalibratedAreaUnits = (handles, image): string => {
...
+          SUPPORTED_LENGTH_VARIANT.includes(
+            `${region.physicalUnitXDirection},${region.physicalUnitYDirection}`
+          )
+      );
+
+    if (!supportedRegionsMetadata.length) {
+      return { units, areaUnits, scale };
+    }
+
+    const region = supportedRegionsMetadata.find(
```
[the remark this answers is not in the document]
wayfarer3130: The problem is that you return the base/underlying units within that region so if you have an unsupported region type, you use the base units, which could result in displaying the wrong values to the user.

packages/tools/src/utilities/getCalibratedUnits.ts
```diff
@@ -56,7 +75,154 @@ const getCalibratedAreaUnits = (handles, image): string => {
...
+        imageIndex1[1] >= region.regionLocationMinY0 &&
+        imageIndex1[1] <= region.regionLocationMaxY1 &&
+        imageIndex2[0] >= region.regionLocationMinX0 &&
+        imageIndex2[0] <= region.regionLocationMaxX1 &&
+        imageIndex2[1] >= region.regionLocationMinY0 &&
+        imageIndex2[1] <= region.regionLocationMaxY1
+    );
+
+    if (region) {
+      scale = 1 / (region.physicalDeltaX * region.physicalDeltaY * 100);
```
wayfarer3130, line 142: If the scaling is not identical between the regions, then you get into some odd situations.  I think we should only allow the non-US specific tools to run if the units are the same between the two of them, and the physical delta is the same.
Otherwise, you could end up showing an invalid value for a measurement, which is a patient safety issue.

wayfarer3130:
For the region measurements, in order to avoid any incorrect values being shown, I think we should:
1. If all points of the measurement are in the same region and the region has the same X,Y units and same X,Y spacing, then use the region unit and spacing for the region as the primary unit/scaling within that area.
2. If all points of the measurement are NOT in an US region, then use the base units (including user specified scaling)
3. Use px for all measurement values otherwise, except the US region specific measurements, which can handle different X,Y units and spacing

I believe this is the only thing safe to do, otherwise we end up displaying invalid values

wayfarer3130 (approved):

packages/tools/src/utilities/getCalibratedUnits.ts
```diff
@@ -151,18 +151,21 @@ const getCalibratedLengthUnitsAndScale = (image, handles) => {
...
-      region.physicalDeltaX,
-      region.physicalDeltaY,
+      physicalDeltaX,
+      physicalDeltaY,
       EPS
     );
 
     if (isSamePhysicalDelta) {
-      scale = 1 / (region.physicalDeltaX * region.physicalDeltaY * 100);
+      scale = 1 / (physicalDeltaX * physicalDeltaY * 100);
```
wayfarer3130, line 168: The scale should also use absolute values here - it doesn't make sense to have it -ve here.

PR #934, opened 2023-12-07
fix(WindowLevelTool): fix window level tool for non-pre-scaled PT images and images with small dynamic range

pwespi:
### Context

In the window level tool, there is a special `getPTScaledNewRange` method implemented which is applied for all PT images, however it should only be used for pre-scaled (SUV) PT images.

Furthermore the default multiplier is too large for images with small dynamic range.  

### Changes & Results

With the changes in this PR `getPTScaledNewRange` is only used for pre-scaled PT images. Furthermore the calculated ratio is used as multiplier also for images with small dynamic range instead of the default multiplier.

before:

https://github.com/cornerstonejs/cornerstone3D/assets/24232962/46e02e2b-8d72-4afe-8feb-147b157f8110

after:

https://github.com/cornerstonejs/cornerstone3D/assets/24232962/aa24299e-3fad-4f3b-bb64-d7091ef466c8

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] "OS: macOS 
- [x] "Node version: 18 
- [x] "Browser: Chrome
  


pwespi:
Thank you for your feedback and for your great work on this project in general! 

> I'm surprised you have a non scaled PT and still the window level is at 1 1 

In PT when the unit is PROPCNTS (proportional to counts) or PROPCPS (proportional to counts/sec), the reconstruction algorithm is free to apply any factor to all pixel values (because of the *proportional* part). Therefore you can end up with and image with window level 1 1 even without scaling to SUV values.

> I think your changes will break the cases where PT is unscaled and window level is high as a result
> 
> Like the one here
> 
> https://viewer-dev.ohif.org/viewer?StudyInstanceUIDs=1.3.6.1.4.1.14519.5.2.1.7009.2403.334240657131972136850343327463

I tested the study you linked and it works as I would expect:

https://github.com/cornerstonejs/cornerstone3D/assets/24232962/962524e1-2f45-42be-880b-7f104eafeb13

Instead of only y-axis window-center it also supports the x-axis window-width operation and the multiplier makes sense. Basically it behaves like other modalities, which is what I would expect.

sedghi:
So by design we limited the XAxis multiplier to not work on PT since that is usually the expected behvaiour in viewers. @salimkanoun Maybe you know more on this?

salimkanoun:
humm i don't think there are absolute rules here.
However the proposed PR is pretty good in my opinion, make sense to use the PET rule only for PET that has been prescaled and then fallback to a common rule

sedghi:
@salimkanoun How about the left and right manipulation? since right now we only allow brightness change with up and down interaction

salimkanoun:
Well in PET the thing to avoid is having a minimal value of the lut under 0, as value are always positive. 
So usually there only the width change witch correspond to changer the upper bound of the VOI settings.

Changing the center could work but with the condition the lower bound of the VOI is always >=0 (and so refuse value < 0).

What I can say is that I saw manufacturer allowing both direction for VOI tools but the add value wasn't too much (and more prone to loose the contrast you want).

So i'm not sure about what is the best...

My personal opinion : It's easier interaction to keep only one direction and let user choose other VOI settings using another GUI with Min and Max of the LUT clearly displayed (like colorbar you showed me).

However this PR only apply to non prescaled PET so it is not a major breaking change. To me if the PET is not prescaled treating it like all other images is ok for me.

packages/tools/src/tools/WindowLevelTool.ts
```diff
@@ -190,12 +191,7 @@ class WindowLevelTool extends BaseTool {
 
     const ratio = imageDynamicRange / DEFAULT_IMAGE_DYNAMIC_RANGE;
 
-    let multiplier = DEFAULT_MULTIPLIER;
-
-    if (ratio > 1) {
-      multiplier = Math.round(ratio);
-    }
-    return multiplier;
+    return ratio > 1 ? Math.round(ratio) : ratio;
```
sedghi, line 194: Apologies for the delay, I've been occupied with various tasks. I believe treating non-scaled PT like any other image is a good idea and I'm completely fine with it. However, I noticed that we are also modifying the `multiplyer` for all images. Is there a specific reason for this change?

pwespi: Thanks for the feedback.

For images with small dynamic range (which result in `ratio < 1` in this code), the previous multiplier was too big. See the before.mov video in my original post which shows that the Window Tool is basically unusable for images with small dynamic range. This is the case for all modalities, not just PT, that's why I changed it here.

With this change the multiplier is always based on the calculated dynamic range, and for `ratio > 1` the behavior is the same as before.

sedghi: Ok makes sense, I pushed some test fixes and will merge after

sedghi (approved):

PR #937, opened 2023-12-11
feat(tool): Bidirectional creation on largest segment slice

wayfarer3130:
### Context

Finds the largest bidirectional on a given segmentation.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  


packages/core/examples/programaticPanZoom/index.ts
```diff
@@ -161,7 +161,7 @@ addButtonToToolbar({
       renderingEngine.getViewport(viewportId)
     );
     viewport.setDisplayArea({
-      imageArea: [1, 1],
+      imageArea: [2, 2],
```
sedghi, line 164: How come the imageArea is 2? I thought this is 0 to 1 always

cspell.json
```diff
@@ -0,0 +1,19 @@
...
+        "kitware",
+        "Labelmap",
+        "segmentations",
+        "toolgroup",
+        "Viewports",
+        "wado"
+    ],
+    "ignoreWords": [],
+    "import": []
+}
```
sedghi, line 19: please add these to your user settings and not workspace, so that we don't get these

packages/tools/examples/segmentBidirectionalTool/index.ts
```diff
@@ -0,0 +1,435 @@
...
+  'Here we demonstrate automatic creation of the largest bidirectional tool which will fit the segment'
+);
+
+const size = '900px';
+const content = document.getElementById('content');
+const viewportGrid = document.createElement('div');
+
+viewportGrid.style.display = 'flex';
+viewportGrid.style.display = 'flex';
+viewportGrid.style.flexDirection = 'row';
```
sedghi, line 75: You don't need a flex it is only one viewport as I see

packages/tools/examples/segmentBidirectionalTool/createBidirectionalToolData.ts
```diff
@@ -0,0 +1,50 @@
...
+
+export default function createBidirectionalToolData(
+  bidirectionalData,
+  viewport
+) {
+  const {
+    handle0,
+    handle1,
+    handle2,
+    handle3,
```
sedghi, line 11: can you get points and don't name handle0, handle1, etc. here?

packages/tools/examples/segmentBidirectionalTool/index.ts
```diff
@@ -0,0 +1,435 @@
...
+      const viewport = renderingEngine.getViewport(viewportId1);
+      const bidirectionalToolData = createBidirectionalToolData(
+        bidirectional,
+        viewport
+      );
+      const { FrameOfReferenceUID, referencedImageId } = bidirectional;
+      annotation.state.addAnnotation(
+        bidirectionalToolData,
+        FrameOfReferenceUID
+      );
```
sedghi, line 217: I understand that we don't want to include this in the library, but I believe that the contourAndFindLargestBidirectional function is actually a helpful utility. Instead of being an example, it could be categorized as a tools/utility/measurement or something internal.

packages/tools/examples/segmentBidirectionalTool/index.ts
```diff
@@ -0,0 +1,435 @@
...
+  initDemo,
+  createImageIdsAndCacheMetaData,
+  setTitleAndDescription,
+  addDropdownToToolbar,
+  addSliderToToolbar,
+  setCtTransferFunctionForVolumeActor,
+  addButtonToToolbar,
+} from '../../../../utils/demo/helpers';
+import createBidirectionalToolData from './createBidirectionalToolData';
+import contourAndFindLargestBidirectional from './findLargestBidirectional';
```
sedghi, line 21: this is odd importing, import A from './B' but I prefer this actually to be inside the tools/utilities/measuremnts or something

packages/tools/examples/segmentBidirectionalTool/findLargestBidirectional.ts
```diff
@@ -0,0 +1,218 @@
...
+
+const vtkUtils = {
+  vtkImageMarchingSquares,
+  vtkDataArray,
+  vtkImageData,
+};
+
+const EPSILON = 1e-2;
+
+export default function contourAndFindLargestBidirectional(segmentations) {
```
sedghi, line 18: please add documentation and types when you move this down into the library

sedghi: If the only problem with not having this in the library is the import of adapters above, we can improve it by refactoring and moving the relevant part from adapters down into tools. We can also add the necessary dependency, which should be fine.

packages/tools/examples/segmentBidirectionalTool/findLargestBidirectional.ts
```diff
@@ -0,0 +1,218 @@
...
+  const { generateContourSetsFromLabelmap } = adaptersRT.Cornerstone3D.RTSS;
+
+  console.time('contour');
```
sedghi, line 21: please remove

packages/tools/examples/segmentBidirectionalTool/findLargestBidirectional.ts
```diff
@@ -0,0 +1,218 @@
...
+      sliceContour,
+      maxBidirectional,
+      isInSegment
+    );
+    const durationMs = Date.now() - startTimeMs;
+    if (!bidirectional) {
+      if (durationMs > 250) {
+        console.log('Skipped slice bidirectional took', durationMs);
+      }
+      continue;
```
sedghi, line 89: so this is like if it takes more than 250ms we skip it? seems odd

packages/tools/examples/segmentBidirectionalTool/findLargestBidirectional.ts
```diff
@@ -0,0 +1,218 @@
...
+    const length1 = vec3.distance(pointI, point);
+    const length2 = vec3.distance(point, pointJ);
+    if (Math.abs(distance - length1 - length2) < 5 * EPSILON) {
+      return true;
+    }
+  }
+  return false;
+}
+
+function createBidirectionalForSlice(sliceContour, currentMax, isInSegment) {
```
sedghi, line 125: please add documentation on the logic here

packages/tools/examples/segmentBidirectionalTool/findLargestBidirectional.ts
```diff
@@ -0,0 +1,218 @@
...
+
+const distanceSq = (p1, p2) => lengthSq(vec3.sub(vec3.create(), p1, p2));
+
+/**
+ * Determines if there is a point in points other than pointI or pointJ which
+ * is on the line between I and J.
+ */
+function isCrossing(points, pointI, pointJ, distance) {
+  for (const point of points) {
+    if (point === pointI || point === pointJ) {
```
sedghi, line 113: This will never be true, since you are comparing objects/arrays

packages/tools/examples/segmentBidirectionalTool/findLargestBidirectional.ts
```diff
@@ -0,0 +1,218 @@
...
+  const { maxMinor: currentMaxMinor, maxMajor: currentMaxMajor } =
+    currentMax || { maxMajor: 0, maxMinor: 0 };
+  let maxMajor = currentMaxMajor * currentMaxMajor;
+  let maxMinor = currentMaxMinor * currentMaxMinor;
+  let maxMajorPoints;
+  for (let i = 0; i < points.length; i++) {
+    for (let j = i + 1; j < points.length; j++) {
+      const pointI = points[i];
+      const pointJ = points[j];
+      const distance2 = distanceSq(pointI, pointJ);
```
sedghi, line 136: I usually avoid IJ when dealing with ijk vs lps etc., so I recommend firstPoint, secondPoint

packages/tools/examples/segmentBidirectionalTool/findLargestBidirectional.ts
```diff
@@ -0,0 +1,218 @@
...
+  }
+  maxMinor = Math.sqrt(maxMinor);
+  const handle2 = points[maxMinorPoints[0]];
+  const handle3 = points[maxMinorPoints[1]];
+
+  const bidirectional = {
+    handle0,
+    handle1,
+    handle2,
+    handle3,
```
sedghi, line 212: since there is no order, how about we return points, if there is order then we should name them major or minor

sedghi (commented):
Great job! Please take a moment to review my comments. I believe this could be considered as a standalone tool called "LongestBidirectionalMeasurement". Just like other tools, it can be activated, disabled, or enabled. It extends the bidirectional feature and ensures that only one bidirectional measurement is kept per segment. Essentially, it operates on the active segmentation and can be set as "Activated" for calculation purposes.

The purpose behind this is to maintain only one annotation per segment as the segmentation is modified. If a new slice becomes the longest in terms of length, the previous winner should be removed and replaced with the new one. This requires state management, which is where annotation tools come in.

We can handle it similarly to other annotation tools. When "onSetToolActive" is triggered, it performs the necessary calculations. If the segmentation data is modified, it is the responsibility of the user/developer to disable and re-activate the tool using "setToolDisabled" and "setToolActive" respectively. Alternatively, we can introduce a new method called "update" or something similar.

cspell.json
wayfarer3130: If we are going to consistently use the spelling, then we need a shared dictionary with common words specific to CS3D - would like to reconsider the suggestion to have that not be shared.

packages/core/examples/programaticPanZoom/index.ts
wayfarer3130: I wanted to demonstrate minifying the area so you can actually see the affect.  It isn't common to use > 1, but sometimes it is important for ensuring there is an area for rendering important image overlay data - like ECG information on an ECG display, or leaving room for the scroll bar for the video display.

packages/tools/examples/segmentBidirectionalTool/index.ts
wayfarer3130: I wasn't sure where to put it initially, so I figured it would move once we decided where it goes.  Maybe adapters is not a bad place to put it actually - alongside the other code?

packages/tools/examples/segmentBidirectionalTool/findLargestBidirectional.ts
wayfarer3130: Not so, I'm compare two objects to see that the object doesn't get compared to itself.  Hmm, can actually start with position of pointI + 1 - will skip 50% of the comparisons on average, definitely worth doing.

cspell.json
sedghi: Since this is extension specific I don't like it, there is no standard spelling check, so I can imaging someone else use another extension and basically add another config file to the repo, since it is different. Ideally you should add these to your user settings on the vscode

CleanShot 2023-12-13 at 08 30 16

packages/tools/examples/segmentBidirectionalTool/index.ts
sedghi: Read my main comment, I think this is a tool

packages/tools/examples/segmentBidirectionalTool/findLargestBidirectional.ts
wayfarer3130: I think we should leave this in for a while - the timing on this can be fairly variable, and until we are reasonably confident in the implementation, I'd like to leave it.

packages/tools/examples/segmentBidirectionalTool/findLargestBidirectional.ts
wayfarer3130: No, if it takes mroe than 250 ms, we log a message indicating it is slow.  I could adjust the others to be like that, even with warnings.

packages/tools/src/utilities/rtstruct/getRTROIObservationsSequence.js
```diff
@@ -0,0 +1,8 @@
+export default function getRTROIObservationsSequence(toolData, index) {
+  return {
+    ObservationNumber: index + 1,
+    ReferencedROINumber: index + 1,
+    RTROIInterpretedType: 'Todo: type',
+    ROIInterpreter: 'Todo: interpreter',
+  };
+}
```
sedghi, line 8: We don't need the RT stuff in tools, rather just the labelmap to contour. RT can remain in the adapter, and in adapters we only call the the converters and build the RT scheme

packages/tools/src/utilities/rtstruct/getPatientModule.js
```diff
@@ -0,0 +1,33 @@
+export default function getPatientModule(imageId, metadataProvider) {
```
sedghi, line 1: same comment, this is really RT specific and should be in adapters

packages/core/src/utilities/createSigmoidRGBTransferFunction.ts
```diff
@@ -1,7 +1,7 @@
 import vtkColorTransferFunction from '@kitware/vtk.js/Rendering/Core/ColorTransferFunction';
 import vtkDataArray from '@kitware/vtk.js/Common/Core/DataArray';
 import { VOIRange } from '../types';
-import { windowLevel as windowLevelUtil } from '.';
+import * as windowLevelUtil from './windowLevel';
```
sedghi, line 4: I wonder how we can really identify these and lint error it

utils/ExampleRunner/template-config.js
```diff
@@ -88,7 +88,7 @@ module.exports = {
   devServer: {
     hot: true,
     open: false,
-    port: 3000,
+    port: ${process.env.CS3D_PORT || 3000},
```
sedghi, line 91: nice for now

packages/tools/src/utilities/index.ts
```diff
@@ -26,6 +26,7 @@ import { pointToString } from './pointToString';
 import annotationFrameRange from './annotationFrameRange';
 
 // name spaces
+import * as rtstruct from './rtstruct';
```
sedghi, line 29: we should keep rt stuff in adapters, only contour converters in tools

packages/tools/src/utilities/rtstruct/getRTROIObservationsSequence.js
wayfarer3130: Some of the stuff needs to be in tools, or is properly in tools because it is a tool, so leaving htose ones.

packages/core/src/utilities/createSigmoidRGBTransferFunction.ts
wayfarer3130: If we just move to rollup it identifies these.  I found this one with rollup - plus 140 more, but only fixed one.

packages/tools/src/eventDispatchers/keyboardEventHandlers/keyDown.ts
```diff
@@ -34,7 +34,9 @@ export default function keyDown(evt: KeyDownEventType): void {
   if (activeToolsWithEventBinding?.size) {
     const { element } = evt.detail;
     for (const [key, value] of [...activeToolsWithEventBinding.entries()]) {
-      key[value.method](element);
+      const method =
+        typeof value.method === 'function' ? value.method : key[value.method];
+      method.call(key, element, value);
```
sedghi, line 39: can you add a comment what is this in the code

packages/tools/src/stateManagement/annotation/annotationState.ts
```diff
@@ -126,6 +126,9 @@ function getNumberOfAnnotations(
  * @param annotationUID - The unique identifier for the annotation.
  */
 function removeAnnotation(annotationUID: string): void {
+  if (!annotationUID) {
+    return;
```
sedghi, line 130: I think we should throw meaningful error here instead of return

packages/tools/src/utilities/rtstruct/RectangleROIStartEndThreshold.ts
```diff
@@ -0,0 +1,60 @@
+// comment
+class RectangleROIStartEndThreshold {
```
sedghi, line 2: I don't think you need this here, it should be in the adapters

packages/tools/src/utilities/rtstruct/index.ts
```diff
@@ -0,0 +1,14 @@
+import contourFinder from './contourFinder';
+import mergePoints from './mergePoints';
+import detectContourHoles from './detectContourHoles';
+import { generateContourSetsFromLabelmap } from './generateContourSetsFromLabelmap';
+
+import AnnotationToPointData from './AnnotationToPointData';
+
+export {
```
sedghi, line 8: can you rename the folder to utiltilies/contours instead of rtStruct

packages/tools/examples/segmentBidirectionalTool/index.ts
```diff
@@ -0,0 +1,465 @@
...
+      },
+    ],
+    segmentData: new Map(),
+  },
+};
+
+// ======== Set up page ======== //
+setTitleAndDescription(
+  'Segment bidirectional tool',
+  'Here we demonstrate automatic creation of the largest bidirectional tool which will fit the segment'
```
sedghi, line 75: can you expand on this description and tell the different behaviours for segmentIndex 1, 2, 3 (which includes 1,2) as well

packages/tools/examples/segmentBidirectionalTool/index.ts
```diff
@@ -0,0 +1,465 @@
...
+
+viewportGrid.appendChild(element1);
+
+content.appendChild(viewportGrid);
+
+const instructions = document.createElement('p');
+instructions.innerText = `
+  Left Click: Use selected Segmentation Tool.
+  Use button to generate new bidirectional on largest slice.
+  Note that segment index 3 is consider to be 3,2 or 1 combined.
```
sedghi, line 99: move the explanation of segment index 3 to top, and in the drop down also say 3,2,1 instead of 3

packages/tools/examples/segmentBidirectionalTool/index.ts
```diff
@@ -0,0 +1,465 @@
...
+  utilities: cstUtils,
+} = cornerstoneTools;
+
+const { roundNumber } = cstUtils;
+
+const { MouseBindings, KeyboardBindings } = csToolsEnums;
+const { ViewportType } = Enums;
+const { segmentation: segmentationUtils } = cstUtils;
+let renderingEngine;
+const viewportId = 'CT_AXIAL';
```
sedghi, line 50: Could you enhance this example by adding two side-by-side viewports? On the left, we'll have the stack viewport CT, and on the right, the volume (acquisition) PT or something similar. If we also have support for reconstructions, perhaps we can make the right one SAG PT; otherwise, it's alright to keep it as just the acquisition volume. The purpose behind this is to ensure that we test on stack seg from now on for each scenario.

packages/tools/examples/segmentBidirectionalTool/index.ts
```diff
@@ -0,0 +1,465 @@
...
+
+const actionConfiguration = {
+  contourBidirectional: {
+    method: cstUtils.segmentation.segmentContourAction,
+    bindings: [
+      {
+        key: 'c',
+      },
+    ],
+    segmentData: new Map(),
```
sedghi, line 68: can you go a level deeper, so that we have an architecture here? so instead of what you have how about

```
const actionConfiguration = {
  contourBidirectional: {
    method: cstUtils.segmentation.segmentContourAction,
    bindings: [
      {
        key: 'c',
      },
    ],
    data: {
      segmentData: new Map(),
      // later we can put more things here or other actions can put more here
    }
  },
};
```

packages/tools/src/utilities/segmentation/segmentContourAction.ts
```diff
@@ -0,0 +1,161 @@
...
+  if (!segmentationsList.length) {
+    return;
+  }
+  const segmentationId =
+    configuration.segmentationId || segmentationsList[0].segmentationId;
+  const segmentIndex =
+    segmentation.segmentIndex.getActiveSegmentIndex(segmentationId);
+  if (!segmentIndex) {
+    return;
+  }
```
sedghi, line 153: can we get the segmentIndex from configuration as well? if not default to active

packages/tools/src/utilities/segmentation/contourAndFindLargestBidirectional.ts
```diff
@@ -0,0 +1,294 @@
...
+/**
+ * Generates a contour object over the segment, and then uses the contouring to
+ * find the largest bidirectional object that can be applied within the acquisition
+ * plane that is within the segment index, or the contained segment indices.
+ *
+ * @param segmentations.segments - a list of segments to apply the contour to.
+ * @param segmentations.segments.containedSegmentIndices - a set of segment indexes equivalent to the primary segment
+ * @param segmentations.segments.label - the label for the segment
+ * @param segmentations.segments.color - the color to use for the segment label
+ */
```
sedghi, line 32: I don't understand why this is plural (segmentations), does it work on multiple segmentations at the same time (I don't think based on the code I see), so lets rename it

packages/tools/src/utilities/segmentation/contourAndFindLargestBidirectional.ts
```diff
@@ -0,0 +1,294 @@
...
+  if (maxBidirectional) {
+    Object.assign(maxBidirectional, segment);
+  }
+  return maxBidirectional;
+}
+
+const lengthSq = (point) =>
+  point[0] * point[0] + point[1] * point[1] + point[2] * point[2];
+
+const distanceSq = (p1, p2) => lengthSq(vec3.sub(vec3.create(), p1, p2));
```
sedghi, line 157: we already have these utilities, please use them

packages/tools/src/utilities/segmentation/contourAndFindLargestBidirectional.ts
```diff
@@ -0,0 +1,294 @@
...
+  let maxMajorPoints;
+  for (let index1 = 0; index1 < points.length; index1++) {
+    for (let index2 = index1 + 1; index2 < points.length; index2++) {
+      const point1 = points[index1];
+      const point2 = points[index2];
+      const distance2 = distanceSq(point1, point2);
+      if (distance2 < maxMajor) {
+        continue;
+      }
+      if (distance2 === maxMajor && maxMajorPoints) {
```
sedghi, line 219: this should use isEqual to include EPS

packages/tools/src/utilities/segmentation/segmentContourAction.ts
```diff
@@ -0,0 +1,161 @@
...
+  existingSegments.forEach((annotation) => {
+    const segments = [];
+    const { segment: updateSegment } = annotation.data;
+    const { segmentIndex, segmentationId } = updateSegment;
+    segments[segmentIndex] = updateSegment;
+    annotationState.removeAnnotation(annotation.annotationUID);
+    const bidirectionalData = contourAndFindLargestBidirectional({
+      ...segmentationsList.find(
+        (segmentation) => segmentation.segmentationId === segmentationId
+      ),
```
sedghi, line 89: you don't need the ...

packages/tools/src/utilities/segmentation/segmentContourAction.ts
```diff
@@ -0,0 +1,161 @@
...
+    return;
+  }
+  const FrameOfReferenceUID = enabledElement.viewport.getFrameOfReferenceUID();
+  const segmentationsList = segmentation.state.getSegmentations();
+  const { segmentIndex, segmentationId } = segment;
+  const bidirectionals = annotationState.getAnnotations(
+    this.toolName || BidirectionalTool.toolName,
+    FrameOfReferenceUID
+  );
+  let hasExistingActiveSegment = false;
```
sedghi, line 56: please prefer more meaningul names such as `annotationItem` instead of `it`

sedghi (commented):
please see my comments

packages/adapters/src/adapters/Cornerstone3D/RTStruct/index.ts
```diff
@@ -1,9 +1,11 @@
-import { generateContourSetsFromLabelmap } from "./utilities/generateContourSetsFromLabelmap";
+import { utilities } from "@cornerstonejs/tools";
```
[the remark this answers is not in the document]
wayfarer3130: It is already there, albeit as a dev dependency.

packages/tools/src/stateManagement/annotation/annotationState.ts
wayfarer3130: We already have a check below that handles this situation, I just wanted a specific on if the annotationUID isn't defined so it is really obvious, but I don't want to change the behaviour that was pre-existing that you can remove something that doesn't exist.

packages/tools/src/utilities/rtstruct/RectangleROIStartEndThreshold.ts
wayfarer3130: I don't know why we have any tools in the adapters package - this is required for the RectangleROIStartEndThresholdTool, which is definitely a tool, and belongs in the tools package.

packages/tools/src/utilities/segmentation/contourAndFindLargestBidirectional.ts
wayfarer3130: What are they called?  I couldn't find them.

packages/tools/src/utilities/segmentation/segmentContourAction.ts
wayfarer3130: How would it work otherwise?  I need a new object since I don't want to modify the original one, and it has to be the extension of the found one?

packages/tools/src/utilities/segmentation/contourAndFindLargestBidirectional.ts
wayfarer3130: THat would need several more multiplies since it would need to use a multiplicative epsilon.  I agree ideally it would include EPS, but I don't think the performance hit is worthwhile.  I can add a simple + EPSILON in a few places if you want - that would get most of the relevant differences.

packages/tools/src/utilities/segmentation/contourAndFindLargestBidirectional.ts
wayfarer3130: Found them.

packages/tools/src/utilities/rtstruct/RectangleROIStartEndThreshold.ts
wayfarer3130: As agreed, this belongs in the tools package

sedghi (approved):

PR #941, opened 2023-12-12
feat(livewire): livewire tool

lscoder:
Livewire tool that uses Laplacian of Gaussian to detect borders and compute the shortest path.

Implementation based on **LiveWire Segmentation implemented in JavaScript**.
A big thanks to them for making the license MIT.
https://code.google.com/archive/p/livewire-javascript/

**Interactive Segmentation with Intelligent Scissors**
https://courses.cs.washington.edu/courses/cse455/09wi/readings/seg_scissors.pdf

https://github.com/cornerstonejs/cornerstone3D/assets/931820/6f0f40e8-e930-420c-822b-d0570e99f269


packages/tools/src/utilities/livewire/LivewireScissors.ts
```diff
@@ -0,0 +1,589 @@
...
+    const p22 = data[index(y, x)];
+    const p23 = data[index(y, x + 1)];
+    const p24 = data[index(y, x + 2)];
+    const p31 = data[index(y + 1, x - 1)];
+    const p32 = data[index(y + 1, x)];
+    const p33 = data[index(y + 1, x + 1)];
+    const p42 = data[index(y + 2, x)];
+
+    // Laplacian of Gaussian
+    let lap = 0; // -16 * p22;
```
lscoder, line 270: Remove comment

packages/tools/src/utilities/livewire/LivewireScissors.ts
```diff
@@ -0,0 +1,589 @@
...
+    this.startPoint = startPoint;
+    this.costs[startPointIndex] = 0;
+    this.priorityQueueNew.push(startPointIndex);
+  }
+
+  /**
+   * Runs Dijsktra until it finds a path from the start point to the target
+   * point. Once it reaches the target point all the state is preserved in order
+   * to save processing time the next time the method is called for a new target
+   * point. The search restarts from scratch only when setStartPoint is called.
```
lscoder, line 97: `setStartPoint` was renamed to `startSearch`

packages/tools/src/utilities/livewire/LivewireScissors.ts
```diff
@@ -0,0 +1,589 @@
...
+
+    this.visited = new Array(numPixels);
+    this.parents = new Uint32Array(numPixels);
+    this.costs = new Float32Array(numPixels);
+  }
+
+  public startSearch(startPoint: Types.Point2): void {
+    const startPointIndex = this._getPointIndex(startPoint[1], startPoint[0]);
+
+    this._reset();
```
lscoder, line 87: Now that `reset()` is called only by this function it makes sense to add that code here and delete `reset()`

packages/tools/examples/livewireContour/index.ts
```diff
@@ -0,0 +1,180 @@
...
+content.appendChild(element);
+
+const info = document.createElement('div');
+content.appendChild(info);
+
+const addInstruction = (text) => {
+  const instructions = document.createElement('p');
+  instructions.innerText = `- ${text}`;
+  info.appendChild(instructions);
+};
```
sedghi, line 60: Can you make this a general utils that you import here? thanks

packages/tools/src/tools/annotation/LivewireContourTool.ts
```diff
@@ -0,0 +1,772 @@
...
+  AnnotationCompletedEventDetail,
+  AnnotationModifiedEventDetail,
+} from '../../types/EventTypes';
+import { StyleSpecifier } from '../../types/AnnotationStyle';
+
+import { LivewireScissors } from '../../utilities/livewire/LivewireScissors';
+import { LivewirePath } from '../../utilities/livewire/LiveWirePath';
+
+const { getViewportIdsWithToolToRender } = viewportFilters;
+const CLICK_CLOSE_CURVE_SQR_DIST = 10 ** 2;
```
sedghi, line 47: what is the unit, please add docs

packages/tools/examples/livewireContour/index.ts
```diff
@@ -0,0 +1,180 @@
...
+
+  // Create a stack viewport
+  const viewportInput = {
+    viewportId,
+    type: ViewportType.STACK,
+    element,
+    defaultOptions: {
+      background: <Types.Point3>[0.2, 0, 0.2],
+    },
+  };
```
sedghi, line 154: please make this example a two viewport examples, left viewport: stack right viewport: reconstructed (sag, cor)

utils/ExampleRunner/example-info.json
```diff
@@ -224,6 +224,10 @@
       "advancedColorBar": {
         "name": "Advanced Colorbar",
         "description": "Demonstrates how to add an interactive colobar to stack and volume viewports with PT/CT volumes"
+      },
+      "livewire": {
+        "name": "Livewire",
```
sedghi, line 229: should be called livewireContour, i tried it it doesn't work

sedghi (commented):
see my comments, great work

packages/tools/examples/livewireContour/index.ts
lscoder: Done

utils/ExampleRunner/example-info.json
lscoder: Updated

packages/tools/src/tools/annotation/LivewireContourTool.ts
lscoder: Unit added (comment)

packages/tools/examples/livewireContour/index.ts
lscoder: Done

packages/tools/src/tools/annotation/LivewireContourTool.ts
```diff
@@ -0,0 +1,772 @@
...
+
+    const defaultActor = (<Types.IStackViewport>viewport).getDefaultActor();
+    const { actor } = defaultActor;
+    const vtkImageData = actor.getMapper().getInputData();
+    const dimensions = vtkImageData.getDimensions();
+    const [imgWidth, imgHeight] = dimensions;
+
+    // TODO: See what to do when working with volumes because imagePos should be
+    // the index in volume space
+    const imagePos = csUtils.transformWorldToIndex(vtkImageData, worldPos);
```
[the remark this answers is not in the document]
lscoder: This code was refactored after adding support for volume viewports because now there a function to convert points to/from "slice" space which does not change anything for stacks but allow us to take one slice from a volume viewport and map (0,0) and (width-1, height-1) to top-left and bottom-right corners instead of working on index space (ijk, full volume).

packages/core/src/utilities/getCurrentVolumeViewportSlice.ts
```diff
@@ -0,0 +1,168 @@
...
+import { IVolumeViewport, Point2, Point3 } from '../types';
+
+/**
+ * Get the image data for the current slice rendered on the viewport.
+ * It does not work for oblique views.
+ * @param viewport - Volume viewport
+ * @returns Slice image dataand  matrices to convert from volume
+ *   to slice and vice-versa
+ */
+function getCurrentVolumeViewportSlice(viewport: IVolumeViewport) {
```
sedghi, line 12: I feel like we can generalize it to be `getCurrentViewImageData(viewport)`, since for stack viewport we can grab that from .getImageData right?

Also we should throw error for oblique

lscoder: That method does not have any stack-related code.

lscoder: It throws an error a few lines below

packages/core/src/utilities/getCurrentVolumeViewportSlice.ts
```diff
@@ -0,0 +1,168 @@
...
+    const worldPoint = transformIndexToWorld(vtkImageData, volPoint);
+
+    return viewport.worldToCanvas(worldPoint);
+  };
+
+  // Convert coordinates from canvas to index space
+  const canvasToIJK = (canvasPoint: Point2) => {
+    const worldPoint = viewport.canvasToWorld(canvasPoint);
+    return transformWorldToIndex(vtkImageData, worldPoint);
+  };
```
sedghi, line 30: extract these to their own utilities since they are useful. Something like transformIJKToCanvas

packages/core/src/utilities/getCurrentVolumeViewportSlice.ts
```diff
@@ -0,0 +1,168 @@
...
+  // because points such as (0,0), (1,0) and (0,1) may be converted to the same
+  // ijk index when the image is zoomed in.
+  const ijkOriginPoint = canvasToIJK([0, 0]);
+  const ijkRowPoint = canvasToIJK([canvasWidth, 0]);
+  const ijkColPoint = canvasToIJK([0, canvasHeight]);
+
+  // Subtract the points to get the row and column vectors in index space
+  const ijkRowVec = vec3.sub(vec3.create(), ijkRowPoint, ijkOriginPoint);
+  const ijkColVec = vec3.sub(vec3.create(), ijkColPoint, ijkOriginPoint);
+  const ijkDepthVec = vec3.cross(vec3.create(), ijkRowVec, ijkColVec);
```
sedghi, line 45: ijkSliceVec instead of ijkDepth

packages/core/src/utilities/getCurrentVolumeViewportSlice.ts
```diff
@@ -0,0 +1,168 @@
...
+  );
+  const maxIJKColVec = Math.max(
+    Math.abs(ijkColVec[0]),
+    Math.abs(ijkColVec[1]),
+    Math.abs(ijkColVec[2])
+  );
+
+  // Using glMatrix.equals() because the number may be not exactly equal to
+  // 1 due to rounding issues
+  if (!glMatrix.equals(1, maxIJKRowVec) || !glMatrix.equals(1, maxIJKColVec)) {
```
sedghi, line 67: probably we should involve an epsilon here because of floating point errors

packages/core/src/utilities/getCurrentVolumeViewportSlice.ts
```diff
@@ -0,0 +1,168 @@
...
+  for (let y = 0; y < sliceHeight; y++) {
+    vec3.copy(ijkPixelCol, ijkPixelRow);
+
+    for (let x = 0; x < sliceWidth; x++) {
+      const volumePixelIndex =
+        ijkPixelCol[2] * pixelsPerSlice +
+        ijkPixelCol[1] * dimensions[0] +
+        ijkPixelCol[0];
+
+      sliceData[slicePixelIndex++] = scalarData[volumePixelIndex];
```
sedghi, line 146: Shouldn't we check if it is in bound for scalarData length first?

packages/tools/src/tools/annotation/LivewireContourTool.ts
```diff
@@ -0,0 +1,785 @@
...
+      width = viewportImageData.dimensions[0];
+      height = viewportImageData.dimensions[1];
+
+      worldToSlice = (point: Types.Point3) => {
+        const ijkPoint = csUtils.transformWorldToIndex(vtkImageData, point);
+        return [ijkPoint[0], ijkPoint[1]];
+      };
+
+      sliceToWorld = (point: Types.Point2) =>
+        csUtils.transformIndexToWorld(vtkImageData, [point[0], point[1], 0]);
```
sedghi, line 141: seems like slice is index, so why call it slice?

sedghi (commented):
it looks fantastic! My main suggestion is regarding the completion process of the tool. It's a bit unusual and not quite cornerstony. 

I think we should handle it similarly to the spline tool. When a user double-clicks, the last point should be added and the contour should automatically be completed. Currently, the user has to go back to the starting point, which can be inconvenient.

lscoder:
> it looks fantastic! My main suggestion is regarding the completion process of the tool. It's a bit unusual and not quite cornerstony.
> 
> I think we should handle it similarly to the spline tool. When a user double-clicks, the last point should be added and the contour should automatically be completed. Currently, the user has to go back to the starting point, which can be inconvenient.

I made some changes and it is now possible to close the path with double click.

sedghi (approved):

PR #962, opened 2023-12-20
fix(annotation-tools): pointInShapeCallback now returns the correct array of points

Celian-abd:
### Context
The pointInShapeCallback method was returning a array of the same point.

### Changes & Results
pointInShapeCallback was returning the first reference of the `currentPos` everytime and not the the actual values.
- Changed it to the actual values.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  


packages/tools/src/utilities/pointInShapeCallback.ts
```diff
@@ -134,7 +134,12 @@ export default function pointInShapeCallback(
             value = scalarData[index];
           }
 
-          pointsInShape.push({ value, index, pointIJK, pointLPS: currentPos });
+          pointsInShape.push({
+            value,
+            index,
+            pointIJK,
+            pointLPS: [currentPos[0], currentPos[1], currentPos[2]],
```
sedghi, line 141: Can't you just do `pointLPS: currentPos` I'm afraid this is adding extra indexing steps to the loop that runs a lot of time

salimkanoun: It was the bug we are solving, the problem is currentPos was add as a reference  and not as a value. So all point had the same reference value value so at the end of the computation we have the last value of the last update in all points done by the last call of  : 
// Increment currentPos by rowStep for the next iteration
        vec3.add(currentPos, currentPos, rowStep); 

So we took individual indexes to be pass by value and not by reference in the final PointInShape

sedghi, line 141: Can we do `currentPos.slice()` at least?

PR #966, opened 2023-12-21
feat: Added contour roi interpolation tool

Belbin-GK:
### Context
* Creates a contour tool that allows to create interpolated contour ROIs between contour drawn slices by [FlyWheel.io](https://flywheel.io/)
* This allows to quickly create annotation in multiple slices for a tissue. Upcoming sculpt tool along with this feature will make editing the interpolated shapes easier.

### Changes & Results
* Add new interpolated contour ROI tool.
* Introduced 'autoGenerated' property in an annotation to indicate whether the annotation was manually created or automatically generated.
* Added events for annotation label changes and contour interpolation completion.
* Allows suppressing event triggers on adding/removing automatically generated annotations.
* Allows separate tool style for auto generated annotations if needed.

### Testing
* Open example file named 'contourROITool'.
* Follow the instructions given in example file to generate interpolated contour ROI.
* There is a ContourROITool_test file to run tests on the tool.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] "OS: Windows 10 with WSL (Ubuntu) 
- [x] "Node version: 16.14.2
- [x] "Browser: Chrome 120.0.6099.110
  


packages/tools/src/tools/annotation/ContourROITool.ts
```diff
@@ -0,0 +1,124 @@
...
+import PlanarFreehandROITool from './PlanarFreehandROITool';
+import { getViewportIdsWithToolToRender } from '../../utilities/viewportFilters';
+import triggerAnnotationRenderForViewportIds from '../../utilities/triggerAnnotationRenderForViewportIds';
+import InterpolationManager from '../../utilities/contourROITool/InterpolationManager';
+import { InterpolationROIAnnotation } from '../../types/ToolSpecificAnnotationTypes';
+import * as annotation from '../../stateManagement/annotation';
+
+const { getImageSliceDataForVolumeViewport } = csUtils;
+
+class ContourROITool extends PlanarFreehandROITool {
```
sedghi, line 19: Oh, interesting! I was actually expecting PlanarFreeHandROITool to be the more advanced tool that extends the ContourROITool. Could we possibly get some documentation explaining why ContourROITool extends PlanarFreehand?

packages/tools/src/enums/Events.ts
```diff
@@ -96,6 +96,30 @@ enum Events {
...
+   */
+  ANNOTATION_LABEL_CHANGE = 'CORNERSTONE_TOOLS_ANNOTATION_LABEL_CHANGE',
+
+  /**
+   * Triggers on the eventTarget when an annotation interpolation process completed.
+   * Make use of {@link EventTypes.AnnotationInterpolationCompletedEventType | Annotation Interpolation process Completed Event Type}
+   * for typing your event listeners for this annotation interpolation complete event, and see what
+   * event detail is included in {@link EventTypes.AnnotationInterpolationCompletedEventDetail | Annotation interpolation process Event Detail}.
+   */
+  ANNOTATION_INTERPOLATION_PROCESS_COMPLETED = 'CORNERSTONE_TOOLS_ANNOTATION_INTERPOLATION_PROCESS_COMPLETED',
```
wayfarer3130, line 113: Could this be named something generic, so the same event could be used for other long-running annotation processing - maybe just:
ANNOTATION_PROCESSING_COMPLETED
Or maybe even have a few of them, so you can show progress as well as completed, but all against a single event, eg
ANNOTATION_PROCESSING
and then allow for a progress/status update to occur.
That is such a common pattern that it would be really nice to support it generically.
@sedghi - how do you think this should be done, as a final event or a full progress, and is it ok to use for other annotation processing types?

packages/tools/src/utilities/contourROITool/Interpolation/getToolData.ts
```diff
@@ -0,0 +1,47 @@
...
+  const { viewport } = eventData;
+  const camera = viewport.getCamera();
+  const FrameOfReferenceUID = viewport.getFrameOfReferenceUID();
+  const { viewPlaneNormal, viewUp } = camera;
+
+  const annotation: InterpolationROIAnnotation = {
+    highlighted: false,
+    interpolationUID: referencedToolData.interpolationUID,
+    invalidated: true,
+    autoGenerated: true,
```
sedghi, line 14: so is this `getAutogeneratedToolData`?

packages/tools/src/eventDispatchers/annotationInterpolationEventDispatcher.ts
```diff
@@ -0,0 +1,38 @@
+import { eventTarget } from '@cornerstonejs/core';
+import Events from '../enums/Events';
+import InterpolationManager from '../utilities/contourROITool/InterpolationManager';
+
+const enable = function (element: HTMLDivElement) {
+  eventTarget.addEventListener(
+    Events.ANNOTATION_LABEL_CHANGE,
+    InterpolationManager.handleAnnotationLabelChange as EventListener
+  );
```
sedghi, line 9: I believe you are pushing for annotations and labels as being segmentations. However, we are advocating for another method that involves contour segmentation representation, which will serve as the foundation for grouping contours (being able to color, fill, etc. etc.). The PR (https://github.com/cornerstonejs/cornerstone3D/pull/983) will be merged shortly, so it might be a good idea to incorporate that concept into your PR. Let's definitely schedule a meeting to discuss this further. I believe email communications have already begun to arrange a suitable time.

packages/adapters/src/adapters/Cornerstone/ContourRoi.js
```diff
@@ -0,0 +1,82 @@
+import { utilities } from "dcmjs";
+
+import MeasurementReport from "./MeasurementReport";
+import CORNERSTONE_4_TAG from "./cornerstone4Tag";
+
+const { Polyline: TID300Polyline } = utilities.TID300;
+
+class ContourRoi {
```
wayfarer3130, line 8: How does this differ from FreehandROI?  Would like to end up with just a single representation for both types if we can figure that out.

packages/tools/src/stateManagement/annotation/annotationState.ts
```diff
@@ -84,15 +86,17 @@ function addAnnotation(
...
-  // if the annotation manager selector is an element, trigger the
-  // annotation added event for that element.
-  if (annotationGroupSelector instanceof HTMLDivElement) {
-    triggerAnnotationAddedForElement(annotation, annotationGroupSelector);
-  } else {
-    // if no element is provided, render all viewports that have the
-    // same frame of reference.
-    // Todo: we should do something else here for other types of annotation managers.
-    triggerAnnotationAddedForFOR(annotation);
+  if (!suppressEvents) {
```
wayfarer3130, line 89: I'm really not keen on suppressing events here - it means that another viewport depending on the events that happens to be displaying a slice which would get included will not end up being displayed.  As well, OHIF and other viewers based on CS3D may need the events to maintain their internal state.

wayfarer3130: It would be better to use an extra flag that someone can check to see that these events are intermediate or interpolated events - you might add another partial set of values instead of the suppress events flag, and within that include the status of what is happening on the interpolation events.  Then, anyone listening who knows nothing about them can just act as normal, and any place where you need to not do the default handling, you can just check the extra flags.

packages/tools/src/stateManagement/annotation/config/getState.ts
```diff
@@ -7,7 +7,7 @@ import { AnnotationStyleStates } from '../../../enums';
  * Given a Annotation object, return the annotationStyle State that it
  * should be in based on its data
  * @param annotation - The annotation that we want to style.
- * @returns The state of the annotation whether it is Default, Highlighted, Locked, or Selected.
+ * @returns The state of the annotation whether it is Default, Highlighted, Locked, Selected, or AutoGenerated.
```
wayfarer3130, line 10: What is the state diagram here?  That is, what happens if you then select an auto generated object - does it become highlighted?  What happens if you then deselect it?  How does it know to go to AutoGenerated and not Default?  What happens if you then lock an auto generated item?  
@sedghi  - is there a way to add styles for specific tools that take an extra parameter to use a different set of styles?  Perhaps using a different toolName would work?  That would then allow you to convert between autogenerated and regular ones, and choose whether or not you can edit it after auto generation without tweaking it?

packages/tools/src/stateManagement/annotation/config/getState.ts
```diff
@@ -22,6 +22,10 @@ function getState(annotation?: Annotation): AnnotationStyleStates {
     if (isAnnotationLocked(annotation)) {
       return AnnotationStyleStates.Locked;
     }
+
+    if (annotation.data && annotation.autoGenerated) {
```
wayfarer3130, line 26: The positioning of this says what the state diagram is - but I think you need to be specific about it up above, as it wasn't clear without getting this far into the weeds.

packages/tools/src/store/addEnabledElement.ts
```diff
@@ -13,6 +13,7 @@ import {
   touchToolEventDispatcher,
   keyboardToolEventDispatcher,
   imageSpacingCalibratedEventDispatcher,
+  annotationInterpolationEventDispatcher,
```
wayfarer3130, line 16: Can we call this annotationGenerationEventDispatcher please?  Then, it will work for other types of generations, such as single click/call remote web server for AI generation/response with auto-generated events - possibly on multiple series.

packages/tools/src/tools/annotation/ContourROITool.ts
```diff
@@ -0,0 +1,124 @@
...
+    this.configuration.interpolation.interpolateOnEdit = false;
+    this.configuration.makeClockWise = true;
+    InterpolationManager.addTool(ContourROITool.toolName);
+
+    // Register event loops and rendering logic, which are stored in different
+    // Files due to their complexity/size.
+    registerContourDrawLoop(this);
+  }
+
+  getSliceData(viewport, referencedImageId): Types.ImageSliceData {
```
wayfarer3130, line 33: Need docs here - I was thinking you were returning image pixel values, but what you are actually returning is number of slices etc.  I'd suggest an external utilities function that gets this data.  I had calling it slice data, but that is kind of what it is.
Maybe just enhance getImageSliceDataForVolumeViewport with something that handles Stack as well, call it getImageIndexSlices
Then, more implementations can be added to your new location.

packages/tools/src/tools/annotation/ContourROITool.ts
```diff
@@ -0,0 +1,124 @@
...
+      viewUp
+    );
+
+    const viewportIdsToRender = getViewportIdsWithToolToRender(
+      element,
+      this.getToolName()
+    );
+
+    const FrameOfReferenceUID = viewport.getFrameOfReferenceUID();
+    const sliceData: Types.ImageSliceData = this.getSliceData(
```
wayfarer3130, line 75: You don't need to declare the type here - it is implicit based on the return type.

packages/tools/src/tools/annotation/planarFreehandROITool/closedContourEditLoop.ts
```diff
@@ -225,6 +225,10 @@ function finishEditAndStartNewEdit(evt: EventTypes.InteractionEventType): void {
 
   annotation.data.polyline = worldPoints;
   annotation.data.isOpenContour = false;
+  // If any manual update, triggered on an annotation, then it will be treated as non-autogenerated.
```
wayfarer3130, line 228: This bit of detail needs to go into the top level documentation on the contour tool, since it is fairly important to know the behaviour.

packages/tools/src/types/AnnotationTypes.ts
```diff
@@ -11,6 +11,8 @@ type Annotation = {
   isVisible?: boolean;
   /** Has annotation data been invalidated (e.g., as a result of mouse interactions) */
   invalidated?: boolean;
+  /** If the annotation is auto generated from other annotations*/
```
wayfarer3130, line 14: How about:
```
/** 
 * If the contour is auto generated, eg via interpolation or a remote process
 */
```
 
That may or may not be done from other annotations, and although the other lines are in the wrong format, the docs
generator won't pick those up, so having it in the full format is (unfortunately) needed.

packages/tools/src/types/ToolSpecificAnnotationTypes.ts
```diff
@@ -300,6 +300,22 @@ export interface PlanarFreehandROIAnnotation extends Annotation {
   };
 }
 
+export interface InterpolationROIAnnotation
```
wayfarer3130, line 303: Why the name difference from the tool name?

packages/tools/src/tools/annotation/ContourROITool.ts
wayfarer3130, line 19: Could we call this InterpolatationROITool since you call the types interpolation?
Or maybe InterpolationFreehandROITool to mirror  `PlanarFreehandROITool`

packages/tools/src/utilities/contourROITool/Interpolation/freehandInterpolate/interpolate.ts
```diff
@@ -0,0 +1,656 @@
...
+  }
+
+  return {
+    c1Interp,
+    c2Interp,
+  };
+}
+
+/**
+ * _shiftSuperSampledContourInPlace - Shifts the indicies of c1i around to
```
wayfarer3130, line 391: Could you move the helper functions like this one into a utilities.interpolation package?  They are extremely helpful, and having them be separate files allows for other uses of them.
I really like what you are doing here, just that I want to use it in more places.  Consider being able to interpolate between an ellipse and a freeform, or a contour generated on threshold values, with the threshold guiding the interpolation.  lots of very useful use cases here.

packages/tools/src/utilities/contourROITool/Interpolation/freehandInterpolate/interpolate.ts
```diff
@@ -0,0 +1,656 @@
...
+ *                                  superfluous nodes removed.
+ * @param zInterp - Number, The z- coordinate of the desired
+ *                                  interpolation.
+ * @param c1HasMoreNodes - boolean, True if c1 has more original nodes
+ *                                  than c2.
+ * @returns object, The interpolated contour at z=zInterp.
+ */
+function _generateInterpolatedOpenContour(c1ir, c2ir, zInterp, c1HasMoreNodes) {
+  const cInterp = {
+    x: [],
```
wayfarer3130, line 288: Consider using TypedArray values here instead - you are doing quite a lot of manipulation, and the typed arrays are noticeably faster for that sort of thing, not to mention more space efficient.

Belbin-GK:
Since already one MR merged related to this MR. I'am closing this MR.

PR #967, opened 2023-12-21
feat(SculptorTool): Add sculptor tool

AbdullaJaseel:
### Context

- This PR provides sculptor tool by [FlyWheel.io](https://flywheel.io/), which helps to sculpt and edit freehand ROIs (PlanarFreehandROI, PlanarFreehandContourSegmentationTool).

### Changes & Results

- Added new tool called sculptor tool.
- With the Sculptor Tool, users can now effortlessly mold and refine their freehand ROI annotations (PlanarFreehandROI, PlanarFreehandContourSegmentationTool).

### Testing

- Open example file named 'SculptorTool'
- Follow the instructions given in example file to use sculptor tool on planar freehand ROI .
- There is a SculptorTool_test file to run tests on the tool.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] "OS: Windows 10 with WSL (Ubuntu) 
- [x] "Node version: 16.14.2
- [x] "Browser: Chrome 120.0.6099.110
  


packages/tools/src/utilities/contourROITool/Interpolation/InterpolationTypes.ts
```diff
@@ -0,0 +1,18 @@
+import { Types } from '@cornerstonejs/core';
```
sedghi, line 1: can you please remove all the interpolation code from this PR?

sedghi (commented):
we should remove the interpolation and non-tool specific logic from this PR

wayfarer3130:
Can you add your example to example-info.json, underneath the interpolation example?  That way it can be run directly from your PR build to see the status.

packages/tools/examples/freehandROISculptorTool/index.ts
```diff
@@ -0,0 +1,283 @@
...
+    title: 'Sculpt',
+    onClick: () => {
+      toolGroup.setToolActive(FreehandROISculptorTool.toolName, {
+        bindings: [
+          {
+            mouseButton: MouseBindings.Primary, // Left Click
+          },
+        ],
+      });
+      toolGroup.setToolPassive(PlanarFreehandROITool.toolName);
```
wayfarer3130, line 117: This is a bit disturbing here to have to set these to passive.  The normal set tool active should deal with setting other tools using the same binding to be setup appropriately.  At very least you need documentation on why this is getting set the way it is, and ideally some configuration which could be set on the tool to allow automating this without necessarily needing to know what all tool groups need to be updated.

wayfarer3130: You are just setting all the other tools passive here, rather than knowing which one was previously active.  Could
you follow the standard pattern and keep track of which tool is active and update that setting as required?  That way
you will be able to support more tools than just these ones, or more configurations for them - like spline.

packages/tools/examples/freehandROISculptorTool/index.ts
```diff
@@ -0,0 +1,283 @@
...
+  // Set the initial state of the tools.
+  toolGroup.setToolActive(PlanarFreehandROITool.toolName, {
+    bindings: [
+      {
+        mouseButton: MouseBindings.Primary, // Left Click
+      },
+    ],
+  });
+
+  toolGroup.setToolActive(PanTool.toolName, {
```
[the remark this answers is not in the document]
AbdullaJaseel: imported addManipulationBindings and used that to register the defaults for zoom/pan etc.

packages/tools/examples/freehandROISculptorTool/index.ts
AbdullaJaseel: Updated

packages/tools/src/tools/FreehandROISculptorTool.ts
```diff
@@ -0,0 +1,685 @@
...
+      this._commonData.viewportIdsToRender
+    );
+  }
+
+  /**
+   * Returns interactable freehand ROI annotations
+   *
+   * @param element - The viewport element
+   */
+  _filterInteractableFreehandRoiAnnotationsForElement(element: HTMLDivElement) {
```
wayfarer3130, line 146: Can we call this _filterSculptableAnnotationsForElement(...)
Just that we may well have more than just Freehand

packages/tools/src/tools/FreehandROISculptorTool.ts
```diff
@@ -0,0 +1,685 @@
...
+      viewportId,
+      renderingEngineId
+    );
+
+    const toolInstance = toolGroup.getToolInstance(config.referencedToolName);
+
+    config.referencedToolNames.forEach((referencedToolName: string) => {
+      const annotation = getAnnotations(referencedToolName, element);
+      if (annotation) {
+        interactableAnnotation = [...interactableAnnotation, ...annotation];
```
wayfarer3130, line 162: Could you return Map<annotationUID,ContourAnnotation> here please?  That should be declared in the return type.

packages/tools/src/tools/FreehandROISculptorTool.ts
```diff
@@ -0,0 +1,685 @@
...
+import {
+  hideElementCursor,
+  resetElementCursor,
+} from '../cursors/elementCursor';
+import { StyleSpecifier } from '../types/AnnotationStyle';
+import { getStyleProperty } from '../stateManagement/annotation/config/helpers';
+import { triggerAnnotationModified } from '../stateManagement/annotation/helpers/state';
+
+type CommonData = {
+  selectedIndex: number | null;
```
wayfarer3130, line 25: Please store annotationUID here instead of selected index.  The way the annotations are used, I'm concerned that we could get into a situation where   we start applying the freehand to a different annotation.  Not saying that is bad, just something to consider.

wayfarer3130:
Found a bug here that is causing issues - 
image

wayfarer3130:
Really useful to have this tool.

Is there a way to make the sculpt shape a quarter moon - that is the red bit below, aligned to the image side closest to where it starts.  Consider making it drawn with ellipses whose major axis is four times the distance from the annotation you are marking up, and the minor axes are twice and one times the axis to the nearest tool.

That way, one can push against the freeform with either a convex or concave surface.  For a convex surface, start the same way you currently do, while for a convex surface, start with the surface on the other side, and then move it around to the side you want to edit.  Maybe ellipses don't work, because of clipping annotations nearby, but it would be nice to see this tool be a bit more flexible.  Could also add a keyboard shortcut (I can show you how to do that) to swap between shapes.

image

sedghi:
> Found a bug here that is causing the annotation to have a tail on it. Not saying you necessarily have to fix this one, but the fix would be to check to see if a contour overlaps itself, and to exclude that bit of contour.

I know this was a bug in cornerstone legacy and since this tool is a port from legacy it is here as well

@wayfarer3130

packages/tools/src/tools/FreehandROISculptorTool.ts
```diff
@@ -0,0 +1,685 @@
...
+  constructor(
+    toolProps: PublicToolProps = {},
+    defaultToolProps: ToolProps = {
+      supportedInteractionTypes: ['Mouse', 'Touch'],
+      configuration: {
+        maxToolSize: null,
+        minSpacing: 1,
+        referencedToolNames: [
+          'PlanarFreehandROI',
+          'PlanarFreehandContourSegmentationTool',
```
wayfarer3130, line 59: Could you switch your tool selection to a pull down in the demo so you actually can test with planar segmentation?

packages/tools/src/tools/FreehandROISculptorTool.ts
```diff
@@ -0,0 +1,685 @@
...
+
+    if (this.toolSize && config.maxToolSize) {
+      return;
+    }
+
+    const eventData = evt.detail;
+    const element = eventData.element;
+    const minDim = Math.min(element.clientWidth, element.clientHeight);
+    const sampleArea = Math.pow(minDim / 5, 2);
+    const maxRadius = Math.pow(sampleArea / Math.PI, 0.5);
```
wayfarer3130, line 186: Math.sqrt is about 10% faster than Math.pow, according to my tests, and I find it easier to read, and (minDim*minDim)/25 is also faster because it doesn't need the function call.

packages/tools/src/tools/FreehandROISculptorTool.ts
```diff
@@ -0,0 +1,685 @@
...
+  }
+
+  /**
+   * Sculpts the freehand ROI with the circular freehandSculpter tool, moving,
+   * adding and removing handles as necessary.
+   *
+   * @param eventData - Data object associated with the event.
+   * @param points - Array of points
+   */
+  _sculpt(eventData: any, points: Array<Types.Point3>): void {
```
wayfarer3130, line 199: Can you make this method protected sculpt( ...)
It may well be desirable to extend your class in the future.

packages/tools/src/tools/FreehandROISculptorTool.ts
```diff
@@ -0,0 +1,685 @@
...
+  }
+
+  /**
+   * Pushes the points radially away from the mouse if they are
+   * contained within the circle defined by the freehandSculpter's toolSize and
+   * the mouse position.
+   *
+   * @param viewport
+   */
+  _pushHandles(viewport: Types.IViewport) {
```
wayfarer3130, line 336: Again, can you make this protected pushHandles so it could be extended or re-used?

packages/tools/src/tools/FreehandROISculptorTool.ts
```diff
@@ -0,0 +1,685 @@
...
+      currentCanvasPoint,
+      nextCanvasPoint
+    );
+
+    if (distanceToNextHandle > maxSpacing) {
+      indicesToInsertAfter.push(i);
+    }
+  }
+
+  _getNextHandleIndex(i: number, length: number): number {
```
wayfarer3130, line 607: This is better done as a function up top:
const contourIndex = (i:number, length: number) => (i+length-1) % (length-1)
Then, you can use it like this:
contourIndex(position+1,length)   
contourIndex(position-1,length)
to get next/previous indices, as well as other indices.  I've been meaning to add that as a utility.

packages/tools/src/tools/FreehandROISculptorTool.ts
```diff
@@ -0,0 +1,685 @@
...
+    if (i === length - 1) {
+      return 0;
+    }
+
+    return i + 1;
+  }
+
+  _getNextHandleIndexBeforeInsert(insertIndex: number, length: number): number {
```
wayfarer3130, line 615: This one can just use the baseline contourIndex function, as it is identical to that - but it also works for negative values (which can be very handy).

packages/tools/src/tools/FreehandROISculptorTool.ts
```diff
@@ -0,0 +1,685 @@
...
+        AnnotationStyleStates.Highlighted,
+        this.mode
+      );
+    }
+
+    const circleUID = '0';
+
+    drawCircleSvg(
+      svgDrawingHelper,
+      '',
```
wayfarer3130, line 673: Could you use the tool name here - that way you won't accidentally update someone else's drawing.  I know it is annotationUID, but it isn't needed.

packages/tools/src/tools/FreehandROISculptorTool.ts
```diff
@@ -0,0 +1,685 @@
...
+  mouseLocation: Types.Point2 | undefined;
+};
+
+class FreehandROISculptorTool extends BaseTool {
+  static toolName: string;
+  private _commonData?: CommonData = {
+    selectedIndex: null,
+    viewportIdsToRender: [],
+    isEditingOpenContour: false,
+    mouseLocation: undefined,
```
wayfarer3130, line 37: Please call this canvasLocation - the problem is mouse location is not necessarily a mouse (could be touch), nor is it always in canvas coordinates - in fact I don't know what coordinate system this is in, so if it is in a different coordinate (say world), then name it appropriately.

packages/tools/src/tools/FreehandROISculptorTool.ts
```diff
@@ -0,0 +1,685 @@
...
+type CommonData = {
+  selectedIndex: number | null;
+  viewportIdsToRender: any[];
+  isEditingOpenContour: boolean;
+  mouseLocation: Types.Point2 | undefined;
+};
+
+class FreehandROISculptorTool extends BaseTool {
+  static toolName: string;
+  private _commonData?: CommonData = {
```
wayfarer3130, line 33: Seems to me this should start out null, and then you fill it in the first time you create a sculpt tool?  If it is never null, then you can remove the ? after the commonData.

wayfarer3130 (changes requested):
Looks very nice - a few changes, but other than the one to allow pushing with a contour polyline instead of a straight distance, I don't think there is a huge amount to fix.

packages/tools/src/tools/FreehandROISculptorTool.ts
AbdullaJaseel: Updated example file, now the sculpt tool can be tested with planar segmentation.

packages/tools/src/tools/FreehandROISculptorTool.ts
AbdullaJaseel: replaced Math.pow

packages/tools/src/tools/FreehandROISculptorTool.ts
AbdullaJaseel: Updated

packages/tools/src/tools/FreehandROISculptorTool.ts
AbdullaJaseel: Updated

packages/tools/src/tools/FreehandROISculptorTool.ts
```diff
@@ -0,0 +1,685 @@
...
+      Events.MOUSE_CLICK,
+      this._endCallback as EventListener
+    );
+    element.removeEventListener(
+      Events.MOUSE_DRAG,
+      this._dragCallback as EventListener
+    );
+  }
+
+  _checkSpacing(
```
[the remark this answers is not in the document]
AbdullaJaseel: Added docs and renamed the function

packages/tools/src/tools/FreehandROISculptorTool.ts
AbdullaJaseel: Updated

packages/tools/src/tools/FreehandROISculptorTool.ts
AbdullaJaseel: Updated

packages/tools/src/tools/FreehandROISculptorTool.ts
AbdullaJaseel: Updated

packages/tools/src/tools/FreehandROISculptorTool.ts
AbdullaJaseel: Below function `filterInteractableAnnotationsForElement` and this function usages expect annotations to be in array format. Changing this expectation would result in unnecessary conversions between map and array data structures.

packages/tools/src/tools/FreehandROISculptorTool.ts
```diff
@@ -0,0 +1,685 @@
...
+   * contained within the circle defined by the freehandSculpter's toolSize and
+   * the mouse position.
+   *
+   * @param viewport
+   */
+  _pushHandles(viewport: Types.IViewport) {
+    const { points, toolSize, mouseCanvasPoint } = this._sculptData;
+    const pushedHandles = { first: undefined, last: undefined };
+
+    for (let i = 0; i < points.length; i++) {
```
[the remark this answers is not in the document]
AbdullaJaseel: The intersection between two polyline paths would take more time to process. So we prefer this logic to be provided by each user defined sculpt tool shape. This way each shape can optimize the processing for points. We have separated the circle shape independent of the sculptor tool as ISculptToolShape interface.

AbdullaJaseel:
@wayfarer3130 @sedghi 
Could you please let us know if any other changes are needed? Please approve if the changes are okay.

sedghi:
@AbdullaJaseel Were you able to resolve the issue mentioned here about the edges by any chance?

https://github.com/cornerstonejs/cornerstone3D/pull/967#issuecomment-1921351873

AbdullaJaseel:
> @AbdullaJaseel Were you able to resolve the issue mentioned here about the edges by any chance?
> 
> [#967 (comment)](https://github.com/cornerstonejs/cornerstone3D/pull/967#issuecomment-1921351873)

@sedghi 
We have not implemented using polygon shape as sculpt tool with this PR. This would be less performant than current circle sculpting as it would require to compare many points from both shapes and need to handle many edge cases. Also we think a tool rotation/tilt behavior also needed when implementing random shape for sculpt tool.
 
To handle the requirement later, we have now separated the sculpt tool shape as a strategy as some shapes can provide optimized way of updating points. This way developers can implement any ISculptToolShape and register and activate the shapes for sculpting.

packages/tools/src/tools/SculptorTool.ts
```diff
@@ -0,0 +1,590 @@
...
+    this.registerShapes(CircleSculptCursor.shapeName, CircleSculptCursor);
+    this.setToolShape(this.configuration.toolShape);
+  }
+
+  /**
+   * Register different tool shapes for sculptor tool
+   * @param shapeName name of shape
+   * @param shapeClass shape class
+   */
+  registerShapes(shapeName: string, shapeClass: any): void {
```
wayfarer3130, line 80: I believe the declaration you want here is:
<T extends ShapeClass>registerShapes(shapeName: string, shapeClass: new () => T)
I'm not 100% sure, and you might not have a class for ShapeClass, although you probably do need an interface for it.
If that doesn't work, then just note it and I will approve the PR.

packages/tools/src/tools/SculptorTool.ts
```diff
@@ -0,0 +1,590 @@
...
+
+    const enabledElement = getEnabledElement(element);
+    const { viewport } = enabledElement;
+    const cursorShape = this.registeredShapes.get(this.selectedShape);
+
+    this.sculptData = {
+      mousePoint: eventData.currentPoints.world,
+      mouseCanvasPoint: eventData.currentPoints.canvas,
+      points,
+      maxSpacing: cursorShape.getMaxSpacing(config.minSpacing),
```
wayfarer3130, line 131: You should be able to declare an inline interface that has getMaxSpacing and has pushHandles on it

packages/tools/src/tools/SculptorTool.ts
```diff
@@ -0,0 +1,590 @@
...
+    sculptableAnnotations =
+      toolInstance.filterInteractableAnnotationsForElement(
+        element,
+        sculptableAnnotations
+      );
+
+    return sculptableAnnotations;
+  }
+
+  private configureToolSize(evt: EventTypes.InteractionEventType): void {
```
wayfarer3130, line 247: Can you add a comment saying this calls the shape specific configure tool size?  I had to read the code to figure out why this one was different from the earlier tool size method

packages/tools/src/tools/SculptorTool.ts
```diff
@@ -0,0 +1,590 @@
...
+    const points = activeAnnotation.data.contour.polyline;
+
+    this.sculpt(eventData, points);
+  }
+
+  /**
+   * Attaches event listeners to the element such that is is visible, modifiable, and new data can be created.
+   * @param element - - The viewport element to attach event listeners to.
+   */
+  private activateSculpt(element: HTMLDivElement): void {
```
wayfarer3130, line 465: Can we use the generic names on this - which are:
_activateModify
_deactivateModify
See AdvancedMagnifyTool for one example.

Also, the methods should be field methods -that causes the bindings to be held, so that when you add it as a listener, you don't need to rebind it, that is:
private _activateModify = (element: HTMLDivElement) => {
....};

packages/tools/src/tools/SculptorTool.ts
```diff
@@ -0,0 +1,590 @@
...
+          'PlanarFreehandROI',
+          'PlanarFreehandContourSegmentationTool',
+        ],
+        toolShape: 'circle',
+        referencedToolName: 'PlanarFreehandROI',
+      },
+    }
+  ) {
+    super(toolProps, defaultToolProps);
+    this.endCallback = this.endCallback.bind(this);
```
wayfarer3130, line 69: This should be done by declaration instead - will comment int he right spot.  That avoids a separate bind.

packages/tools/src/tools/SculptorTool.ts
```diff
@@ -0,0 +1,590 @@
...
+
+    triggerAnnotationModified(activeAnnotation, element);
+  };
+
+  /**
+   * Event handler for MOUSE_DRAG during the active loop.
+   *
+   * @param evt - The event
+   */
+  private dragCallback(evt: EventTypes.InteractionEventType): void {
```
wayfarer3130, line 440: Bind this with
private dragCallback: (evt: EventTypes.InteractionEventType) => {
....

That avoids needing to re-assign the value in the constructor.

packages/tools/src/tools/SculptorTool/CircleSculptCursor.ts
```diff
@@ -0,0 +1,174 @@
+import type { Types } from '@cornerstonejs/core';
+import { getEnabledElement } from '@cornerstonejs/core';
+import { ISculptToolShape } from './ISculptToolShape';
+import { distanceFromPoint, SculptData } from '../SculptorTool';
+import { drawCircle as drawCircleSvg } from '../../drawingSvg';
+import { point } from '../../utilities/math';
+import { SVGDrawingHelper, EventTypes, Annotation } from '../../types';
+
+export interface PushedHandles {
```
wayfarer3130, line 9: This one should be a type - that is more generic in typescript, whereas interface is really only used when you need to declare extends ... or implements in a class.

packages/tools/src/tools/SculptorTool/CircleSculptCursor.ts
```diff
@@ -0,0 +1,174 @@
...
+  first?: number;
+  last?: number;
+}
+
+class CircleSculptCursor implements ISculptToolShape {
```
wayfarer3130, line 14: Need at least a short comment here.

packages/tools/src/tools/SculptorTool/CircleSculptCursor.ts
```diff
@@ -0,0 +1,174 @@
...
+    const element = eventData.element;
+    const minDim = Math.min(element.clientWidth, element.clientHeight);
+    const sampleArea = (minDim * minDim) / 25;
+    const maxRadius = Math.sqrt(sampleArea / 3.14);
+
+    toolInfo.toolSize = maxRadius;
+    toolInfo.maxToolSize = maxRadius;
+  }
+
+  UpdateToolsize(
```
wayfarer3130, line 85: Can you lower initial case this function name?

packages/tools/src/tools/SculptorTool/CircleSculptCursor.ts
wayfarer3130: Generally a little bit more documentation would make this easier to follow.

packages/tools/src/tools/SculptorTool/ISculptToolShape.ts
```diff
@@ -0,0 +1,61 @@
+import type { Types } from '@cornerstonejs/core';
+import { SVGDrawingHelper, EventTypes, Annotation } from '../../types';
+import { PushedHandles } from './CircleSculptCursor';
+import { SculptData } from '../SculptorTool';
+
+/**
+ * This interface defines a contract for implementing various shapes within sculptor tool.
+ * Classes such as `CircleSculptCursor` adhere to this interface, providing specific implementations for sculptor tools to utilize circle shapes during sculpting operations.
+ */
+export interface ISculptToolShape {
```
wayfarer3130, line 10: This belongs in the tools/types/ package

packages/tools/src/tools/SculptorTool/ISculptToolShape.ts
```diff
@@ -0,0 +1,61 @@
...
+   * @param canvasLocation - Current canvas location
+   * @param options - Options for drwing shapes
+   */
+  renderShape(
+    svgDrawingHelper: SVGDrawingHelper,
+    canvasLocation: Types.Point2,
+    options: any
+  ): void;
+  /**
+   * Pushes the points radially away from the mouse if they are
```
wayfarer3130, line 23: These are good comments - suggest just copying them to the Circle instance.

packages/tools/src/tools/SculptorTool/ISculptToolShape.ts
```diff
@@ -0,0 +1,61 @@
+import type { Types } from '@cornerstonejs/core';
+import { SVGDrawingHelper, EventTypes, Annotation } from '../../types';
+import { PushedHandles } from './CircleSculptCursor';
+import { SculptData } from '../SculptorTool';
+
+/**
+ * This interface defines a contract for implementing various shapes within sculptor tool.
+ * Classes such as `CircleSculptCursor` adhere to this interface, providing specific implementations for sculptor tools to utilize circle shapes during sculpting operations.
+ */
```
wayfarer3130, line 9: The constructor for this would look like:

<T extends IScultToolShape>registerShape(name: string, toolClass: new () => T)

wayfarer3130 (commented):
Just cosmetic/docs changes now - the bones of this look quite good, but the costmetic changes make it easier to include as part of a larger project.  Would like to see this get updated and included soon.

wayfarer3130:
I'm going to apply the comments and then merge this PR

wayfarer3130:
@AbdullaJaseel  - I made the remaining type/documentation changes, and have approved this now - if you are ok too, then I will merge.

wayfarer3130 (approved):

AbdullaJaseel:
@wayfarer3130 I have found and resolved a problem related to the updateToolSize function usage. Everything else looks good. Thank you for your update

Issue #492, opened 2023-03-18
Is Volume Viewport CPU memory redundant?


Ouwen:
This is related to https://github.com/cornerstonejs/cornerstone3D-beta/issues/452

Looking at the code, it looks like a 3D texture is created by the following chain of vtk calls:

https://github.com/Kitware/vtk-js/blob/00a46c083f4a637eb9433ea06bfc3b63ba262189/Sources/Rendering/OpenGL/Texture/index.js#L1408-L1415

https://github.com/Kitware/vtk-js/blob/00a46c083f4a637eb9433ea06bfc3b63ba262189/Sources/Rendering/OpenGL/Texture/index.js#L1301-L1338

In cornerstone3D, the streaming viewport will use texSubImage3D to send frame offsets
https://github.com/cornerstonejs/cornerstone3D-beta/blob/47d0671b38e22c6509a1f02cce0489b6512e33fd/packages/core/src/RenderingEngine/vtkClasses/vtkStreamingOpenGLTexture.js#L160-L164

It doesn't seem like 3D requires one continuous memory array in order to display the image. Instead would it be possible to keep individual slices same as the stack viewport and feed in calculated offsets for the 3D case?

This would allow CPU memory to be relatively light and not duplicate data kept in the GPU memory.

Ouwen:
@sedghi @JamesAPetts?

Ouwen:
<img width="1397" alt="image" src="https://user-images.githubusercontent.com/5455421/226086406-5af6b1ea-b43a-4cfc-a5a6-0db066700297.png">

Here is a quick POC on the `yarn run example volumeAPI`
we can run the following in the console and still have most of the volume functionality:

```
const v = cornerstone.cache.getVolume('cornerstoneStreamingImageVolume:CT_VOLUME_ID')
delete v
cornerstone.getEnabledElements()[0].viewport.getDefaultActor().actor.getMapper().getInputData().getPointData().getScalars().delete()
```

swederik:
That's a very interesting idea. So you're saying that we could just keep
the individual slices separate and only build the volume for WebGL when
it's being used in the GPU.

I think it's not exactly redundant because with the volume loader we stream
images directly into the volume buffer so the original slice should no
longer be an individual entry in memory (unless we split the volume up when
we hit the cache limit). Unless something has changed, we shouldn't have
two copies of the same pixel data for the image and for the volume.

Not building a single volume on the CPU side would be interesting because
[...]

PR #969, opened 2023-12-22
performance(memory): Optimize caching and refactor stack segmentation for memory efficiency

sedghi:
### Context

Kind of address this https://github.com/cornerstonejs/cornerstone3D/issues/492 

This PR has couple of good stuff

**1. Cache Optimizations via config enableCacheOptimization: true by default - reducing memory footprint by half going from stack to MPR and vice versa** 

Perfecting the memory consumption going to MPR from a Stack.

Basically, when we have a stack viewport (image stack) and we convert it to a volume the volume will be cached as well as the stack. However, if we can optimize this by going back to the image cache and create a view at the correct offset of the bigger volume array buffer, this will save memory. This will get enabled if cornerstone3D is configured to use SharedArrayBuffer by default 

Before
CleanShot 2024-01-02 at 10 37 49

After
CleanShot 2024-01-02 at 10 35 25

**2. Moving the decaching down, since not every volume is streaming volume but a volume should be able to decache into multiple images**

**4. Utilities for going to Stack Viewport to Volume Viewport and Vice versa**

This utility will make sure the viewports are properly handled

**5. Utilities for going from Stack Segmentation to Volume Segmentation with an example** 

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  


common/reviews/api/core.api.md
```diff
@@ -1184,6 +1224,11 @@ interface IGeometryLoadObject {
 
 // @public (undocumented)
 interface IImage {
+    // (undocumented)
+    bufferView?: {
+        buffer: ArrayBuffer;
+        offset?: number;
```
wayfarer3130, line 1230: Please require the offset if you are using a bufferView at all - that just makes things a bit more consistent, and adding in 0 is pretty easy.

common/reviews/api/core.api.md
```diff
@@ -1212,6 +1257,8 @@ interface IImage {
     // (undocumented)
     height: number;
     // (undocumented)
+    imageFrame?: any;
```
wayfarer3130, line 1260: Please ensure this is documented as being 1 based, since it is sometimes hard to remember which one is which.

packages/core/src/cache/classes/ImageVolume.ts
```diff
@@ -158,6 +225,453 @@ export class ImageVolume implements IImageVolume {
...
+      columnPixelSpacing: spacing[0],
+      rowPixelSpacing: spacing[1],
+      invert,
+      photometricInterpretation,
+    };
+  }
+
+  /**
+   * Converts imageIdIndex into frameIndex which will be the same
+   * for 3D volumes but different for 4D volumes
```
wayfarer3130, line 453: Add note saying that indices are 0 based, unlike frameNumber

packages/core/src/eventTarget.ts
```diff
@@ -12,6 +12,20 @@ class CornerstoneEventTarget implements EventTarget {
     this.listeners = {};
   }
 
+  public addEventListenerOnce(type, callback) {
```
wayfarer3130, line 15: Thank you - this is nice to have.

packages/core/src/types/Cornerstone3DConfig.ts
```diff
@@ -51,6 +51,18 @@ type Cornerstone3DConfig = {
...
   };
+  /**
+   * This flag controls whether to enable cache optimization or not. Basically,
+   * when we have a stack viewport (image stack) and we convert it to a volume
+   * the volume will be cached as well as the stack. However, if we can optimize this
+   * by going back to the image cache and create a view at the correct offset
+   * of the bigger volume array buffer, this will save memory. This will get enabled
+   * if cornerstone3D is configured to use SharedArrayBuffer, the reason is that
+   * when we modify the image cache then the images are referring to a different
+   * buffer (SharedArrayBuffer) and sometimes in some system we don't want to enable
```
wayfarer3130, line 62: Just say that some systems don't support shared array buffer.

packages/core/src/types/IImage.ts
```diff
@@ -120,6 +120,12 @@ interface IImage {
 
   imageQualityStatus?: ImageQualityStatus;
   calibration?: IImageCalibration;
+  imageFrame?: any;
+
+  bufferView?: {
+    buffer: ArrayBuffer;
+    offset?: number;
```
wayfarer3130, line 127: Make the offset required if the bufferView is defined.
Consider including the string representation of the view as well (eg 'Uint8array')

packages/core/src/utilities/cacheUtils.ts
```diff
@@ -0,0 +1,107 @@
...
+    return;
+  }
+
+  const scalarData = volume.getScalarData();
+
+  volume.imageCacheOffsetMap.size > 0
+    ? _processImageCacheOffsetMap(volume, scalarData)
+    : _processVolumeImages(volume, scalarData);
+
+  console.log(
```
wayfarer3130, line 54: Remove - at least for commit.  Really we need to add a log utility so that we can leave debug information in without hitting it all the time.

packages/core/src/utilities/cacheUtils.ts
```diff
@@ -0,0 +1,107 @@
...
+    if (!image) {
+      return;
+    }
+
+    _updateImageWithScalarDataView(image, scalarData, offset);
+    cache.decrementImageCacheSize(image.sizeInBytes);
+  });
+}
+
+function _processVolumeImages(volume, scalarData) {
```
wayfarer3130, line 71: Docs please

packages/core/src/utilities/convertVolumeToStackViewport.ts
```diff
@@ -0,0 +1,99 @@
...
+  const purgeFromCache = isDerivedFromStack
+    ? allImagesAlreadyCached
+      ? true
+      : false
+    : false;
+
+  const volumeUsedInOtherViewports =
+    renderingEngine
+      .getVolumeViewports()
+      .filter((vp) => vp.hasVolumeId(volumeId)).length > 0;
```
wayfarer3130, line 83: this can be done with find, which stops on the first element found, and does not create a nested list.

packages/core/src/utilities/convertVolumeToStackViewport.ts
```diff
@@ -0,0 +1,99 @@
...
+
+  volume.decache(purgeFromCache && !volumeUsedInOtherViewports);
+
+  const stack = volume.imageIds.reverse();
```
wayfarer3130, line 87: Use toReversed instead, so you don't mutate in place.

packages/core/src/utilities/convertVolumeToStackViewport.ts
```diff
@@ -0,0 +1,99 @@
...
+  // their copy of the image back
+  // 2. It was actually a native volume and we need to decache it, this is a bit more
+  // complicated since then we need to decide on the imageIds for it to get
+  // decached to
+  const isDerivedFromStack = volume.imageCacheOffsetMap.size > 0;
+  const allImagesAlreadyCached = volume.imageIds.every((imageId) =>
+    cache.getImage(imageId)
+  );
+
+  const purgeFromCache = isDerivedFromStack
```
wayfarer3130, line 74: Why not isDerivedFromStack && allImagesAlreadyCached?  I found it hard to read as a multi-level conditional.

packages/core/src/utilities/convertVolumeToStackViewport.ts
```diff
@@ -0,0 +1,99 @@
...
+    viewportId?: string;
+    background?: Types.Point3;
+  };
+}): Promise<Types.IVolumeViewport> {
+  const volumeViewport = viewport;
+  const { id, element } = volumeViewport;
+  const renderingEngine = viewport.getRenderingEngine();
+  const imageIdIndex = viewport.getCurrentImageIdIndex();
+
+  const { background } = options || { volumeId: uuidv4() };
```
wayfarer3130, line 33: Why do you generate a volumeId here and then throw it away?  Seems very odd.  You aren't even allowed to specify  volumeId in the options type.

packages/core/src/utilities/convertVolumeToStackViewport.ts
```diff
@@ -0,0 +1,99 @@
...
+ * @param params - The parameters for the conversion.
+ * @param params.viewport - The volume viewport to convert.
+ * @param params.options - The conversion options.
+ * @param [params.options.viewportId] - The new stackViewportId, If not provided, the volume viewport id will be used.
+ * @param [params.options.background] - The background color of the stack viewport.
+ * @param [params.options.decache] - Whether to decache the volume. Defaults to false.
+ *
+ * @returns The converted stack viewport.
+ */
+async function convertVolumeToStackViewport({
```
wayfarer3130, line 18: Should this throw if the viewport is showing a non-acquisition plane modality?

packages/core/src/utilities/convertVolumeToStackViewport.ts
```diff
@@ -0,0 +1,99 @@
...
+  volume.decache(purgeFromCache && !volumeUsedInOtherViewports);
+
+  const stack = volume.imageIds.reverse();
+  await stackViewport.setStack(
+    stack,
+    Math.max(volume.imageIds.length - imageIdIndex - 1, 0)
+  );
+
+  // Render the image
+  volumeViewport.render();
```
wayfarer3130, line 94: Why are you rendering the volume viewport?  Won't that be the wrong one now?

packages/streaming-image-volume-loader/src/BaseStreamingImageVolume.ts
```diff
@@ -351,6 +241,17 @@ export default class BaseStreamingImageVolume
     // it is either cachedImage or cachedVolume
     const isFromImageCache = !!cachedImage;
 
+    if (isFromImageCache) {
```
wayfarer3130, line 244: Can you check the target buffer too?  If that isn't set, you shouldn't be doing anything here.

common/reviews/api/core.api.md
sedghi: it is the imageFrame object from the CSWIL

packages/core/src/utilities/convertVolumeToStackViewport.ts
sedghi: I see it is pretty recent

CleanShot 2024-01-02 at 12 12 49

I will write it manually

packages/tools/examples/stackToVolumeWithSegmentations/index.ts
```diff
@@ -0,0 +1,430 @@
...
+  },
+});
+
+function addOrientationDropdownIfVolumeViewport() {
+  const renderingEngine = getRenderingEngine(renderingEngineId);
+  const viewport = renderingEngine.getViewport(viewportId);
+
+  // Check if the current viewport is a VolumeViewport
+  if (viewport.type === ViewportType.ORTHOGRAPHIC) {
+    addDropdownToToolbar({
```
wayfarer3130, line 167: There is no corresponding remove when you drop it back to stack, so you get more and more dropdowns as you add.

packages/core/src/utilities/convertVolumeToStackViewport.ts
sedghi: This is a terrible API, unfortunately, that I will fix in 2.x. Basically, `decache(true)` means remove from cache, `decache(false)` means decache into images. It doesn't make sense, but it is what it is for now.

packages/core/src/utilities/convertVolumeToStackViewport.ts
sedghi: Hmmm, I don't think so, since it is only a `view` and does not matter regarding the underlying images

wayfarer3130:
Can get into a very weird state with the display going orange if you:
yarn run example stacktovolumewithsegmentations
Disable cache/set Fast 5g (50mbit/s, 10 ms latency)
Convert/add segmentation randomly between stack and volume on different slices.
Wait for load to complete
Sometimes, but not always, nothing is being loaded, but images are all gray/blank

Suggest configuring this with the interleaved loading and test with that as well - this should still work, and eventually succeed.

packages/tools/src/eventListeners/segmentation/imageChangeEventListener.ts
```diff
@@ -148,10 +150,25 @@ function _imageChangeEventListener(evt) {
...
+    const { origin: currentOrigin } =
+      viewport.getImageDataMetadata(currentImage);
+
+    // IMPORTANT: We need to make sure that the origin of the segmentation
+    // is the same as the current image origin. This is because due to some
+    // floating point precision issues, when coming from volume to stack
+    // the origin of the segmentation can be slightly different from the
+    // current image origin. This can cause the segmentation to be rendered
+    // in the wrong location.
+    // Todo: This will not work for segmentations that are not in the same
```
wayfarer3130, line 166: I think you mean in the same Frame of Reference?

packages/tools/examples/stackToVolumeWithSegmentations/index.ts
sedghi: There is down below

```js
const orientationDropdown = document.getElementById('orientationDropdown');
    if (orientationDropdown) {
      orientationDropdown.remove();
    }
```

wayfarer3130:
> 3. Refactor the Stack segmentation to accept multiple stack of images
>
> This is building block of enabling rendering of the overlapping segments. The rendering part is not there yet, but will be soon.

I'm not convinced that is a good idea - it makes knowing which segmentation instance to draw into quite difficult.  There are several other options where a segment index can represent multiple segments that seems like a better solution.  The memory consumption of that is also quite a bit less.  Having multiple segmentation images should be handled by having two or more segmentation UIDs, each with a single segmentation image per base image frame.

packages/tools/src/eventListeners/segmentation/segmentationDataModifiedEventListener.ts
```diff
@@ -101,7 +101,10 @@ const onSegmentationDataModified = function (
           const currentSegmentationImageId =
             imageIdReferenceMap.get(currentImageId);
 
-          const segmentationImage = cache.getImage(currentSegmentationImageId);
+          // Todo: fix the logic to support multiple derived imageIds
+          const segImageId = [...currentSegmentationImageId][0];
```
wayfarer3130, line 105: This seems like a bad idea to convert to an array and back - it is relatively slow - you can instead get the iterator if you want and the first element, or add a helper function to do that for you.

packages/tools/src/store/ToolGroupManager/ToolGroup.ts
```diff
@@ -185,6 +185,10 @@ export default class ToolGroup implements IToolGroup {
    * @param renderingEngineId - The rendering engine to use.
    */
   public addViewport(viewportId: string, renderingEngineId?: string): void {
+    if (typeof viewportId !== 'string') {
```
wayfarer3130, line 188: How about !viewportId  since that checks for empty, null and undefined, which seems a better solution here.

sedghi: I was passing the `viewport` object itself, and I wasted 30 minutes on this trying to figure out what was wrong. So, `!viewportId` will not catch that, unfortunately.

packages/tools/src/tools/displayTools/Labelmap/addLabelmapToElement.ts
```diff
@@ -61,10 +61,13 @@ async function addLabelmapToElement(
   } else {
     // We can use the current imageId in the viewport to get the segmentation imageId
     // which later is used to create the actor and mapper.
-    const segmentationImageId = (
+    const segmentationImageIds = (
       labelMapData as LabelmapSegmentationDataStack
     ).imageIdReferenceMap.get(viewport.getCurrentImageId());
 
+    // Todo: fix this to support multiple derived imageIds
```
wayfarer3130, line 68: Can you revert the change and instead allow multiple segmentation UIDs to be associated?  That way they are cleanly separated.  With this change it becomes very hard to know where to do things.

packages/tools/src/tools/segmentation/BrushTool.ts
```diff
@@ -69,7 +69,7 @@ class BrushTool extends BaseTool {
   private _editData: {
     segmentsLocked: number[]; //
     segmentationRepresentationUID?: string;
-    imageIdReferenceMap?: Map<string, string>;
+    imageIdReferenceMap?: Map<string, Set<string>>;
```
wayfarer3130, line 72: It really isn't obvious to me how you would figure out which one is of interest.  I think they should be keyed by the segmentation UID, which would mean you get exactly one value, so that you would then undo this line.

packages/tools/src/tools/segmentation/BrushTool.ts
```diff
@@ -235,11 +233,9 @@ class BrushTool extends BaseTool {
       // and should throw an error or maybe simply just allow circle manipulation
       // and not sphere manipulation
       if (this.configuration.activeStrategy.includes('SPHERE')) {
```
wayfarer3130, line 235: Can you assign an error message on the activeStrategy object that can be thrown instead?  Eg, something like:
if( this.getStrategy().VOLUMETRIC_ERROR_MESSAGE ) {
throw new Error(this.getStrategy().VOLUMETRIC_ERROR_MESSAGE );

That way you don't need to know anything special about the strategy, just that it doesn't work unless volumetric.  I really hate including the string key here as it makes it very brittle on changes or additions.

packages/tools/src/tools/segmentation/strategies/fillCircle.ts
```diff
@@ -98,9 +95,9 @@ function createPointInEllipse(worldInfo: {
 
   const radius = Math.max(xRadius, yRadius, zRadius);
   if (
-    nearOrZero(xRadius, radius) &&
-    nearOrZero(yRadius, radius) &&
-    nearOrZero(zRadius, radius)
+    csUtils.isEqual(xRadius, radius) &&
```
wayfarer3130, line 98: Extract isEqual up on line 19 or so - this function is called enough times that you don't want to dereference the csUtils multiple times.

packages/tools/src/tools/segmentation/strategies/utils/getStrategyData.ts
```diff
@@ -40,7 +40,10 @@ function getStrategyData({ operationData, viewport }) {
     // of that actor at that moment so we have the imageData already
     const actor = viewport.getActor(segmentationRepresentationUID);
     segmentationImageData = actor.actor.getMapper().getInputData();
-    const currentSegmentationImageId = imageIdReferenceMap.get(currentImageId);
+    const currentSegmentationImageIds = imageIdReferenceMap.get(currentImageId);
+
+    // Todo: fix this to support multiple derived imageIds
+    const currentSegmentationImageId = [...currentSegmentationImageIds][0];
```
wayfarer3130, line 46: Leaving this code everywhere without having real implementations seems like a bad idea for putting this in core.  I'd rather see the old code here and then actually fix it once we understand what is actually going to happen.

packages/tools/src/types/LabelmapTypes.ts
```diff
@@ -42,13 +42,10 @@ export type LabelmapSegmentationDataVolume = {
...
-   * This is a Map from referenced imageId to the segmentation (Derived) imageId that
-   * has been used to create the derived labelmap.
-   * Todo: later I guess we can have it as Record<string, metadata> where metadata
-   * can contain a derived image arbitrary information, for use cases such that the labelmap is
-   * derived from another image that is irrelevant to the current viewport.
+   * This is a Map from referenced imageId to the segmentation (Derived) imageIds (can be
+   * multiple) that are associated with it.
    */
-  imageIdReferenceMap: Map<string, string>;
+  imageIdReferenceMap: Map<string, Set<string>>;
```
wayfarer3130, line 48: Ditto - remove this

wayfarer3130 (approved):
I'm ok with the changes as they are now - a couple of things could still be improved, but I think it is good enough to go as is.

PR #972, opened 2023-12-22
feat(tools): Add new CircleStartEndThresholdTool and pointsInVolume statistics for 3D annotations

Celian-abd:
### Context
We created a new tool based on the `RectangleROIStartEndThresholdTool` which is the `CircleROIStartEndThresholdTool`.
We added to both of this tools a new statistic called `pointsInVolume` which is litteraly the points inside the 3D annotation.

### Changes & Results
- Add new tool `CircleROIStartEndThresholdTool`
- Add new statistic for `RectangleROIStartEndThresholdTool` and `CircleROIStartEndThresholdTool` called `pointsInVolume`

### Testing
There is a new example in `tools/examples/CircleROIStartEndThreshold`

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  


packages/tools/src/tools/segmentation/RectangleROIStartEndThresholdTool.ts
```diff
@@ -264,6 +266,78 @@ class RectangleROIStartEndThresholdTool extends RectangleROITool {
...
+      const worldPos1 = data.handles.points[0];
+      const worldPos2 = data.handles.points[3];
+
+      // If image does not exists for the targetId, skip. This can be due
+      // to various reasons such as if the target was a volumeViewport, and
+      // the volumeViewport has been decached in the meantime.
+      if (!imageVolume) {
+        continue;
+      }
+
```
sedghi, line 287: This should go outside or maybe the first line

packages/tools/src/tools/segmentation/CircleROIStartEndThresholdTool.ts
```diff
@@ -0,0 +1,625 @@
...
+  _getStartSliceIndex(
+    imageVolume: Types.IImageVolume,
+    worldPos: Types.Point3,
+    spacingInNormal: number,
+    viewPlaneNormal: Types.Point3
+  ): number | undefined {
+    const numSlicesToPropagate = this.configuration.numSlicesToPropagate;
+
+    const numSlicesToPropagateFromStart = Math.round(numSlicesToPropagate / 2);
+    console.log(numSlicesToPropagateFromStart);
```
sedghi, line 541: remove console log

packages/tools/src/tools/segmentation/CircleROIStartEndThresholdTool.ts
```diff
@@ -0,0 +1,625 @@
...
+        viewUp: <Types.Point3>[...viewUp],
+        FrameOfReferenceUID,
+        referencedImageId,
+        volumeId,
+        spacingInNormal,
+        enabledElement,
+      },
+      data: {
+        label: '',
+        startSlice: newStartIndex,
```
sedghi, line 164: why in the `RectangleROIStartEndThreshold`, the startSlice is `startIndex` which is imageIdIndex, but here is different?

packages/tools/src/tools/segmentation/CircleROIStartEndThresholdTool.ts
```diff
@@ -0,0 +1,625 @@
...
+    imageVolume: Types.IImageVolume
+  ): void {
+    const { data, metadata } = annotation;
+    const { viewPlaneNormal, spacingInNormal } = metadata;
+    const { imageData } = imageVolume;
+    const { startSlice, endSlice } = data;
+    const { points } = data.handles;
+
+    const startIJK = transformWorldToIndex(imageData, points[0]);
+    console.log(startIJK);
```
sedghi, line 378: remove this and other console logs please

packages/tools/src/tools/segmentation/CircleROIStartEndThresholdTool.ts
sedghi: I believe I grasp the reason why, but just to confirm, this tool differs from the RectangleROITool in that it projects both before and after, whereas the rectangle only projects to after. Is this the intended functionality, or is it purely coincidental?

packages/tools/src/tools/segmentation/RectangleROIStartEndThresholdTool.ts
```diff
@@ -276,6 +350,7 @@ class RectangleROIStartEndThresholdTool extends RectangleROITool {
     // Since we are extending the RectangleROI class, we need to
     // bring the logic for handle to some cachedStats calculation
     this._computeProjectionPoints(annotation, imageVolume);
+    this._computePointsInsideVolume(annotation, imageVolume, enabledElement);
```
sedghi, line 353: Could you consider placing it under a configuration? This way, it will add extra steps only when necessary for the use case.

sedghi (commented):
see my comments, thanks

packages/tools/src/tools/segmentation/CircleROIStartEndThresholdTool.ts
salimkanoun: Yes it is intentional, when the physician is going to draw a 3D ROI he is likely to use the center slice to draw the ROI which should propagate upper and lower of the drawn ROI.

It's hard to imagine that the physician is going to to draw a ROI on a upperslice (in which the tumor is not visible if you don't want to miss slices) and expect the ROI to propagate only in lowerslice

packages/tools/src/tools/segmentation/RectangleROIStartEndThresholdTool.ts
```diff
@@ -264,6 +267,78 @@ class RectangleROIStartEndThresholdTool extends RectangleROITool {
     data.cachedStats.projectionPointsImageIds = projectionPointsImageIds;
   }
 
+  _computePointsInsideVolume(annotation, imageVolume, enabledElement) {
```
sedghi, line 270: can you please add a small doc on top of this function on what it actually does?

packages/tools/src/tools/segmentation/RectangleROIStartEndThresholdTool.ts
```diff
@@ -264,6 +267,79 @@ class RectangleROIStartEndThresholdTool extends RectangleROITool {
...
+          null,
+          boundsIJK
+        );
+
+        //@ts-ignore
+        pointsInsideVolume.push(pointsInShape);
+      }
+    }
+    data.cachedStats.pointsInVolume = pointsInsideVolume;
+  }
```
sedghi, line 341: So this will run at every frame (typicall 60fps) when the tool is modified, why do you need to re-calculate that everytime? I think mouse up is a better place for it. I also think maybe even this is a post processing step and you should have it in the app not here in cornerstone3D

Why do you need points in volume?

sedghi (commented):
see my comments, I"m back to open source yay

packages/tools/src/tools/segmentation/RectangleROIStartEndThresholdTool.ts
salimkanoun: The point in Volume is add like we add previously points for 2D ROI, these are usefull to implement in the app, histogramm based calculation, if you have all points value you can calculate histogram based quantification such as quantiles, entropy ... 
In all anotations, Cornerstone handle the basic calculations and there is even the problem that cornerstone trie to guess what it need to display (some SUV normalization logic is in cornerstone, which don't think it is a good partern).

Globally I agree that the rendering library tsuch as cornerstone should not handle calculation logic, I think it would be better to have a quantification class in which we would have a volume ID and an Annotation object, and this class will produce the results and will be called by the app.

However the implementation of annotation is coupled for now as it include the quantification and rendering of results, that's something we improved in this PR by abstracting the calculator and making the rendering modifiable by the app : https://github.com/cornerstonejs/cornerstone3D/pull/723

For this PR, yes the calculation should be done on mouse up, celian will change it. 

But then if you want to uncouple the annotation from the point calculation step it will need to redesgin all anotations to keep a common structure ... As the number of annotation tools is increasing it sounds to be a critical choice, if this change is not done now it will be more difficult to make this change later.

Globally i think the actual design is not perfect but suitable. (and hopfully i'm happy to not be the maintainer of such complicated project that is going to carry all medical imaging ecosystem ^^)

Celian-abd:
I made the change to apply _computePointsInsideVolume only on mouseUp but I had to override the _endCallback from both CircleROITool and RectangleROITool. Should we keep it like this or should we change the callbacks to protected. It would means that we need to change it for every tool as well.

sedghi (approved):
Looks great, thanks

PR #984, opened 2024-01-04
chore(test): Basic changes to have jest running tests

wayfarer3130:
### Context

Some changes to get jest tests running.  This will allow us to ensure tests run correctly in both a node and karma environment, and have a fast starting/IDE debuggable test environment.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  


packages/core/src/utilities/getViewportModality.ts
```diff
@@ -1,13 +1,12 @@
-import { IViewport } from '../types';
-import { StackViewport, VolumeViewport } from '../RenderingEngine';
+import type { IViewport, IStackViewport, IVolumeViewport } from '../types';
 import cache from '../cache';
 
 function getViewportModality(viewport: IViewport, volumeId?: string): string {
-  if (viewport instanceof StackViewport) {
-    return viewport.modality;
+  if ((viewport as IStackViewport).modality) {
```
sedghi, line 5: Later we might add .modality to the volume and this will break, can we properly fix this with some method? check the type?

sedghi (approved):

packages/core/src/utilities/getViewportModality.ts
wayfarer3130: Actually it will be fine if we add .modality to the volume as it will then return direct.  I prefer this implementation since it correctly handles other viewport types like WSI and Video, which all have .modality.

sedghi (approved):

Issue #888, opened 2023-11-14
[Bug] Volume renders incorrectly when webWorkerManager is initialized before rendering


sgielen:
### Describe the Bug

If `cornerstoneDICOMImageLoader.webWorkerManager.initialize` is called before rendering the first volume, the volume is rendered incorrectly.

### Steps to Reproduce

A reproduction scenario is available at: https://sjorsgielen.nl/cornerstone3d/webworkermanager-init.zip

Unpack this .zip file, run yarn install and yarn dev, then browse to the indicated port.

### The current behavior

The volume will be extremely bright (at least for me).
If `initWebWorkers` is set to `false` at the top of `src/main.ts`, it is rendered normally.

### The expected behavior

It does not matter whether the web workers are initialized ahead-of-time or not.

### OS

macOS 14.0

### Node version

20.8.1

### Browser

Safari 17.0

sedghi:
This relates to the same issue you're experiencing with other matters, specifically regarding wadouri, scaling, and metadata concerns.

Let's begin by addressing the scaling problem, and then we can tackle the other issues. Currently, I'm a bit short on time to prioritize this, but I'm more than happy to assist you with reviewing and providing guidance.

sgielen:
Thank you @sedghi! I am currently away for a few weeks, but when I get back, I would like to continue with these. With the scaling issue, do you mean #887? If so, I will start with that one when I am back.

sgielen:
I can't reproduce this anymore now that I'm on Cornerstone3D 2.0, so I believe this has been fixed since then. Probably since the prescaling was enabled by default on main.

PR #990, opened 2024-01-09
fix: Lockup in request pool when returning non promise result

wayfarer3130:
### Context

The wrong loader was being used in the URI loadImage call, and there was a bug in the request pool manager that failed to start requests when the request returned an empty result instead of a promise (eg short circuit)

### Changes & Results

Fix the request pool manager to ensure the counts are decremented even on no response
Fix the loadImage to use the original scheme loader (the options loader was never specified, so it wasn't relevant, and was slightly different in the newer version, so wasn't working)

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  

### DIFF
--- a/common/reviews/api/tools.api.md
+++ b/common/reviews/api/tools.api.md
@@ -3479,6 +3479,9 @@ function pointInEllipse(ellipse: any, pointLPS: any, inverts?: Inverts): boolean
 function pointInShapeCallback(imageData: vtkImageData | Types_2.CPUImageData, pointInShapeFn: ShapeFnCriteria, callback?: PointInShapeCallback, boundsIJK?: BoundsIJK_2): Array<PointInShape>;
 
 // @public (undocumented)
+function pointInSurroundingSphereCallback(imageData: vtkImageData, circlePoints: [Types_2.Point3, Types_2.Point3], callback: PointInShapeCallback, viewport?: Types_2.IVolumeViewport): void;
+
+// @public (undocumented)
 const pointsAreWithinCloseContourProximity: (p1: Types_2.Point2, p2: Types_2.Point2, closeContourProximity: number) => boolean;
 
 // @public (undocumented)
@@ -5361,6 +5364,7 @@ declare namespace utilities {
         getAnnotationNearPoint,
         getAnnotationNearPointOnEnabledElement,
         jumpToSlice,
+        pointInSurroundingSphereCallback,
         viewport,
         cine,
         clip_2 as clip,
--- a/packages/core/src/requestPool/requestPoolManager.ts
+++ b/packages/core/src/requestPool/requestPoolManager.ts
@@ -239,10 +239,17 @@ class RequestPoolManager {
         this.numRequests[type]++;
         this.awake = true;
 
-        requestDetails.requestFn()?.finally(() => {
+        const requestResult = requestDetails.requestFn();
+        if (requestResult?.finally) {
+          requestResult.finally(() => {
+            this.numRequests[type]--;
+            this.startAgain();
+          });
+        } else {
+          // Handle non-async request functions too - typically just short circuit ones
           this.numRequests[type]--;
-          this.startAgain();
-        });
+          i--;
+        }
       }
     }
 
--- a/packages/dicomImageLoader/src/imageLoader/wadouri/loadImage.ts
+++ b/packages/dicomImageLoader/src/imageLoader/wadouri/loadImage.ts
@@ -126,12 +126,7 @@ function loadImageFromDataSet(
         const pixelData = getPixelData(dataSet, frame);
         const transferSyntax = dataSet.string('x00020010');
 
-        imagePromise = createImage(
-          imageId,
-          pixelData,
-          transferSyntax,
-          options
-        );
+        imagePromise = createImage(imageId, pixelData, transferSyntax, options);
       } catch (error) {
         // Reject the error, and the dataSet
         reject({
@@ -177,13 +172,10 @@ function loadImage(
   const parsedImageId = parseImageId(imageId);
 
   options = Object.assign({}, options);
-  let loader = options.loader;
-
-  if (loader === undefined) {
-    loader = getLoaderForScheme(parsedImageId.scheme);
-  } else {
-    delete options.loader;
-  }
+  // The options might have a loader, but it is a loader into the cache, so not
+  // the scheme loader, which is separate
+  delete options.loader;
+  const schemeLoader = getLoaderForScheme(parsedImageId.scheme);
 
   // if the dataset for this url is already loaded, use it, in case of multiframe
   // images, we need to extract the frame pixelData from the dataset although the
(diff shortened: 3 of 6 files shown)


common/reviews/api/tools.api.md
```diff
@@ -5361,6 +5364,7 @@ declare namespace utilities {
         getAnnotationNearPoint,
         getAnnotationNearPointOnEnabledElement,
         jumpToSlice,
+        pointInSurroundingSphereCallback,
```
wayfarer3130, line 5367: This got removed somehow and I'm just adding it back as OHIF still uses it.

packages/core/src/requestPool/requestPoolManager.ts
```diff
@@ -239,10 +239,17 @@ class RequestPoolManager {
         this.numRequests[type]++;
         this.awake = true;
 
-        requestDetails.requestFn()?.finally(() => {
```
wayfarer3130, line 242: There was a leak on the numRequests pool size.

packages/core/src/requestPool/requestPoolManager.ts
```diff
@@ -239,10 +239,23 @@ class RequestPoolManager {
...
+          requestResult.finally(() => {
+            this.numRequests[type]--;
+            this.startAgain();
+          });
+        } else {
+          // Handle non-async request functions too - typically just short circuit ones
           this.numRequests[type]--;
-          this.startAgain();
-        });
+          i--;
```
sedghi, line 257: This is odd, you are decrementing the loop iterator. Are you sure about this?

wayfarer3130: Yes, because I don't want to call startAgain, which would run the entire loop again, but I want to add anything the given number of requests to the request pool.  The send request really on throws or returns undefined when it is an empty request, so for those the request can simply be skipped and we "redo" that index - except as another request.

sedghi: is this part of this fix or is unrelated and potential bug in future, since I don't think the approach is fine. if it is not part of this fix let's worry about that later

sedghi (approved):
Thanks a lot!

sgielen:
@sedghi @wayfarer3130 this PR removes the `wadouri.loadImage()` `loader` option, which was the only way I knew how to load a Cornerstone image from an ArrayBuffer. Now, the `loader` option is ignored. What is now the preferred way to load an image from an ArrayBuffer?

sedghi:
@sgielen So how do we test it?

sgielen:
> @sgielen So how do we test it?

I think we should look at the design first and see where the root issue stems from.

[By design](https://www.cornerstonejs.org/docs/concepts/cornerstone-core/imageloader/), cornerstone3d ImageLoaders receive an ID, start some download, import those bytes into an ImageObject, and return it. The parsing of those download bytes (e.g. DICOM, NIfTI format) is considered to be the ImageLoader responsibility as well, since every format is different. However, Cornerstone still needs generic metadata from the image, so ImageLoaders also register MetadataProviders which can provide the responses to generic questions of many types (width/height, pixel/voxel sizes, LUTs / window settings etc). So, the ImageLoader not only loads pixel data, but also converts metadata into a format recognized by the rest of Cornerstone.

So the two (current) purposes of an ImageLoader are to return image pixels, and to keep its metadata in its cache for as long as the image can be handled.

Now, my particular issue is that this smartness is included in the Cornerstone DICOM image loader, but in a way that makes it difficult to use in another ImageLoader. The metadata cache is included in the WADOURI implementation runtime as a global variable, and the metadata logic code is tightly coupled with this cache variable. This means either my DICOM datasets need to be in its cache, or I need to copy all of the DICOM -> Cornerstone metadata logic into my own project to make it use my own cache. I can't just call any of the parsing functions without depending on the WADOURI cache as well, so my ImageLoader is either too tightly coupled or does not reuse code properly.

So far, I chose to use the WADOURI cache even if I don't use the WADOURI loader, so I don't need to replicate all the code. This is why, in my image loader, I wrap the WADOURI loader, overwriting its `loader` option. Now that it's gone, I probably can work around this by using other parts of the DICOM image loader, such as calling `dataSetCacheManager.load()` myself, but it does still feel a bit wonky and not using the loader as intended. Perhaps this is a good time to look at it from some distance and see what we could improve.

I believe that ideally, all of the DICOM functions in the dicomImageLoader would be separate, testable chunks, to be used by WADOURI, perhaps WADORS and other image loaders that eventually work with DICOM bytes.

If you agree, perhaps I can file an issue for it and I can propose a refactoring strategy.

sedghi:
@sgielen Sorry I meant do you have steps or better code that you can share to reproduce

sgielen:
> @sgielen Sorry I meant do you have steps or better code that you can share to reproduce

Oh, yes, certainly! You can find an example e.g. in #888 (https://sjorsgielen.nl/cornerstone3d/webworkermanager-init.zip). The code in the .zip file downloads DICOMs using `fetch`, as an example of an alternative non-WADOURI method to get DICOMs, then tries to load them using wadouri. It's intended to be a testcase for another issue, but if you set `initWebWorkers=false` at the top of the file then it works well in Cornerstone 1.21.2. Since the `loader` option to WADOURI is gone after this PR, on current latest release / main branch, it will perform a WADOURI request instead of using `fetch`, and that request will of course fail.

So, if the `loader` option is brought back it should work well again on the newest version, but if the `loader` option should not be brought back, we could also use this as a starting point to design a better way to load DICOMs in a third-party ImageLoader.

sgielen:
@sedghi any thoughts as to bringing back `wadouri.loadImage` or implementing an alternative? :)

jlopes90:
@sedghi 

I'm using `options.loader` and now it's removed. Can I create a pull request to return `options.loader`?

Because I'm using `options.loader`, for example:
It's very similar to `wadouri.loadFileRequest`, but I created store manager so I can see the private files inside.

```javascript
/**
 * Load Image From Dicom File
 * @param {string} imageId
 * @returns {object}
 */
function loadImageFromDicomFile(imageId) {

    // Function Loader
    const fnLoader = function (imageId) {
        const index = imageId.replace('dicomfile:', '');

        // Get file
        const file = exStore.manager.get(index); // it's like fileManager

        // Promise
        const promise = new Promise(function (resolve, reject) {

            //
            const fileReader = new FileReader();

            //
            fileReader.onload = function (evt) {
                resolve(evt.target.result);
            };

            //
            fileReader.onerror = function (err) {
                reject(err);
            };

            //
            fileReader.readAsArrayBuffer(file);
        });

        // Return
        return promise;
    };

    //
    return wadouri.loadImage(imageId, {
        loader: fnLoader
    });
}
```

jlopes90:
My opinion in the future, `fileManager `should move to `package/core` because it's not just `dicomfile`, there will also be `imagefile`, `niftifile`, `videofile`, etc.

sgielen:
I have two PRs open where this topic is being progressed:

- https://github.com/cornerstonejs/cornerstone3D/pull/1508
- https://github.com/cornerstonejs/cornerstone3D/pull/1509

PR #1003, opened 2024-01-16
feat(interpolation): Contour segmentation interpolation for freehand and SplineROI

wayfarer3130:
This PR is heavily based on the contour interpolation changes https://github.com/cornerstonejs/cornerstone3D/pull/966/files#diff-b23b56a65802245bafe668e7a0df8d734795f6af2c4326782464014f00c2d4c5
The original PR was for [FlyWheel.io](https://flywheel.io/) by @Belbin-GK 

### Context
* Creates a contour tool that allows to create interpolated contour ROIs between contour drawn slices by [FlyWheel.io](https://flywheel.io/)
* This allows to quickly create annotation in multiple slices for a tissue. Upcoming sculpt tool along with this feature will make editing the interpolated shapes easier.

### Changes & Results
* Add new interpolation for segmentation contours using polyline
* Introduced 'autoGenerated' property in an annotation to indicate whether the annotation was manually created or automatically generated.
* Added events for contour interpolation completion.
* Allows separate styling for autoGenerated results
 
### Testing
* `yarn example interpolationContourSegmentation`.
* Follow the instructions given in example file to generate interpolated contour ROI.
* There is a Interpolation test file to run tests on the tool.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  


packages/adapters/src/adapters/Cornerstone/ContourRoi.js
```diff
@@ -0,0 +1,82 @@
+import { utilities } from "dcmjs";
```
wayfarer3130, line 1: @sedghi - should this file just be deleted for now, until we figure out how to serialize all the contour annotations to SR?

packages/tools/examples/videoContourSegmentation/index.ts
```diff
@@ -52,6 +52,44 @@ const segmentIndexes = [1, 2, 3, 4, 5];
 const segmentVisibilityMap = new Map();
 let activeSegmentIndex = 1;
 
+const configuredTools = new Map<string, any>();
```
wayfarer3130, line 55: @sedghi - I think this should be our standard way of defining sets of tools to include in a list - putting them into a map
allows iterating over them in the original organization, and getting the appropriate configuration out.

packages/tools/src/enums/AnnotationStyleStates.ts
```diff
@@ -11,12 +11,15 @@
...
  *   The annotation has been locked;
+ * AutoGenerated:
+ *   The annotation has been auto generated;
  */
 enum AnnotationStyleStates {
   Default = '',
   Highlighted = 'Highlighted',
   Selected = 'Selected',
   Locked = 'Locked',
+  AutoGenerated = 'AutoGenerated',
```
wayfarer3130, line 22: I'm not sure we end up using this, but I think it is a worthwhile change, so I'm going to leave it in.

packages/tools/src/enums/ChangeTypes.ts
```diff
@@ -0,0 +1,32 @@
+/**
+ * ChangeTypes defines the types of changes occurring on annotation and
+ * segmentation data.
+ */
+enum ChangeTypes {
```
wayfarer3130, line 5: This is enormously helpful for all sorts of detection of types of changes.  I think it is a good overall change.  The question is where and how to document it?

packages/tools/src/init.ts
```diff
@@ -74,6 +75,8 @@ function _addCornerstoneEventListeners(): void {
 
   eventTarget.addEventListener(elementEnabledEvent, addEnabledElement);
   eventTarget.addEventListener(elementDisabledEvent, removeEnabledElement);
+  console.log('Enabling cornerstone event listeners');
+  annotationInterpolationEventDispatcher.enable();
```
wayfarer3130, line 79: Putting the dispatcher here means that it gets registered/unregistered once per lifetime of the CS tools, rather than ending up with multiple copies of the listener.

packages/tools/src/tools/annotation/PlanarFreehandContourSegmentationTool.ts
```diff
@@ -10,6 +10,7 @@ class PlanarFreehandContourSegmentationTool extends PlanarFreehandROITool {
       {
         configuration: {
           calculateStats: false,
+          allowOpenContours: false,
```
wayfarer3130, line 13: This causes the contours to be drawn closed even during initial creation.

packages/tools/src/tools/annotation/SplineROITool.ts
```diff
@@ -316,11 +323,21 @@ class SplineROITool extends ContourSegmentationBaseTool {
...
+        renderingEngineId,
+        changeType: ChangeTypes.Completed,
       };
 
       triggerEvent(eventTarget, eventType, eventDetail);
+    } else {
+      this.triggerAnnotationModified(
+        annotation,
+        enabledElement,
+        ChangeTypes.HandlesUpdated
```
wayfarer3130, line 339: The handles updated change is required to prevent updating the interpolated slices too often - it will get fired when the mouse is released rather than on every little move.

packages/tools/src/tools/annotation/SplineROITool.ts
```diff
@@ -724,25 +743,60 @@ class SplineROITool extends ContourSegmentationBaseTool {
...
+   * Creates new interpolated handles for the spline control given the
+   * polyline data.  This allows creating the spline from polyline data
+   * directly.
+   */
+  protected createInterpolatedSplineControl(annotation) {
+    if (annotation.data.handles.points?.length) {
+      // The interpolation itself created the handles
+      return;
+    }
+    const { polyline } = annotation.data.contour;
```
wayfarer3130, line 756: This code isn't actually active - I tested it, and it works, but it is the code in teh interpolation routine that is being used.  However, this one is required in case the spline comes from some other system which does not calculate handles.

packages/tools/src/tools/base/ContourSegmentationBaseTool.ts
```diff
@@ -207,7 +212,12 @@ abstract class ContourSegmentationBaseTool extends ContourBaseTool {
     let lineOpacity = 1;
     let fillOpacity = 0;
 
-    if (active) {
+    if (autoGenerated) {
```
wayfarer3130, line 215: I dislike getting these in this way - would prefer to have the style getter algorithms directly retrieve the data, but I don't want to make that change in this PR.

packages/tools/src/tools/displayTools/Contour/contourConfig.ts
```diff
@@ -2,13 +2,18 @@ import { ContourConfig } from '../../../types/ContourTypes';
 
 const defaultContourConfig: ContourConfig = {
   renderOutline: true,
+  outlineWidthAutoGenerated: 3,
```
wayfarer3130, line 5: As requested, sets the defaults to the ones in the tools (mostly), with some auto generated enhancements.

packages/tools/src/types/EventTypes.ts
```diff
@@ -116,8 +118,13 @@ type AnnotationAddedEventDetail = {
  * drawing on the viewport.
  */
 type AnnotationCompletedEventDetail = {
```
wayfarer3130, line 120: Would like to drop the annotation events to just changed and removed, with the type of change being included in the change type, but that requires more thinking about when it should occur.

packages/tools/src/utilities/contours/PointsArray.ts
wayfarer3130, line 1: I'm not using this yet - the intent was to start optimizing the contour performance by have a class that deals with converting between various types - I might still complete this for use in the interpolation classes.

packages/tools/src/utilities/contours/interpolation/interpolate.ts
wayfarer3130, line 1: The code here is basically what was submitted.  It perhaps should get broken up into separate pieces, and maybe refactored to work with the points array class for better performance.

packages/tools/src/utilities/polyData/utils.ts
```diff
@@ -1,14 +1,21 @@
...
- * @param points array of number, each point defined by three consecutive numbers
- * @param idx index of the point to retrieve
+ * @param points - array of number, each point defined by three consecutive numbers
+ * @param idx - index of the point to retrieve
  * @returns
  */
-export function getPoint(points, idx) {
-  if (idx < points.length / 3) {
-    return [points[idx * 3], points[idx * 3 + 1], points[idx * 3 + 2]];
+export function getPoint(points, idx): Types.Point3 {
```
wayfarer3130, line 11: This version is more performant than the previous one, but is compatible.

packages/adapters/src/adapters/Cornerstone/ContourRoi.js
sedghi: I thought this is working, you mean you haven't tested it? Yes please delete if you think it is hard to test for now

packages/tools/examples/interpolationContourSegmentation/index.ts
```diff
@@ -0,0 +1,362 @@
...
+  labelText: 'Segment Index',
+  options: { values: segmentIndexes, defaultValue: segmentIndexes[0] },
+  onSelectedValueChange: (nameAsStringOrNumber) => {
+    updateActiveSegmentIndex(Number(nameAsStringOrNumber));
+  },
+});
+
+const toolsNames = [
+  ...interpolationTools.keys(),
+  PlanarFreehandROITool.toolName,
```
sedghi, line 110: Not sure why we need vanilla ROI tools. This example should focus only on segmentation and interpolation IMO

packages/tools/examples/interpolationContourSegmentation/index.ts
```diff
@@ -0,0 +1,362 @@
...
+  // Make the segmentation created as the active one
+  segmentation.activeSegmentation.setActiveSegmentationRepresentation(
+    toolGroupId,
+    segmentationRepresentationUID
+  );
+
+  updateActiveSegmentIndex(1);
+}
+
+function getViewportDataBasedOnAnnotation(annotationData) {
```
sedghi, line 342: don't see used anywhere

packages/tools/examples/interpolationContourSegmentation/index.ts
```diff
@@ -0,0 +1,362 @@
...
+
+  // Store the segmentation representation that was just created
+  [segmentationRepresentationUID] = segmentationRepresentationUIDs;
+
+  // Make the segmentation created as the active one
+  segmentation.activeSegmentation.setActiveSegmentationRepresentation(
+    toolGroupId,
+    segmentationRepresentationUID
+  );
+
```
sedghi, line 338: can you remind me why we should run this? I bet if we don't it would work too

packages/tools/src/utilities/triggerAnnotationRenderForViewportIds.ts
```diff
@@ -10,7 +10,12 @@ export function triggerAnnotationRenderForViewportIds(
   }
 
   viewportIdsToRender.forEach((viewportId) => {
-    const { element } = renderingEngine.getViewport(viewportId);
+    const viewport = renderingEngine.getViewport(viewportId);
+    if (!viewport) {
+      // Happens on shutdown sometimes
+      return;
```
sedghi, line 16: a warn at least?

packages/tools/src/tools/annotation/PlanarFreehandContourSegmentationTool.ts
sedghi: can you handle others as well. Also I'm confused since there is another configuration option called `closed`, what is that?

packages/tools/src/tools/annotation/PlanarFreehandROITool.ts
```diff
@@ -195,19 +195,35 @@ class PlanarFreehandROITool extends ContourSegmentationBaseTool {
...
           knotsRatioPercentageOnEdit: 40,
         },
+        /**
+         * Interpolation is the creation of new segmentations in between the
+         * existing segmentations/indices
+         */
+        interpolation: {
+          enabled: false,
+          // Callback to be made after interpolation is completed
+          postProcess: null,
```
sedghi, line 225: I prefer `onInterpolationComplete` instead of postProcess

packages/tools/src/tools/annotation/PlanarFreehandROITool.ts
```diff
@@ -195,19 +195,35 @@ class PlanarFreehandROITool extends ContourSegmentationBaseTool {
         // The proximity at which we fallback to the simplest grabbing logic for
         // determining what index of the contour to start editing.
         checkCanvasEditFallbackProximity: 6,
+        // For closed contours, make them clockwise
```
sedghi, line 198: requires more doc, on why it is preferable to make clockwise

packages/tools/src/types/EventTypes.ts
```diff
@@ -116,8 +118,13 @@ type AnnotationAddedEventDetail = {
...
+  /** unique id of the viewport */
+  viewportId?: string;
+  /** unique id of the rendering engine */
+  renderingEngineId?: string;
```
sedghi, line 124: actually we should go the other way around and remove the viewportId, and renderingEngineId from the other ones. I really hated that annotationAdded and modified are emitting viewport/renderingEngineId

packages/tools/src/utilities/contours/interpolation/createInterpolatedToolData.ts
```diff
@@ -0,0 +1,58 @@
...
+ * data type.
+ * Note that this object takes ownership of the polyline and handlePoints data, that is,
+ * directly assigns them internally to the result.
+ *
+ * @param polyline - data for the polyline, owned hereafter by the annotation
+ * @param handlePoints - data for the edit handles, if any, owned hereafter by the annotation
+ * @param referencedToolData
+
+ */
+export default function createInterpolatedToolData(
```
sedghi, line 15: It should contain contour or something to shows it is not lengthToolData, you know what I mean?

packages/tools/src/utilities/polyData/utils.ts
sedghi: please type the inputs

packages/tools/src/utilities/segmentation/InterpolationManager/InterpolationManager.ts
```diff
@@ -0,0 +1,173 @@
...
+      annotation,
+      interpolationUID: annotation.interpolationUID,
+    };
+    const isInitializeLabel = !annotation.interpolationUID;
+    // If any update, triggered on an annotation, then it will be treated as non-autogenerated.
+    annotation.autoGenerated = false;
+    if (!isInitializeLabel) {
+      updateRelatedAnnotations(viewportData, true);
+    }
+    if (!annotation.interpolationUID) {
```
sedghi, line 82: I prefer 

if (annotatoin.interpolationUID){
  return
}

// rest of logic

to reduce indentation

sedghi (commented):
Makes a lot of sense, thanks.
One major remaining step is how to finalize with shortcuts, for instance, we should have an action on key (e.g., Enter) to get the active segment index in the current slice and if autogenerated, make it finalized. 

Another requirement is to finalize the segment interpolations (all of them) and someone can link it to a hotkey themselves. Or it can be again an action on the tool as well

Thanks

packages/tools/examples/interpolationContourSegmentation/index.ts
wayfarer3130: Because sometimes one needs to compare the vanilla tool with the interpolated one.  Also, it allows showing additional contours on the same image as they have different behaviour.

packages/tools/src/types/EventTypes.ts
wayfarer3130: How do you want that implemented then?  The current interpolation code depends on having those values.  Is this a change we can make right now given that there might be others depending on having those values?

packages/tools/examples/interpolationContourSegmentation/index.ts
```diff
@@ -0,0 +1,337 @@
...
+    toolGroup.setToolPassive(selectedToolName);
+
+    selectedToolName = <string>newSelectedToolName;
+  },
+});
+
+addButtonToToolbar({
+  title: 'Accept AutoGenerated',
+  onClick: () => {
+    cornerstoneTools.utilities.contours.interpolation.InterpolationManager.acceptAutoGenerated(
```
sedghi, line 142: This should be a much shallower utility something like `cstools.utilities.contours.acceptAutogeneratedInterpolations` and that one can call this dee

packages/tools/examples/interpolationContourSegmentation/index.ts
```diff
@@ -0,0 +1,337 @@
...
+      frameOfReferenceUID
+    );
+    getRenderingEngine(renderingEngineId).render();
```
sedghi, line 145: I prefer using the triggerAnnotationRenderForViewports function to display the purpose. Although it essentially serves the same purpose, it helps readers understand and keeps things tidy.

sedghi: Also I see it is accepting a FOR, which is unusual, this should accept a toolGroup and optional object containing segmentIndex, segmentationId

packages/tools/src/tools/annotation/PlanarFreehandROITool.ts
```diff
@@ -195,19 +195,38 @@ class PlanarFreehandROITool extends ContourSegmentationBaseTool {
...
+          smoothOnEdit: false, // used for edit only
           knotsRatioPercentageOnAdd: 40,
           knotsRatioPercentageOnEdit: 40,
         },
+        /**
+         * Interpolation is the creation of new segmentations in between the
+         * existing segmentations/indices
+         */
+        interpolation: {
+          enabled: false,
```
sedghi, line 225: Could you please explain that in addition to the `enabled` option, it doesn't really make sense to interpolate between two ROIs? If a user needs to do that, they are essentially referring to a segment and segmentation. In such cases, it would be more appropriate for them to create a contourSegmentation and utilize segmentation tools.

packages/tools/src/utilities/planar/filterAnnotationsForDisplay.ts
```diff
@@ -42,7 +42,10 @@ export default function filterAnnotationsForDisplay(
         return false;
       }
 
-      const imageId = annotation.metadata.referencedImageId;
+      const imageId = annotation.metadata.referencedImageId?.replace(
```
sedghi, line 45: I didn't expect this, why?

packages/tools/src/tools/annotation/PlanarFreehandContourSegmentationTool.ts
```diff
@@ -10,6 +10,7 @@ class PlanarFreehandContourSegmentationTool extends PlanarFreehandROITool {
...
         },
```
sedghi, line 14: still not sure what is allowOpenContours compared to closed status

packages/tools/src/types/InterpolationTypes.ts
```diff
@@ -0,0 +1,18 @@
...
+  annotation: InterpolationROIAnnotation;
+  /** The type of event, whether initializing the label or updating it. */
+  interpolationUID: string;
+  /** unique id of the viewport */
+  viewport: Types.IViewport;
+  sliceData: Types.ImageSliceData;
+};
+
+export type ImageInterpolationData = {
+  sliceIndex: number;
```
sedghi, line 16: since the Types.ImageSliceData is 

```
type ImageSliceData = {
  numberOfSlices: number;
  imageIndex: number;
};
```

I prefer imageIndex instead of sliceIndex just to reduce the number of concepts for people to know

packages/tools/src/types/InterpolationTypes.ts
```diff
@@ -0,0 +1,18 @@
...
+  annotations?: Annotation[];
```
sedghi, line 17: Shouldn't this be InterpolationROIAnnotation[]?

packages/tools/src/utilities/getMatchingViewport.ts
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * Finds a matching viewport in terms of the orientation of the annotation data
+ * and the frame of reference.  This doesn't mean the annotation IS being displayed
+ * in the viewport, just that it could be by navigating the slice, and/or pan/zoom,
+ * without changing the orientation.
+ *
+ * @param annotation - to find a viewport that it could display in
+ * @returns The viewport to display in
+ */
+export default function getMatchingViewport(annotation: Annotation) {
```
sedghi, line 15: cornerstoneTools.utilities.getMatchingViewport is too general, can we say `getMatchingViewportForAnnotation` or `getViewportForAnnotation` or `getAnnotationViewport`.

packages/tools/src/utilities/contours/reverseIfAntiClockwise.ts
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * @param points - The points array.
+ * @returns The contour, corrected to be clockwise if appropriate.
+ */
+export default function reverseIfAntiClockwise(points: Types.Point2[]) {
+  const length = points.length;
+  if (!length) {
+    return points;
+  }
+  const xMean = points.map((point) => point[0]).reduce(getSumReducer) / length;
```
sedghi, line 26: for loop is much more performant than map and reduce 

```
let xSum = 0;
for (const point of points) {
  xSum += point[0];
}
const xMean = xSum / length;
```

packages/tools/src/utilities/contours/PointsArray.ts
```diff
@@ -0,0 +1,151 @@
...
+      (point) => point,
+      () => vec2.create() as Types.Point2
+    );
+  }
+}
+
+/**
+ * A version of this based on Types.Point3 and vec3
+ */
+export class PointsArray3 extends PointsArray<Types.Point3> {
```
sedghi, line 110: Why isn't this one called PolyDataPoints3? I noticed that there's PolyDataPoints2 above it. Additionally, we should consider removing these two derived classes. This file has become quite large.

packages/tools/src/utilities/contours/PointsArray.ts
```diff
@@ -0,0 +1,151 @@
...
+ */
+export abstract class PointsArray<T> {
+  data: Float32Array;
+  dimensions = 3;
+  public length = 0;
+
+  constructor(configuration: PolyDataPointConfiguration = {}) {
+    const { initialSize = 1024, dimensions = 3 } = configuration;
+    this.data = new Float32Array(initialSize * dimensions);
+    this.dimensions = dimensions;
```
sedghi, line 24: shouldn't you update the legnth here as well?

packages/tools/src/utilities/contours/PointsArray.ts
```diff
@@ -0,0 +1,151 @@
...
+    }
+  }
+
+  abstract getPoint(index: number, point: T): T;
+
+  /**
+   * Reverse the points in place
+   */
+  public reverse() {
+    const midLength = Math.floor(length / 2);
```
sedghi, line 39: this.length

packages/tools/src/utilities/contours/PointsArray.ts
```diff
@@ -0,0 +1,151 @@
...
+  protected map(f: (value, index: number) => T, factory: (index: number) => T) {
+    const mapData = [];
+    for (let i = 0; i < this.length; i++) {
+      mapData.push(f(this.getPoint(i, factory(i)), i));
+    }
+    return mapData;
+  }
+
+  /** Create an PointsArray3 from the x,y,z individual arrays */
+  public static fromXYZ({ x, y, z }): PointsArray3 {
```
sedghi, line 61: { x, y, z }: { x: number[]; y: number[]; z: number[] }

packages/tools/src/utilities/contours/PointsArray.ts
```diff
@@ -0,0 +1,151 @@
...
+    }
+    array.length = x.length;
+    return array;
+  }
+}
+
+/**
+ * A version of this based on Types.Point2 and vec2
+ */
+export class PolyDataPoints2 extends PointsArray<Types.Point2> {
```
sedghi, line 77: please add some docs above these two classes on what are the use cases

packages/tools/src/utilities/segmentation/InterpolationManager/updateRelatedAnnotations.ts
```diff
@@ -0,0 +1,30 @@
...
+ * @returns null
+ */
+export default function updateRelatedAnnotations(
+  viewportData: InterpolationViewportData,
+  isLabelUpdate = true
+) {
+  const { annotation } = viewportData;
+
+  if (isLabelUpdate) {
+    console.log('isLabelUpdate');
```
sedghi, line 19: remove console log

Also I'm not sure what isLabelUpdate mean, I know flywheel was using it to create a "Segment" but we have moved away with that construct to a more sophisticated contourSegmentation

packages/tools/src/utilities/segmentation/InterpolationManager/updateRelatedAnnotations.ts
```diff
@@ -0,0 +1,30 @@
...
+    delete annotation.interpolationUID;
+    deleteRelatedAnnotations(viewportData);
+  } else {
+    console.log(
```
sedghi, line 23: remove log

packages/tools/src/utilities/segmentation/InterpolationManager/updateRelatedAnnotations.ts
```diff
@@ -0,0 +1,30 @@
...
+import deleteRelatedAnnotations from './deleteRelatedAnnotations';
+
+/**
+ * updatedRelatedContours - Update the same roi contour id contours on modifying
+ * one of them .
+ *
+ * @param eventData - Object.
+ * @returns null
+ */
+export default function updateRelatedAnnotations(
```
sedghi, line 12: update docs please

packages/tools/src/utilities/contours/interpolation/getInterpolationDataCollection.ts
```diff
@@ -0,0 +1,34 @@
+import getInterpolationData from './getInterpolationData';
+import type {
+  InterpolationViewportData,
+  ImageInterpolationData,
+} from '../../../types/InterpolationTypes';
+
+/**
+ * getInterpolationDataCollection - Generates a collection of the 2D
+ * polygons in difference slices that make up the interpolated annotations.
```
sedghi, line 9: does it generate or get? documentation and function name does not match

packages/tools/src/utilities/contours/interpolation/getInterpolationDataCollection.ts
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * @param eventData - Object, cornerstone viewport.
```
sedghi, line 11: wrong param docs

packages/tools/src/utilities/contours/interpolation/getInterpolationData.ts
```diff
@@ -0,0 +1,70 @@
...
+ */
+
+export default function getInterpolationData(
+  viewportData: InterpolationViewportData,
+  filterParams = [],
+  onlyAnnotationImage = false
+) {
+  const { viewport, sliceData, annotation } = viewportData;
+  const interpolationDatas: ImageInterpolationData[] = [];
+  const annotations = annotationStateManagement.state.getAnnotations(
```
sedghi, line 25: how about the getAnnotations import instead sicne you only need that

packages/tools/src/utilities/contours/interpolation/findInterpolationList.ts
```diff
@@ -0,0 +1,216 @@
...
+ * @param toolData - object, The tool data of the roi contour.
+ * @returns object, An object containing the ROIContourData and the
+ * interpolationList.
+ */
+
+function generateInterpolationData(
+  toolData,
+  viewportData: InterpolationViewportData
+) {
+  console.log(
```
sedghi, line 18: remove logs please

packages/tools/src/utilities/contours/interpolation/findInterpolationList.ts
```diff
@@ -0,0 +1,216 @@
...
+
+  for (let i = interpolationData.length - 1; i >= 0; i--) {
+    if (interpolationData[i].annotations) {
+      extent.push(i);
+      break;
+    }
+  }
+
+  return extent;
+}
```
sedghi, line 83: function name, docs does not match

packages/tools/src/utilities/contours/interpolation/findInterpolationList.ts
```diff
@@ -0,0 +1,216 @@
...
+/**
+ * _getSlicePositionOfToolData - Finds the slice that was edited.
+ *
+ * @param ROIContourData - description
+ * @param polygonUid - description
+ * @returns description
+ */
+function _getSlicePositionOfToolData(interpolationData, annotationUID) {
+  for (let i = 0; i < interpolationData.length; i++) {
+    if (interpolationData[i].annotations) {
```
sedghi, line 94: can we do something like 

const sliceAnnotations = interpolationData[i].annotations

sicne you use it twice

packages/tools/src/utilities/contours/interpolation/findInterpolationList.ts
```diff
@@ -0,0 +1,216 @@
...
+ * @param sliceIndex - Number, The slice index.
+ * @param extent - Number[], The extent of slice occupancy of the
+ *                                  ROIContour.
+ * @param ROIContourData - object[],  Data for the slice location of contours
+ *                                  for the ROIContour.
+ * @param interpolationList - object[],  The list of contours to be interpolated.
+ * @returns null
+ */
+function _appendInterpolationList(contourPair, interpolationList, i) {
+  if (!interpolationList[contourPair[0]]) {
```
sedghi, line 138: some refactoring again

const startIndex = contourPair[0]

packages/tools/examples/interpolationContourSegmentation/index.ts
wayfarer3130: It accepts a annotation group selector, the same as every other annotation selector/change object - that can be a viewport or an FOR.
Then, the second parameter allows specifying the tool names to apply to (really only a single name is expected here, if you want to do it for exactly one tool, but in general a list works), and a segmentIndex and segmentationId

packages/tools/src/utilities/planar/filterAnnotationsForDisplay.ts
wayfarer3130: Added a comment, but basically we should determine whether or not to display given the targetId, as that is actually specifying where it gets displayed.  I can remove this for now if you prefer, and simply remove the imageId prefix in the interpolation code for now, and we can discuss how to really handle referencedImageId/targetId differences.

packages/tools/src/tools/annotation/PlanarFreehandContourSegmentationTool.ts
wayfarer3130: Add docs there.

packages/tools/src/types/InterpolationTypes.ts
wayfarer3130: The problem is that it isn't imageIndex, nor is numberOfSlices the same as the number of images.  This has to work for non-acquisition plane volumes as well as video (which has a a frame number and a frame time, but those aren't quite image indices)

packages/tools/src/types/InterpolationTypes.ts
wayfarer3130: Yes, most likely it should but the casting isn't quite right to do that now.  We can talk about whether we want to fix that.

packages/tools/src/utilities/contours/PointsArray.ts
wayfarer3130: No, there isn't any data in it yet.

packages/tools/src/utilities/contours/interpolation/interpolate.ts
```diff
@@ -0,0 +1,686 @@
+import { triggerEvent } from '@cornerstonejs/core';
+import type { Types } from '@cornerstonejs/core';
+import { vec3 } from 'gl-matrix';
+
+import createPolylineToolData from './createPolylineToolData';
+import generateInterpolationData from './findInterpolationList';
+import type { InterpolationViewportData } from '../../../types/InterpolationTypes';
+import { InterpolationROIAnnotation } from '../../../types/ToolSpecificAnnotationTypes';
+import { AnnotationInterpolationCompletedEventDetail } from '../../../types/EventTypes';
```
sedghi, line 9: import types for these two as well

packages/tools/src/utilities/segmentation/InterpolationManager/deleteRelatedAnnotations.ts
```diff
@@ -0,0 +1,120 @@
...
+  const referencedSliceIndex = annotation.metadata.referencedSliceIndex;
+  let nextNonInterpolatedIndex = null;
+  for (
+    let i = referencedSliceIndex + 1, j = referencedSliceIndex - 1;
+    i < interpolationAnnotations.length, j >= 0;
+    i++, j--
+  ) {
+    let interpolationAnnIndex = interpolationAnnotations.findIndex(
+      (ann) => ann.metadata.referencedSliceIndex === i
+    );
```
sedghi, line 34: Let's consider creating a precached map for these slice indices. I encountered a similar performance issue in the segmentation import, and addressing it with a map outside of the for loop resulted in a significant 3x performance improvement.

packages/tools/src/utilities/segmentation/InterpolationManager/deleteRelatedAnnotations.ts
```diff
@@ -0,0 +1,120 @@
...
+    let interpolationAnn = interpolationAnnotations[interpolationAnnIndex];
+    if (interpolationAnn && !interpolationAnn.autoGenerated) {
+      nextNonInterpolatedIndex = interpolationAnnIndex;
+      break;
+    }
+
+    interpolationAnnIndex = interpolationAnnotations.findIndex(
```
sedghi, line 41: same

packages/tools/src/utilities/segmentation/InterpolationManager/deleteRelatedAnnotations.ts
```diff
@@ -0,0 +1,120 @@
...
+  viewportData,
+  direction = 1
+) {
+  const removedAnnotations = [];
+  for (
+    let i = referencedSliceIndex + direction;
+    i < interpolationAnnotations.length && i > -1;
+    i += direction
+  ) {
+    const interpolationAnnIndex = interpolationAnnotations.findIndex(
```
sedghi, line 89: same

packages/tools/src/utilities/segmentation/InterpolationManager/InterpolationManager.ts
```diff
@@ -0,0 +1,209 @@
...
+import ChangeTypes from '../../../enums/ChangeTypes';
+import getMatchingViewport from '../../getMatchingViewport';
+
+const { uuidv4 } = csUtils;
+
+export default class InterpolationManager {
+  static toolNames = [];
+
+  static addTool(toolName: string) {
+    if (-1 === this.toolNames.indexOf(toolName)) {
```
sedghi, line 24: this.toolNames.includes(toolName)

packages/tools/src/utilities/segmentation/InterpolationManager/InterpolationManager.ts
```diff
@@ -0,0 +1,209 @@
...
+  }
+
+  static handleAnnotationCompleted = (evt: AnnotationCompletedEventType) => {
+    const annotation = evt.detail.annotation as InterpolationROIAnnotation;
+    if (!annotation?.metadata) {
+      return;
+    }
+    const { toolName } = annotation.metadata;
+
+    if (-1 === this.toolNames.indexOf(toolName)) {
```
sedghi, line 77: same

sedghi (commented):
Please take a look at my comments. We're almost there, and I believe there are many areas where we can improve performance.

packages/tools/src/utilities/contours/PointsArray.ts
sedghi: You have data which is all zero

sedghi (approved):
Thanks Flywheel team for the initial work and your alley-oop and great work @wayfarer3130 to slam dunk it into our repo

PR #1029, opened 2024-01-23
feat(contouSeg): append/remove

lscoder:
https://github.com/cornerstonejs/cornerstone3D/assets/931820/444df8e3-8bb8-419a-92cc-c5c7ab384073

https://github.com/cornerstonejs/cornerstone3D/assets/931820/2b7bbedf-2f2f-438a-9bc1-964d777325cf


packages/tools/examples/splineContourSegmentationTools/index.ts
```diff
@@ -366,6 +369,9 @@ async function run() {
     ],
   });
 
+  // Spline curves may be converted into freehand contours when they overlaps (append/remove)
+  toolGroup.setToolPassive(PlanarFreehandContourSegmentationTool.toolName);
```
sedghi, line 373: I see why you are doing it here, but in fact this should be handled internally

packages/tools/src/types/EventTypes.ts
```diff
@@ -118,6 +118,10 @@ type AnnotationAddedEventDetail = {
  * drawing on the viewport.
  */
 type AnnotationCompletedEventDetail = {
+  /** unique id of the viewport */
+  viewportId: string;
+  /** unique id of the rendering engine */
+  renderingEngineId: string;
```
sedghi, line 124: I actually don't like this. The annotation completed event shouldn't include information about the viewport/renderingEngineIds. Can you make your way around by using some utility? like with `packages/tools/src/utilities/getViewportForAnnotation.ts`

packages/tools/src/utilities/math/aabb/intersectAABB.ts
```diff
@@ -0,0 +1,13 @@
+import { Types } from '@cornerstonejs/core';
+
+export default function intersectAABB(
```
sedghi, line 3: docs please

packages/tools/src/eventListeners/annotations/contourSegmentation/areContoursFromSameSegmentIndex.ts
```diff
@@ -0,0 +1,15 @@
+import { ContourSegmentationAnnotation } from '../../../types/ContourSegmentationAnnotation';
+
+export default function areContoursFromSameSegmentIndex(
+  firstAnnotation: ContourSegmentationAnnotation,
+  secondAnnotation: ContourSegmentationAnnotation
+) {
+  const { segmentation: firstSegmentation } = firstAnnotation.data;
+  const { segmentation: secondSegmentation } = secondAnnotation.data;
+
+  return (
```
sedghi, line 10: You should also chexck firstSegmentation.segmentationId is the same as the second one, no?

packages/tools/src/utilities/math/polyline/lineSegmentsIntersect.ts
```diff
@@ -0,0 +1,114 @@
...
+// intersections are calculated. Also that one may return [NaN, NaN] for
+// collinear points.
+
+/**
+ * Checks whether the line (`p1`,`q1`) intersects the line (`p2`,`q2`) via an
+ * orientation algorithm.
+ *
+ * Credit and details: geeksforgeeks.org/check-if-two-given-line-segments-intersect/
+ */
+function lineSegmentsIntersect(
```
sedghi, line 17: can you pleaes update the docs with proper param documentation, what is p1, q1 what is p2 q2

packages/tools/src/utilities/math/polyline/intersectPolyline.ts
```diff
@@ -0,0 +1,30 @@
...
+
+    if (intersectionPointIndexes?.length === 2) {
+      return true;
+    }
+  }
+
+  return false;
+}
+
+export { intersectPolyline as default };
```
sedghi, line 30: docs

packages/tools/src/utilities/math/polyline/lineSegmentsIntersect.ts
sedghi: I prefer another name suggesting the boolean nature of the output, something like areLineSegmentsIntersecting or similar

packages/tools/src/utilities/math/polyline/intersectPolyline.ts
```diff
@@ -0,0 +1,30 @@
...
+    const intersectionPointIndexes = getFirstLineSegmentIntersectionIndexes(
+      targetPolyline,
+      sourceP1,
+      sourceP2
+    );
+
+    if (intersectionPointIndexes?.length === 2) {
+      return true;
+    }
+  }
```
sedghi, line 25: handle the first and last one outside the loop so that you remove that check inside each iteration


for (let i = 0, sourceLen = sourcePolyline.length; i < sourceLen; i++) {
    const sourceP1 = sourcePolyline[i];
    const sourceP2 = sourcePolyline[i+1]

    const intersectionPointIndexes = getFirstLineSegmentIntersectionIndexes(
      targetPolyline,
      sourceP1,
      sourceP2
    );

    if (intersectionPointIndexes?.length === 2) {
      return true;
    }
  }

  const sourceP1 = sourcePolyline[sourceLen - 1];
const sourceP2 = sourcePolyline[0];


...

packages/tools/src/utilities/math/polyline/getNormal3.ts
```diff
@@ -0,0 +1,37 @@
+import { vec3 } from 'gl-matrix';
+import { Types } from '@cornerstonejs/core';
+
+/**
+ * Calculate the normal of a 3D planar polyline
+ * @param polyline - Planar polyline in 3D space
+ * @returns Normal of the 3D planar polyline
+ */
+export default function getNormal3(polyline: Types.Point3[]): Types.Point3 {
```
sedghi, line 9: I see this and the getNormal2 shares a lot of logic, maybe refactor the code

packages/tools/src/utilities/math/polyline/getNormal3.ts
```diff
@@ -0,0 +1,37 @@
...
+    (acc, cur) => vec3.add(acc, acc, cur),
+    vec3.create()
+  );
+
+  vec3.scale(meanPoint, meanPoint, 1 / polyline.length);
+
+  for (let i = 0, len = polyline.length; i < len; i++) {
+    const p1 = polyline[i];
+    // Using ternary instead of % (mod) operator to make it faster
+    const p2Index = i === len - 1 ? 0 : i + 1;
```
sedghi, line 25: same comment

packages/tools/src/utilities/math/polyline/getLineSegmentsIntersection.ts
```diff
@@ -0,0 +1,46 @@
...
+ * https://en.wikipedia.org/wiki/Line%E2%80%93line_intersection#Given_two_points_on_each_line
+ */
+export default function getLineSegmentsIntersection(
+  p1: Types.Point2,
+  q1: Types.Point2,
+  p2: Types.Point2,
+  q2: Types.Point2
+): Types.Point2 {
+  const denominator =
+    (q2[1] - p2[1]) * (q1[0] - p1[0]) - (q2[0] - p2[0]) * (q1[1] - p1[1]);
```
sedghi, line 29: q1[0] - p1[0] is re-calculated 3 times, and similarly others can be improved by caching

packages/tools/src/utilities/math/polyline/getLineSegmentIntersectionsIndexes.ts
```diff
@@ -0,0 +1,30 @@
...
+  closed = true
+): Types.Point2[] {
+  const intersections: Types.Point2[] = [];
+  const numPoints = polyline.length;
+  const maxI = numPoints - (closed ? 1 : 2);
+
+  for (let i = 0; i <= maxI; i++) {
+    const p2 = polyline[i];
+    // Do not use % operator for better performance
+    const j = i === numPoints - 1 ? 0 : i + 1;
```
sedghi, line 21: same comment (to handle the last -> first) outside to remove condition inside the loop

packages/tools/src/utilities/math/polyline/getClosestLineSegmentIntersection.ts
```diff
@@ -0,0 +1,67 @@
+import { vec2 } from 'gl-matrix';
+import type { Types } from '@cornerstonejs/core';
+import lineSegmentsIntersect from './lineSegmentsIntersect';
+
+/**
+ * Checks whether the line (`p1`,`q1`) intersects any of the other lines in the
+ * `points`, and returns the closest value.
+ */
+export default function getClosestLineSegmentIntersection(
```
sedghi, line 9: please add proper param docs

packages/tools/src/utilities/math/polyline/getClosestLineSegmentIntersection.ts
```diff
@@ -0,0 +1,67 @@
...
+  points: Types.Point2[],
+  p1: Types.Point2,
+  q1: Types.Point2,
+  closed = true
+): { segment: Types.Point2; distance: number } | undefined {
+  let initialI;
+  let j;
```
sedghi, line 16: let lastPointIndex instead makes this code more readable, similarly let currentPointIndex

packages/tools/src/utilities/math/polyline/containsPoint.ts
```diff
@@ -0,0 +1,63 @@
...
+  // Test intersection agains [end, start] line segment if it should be closed
+  const shouldClose = !(closed === undefined ? isClosed(polyline) : closed);
+  const maxSegmentIndex = polyline.length - (shouldClose ? 1 : 2);
+
+  for (let i = 0; i <= maxSegmentIndex; i++) {
+    const p1 = polyline[i];
+
+    // Calculating the next point index without using % (mod) operator like in
+    // `(i + 1) % numPolylinePoints` to make it 20% faster
+    const p2Index = i === numPolylinePoints - 1 ? 0 : i + 1;
```
sedghi, line 36: same

packages/tools/src/utilities/math/polyline/combinePolyline.ts
```diff
@@ -0,0 +1,327 @@
...
+      //   - [(0, 0), (1, 1)] x [(1, 1), (1, 2)]
+      //   - [(0, 1), (2, 1)] x [(1, 1), (1, 2)]
+      const intersectionCoordinate = getLineSegmentsIntersection(
+        p1,
+        q1,
+        p2,
+        q2
+      ) as Types.Point2;
+
+      const targetStartPointDistSquared = math.point.distanceToPointSquared(
```
sedghi, line 104: can't we get away with distanceToPoint (not squared) to avoid squaring operation, since you need it for sort only the non-squared version is as good as the squared version

packages/tools/src/utilities/math/polyline/combinePolyline.ts
```diff
@@ -0,0 +1,327 @@
...
+        visited: false,
+        next: null,
+      };
+
+      // Intersection point to be added to the source polyline list.
+      // At this point there is no way to know if the point is entering or
+      // exiting the intersection region but that is not going to be used
+      // hence it is set to "unknown".
+      const sourceEdgePoint: PolylineIntersectionPoint = {
+        ...targetEdgePoint,
```
sedghi, line 140: structuredClone is faster than spread 

https://www.measurethat.net/Benchmarks/Show/17968/0/structuredclone-vs-spread-object

packages/tools/src/utilities/math/polyline/combinePolyline.ts
```diff
@@ -0,0 +1,327 @@
...
+
+  // Source polyline contains target polyline
+  if (!startPoint) {
+    return targetPolyline.slice();
+  }
+
+  const mergedPolyline = [startPoint.coordinates];
+  let currentPoint = startPoint.next;
+
+  while (currentPoint !== startPoint) {
```
sedghi, line 259: This is most likely the infinite loop you showed me

packages/tools/src/utilities/math/polyline/combinePolyline.ts
```diff
@@ -0,0 +1,327 @@
+import { Types } from '@cornerstonejs/core';
+import * as math from '..';
```
sedghi, line 2: This will probably lead to recursive imports that some bundlers can't handle

packages/tools/src/eventListeners/annotations/contourSegmentation/contourSegmentationCompleted.ts
```diff
@@ -0,0 +1,216 @@
...
+      );
+      triggerAnnotationRenderForViewportIds(
+        renderingEngine,
+        viewportIdsToRender
+      );
+    }
+  });
+}
+
+export default function contourSegmentationCompletedListener(
```
sedghi, line 173: I expect the default function to be the first thing in a file when I open it instead of noisy functions and then the default. Also docs please

packages/tools/src/eventListeners/annotations/contourSegmentation/contourSegmentationCompleted.ts
```diff
@@ -0,0 +1,216 @@
...
+  evt: AnnotationCompletedEventType
+) {
+  const sourceAnnotation = evt.detail
+    .annotation as ContourSegmentationAnnotation;
+  const { renderingEngineId, viewportId } = evt.detail;
+  const renderingEngine = getRenderingEngine(renderingEngineId);
+  const viewport = renderingEngine.getViewport(viewportId);
+
+  if (!isContourSegmentationAnnotation(sourceAnnotation)) {
```
sedghi, line 182: can do this sooner

packages/tools/src/eventListeners/annotations/contourSegmentation/contourSegmentationCompleted.ts
```diff
@@ -0,0 +1,216 @@
...
+  const contourSegmentationAnnotations =
+    getValidContourSegmentationAnnotations(sourceAnnotation);
+
+  if (!contourSegmentationAnnotations.length) {
+    return;
+  }
+
+  const sourcePolyline = sourceAnnotation.data.contour.polyline.map((point) =>
+    viewport.worldToCanvas(point)
+  );
```
sedghi, line 195: for loop for performance

packages/tools/src/eventListeners/annotations/contourSegmentation/contourSegmentationCompleted.ts
```diff
@@ -0,0 +1,216 @@
...
+
+  const { element } = viewport;
+  const enabledElement = getEnabledElement(element);
+  const { renderingEngine } = enabledElement;
+  const { metadata, data } = targetAnnotation;
+  const { handles, segmentation } = data;
+  const { textBox } = handles;
+
+  newPolylines.forEach((newPolyline) => {
+    const polyline = newPolyline.map((p) => viewport.canvasToWorld(p));
```
sedghi, line 123: for loop for performance

sedghi (commented):
Looks great

I tried it locally and found an unusual bug in interactions that is reproducible, it only happens with the other contour is in the north of the main contour, hmmm

https://github.com/cornerstonejs/cornerstone3D/assets/7490180/1140b82c-bbfe-4c89-b191-18424171fb3e

But otherwise seems to work

packages/tools/src/utilities/math/polyline/combinePolyline.ts
lscoder: Yes but the problem is related to `getSourceAndTargetPointsList` that is responsible for creating the linked list.

packages/tools/src/utilities/math/polyline/combinePolyline.ts
lscoder: structuredClone_vs_spread

packages/tools/src/utilities/math/polyline/combinePolyline.ts
lscoder: I didn't get. Are you saying to call `distanceToPoint` instead of `distanceToPointSquared`? But `distanceToPoint` calls `distanceToPointSquared` internally and it also takes the square root that is one of the most costly operator.

Time to process 10M numbers:
 - (n * n) = 161ms
 - (n ** 2) = 199ms
 - Math.pow(n, 2) = 29529ms
 - Math.sqrt(n) = 31946ms

sedghi: What I mean is you only need this distance for sorting right? Then you don't need squared or squared root, you can just use the products and sum

packages/tools/src/utilities/math/polyline/intersectPolyline.ts
lscoder: Why? Performance? I cloned/refactored the function and I do not see any difference when calculating the intersection between two polylines with 5k points each (25M intersections calculated for the naive implementation). I wouldn't like to duplicate the code for no improvement.

Screenshot_20240124_151915

Screenshot_20240124_151820

packages/tools/src/utilities/math/polyline/intersectPolyline.ts
lscoder: Updated

packages/tools/src/utilities/math/polyline/lineSegmentsIntersect.ts
lscoder: Updated

packages/tools/src/types/EventTypes.ts
lscoder: I had never used `getViewportForAnnotation`. That's good to know!
Updated

packages/tools/src/utilities/math/point/distanceToPointSquared.ts
```diff
@@ -16,6 +16,11 @@ export default function distanceToPointSquared(p1: Point, p2: Point): number {
...
   const [x1, y1, z1 = 0] = p1;
   const [x2, y2, z2 = 0] = p2;
+  const dx = x2 - x1;
+  const dy = y2 - y1;
+  const dz = z2 - z1;
 
-  return Math.pow(x1 - x2, 2) + Math.pow(y1 - y2, 2) + Math.pow(z1 - z2, 2);
+  // Time to square 10M numbers:
+  //   (n * n) = 161ms | (n ** 2) = 199ms | Math.pow(n, 2) = 29529ms
+  return dx * dx + dy * dy + dz * dz;
```
lscoder, line 25: We should never use `Math.pow(n, 2)` since it is 183x slower than a simple `n*n`

packages/tools/examples/splineContourSegmentationTools/index.ts
lscoder: The contour annotations are not converted into freehand segmentation tool when it is not available and a warning is logged on console.

packages/tools/src/utilities/math/polyline/combinePolyline.ts
lscoder: as we discussed there are no changes to be done.

packages/tools/src/tools/annotation/SplineROITool.ts
```diff
@@ -734,11 +764,11 @@ class SplineROITool extends ContourSegmentationBaseTool {
       annotationStyle.textbox
     );
 
-    if (this.fireChangeOnUpdate) {
-      this.triggerAnnotationModified(
+    if (this.fireChangeOnUpdate?.annotationUID === annotationUID) {
```
lscoder, line 767: It was triggering multiple events only for the first annotation

packages/tools/src/utilities/contours/reverseIfAntiClockwise.ts
```diff
@@ -19,32 +9,11 @@ function getSumReducer(total: number, num: number): number {
...
-    return points;
-  }
-  let xSum = 0;
-  for (const point of points) {
-    xSum += point[0];
-  }
-  const xMean = xSum / length;
-
-  let checkSum = 0;
+  const signedArea = getSignedArea(points);
```
lscoder, line 12: The previous function was a bit slower and it was also looping through all points twice. `checksum` was actually the double of the signed area.

packages/tools/src/utilities/math/polyline/combinePolyline.ts
```diff
@@ -0,0 +1,350 @@
...
+  const sourceIntersectionsCache = new Map<
+    number,
+    PolylineIntersectionPoint[]
+  >();
+
+  const isFisrtPointInside = containsPoint(sourcePolyline, targetPolyline[0]);
+
+  let intersectionPointDirection = isFisrtPointInside
+    ? PolylinePointDirection.Exiting
+    : PolylinePointDirection.Entering;
```
lscoder, line 81: @sedghi This was the problem with that first point inside the polyline. It may also be on the edge but that requires a bigger change for a much lower probability of happening.

packages/tools/src/utilities/math/polyline/getArea.ts
```diff
@@ -7,7 +7,7 @@ import type { Types } from '@cornerstonejs/core';
  * the points are in canvas, then the result is in canvas pixels ^2; If they are
  * in mm, then the result is in mm^2; etc.
  */
-export default function calculateAreaOfPoints(points: Types.Point2[]): number {
+export default function getArea(points: Types.Point2[]): number {
```
lscoder, line 10: Renamed only to have `getArea` and `getSignedArea` functions

packages/tools/src/utilities/math/polyline/getSignedArea.ts
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * This functions has a runtime very close to `getArea` and it is recommended to
+ * be called only if you need the area signal (eg: calculate polygon normal). If
+ * you do not need the area signal you should always call `getArea`.
+ *
+ *
+ * @param polyline - Polyline points (2D)
+ * @returns Area of the polyline (with signal)
+ */
+export default function getSignedArea(polyline: Types.Point2[]): number {
```
lscoder, line 15: After refactoring the function and removing any `glMatrix` function call it is 2-5x faster (2x faster  for 1k points, 4x faster for 7k points, 5x faster for 20k points) and it is <4% slower than `getArea` (it was faster ~15% of the time when some tests)

packages/tools/src/utilities/math/polyline/getNormal3.ts
lscoder: Refactored to remove all calls to `glMatrix` functions which makes a huge difference (4x faster) when processing thousands of points.

packages/tools/src/eventListeners/annotations/contourSegmentation/isContourSegmentationAnnotation.ts
```diff
@@ -0,0 +1,8 @@
+import { Annotation } from '../../../types';
+import { ContourSegmentationAnnotation } from '../../../types/ContourSegmentationAnnotation';
+
+export default function isContourSegmentationAnnotation(
```
sedghi, line 4: can we move to utilities.contourSegmentation.areSameSegments or something?

packages/tools/src/eventListeners/annotations/contourSegmentation/areContoursFromSameSegmentIndex.ts
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * Check if two contour segmentations are from same segmentId,
+ * segmentationRepresentationUID and segmentIndex.
+ * @param firstAnnotation - First annotation
+ * @param secondAnnotation - Second annotation
+ * @returns True if they are from same segmentId, segmentationRepresentationUID
+ * and segmentIndex or false otherwise.
+ */
+export default function areContoursFromSameSegmentIndex(
```
sedghi, line 11: can we move to utilities.contourSegmentation.areSameSegments or something?

packages/tools/src/eventListeners/annotations/contourSegmentation/isContourSegmentationAnnotation.ts
sedghi, line 4: move I guess

packages/tools/src/eventListeners/annotations/contourSegmentation/areCoplanarContours.ts
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * A plane may be represented by a normal and a distance then to know if they
+ * are coplanar we need to:
+ *   - check if the normals of the two annotations are pointing to the same
+ *   direction or to opposite directions (dot product equal to 1 or -1
+ *   respectively)
+ *   - Get one point from each polyline and project it onto the normal to get
+ *   the distance from the origin (0, 0, 0).
+ */
+export default function areCoplanarContours(
```
sedghi, line 15: move

packages/tools/src/eventListeners/annotations/contourSegmentation/contourSegmentationCompleted.ts
```diff
@@ -0,0 +1,276 @@
...
+        isContourSegmentationAnnotation(targetAnnotation) &&
+        areContoursFromSameSegmentIndex(targetAnnotation, sourceAnnotation) &&
+        areCoplanarContours(targetAnnotation, sourceAnnotation)
+    );
+
+    return validAnnotations.concat(toolAnnotations);
+  }, []);
+}
+
+function getTargetAnnotation(
```
sedghi, line 156: findIntersectingContour

sedghi: and also a utility in the utilities

packages/tools/src/eventListeners/annotations/contourSegmentation/areContoursFromSameSegmentIndex.ts
lscoder: Moved

packages/tools/src/eventListeners/annotations/contourSegmentation/isContourSegmentationAnnotation.ts
lscoder: Moved

packages/tools/src/eventListeners/annotations/contourSegmentation/areCoplanarContours.ts
lscoder: Moved

packages/tools/src/eventListeners/annotations/contourSegmentation/contourSegmentationCompleted.ts
lscoder: Renamed

sedghi (approved):
Looks great as always, thanks!

PR #1035, opened 2024-01-24
feat: Livewire editing

wayfarer3130:
### Context

Adding basic livewire editing.   This is a first step towards livewire interpolation between slices.

### Changes & Results

Adding ability to edit livewire handles and auto-recompute livewire data.

### Testing

Run the interpolationContourSegmentation example and test the livewire tools with editing.
Note that I can get into a situation where it does weird things on the edit, still looking into that.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  


packages/tools/examples/interpolationContourSegmentation/index.ts
wayfarer3130, line 1: Adding a few changes to make livewire testing with interpolation convenient - sizing and positioning basically.

packages/tools/src/utilities/contours/interpolation/interpolate.ts
```diff
@@ -167,8 +167,9 @@ function _linearlyInterpolateContour(
   const handleCount = Math.round(
     Math.max(
       8,
-      interpolationData.get(startIndex)[0].data.handles.points.length * 2,
-      interpolationData.get(endIndex)[0].data.handles.points.length * 2
+      interpolationData.get(startIndex)[0].data.handles.points.length,
```
wayfarer3130, line 170: Needed to fix this, otherwise the number of handles grows without bounds.

packages/tools/src/types/ContourAnnotation.ts
```diff
@@ -7,6 +7,13 @@ export type ContourAnnotationData = {
       polyline: Types.Point3[];
       closed: boolean;
     };
+    handles?: {
+      /**
+       * An index of the handle to contour points.  This allows associating
+       * the handle data to the polyline position data.
+       */
+      handleContourIndex?: number[];
```
sedghi, line 15: @lscoderwhat do you think about this?

packages/tools/src/tools/annotation/LivewireContourTool.ts
```diff
@@ -26,18 +26,23 @@ import { LivewireContourAnnotation } from '../../types/ToolSpecificAnnotationTyp
...
 import ContourSegmentationBaseTool from '../base/ContourSegmentationBaseTool';
 
+const { isEqual } = csUtils;
+
 const CLICK_CLOSE_CURVE_SQR_DIST = 10 ** 2; // px
 
 class LivewireContourTool extends ContourSegmentationBaseTool {
   public static toolName: string;
   private scissors: LivewireScissors;
+  /** The scissors from the right handle, used for editing */
```
sedghi, line 44: what is right handle?

packages/tools/src/utilities/livewire/LivewireScissors.ts
```diff
@@ -102,6 +104,48 @@ export class LivewireScissors {
...
+    let removeCount = 0;
+    for (
+      let i = pathPoints.length - 1;
+      i > pathPoints.length - count && i > 0;
+      i--
+    ) {
+      const weighted = this._getWeightedDistance(
+        pathPoints[i],
+        pathPoints[i - 1]
+      );
```
lscoder, line 138: I thought this function should be in `LivewirePath` but there is a dependency to `_getWeightedDistance` that does not allow it to be move :disappointed:

packages/tools/src/types/ContourAnnotation.ts
lscoder: Where is it being used?

packages/tools/src/tools/annotation/LivewireContourTool.ts
```diff
@@ -133,8 +127,8 @@ class LivewireContourTool extends ContourSegmentationBaseTool {
 
       sliceToWorld = (point: Types.Point2) => {
         const ijkPoint = vec3.transformMat4(
-          [0, 0, 0],
-          [point[0], point[1], 0],
+          vec3.create(),
+          vec3.fromValues(point[0], point[1], 0),
```
lscoder, line 131: I used normal arrays in this case and a few lines above because I've measured the performance in the past and converting a bunch of points was way slower when creating new vectors with `vec3.create()` because it allocates a new buffer for every point. Just make sure this does not degrades the performance.

packages/tools/src/tools/annotation/LivewireContourTool.ts
lscoder: Yeah... the name is a bit confusing. After reading the code I saw that `handle-1` is called `left` and `handle+1` is called `right`. I thought about `editScissors` but I'm still not happy with it.

packages/tools/src/tools/annotation/LivewireContourTool.ts
```diff
@@ -459,38 +514,108 @@ class LivewireContourTool extends ContourSegmentationBaseTool {
...
+        confirmedPathRight.addPoints(
+          polyline.slice(rightIndex, polyline.length).map(worldToSlice)
+        );
+      }
+      // for (let i = 0; i < leftIndex; i++) {
+      //   confirmedPath.addPoint(worldToSlice(polyline[i]));
+      // }
+      // for (let i = rightIndex + 1; i < polyline.length; i++) {
+      //   confirmedPathRight.addPoint(worldToSlice(polyline[i]));
+      // }
```
lscoder, line 561: Delete

packages/tools/src/types/ContourAnnotation.ts
wayfarer3130: Removing momentarily - I'm not using it right now.

packages/tools/src/tools/annotation/LivewireContourTool.ts
wayfarer3130: They are the left and right handles relative to a clockwise direction set of handles.  Could alternatively call them next and previous if you prefer.

packages/tools/src/utilities/livewire/LivewireScissors.ts
wayfarer3130: It specifically requires the livewire scissors data.  There are going to be other additions to livewire scissors to deal with the interpolation changes.

lscoder (approved):

PR #1037, opened 2024-01-25
fix(WindowLevelTool): attempt to set zero width color range

Blackman99:
### Context

When using mouse with `WindowLevelTool`. When the `voiRange.upper` is equal to or less than `voiRange.lower`, would cause `attempt to set zero width color range` in the console.

### Changes & Results

Add a condition to ensure `voiRange.upper` is larger than `voiRange.lower`

### Testing

* Before
cs-window-level-before-fix
* After
cs-window-level-after-fix

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] OS: Ubuntu 23.10
- [x] Node version: 20.10.0
- [x] Browser: Chrome 120.0.6099.225

### DIFF
--- a/packages/tools/src/tools/WindowLevelTool.ts
+++ b/packages/tools/src/tools/WindowLevelTool.ts
@@ -98,6 +98,11 @@ class WindowLevelTool extends BaseTool {
       });
     }
 
+    // If the range is not valid. Do nothing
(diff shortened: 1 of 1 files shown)


sedghi (approved):
Thanks!

PR #1038, opened 2024-01-25
fix: Capture error messages thrown by vtk

baokeyu123:
### Context
Due to vtk.js  version update.
DataArray.js add  conditional judgment 
```javascript
  if (model.size % model.numberOfComponents !== 0) {
    throw new RangeError('model.size is not a multiple of model.numberOfComponents');
  }
```
i think that we should add condition to capture error message in Vtk.

image

### Changes & Results

Capture error information by dispatch  `Events.IMAGE_LOAD_ERROR`

### Testing

after:
image

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] "OS: 
- [x] "Node version: 
- [x] "Browser:
  


sedghi:
I think this is related PR https://github.com/cornerstonejs/cornerstone3D/pull/1043
Do you actually have any data to test this? Unfortunately, I don't have any data to test for 1043.

baokeyu123:
> Do you actually have any data to test this? Unfortunately, I don't have any data to test for 1043

yes. dicom tag(00280004) is RGB.
But `PR1043` is only for the changes made to RGB DICOM. I have not tested all color(such as RGBA,MONOCHROME2,MONOCHROME1,YBR_FULL).
createVTKImageData error information can be captured through PR code.

Blackman99:
This is exactly what we're facing too.

sedghi:
If someone can provide test data, I would be happy to test and merge this. Unfortunately, I cannot blindly approve it without the necessary information.

Blackman99:
The current `Promise.allSettled` function would always return a array, the item of which is either a fullfilled or a rejected promise. The problem is the current code do nothing even if there are some **rejected results**. It would cause the user get nothing when some error happened.

This PR add the error handle part which can handle any error happed during `getLoaderImageOptions` and dispatch the `IMAGE_LOAD_ERROR` event by `eventTarget`.  So the user can use the `eventTarget.addEventListener` to get error message

baokeyu123:
> If someone can provide test data, I would be happy to test and merge this. Unfortunately, I cannot blindly approve it without the necessary information.

Firstly,based on `dicomImageLoader`'s wadors to load Dicom image（cornerstoneDICOMImageLoader.wadors.metaDataManager.add(imageId，metadata)).
Secondly,accept set  `multipart/related;type=application/dicom`   in request header.
Finally,make sure dicom tag (0028004) is RGB.
Regardless of `PR1038` and `PR1043`, the problem can be reproduced.
I don't think it can be reproduced simply by giving you the .dcm file.I have to think about how to help you reproduce the problem.

sedghi (approved):
Makes sense thanks

PR #1042, opened 2024-01-26
fix nifti issues

CYB3RL1F3:
### Context

Impossible to scroll and use annotation with niftis because of a delta of the volumeId. This PR fixes it.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  

### DIFF
--- a/packages/tools/src/tools/MIPJumpToClickTool.ts
+++ b/packages/tools/src/tools/MIPJumpToClickTool.ts
@@ -5,6 +5,7 @@ import { getPointInLineOfSightWithCriteria } from '../utilities/planar';
 import jumpToWorld from '../utilities/viewport/jumpToWorld';
 import { PublicToolProps, ToolProps } from '../types';
 import { getToolGroupForViewport } from '../store/ToolGroupManager';
+import { getVolumeId } from '../utilities/getVolumeId';
 
 /**
  * On a Maximum Intensity Projection (MIP) viewport, MIPJumpToClickTool allows the
@@ -54,7 +55,7 @@ class MIPJumpToClickTool extends BaseTool {
       );
     }
 
-    const volumeId = targetId.split(/volumeId:/)[1];
+    const volumeId = getVolumeId(targetId);
 
     // 3. Criteria function to search for the point (maximum intensity)
     let maxIntensity = -Infinity;
--- a/packages/tools/src/tools/MIPJumpToClickTool.ts
+++ b/packages/tools/src/tools/MIPJumpToClickTool.ts
@@ -54,7 +54,7 @@ class MIPJumpToClickTool extends BaseTool {
       );
     }
 
-    const volumeId = targetId.split(/volumeId:|\?/)[1];
+    const volumeId = targetId.split(/volumeId:/)[1];
 
     // 3. Criteria function to search for the point (maximum intensity)
     let maxIntensity = -Infinity;
--- a/packages/tools/src/utilities/getVolumeId.ts
+++ b/packages/tools/src/utilities/getVolumeId.ts
@@ -1,8 +1,13 @@
 export const getVolumeId = (targetId: string) => {
-  const str = targetId.substring('volumeId:'.length);
+  const prefix = 'volumeId:';
+  const str = targetId.includes(prefix)
+    ? targetId.substring(prefix.length)
+    : targetId;
+
   if (str.startsWith('nifti:')) {
     return str;
   }
+
   const index = str.indexOf('?');
   return index === -1 ? str : str.substring(0, index);
 };
--- a/packages/tools/src/utilities/getVolumeId.ts
+++ b/packages/tools/src/utilities/getVolumeId.ts
@@ -1,6 +1,8 @@
 export const getVolumeId = (targetId: string) => {
-  if (targetId.includes('nifti:')) {
-    return targetId.split(/volumeId:/)[1];
+  const str = targetId.substring('volumeId:'.length);
+  if (str.startsWith('nifti:')) {
+    return str;
   }
-  return targetId.split(/volumeId:|\?/)[1];
+  const index = str.indexOf('?');
+  return index === -1 ? str : str.substring(0, index);
 };
--- a/packages/tools/src/utilities/getVolumeId.ts
+++ b/packages/tools/src/utilities/getVolumeId.ts
@@ -1,6 +1,6 @@
 export const getVolumeId = (targetId: string) => {
-  if (targetId.includes('volumeId:')) {
-    return targetId.split(/volumeId:|\?/)[1];
+  if (targetId.includes('nifti:')) {
+    return targetId.split(/volumeId:/)[1];
   }
-  return targetId;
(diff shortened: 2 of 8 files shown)


wayfarer3130:
Please refactor the code to use a utility to extract the volumeId, with samples from BaseVolumeViewport including the sliceIndex etc that identifies the actual slice to view this on - the samples should be in a unit test (jest, not karma) that also includes samples of your nifti volumeIds that are causing.  That should pass the PR checks and then I can take a look at this again.

wayfarer3130 (changes requested):
See notes

packages/tools/src/tools/MIPJumpToClickTool.ts
```diff
@@ -54,7 +54,7 @@ class MIPJumpToClickTool extends BaseTool {
       );
     }
 
-    const volumeId = targetId.split(/volumeId:|\?/)[1];
+    const volumeId = targetId.split(/volumeId:/)[1];
```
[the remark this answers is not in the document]
CYB3RL1F3: I need the "?" because the URL of niftis have query strings... 
Please find another solution!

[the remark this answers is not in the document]
CYB3RL1F3: Please fix it!! Still not working on the version 51...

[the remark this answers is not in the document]
wayfarer3130: Move the extraction of volumeId into a utility, add a unit test for it with examples of both the  target id from nifti and the target id from the base image volume changes, then we can look at how to fix this.  Without having examples of what the data looks like, I can't just "fix this".  But, if we have it as a utility, there will be one spot to fix things, and with the unit tests, we can have the fix be a lot more stable over time.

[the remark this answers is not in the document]
CYB3RL1F3: Done

packages/tools/test/getVolumeId_test.js
```diff
@@ -0,0 +1,13 @@
+import { getVolumeId } from '../../tools/src/utilities/getVolumeId';
+
+describe('getVolumeId', () => {
+  it('should extract volumeId from target', () => {
+    const targetId = 'volumeId:aaaa';
+    expect(getVolumeId(targetId)).toEqual('aaaa');
+  });
+
+  it('should extract volumeId from a complex nifti url', () => {
+    const targetId = 'nifti:https://nifti.com?xoxo';
```
wayfarer3130, line 10: Are you sure this one isn't  'volumeId:nifti:https://nifti.com?xoxo'   Otherwise this would not have worked previously since
'nifti:https://nifti.com?xoxo'.split(/volumeId:/)[1] 
returns undefined

packages/tools/test/getVolumeId_test.js
```diff
@@ -0,0 +1,13 @@
+import { getVolumeId } from '../../tools/src/utilities/getVolumeId';
+
+describe('getVolumeId', () => {
+  it('should extract volumeId from target', () => {
```
wayfarer3130, line 4: Can we add this test with arguments, eg volumeId:aaaa?sliceIndex=4&focalPoint=3,2,5.5
It should still extract aaaa

packages/tools/src/utilities/getVolumeId.ts
```diff
@@ -0,0 +1,6 @@
+export const getVolumeId = (targetId: string) => {
```
wayfarer3130, line 1: How about:
```
function getVolumeId(str) {
... if( str.startsWith('volumeId:') ) {
...   str = str.substring(9);
... }
... if( str.startsWith('nifti:') ) {
...   return str;
... }
... const query = str.indexOf('?');
... return query===-1 ? str : str.substring(0,query);
... }
```

wayfarer3130 (commented):
Almost there, just a couple small tweaks and that will do it.

packages/tools/test/getVolumeId_test.js
CYB3RL1F3: At which moment "volumeId:" is added?

packages/tools/test/getVolumeId_test.js
CYB3RL1F3: Done

CYB3RL1F3:
@wayfarer3130 or @sedghi could you checkout this and merge please?

packages/tools/src/utilities/getVolumeId.ts
```diff
@@ -0,0 +1,8 @@
...
```
wayfarer3130, line 1: You need to check that targetId starts with volumeId first, otherwise this code does weird things on imageId substrings.

packages/tools/test/getVolumeId_test.js
wayfarer3130, line 1: Please make this getVolumeId.jest.js  so that it runs as unit test - with the appropriate imports for describe, it and so on.
yarn jest
will run this inside the package or
yarn test:unit
will run all the unit tests from the top level.

packages/tools/test/getVolumeId_test.js
```diff
@@ -0,0 +1,15 @@
+import { getVolumeId } from '../../tools/src/utilities/getVolumeId';
```
wayfarer3130, line 1: This can be ../src/utilities/getVolumeId

packages/tools/src/utilities/getVolumeId.ts
CYB3RL1F3: It already does weird things on imageId that's the reason of the PR...
This PR is tiring me for a little fix of a regression introduced 3 weeks ago!!

CYB3RL1F3: It's already tested anyway...

CYB3RL1F3: @wayfarer3130 could you write the code that makes me working with a nifti:xxxx.com/xxx?xxx=xxx&xxx=xxx that is the volumeId I get in the project. Thanks.

packages/tools/test/getVolumeId_test.js
CYB3RL1F3: I don't understand what you ask. All files are named _test.js

packages/tools/test/getVolumeId_test.js
CYB3RL1F3: Your suggestion doesn't work.

packages/tools/src/utilities/getVolumeId.ts
CYB3RL1F3: It's up to date now, could you merge it or fix the problem please?? Thanks!

CYB3RL1F3:
Could you merge the PR or fix the problem by your side please ? Thank you!

CYB3RL1F3:
Up! Please!

wayfarer3130 (approved):

PR #1043, opened 2024-01-26
fix(dicom-image-loader): add rgba condition interception for canvas createImageData API

Blackman99:
### Context

The canvas context API `createImageData ` would always return a `ImageData `. And its `data` field is always a rgba color mode `Uint8ClampedArray `, like this:
image

In the newest vtk package which has already been upragded in cornserstone repo would throw a error like this:
image

As a result, the image would not render.

This PR fix it.

### Changes & Results

Add a condition, if not using rgba, then replace the data with a new Uint8ClampedArray with  `imageFrame.samplesPerPixel * columns * rows`  length

### Testing
I've tested on my private project. I'm sorry I can't share it with others because the image 
 belongs to company.

Sorr for the inconvenience.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] OS: Ubuntu 23.10
- [x] Node version: 20.10.0
- [x] Browser: Chrome 120.0.6099.227

### DIFF
--- a/packages/dicomImageLoader/src/imageLoader/createImage.ts
+++ b/packages/dicomImageLoader/src/imageLoader/createImage.ts
@@ -238,11 +238,18 @@ function createImage(
           canvas.height = imageFrame.rows;
           canvas.width = imageFrame.columns;
           const context = canvas.getContext('2d');
-          const imageData = context.createImageData(
+          let imageData = context.createImageData(
             imageFrame.columns,
             imageFrame.rows
           );
-
+          if (!useRGBA) {
+            imageData = {
+              ...imageData,
+              data: new Uint8ClampedArray(
+                3 * imageFrame.columns * imageFrame.rows
+              ),
+            };
+          }
           convertColorSpace(imageFrame, imageData.data, useRGBA);
           imageFrame.imageData = imageData;
           imageFrame.pixelData = imageData.data;
--- a/packages/dicomImageLoader/src/imageLoader/createImage.ts
+++ b/packages/dicomImageLoader/src/imageLoader/createImage.ts
@@ -246,7 +246,9 @@ function createImage(
             imageData = {
               ...imageData,
               data: new Uint8ClampedArray(
-                3 * imageFrame.columns * imageFrame.rows
+                imageFrame.samplesPerPixel *
+                  imageFrame.columns *
+                  imageFrame.rows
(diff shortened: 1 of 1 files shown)


Sergey1888888:
I have the same problem. Would be nice if it will be merged and fix in my project too. Because now I have to using hack with replacing pixelData and updating actors for viewports.

sedghi:
Can you describe the image that needs to be tested?

vnosikov:
I can't provide a sample image because it is a corporate policy not to provide them, but I confirm that there is a problem after vtk added that check and fixes like that solve it

Blackman99:
> Can you describe the image that needs to be tested?

A dicom file with rgb color series image would do

baokeyu123:
I think that this is root cause about `PR1038`. when i show the dicom image tag(0028,0004)  is rgb, this problem will arise.
because `context.createimagedata` is based on rgba,its length doesn't divisible by 3.
due to vtk.js upgrade,It adds a conditional judgment about whether you can divide by 3.
Therefore, rgb images that are not evenly divided into 3 and 4 can not be displayed normally and without any hint.

baokeyu123:
I think that this is root cause about `PR1038`. when i show the dicom image tag(0028,0004) is rgb, this problem will arise.
because `context.createimagedata` is based on rgba,its length doesn't divisible by 3.
due to vtk.js upgrade,It adds a conditional judgment about whether you can divide by 3.
Therefore, except for common multiples of 3 and 4, rgb DICOM cannot display images and will not give any prompts.

baokeyu123:
@Blackman99  It's something only `DS` can do 😃😃😃

baokeyu123:
@Blackman99 Hey man,i suggest changing`3 * imageFrame.columns * imageFrame.rows` to `imageFrame.samplesPerPixel* imageFrame.columns * imageFrame.rows`.

Blackman99:
> @Blackman99 Hey man,i suggest changing`3 * imageFrame.columns * imageFrame.rows` to `imageFrame.samplesPerPixel* imageFrame.columns * imageFrame.rows`.

Thanks! Change has been updated

sedghi (approved):
Apologies for the delay in my review. I have been (and still am) occupied with other tasks. However, I appreciate the fix you provided.

PR #1055, opened 2024-02-01
fix: Refactor image splitting logic for 4D datasets

JoostJM:
### Context

Related to #1048 and OHIF PR [#3664](https://github.com/OHIF/Viewers/pull/3664). Both deal with the issue of adding support for 4D datasets to OHIF Viewer.

### Changes & Results

Refactor splitting datasets in `packages/streaming-image-volume-loader/helpers/splitImageIdsBy4dTags`.
- First group by imagePositionPatient. Allows checking whether the dataset is potentially a multi-stack (4D) dataset (i.e. equal and > 1 number of frames for each position)
- Check ability of predefined tags to split the dataset (i.e. unique values within 1 position, but corresponding instances for any one value in all positions).
- By checking tags on the position-grouped instances, early stopping is possible (e.g. when duplicate values are encountered, or when a set of values does not match the first such set).
- Add custom functions for parsing DWI b values stored in private tags for Philips, Siemens and GE datasets.

### Checklist

ToDo:
- [ ] Frame timepoint values are currently not provided, only used to sort the array of arrays. Therefore, the timepoints cannot be displayed downstream.
- [ ] Certain tags are implemented and should potentially work, but have not been tested.
- [ ] There exist use cases where a series with a single SeriesInstanceUID contains multiple 4D stacks identified by different ImageType values. To correctly display them, they would need to be grouped by both ImageType and a Frame timepoint identifier tag.

#### PR

#### Code

- [ ] My code has been well-documented (function documentation, inline comments,
  etc.)

  additions or removals.

#### Tested Environment

- [x] "OS: Linux 22.04.3 
- [x] "Node version: 20.11.0 
- [x] "Browser:
  - Chrome 121.0.6167.85
  - Firefox 122.0
  

### DIFF
--- a/packages/streaming-image-volume-loader/src/helpers/splitImageIdsBy4DTags.ts
+++ b/packages/streaming-image-volume-loader/src/helpers/splitImageIdsBy4DTags.ts
@@ -1,21 +1,20 @@
 import { metaData } from '@cornerstonejs/core';
 
-// TODO: add support for other 4D tags as listed below
+// TODO: Test remaining implemented tags
 // Supported 4D Tags
-//   (0018,1060) Trigger Time                   [NOK]
-//   (0018,0081) Echo Time                      [NOK]
-//   (0018,0086) Echo Number                    [NOK]
+//   (0018,1060) Trigger Time                   [Implemented, not tested]
+//   (0018,0081) Echo Time                      [Implemented, not tested]
+//   (0018,0086) Echo Number                    [Implemented, not tested]
 //   (0020,0100) Temporal Position Identifier   [OK]
 //   (0054,1300) FrameReferenceTime             [OK]
+//   (0018,9087) Diffusion B Value              [OK]
+//   (2001,1003) Philips Diffusion B-factor     [OK]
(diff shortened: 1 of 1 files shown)


sedghi:
This seems to have broken our PT 4D support, check this example, and compare to the current main branch example https://www.cornerstonejs.org/live-examples/dynamiccinetool

You can run the example `yarn run example dynamiccinetool`

JoostJM:
@sedghi I found the breaking error. Apperently in this example I'm unable to retrieve the ImagePositionPatient. This results in the data being considered as "unfit for 4D volume". I'll follow up to see why it is failing...

JoostJM:
@sedghi, I was able to fix the example by changing the way I retrieve the imagePositionPatient. This works both in the cornerstone examples as in OHIF (tested).

However, this bug has led me to believe it may still fail for the other tags implemented (such as TemporalPositionIdentifier and the B Value tags). I will try to investigate this tomorrow. Do you have any examples in cornerstone of 4D data that needs to be split using TemporalPositionIdentifier  (0020,0100)?

sedghi:
No unfortunately, we didn't test that

sedghi (approved):

PR #1061, opened 2024-02-07
feat(contourSeg): polyline performance improvements

lscoder:
### Context

Adding contour annotations may result in polylines with thousands of points making the viewer much slower to render the annotations and also process them (append, remove, create holes, interpolate, etc.). 

### Changes & Results

Implemented the Ramer–Douglas–Peucker algorithm that simplifies polylines based on an a given distance `epsilon` reducing the amount of points in a polyline.

Two polylines with 6k (source) and 7k (target) points were used on the test below to measure the time it would take to merge them before and after simplifying the polylines.

perf_polyline-simplification_merge

A good way to generate contours with +50k points is zooming all the way out before creating a freehand segmentation contour. Both polylines got simplified to new polylines with ~200 points or less each with no visual changes on the viewport. I know this is not a real example but that helped to measure the time it would take to process the contours.

https://github.com/cornerstonejs/cornerstone3D/assets/931820/ce4e68b5-d714-479a-8113-52aaaefd6612

### Testing

Draw contours (spline, freehand, livewire) and make sure they are rendered and processed (append, remove, hole, interpolation, etc) correctly. A performance improvement is very noticeable when drawing multiple freehand ROIs.

### DIFF
--- a/packages/tools/src/tools/annotation/SplineROITool.ts
+++ b/packages/tools/src/tools/annotation/SplineROITool.ts
@@ -56,7 +56,6 @@ import { LinearSpline } from './splines/LinearSpline';
 import { CatmullRomSpline } from './splines/CatmullRomSpline';
 import { BSpline } from './splines/BSpline';
 import ContourSegmentationBaseTool from '../base/ContourSegmentationBaseTool';
-import updateContourPolyline from '../../utilities/contours/updateContourPolyline';
 
 const SPLINE_MIN_POINTS = 3;
 const SPLINE_CLICK_CLOSE_CURVE_DIST = 10;
@@ -121,6 +120,18 @@ class SplineROITool extends ContourSegmentationBaseTool {
          * modifier must be pressed when the first point of a new contour is added.
          */
         contourHoleAdditionModifierKey: KeyboardBindings.Shift,
+        /**
+         * The polyline may get processed in order to reduce the number of points
+         * for better performance and storage.
+         */
+        decimate: {
+          enabled: true,
+          /** A maximum given distance 'epsilon' to decide if a point should or
+           * shouldn't be added the resulting polyline which will have a lower
+           * number of points for higher `epsilon` values.
+           */
+          epsilon: 0.1,
+        },
         spline: {
           configuration: {
             [SplineTypesEnum.Cardinal]: {
@@ -693,7 +704,7 @@ class SplineROITool extends ContourSegmentationBaseTool {
         const spline = this._updateSplineInstance(element, annotation);
         const splinePolylineCanvas = spline.getPolylinePoints();
 
-        updateContourPolyline(
+        this.updateContourPolyline(
           annotation,
           {
             points: splinePolylineCanvas,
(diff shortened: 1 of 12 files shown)


packages/tools/src/utilities/math/polyline/simplify.ts
```diff
@@ -0,0 +1,105 @@
...
+ * https://rosettacode.org/wiki/Ramer-Douglas-Peucker_line_simplification
+ * https://karthaus.nl/rdp/
+ *
+ * @param polyline - Polyline to simplify
+ * @param epsilon - A maximum given distance 'epsilon' to decide if a point
+ * should or shouldn't be added the the simplified polyline version. In each
+ * iteration the polyline is split into two polylines and the distance of each
+ * point from those new polylines are checked against the line that connects
+ * the first and last points.
+ * @returns Simplified polyline
```
sedghi, line 20: is this our implementatoin? is the used code open permissive license?

packages/tools/src/utilities/contours/updateContourPolyline.ts
```diff
@@ -14,6 +14,12 @@ import {
  * @param viewport - Viewport
  * @param polylineData - Polyline data (points, winding direction and closed)
  * @param transforms - Methods to convert points to/from canvas and world spaces
+ * @param options - Options
```
sedghi, line 17: I don't see any tool using this option. I expected the simplify object being part of the tools configuration?

sedghi (commented):
see my comments

packages/tools/src/utilities/math/polyline/simplify.ts
lscoder: This is 100% my implementation.

packages/tools/src/utilities/contours/updateContourPolyline.ts
lscoder: Updated spline, freehand and livewire configs

packages/tools/src/tools/annotation/SplineROITool.ts
```diff
@@ -121,6 +120,18 @@ class SplineROITool extends ContourSegmentationBaseTool {
          * modifier must be pressed when the first point of a new contour is added.
          */
         contourHoleAdditionModifierKey: KeyboardBindings.Shift,
+        /**
+         * The polyline may get processed in order to reduce the number of points
+         * for better performance and storage.
+         */
+        decimate: {
+          enabled: true,
```
sedghi, line 128: I prefer enabled to be false as default for now for all tools

packages/tools/src/tools/base/ContourBaseTool.ts
```diff
@@ -210,6 +212,31 @@ abstract class ContourBaseTool extends AnnotationTool {
...
+      points: Types.Point2[];
+      closed?: boolean;
+      targetWindingDirection?: ContourWindingDirection;
+    },
+    transforms: {
+      canvasToWorld: (point: Types.Point2) => Types.Point3;
+    }
+  ) {
+    const decimateConfig = this.configuration?.decimate || {};
+    console.log('>>>>> decimate', {
```
sedghi, line 227: remove

packages/tools/src/tools/annotation/SplineROITool.ts
lscoder: Updated

sedghi (approved):

PR #1065, opened 2024-02-08
fix: segmentSelect issue + segmentation modified event

lscoder:
### Context

- `segmentSelect` example was not working as intended
- `SEGMENTATION_DATA_MODIFIED` was not being triggered for `PlanarFreehandContourSegmentationTool`

### Changes & Results

- changed how `contourSegmentationCompletedListener` finds a viewport with PlanarFreehandContourSegmentationTool tool enabled
- added an event listener to run when an annotation is deleted to update the annotationUIDsMap that contains contour UIDs in each segmentation
- added the contour UID of each annotation created after appending/removing annotations

### DIFF
--- a/packages/tools/src/utilities/getViewportsForAnnotation.ts
+++ b/packages/tools/src/utilities/getViewportsForAnnotation.ts
@@ -1,0 +1,31 @@
+import { getEnabledElements, utilities as csUtils } from '@cornerstonejs/core';
+import type { Annotation } from '../types';
+
+const { isEqual } = csUtils;
+
+/**
+ * Finds a all matching viewports in terms of the orientation of the annotation data
+ * and the frame of reference. This doesn't mean the annotation IS being displayed
+ * on these viewports, just that it could be by navigating the slice, and/or pan/zoom,
+ * without changing the orientation.
+ *
+ * @param annotation - Annotation to find the viewports that it could display in
+ * @returns All viewports to display in
+ */
+export default function getViewportsForAnnotation(annotation: Annotation) {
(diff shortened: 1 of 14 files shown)


packages/tools/src/utilities/getViewportsForAnnotation.ts
```diff
@@ -0,0 +1,31 @@
...
+/**
+ * Finds a all matching viewports in terms of the orientation of the annotation data
+ * and the frame of reference. This doesn't mean the annotation IS being displayed
+ * on these viewports, just that it could be by navigating the slice, and/or pan/zoom,
+ * without changing the orientation.
+ *
+ * @param annotation - Annotation to find the viewports that it could display in
+ * @returns All viewports to display in
+ */
+export default function getViewportsForAnnotation(annotation: Annotation) {
```
sedghi, line 15: we have this already 

packages/tools/src/utilities/getViewportForAnnotation.ts

sedghi (commented):
see my comment thanks

packages/tools/src/utilities/getViewportsForAnnotation.ts
lscoder: ViewportS (plural)
Later it looks for the viewport that is in a toolGroup with freehand tool

lscoder: Pushed a change to make `getViewportForAnnotation` call `getViewportsForAnnotation` removing duplicated code

sedghi (approved):

PR #1066, opened 2024-02-08
fix: Annotation view detection is separated out into multiple places breaking video interpolation

wayfarer3130:
### Context

There are a number of places where it is checked if a view can or should show a given view position, but they are all slightly different and incompatible.  The PR centralizes them into the viewports themselves, and as a benefit, that brings about a speed increase when there are many annotations/viewports.

### Changes & Results

Moved the view detection into the Viewport classes
Fixed VIdeoViewport to ensure it fires image rendered/changed.

### Testing

Run the various contour and segmentation annotation views on both video and stack.  They should all work (note volume does NOT work right now due to a bug in volume)

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  


packages/tools/src/eventListeners/annotations/contourSegmentation/contourSegmentationCompleted.ts
```diff
@@ -64,13 +66,22 @@ export default async function contourSegmentationCompletedListener(
     return;
   }
 
-  if (!isFreehandContourSegToolRegistered(viewport)) {
+  if (
+    !isFreehandContourSegToolRegistered(
+      viewport,
+      sourceAnnotation.metadata.toolName
```
wayfarer3130, line 72: This is probably wrong - but the check down below for handles corresponds to checking if freeform - since that is the actual condition we want it on, I think that one is ok.

packages/tools/src/types/AnnotationTypes.ts
```diff
@@ -28,41 +28,11 @@ type Annotation = {
...
-    viewPlaneNormal?: Types.Point3;
-    /**
-     * The viewUp on which the tool was drawn.
-     */
-    viewUp?: Types.Point3;
-    /**
-     * The FrameOfReferenceUID
-     */
-    FrameOfReferenceUID: string;
+  metadata: Types.ViewTarget & {
```
wayfarer3130, line 31: Should the viewTarget be a child?  I'd like to do so, but there are SO many places it is set.

packages/core/src/types/IViewport.ts
```diff
@@ -6,19 +6,99 @@ import { ActorEntry } from './IActor';
...
+   * changing the normal etc)
+   */
+  asVolume?: boolean;
+  /**
+   * Use this imageURI for testing - may or may not be the current one.
+   * Should be a straight contains URI for the set of imageIds in any of
+   * the volumes or set of image ids.
+   */
+  imageURI?: string;
+};
```
sedghi, line 53: This is not really clear to me the purpose of these options.

packages/core/src/types/IViewport.ts
```diff
@@ -6,19 +6,99 @@ import { ActorEntry } from './IActor';
...
+  /**
+   * VolumeId that the referencedImageId was chosen from
+   */
+  volumeId?: string;
+  /**
+   * The bounds that are shown.  Allows specifying whether a view includes
+   * particular bounds or not.  This will be in world coordinates.
+   */
+  bounds?: BoundsLPS;
 };
```
sedghi, line 102: I like the where these changes are going, but coming to a conclusion that `target` might not be the best naming choice and `reference` is better. We already have referencedImageId, referencedImageIds and referencedVolumeId, referencedSliceIndex. 

Also if we use reference then `forXYZ` prefix becomes redundant as it is implied. Also if you notice 
in your documentation of types you use `to reference...`, `to get a reference for...` a lot which means the explanation
is clearer with `reference` than `target`. e.g., you say `A view target is used to get a reference to a view for a viewport` so ...

So I would suggest we switch from 

`TargetSpecifier` -> `ViewReferenceSpecifier` or `viewReference` then you can rename forFrameOfReference to simple FrameOfReference.

`ViewTarget` -> `ViewReference` or `ViewportReference`

---

also `getViewTarget` to `getViewReference`, since we introduced this recently i think it is fine to change that

sedghi (commented):
see my comments please

packages/core/src/types/IViewport.ts
wayfarer3130: Updated the docs quite a bit.

packages/core/src/RenderingEngine/BaseVolumeViewport.ts
```diff
@@ -534,6 +537,61 @@ abstract class BaseVolumeViewport extends Viewport implements IVolumeViewport {
...
+
+  /**
+   * Find out if this viewport would show this view
+   *
+   * @param options - allows specifying whether the view COULD display this with
+   *                  some modification - either navigation or displaying as volume.
+   * @returns true if the target is compatible with this view
+   */
+  public isReferenceViewable(
+    target: ViewReference,
```
sedghi, line 565: viewRef or viewReference for the param

packages/core/src/RenderingEngine/Viewport.ts
```diff
@@ -1381,6 +1383,29 @@ class Viewport implements IViewport {
     return { widthWorld: maxX - minX, heightWorld: maxY - minY };
   }
 
+  public getViewReference(
+    forTarget: ViewReferenceSpecifier = {}
```
sedghi, line 1387: forTarget -> viewRefSpecifier

packages/core/src/RenderingEngine/Viewport.ts
```diff
@@ -1381,6 +1383,29 @@ class Viewport implements IViewport {
...
+      FrameOfReferenceUID: this.getFrameOfReferenceUID(),
+      cameraFocalPoint,
+      viewPlaneNormal,
+      sliceIndex: forTarget.sliceIndex ?? this.getCurrentImageIdIndex(),
+    };
+    return target;
+  }
+
+  public isReferenceViewable(
+    target: ViewReference,
```
sedghi, line 1400: ditto

packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -2827,9 +2832,57 @@ class StackViewport extends Viewport implements IStackViewport, IImagesLoader {
...
   };
 
-  public getTargetId(specifier: TargetSpecifier = {}): string {
-    const { sliceIndex: imageIdIndex = this.currentImageIdIndex } = specifier;
-    return `imageId:${this.imageIds[imageIdIndex]}`;
+  /**
+   * Checks to see if this target is or could be shown in this viewport
+   */
+  public isReferenceViewable(
+    target: ViewReference,
```
sedghi, line 2839: target -> viewRef

packages/core/src/types/IViewport.ts
```diff
@@ -6,19 +6,112 @@ import { ActorEntry } from './IActor';
...
-  /** True to get a frame of reference UID reference instead of a regular image one */
+  sliceIndex?: number | [number, number];
+  /**
+   * true to get a frame of reference UID reference instead of a regular image one,
+   * or false to not get it when it would otherwise do so.
+   * The default depends on the view type, but in general showing multiple images
+   * combined, the return will be only for frame of reference, while showing a
+   * single one, it will be for frame of reference.
+   */
   forFrameOfReference?: boolean;
```
sedghi, line 27: we can use FrameOfReference here instead of forFOR

packages/core/src/types/IViewport.ts
```diff
@@ -6,19 +6,112 @@ import { ActorEntry } from './IActor';
...
   /** The volumeId to reference */
   volumeId?: string;
+  /**
+   * true to include the bounds, if it would not otherwise do so, or
+   * false to exclude them if it would normally do so.
+   * Bounds are normally only included if the viewport is very magnified
+   * AND the points are included in the request, or the view is for a range
+   * of slices.
+   */
+  includeBounds?: boolean;
```
sedghi, line 39: this seems out of place, the rest has another nature, this is boolean, maybe this should be inside options?

sedghi (commented):
see my comments please

packages/core/src/types/IViewport.ts
wayfarer3130: No, because that isn't a frame of reference, it is indicating to create a view reference that is for a specific frame of reference rather than a specific volume or image.

packages/core/src/RenderingEngine/Viewport.ts
```diff
@@ -1381,6 +1383,45 @@ class Viewport implements IViewport {
...
+      !isEqual(
+        vec3.negate(camera.viewPlaneNormal, camera.viewPlaneNormal),
+        viewPlaneNormal
+      )
+    ) {
+      // Could navigate as a volume to the reference
+      return options?.asVolume === true;
+    }
+    return true;
+  }
```
sedghi, line 1423: this should be a no op abstract method that each children implement in my opinion. I don't like the fact that we have FOR and asVolume in the base viewport

wayfarer3130: But you already have frame of reference and camera as basic concepts - I'm not using anything that isn't already there.  I really want these to be in the base viewports as they will apply for all 4 types (including WSI)  Have a base class is kind of the point of inheritance, and this functionality applies across the board.

sedghi (approved):

PR #1070, opened 2024-02-09
fix: multiple contour interpolation, holes and combinePolyline in contours

wayfarer3130:
### Context

Adds the ability to have multiple contours getting interpolation, by excluding contours already accepted or contours which are co located on the same plane.  Also add holey support to add holes into freeform segmentations.

### Changes & Results
Add lots of little fixes how the selection occurs for interpolation as well as how it occurs for contour combine and hole addition.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  


packages/tools/test/ContourInterpolation_test.js
```diff
@@ -667,7 +667,7 @@ describe('Contours Interpolation: ', () => {
       }
     });
 
-    it('Should successfully edit auto generated contour annotation', function (done) {
+    fit('Should successfully edit auto generated contour annotation', function (done) {
```
sedghi, line 670: remove

packages/tools/src/utilities/math/polyline/getLineSegmentsIntersection.ts
```diff
@@ -30,7 +30,10 @@ export default function getLineSegmentsIntersection(
   const denominator = diffQ2P2[1] * diffQ1P1[0] - diffQ2P2[0] * diffQ1P1[1];
 
   if (denominator == 0) {
-    return;
+    return [
```
sedghi, line 33: @lscoder what do you think

packages/tools/src/eventListeners/annotations/contourSegmentation/contourSegmentationCompleted.ts
```diff
@@ -29,6 +29,7 @@ import { PlanarFreehandContourSegmentationTool } from '../../../tools';
 import type { Annotation } from '../../../types';
 import type { ContourAnnotation } from '../../../types/ContourAnnotation';
 import { ContourWindingDirection } from '../../../types/ContourAnnotation';
+import { triggerAnnotationModified } from '../../../stateManagement/annotation/helpers/state';
```
sedghi, line 32: @lscoder pleaes review changes to this file which you mostly wrote

packages/tools/src/tools/annotation/PlanarFreehandROITool.ts
```diff
@@ -191,6 +191,8 @@ class PlanarFreehandROITool extends ContourSegmentationBaseTool {
           radius: 2,
         },
         allowOpenContours: true,
+        // Allow combining by clicking outside/dragging into this contour
+        allowCombine: true,
```
sedghi, line 195: not sure what is this

sedghi (commented):
see my comments please

packages/tools/src/types/ToolSpecificAnnotationTypes.ts
```diff
@@ -260,18 +260,25 @@ export type PlanarFreehandROIAnnotation = ContourAnnotation & {
 export type PlanarFreehandContourSegmentationAnnotation =
   PlanarFreehandROIAnnotation & ContourSegmentationAnnotationData;
 
-export type InterpolationROIAnnotation = ContourAnnotation & {
-  metadata: {
-    annotationUID?: string;
+export type InterpolationROIAnnotation = ContourAnnotation &
+  ContourSegmentationAnnotationData & {
+    metadata: {
+      annotationUID?: string;
```
lscoder, line 266: I know you didn't change this now but do you know the difference between `annotation.annotationUID` from `Annotation` and `annotation.metadata.annotationUID` from `InterpolationROIAnnotation?

packages/tools/src/eventListeners/annotations/contourSegmentation/contourSegmentationCompleted.ts
```diff
@@ -91,12 +92,12 @@ export default async function contourSegmentationCompletedListener(
   }
 }
 
-function isFreehandContourSegToolRegisteredForViewport(
+/** Indicates if this contour tool can handle combining */
+function isContourToolCombineAvailable(
```
lscoder, line 96: The goal was to check only if freehand contour segmentation is enabled because when we append/remove contour tools they get converted into freehand contour segmentation and if that tool is not available the annotation is "removed" from the viewport. This change now seems to check if the tool used to draw the annotation is available which should always return true, right?

lscoder: It still needs to check if `isFreehandContourSegToolRegisteredForViewport` is available when appending/removing contours

packages/tools/src/tools/annotation/PlanarFreehandROITool.ts
lscoder: This can be removed after adding something like an `if(contourSegUtils.isInterpolatedContourAnnotation(annotation)) return;` in `contourSegmentationCompleted.ts`

packages/tools/src/utilities/math/polyline/getLineSegmentsIntersection.ts
lscoder: Returning only a mid point does not look right because the intersection would be an infinity number of points (range). In which case this happens and why the midpoint would solve the problem?

lscoder: Also the function is returning wrong values:
- ([0, 0], [4, 4], [5, 5], [10, 10]): 4.75 (the lines do not even intersect)
- ([0, 0], [4, 4], [3, 3], [20, 20]): 6.75 (The intersection goes from [3, 3] to [4, 4])

packages/tools/src/types/ToolSpecificAnnotationTypes.ts
wayfarer3130: No idea

packages/tools/src/utilities/math/polyline/getLineSegmentsIntersection.ts
wayfarer3130: That was pre-existing - the lines never needed to intersect to return a value.  Will write a unit test and add a flag to have true intersection or midpoint etc.  Returning undefined isn't handled by either use case, and I don't know enough about those use cases to fix them.

lscoder: Issue resolved with the changes I pushed

PR #1079, opened 2024-02-13
fix: Combine polyline interpolation breaks annotation state data

wayfarer3130:
### Context

If the list returned by getAnnotations is modified, it directly affects the underlying data object, breaking further interpolation.

Also fixes [#1091]

### Changes & Results

Fixed the interpolation retrieve check to take a copy of the data returned from getAnnotations.

### Testing

Create two spline interpolations, one slice apart.  Combine an annotation.  The original annotations shouldn't go blank.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  


packages/tools/src/utilities/contours/interpolation/getInterpolationData.ts
```diff
@@ -45,21 +45,37 @@ export default function getInterpolationData(
...
+      modifiedAnnotations.forEach((annotation) => {
+        const { metadata } = annotation;
+        if (
+          metadata.originalToolName === testToolName &&
+          metadata.originalToolName !== metadata.toolName
+        ) {
+          annotations.push(annotation);
+        }
+      });
+    }
```
lscoder, line 78: It seems that `annotations` already have all modified annotations you are looking for, no? Would this code work instead of having to get all annotations again?

```javascript
const modifiedAnnotations = annotations.filter(({ metadata }) =>
  metadata.originalToolName !== metadata.toolName
)
```

packages/tools/test/ContourInterpolation_test.js
```diff
@@ -667,114 +675,116 @@ describe('Contours Interpolation: ', () => {
...
-      this.stackToolGroup.addViewport(vp.id, this.renderingEngine.id);
-
-      try {
-        vp.setStack(imageIds, 0);
-        this.renderingEngine.render();
-      } catch (e) {
-        done.fail(e);
-      }
-    });
+    // it('Should successfully edit auto generated contour annotation', function (done) {
```
lscoder, line 678: Why these unit tests were removed?

packages/tools/src/utilities/contours/interpolation/getInterpolationData.ts
wayfarer3130: No, that isn't true - annotations has all the annotations with the original tool name, so suppose that is "Spline", then one needs to add to the annotations list all the PlanarFreehand annotations which were originally Spline.
That also needs to work to get all the PlanarFreehand annotations which were originally PlanarFreehand (but are now edited annotations).

packages/tools/test/ContourInterpolation_test.js
wayfarer3130: Random failures due to other changes - haven't figured out what is going on there yet.  I'm adding back the other integration tests, but I don't know why these ones fail.  I THINK what is happening is that the combine code is getting activated.

wayfarer3130:
@lscoder  - I've responded to both your comments.  If that is ok, could you approve it and I will merge?  The defect is currently corrupting internal annotation state, so I'd like to get it this far.

lscoder (approved):

PR #1080, opened 2024-02-13
fix(polySeg): able to handle holes inside contour segmentation

sedghi:
- Move the view target detection into core for consistency
- fix wrong name issue in video viewport
- fix: video contour segmentations displaying and interpolating
- fix: View rendering
- Fix contour segmentation appearing on wrong viewport view
- NPE
- PR changes
- Fix multiple contours issues
- Fix annotation visibility on volume from stack
- fix registration of contours and holey contours
- fix tests
- fix docs
- Fix tests
- PR changes
- Fix interpolation combining
- test: Add unit test for getLineSegmentsIntersection
- Added edge condition test
- fix: return midpoint for overlapping line segments
- fix: invalid reference to getLineSegmentIntersections in jest file
- fix: unit tests
- Fix build
- code review :: renamed 'containsPoint' to 'isPointOnLineSegment'
- api
- Add new utility function getRandomSampleFromArray
- Add support for shift + left click in PolySegWasmContourToVolumeLabelmap
- api

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  


packages/tools/src/utilities/math/polyline/projectTo2D.ts
```diff
@@ -0,0 +1,48 @@
...
+import type { Types } from '@cornerstonejs/core';
+
+/**
+ * Projects a polyline from 3D to 2D by reducing one dimension.
+ *
+ * @param polyline - The polyline to be projected.
+ * @returns An object containing the shared dimension index and the projected polyline in 2D.
+ * @throws Error if a shared dimension index cannot be found for the polyline.
+ */
+export function projectTo2D(polyline: Types.Point3[]) {
```
sedghi, line 11: refactored out

packages/core/src/utilities/getRandomSampleFromArray.ts
```diff
@@ -0,0 +1,20 @@
...
+ * @param array - The input array.
+ * @param size - The number of elements to sample.
+ * @returns A new array containing the randomly sampled elements.
+ */
+export function getRandomSampleFromArray(array, size) {
+  const result = [];
+  const len = array.length;
+  const set = new Set(); // Use a Set to track selected indices for O(1) lookups
+
+  while (result.length < size) {
```
lscoder, line 12: If `size` is greater than `array.length` this function will never return and if the size is greater than 50% the number of "hits" (when the set already contains the number) will increase considerably and we will never know when it will return.

I would implement it in a more predictable way. Example, sort using random + slice N first elements.

lscoder: There is a method in lodash that do something similar:
- clone the array
- loop through all elements and choose a random index to switch with the current element
- take the first N elements

`sampleSize`
https://github.com/lodash/lodash/blob/4.17.15/lodash.js#L9791-L9799

`arraySampleSize`
https://github.com/lodash/lodash/blob/4.17.15/lodash.js#L2430C14-L2432

`shuffleSelf`
https://github.com/lodash/lodash/blob/4.17.15/lodash.js#L6711-L6726

packages/tools/examples/PolySegWasmContourToVolumeLabelmap/index.ts
```diff
@@ -146,6 +146,10 @@ async function run() {
       {
         mouseButton: MouseBindings.Primary,
       },
+      {
+        mouseButton: MouseBindings.Primary, // Shift + Left Click
+        modifierKey: KeyboardBindings.Shift,
+      },
```
lscoder, line 152: I have 90% of the changes to remove this binding. I'll get it done and create a PR after the demo.

packages/tools/src/stateManagement/segmentation/polySeg/Labelmap/convertContourToLabelmap.ts
```diff
@@ -256,14 +256,42 @@ function _getAnnotationMapFromSegmentation(
...
 
-  const annotationUIDsInSegmentMap = new Map<number, Annotation[]>();
+  const annotationUIDsInSegmentMap = new Map<number, any>();
   segmentIndices.forEach((index) => {
     const annotationUIDsInSegment = annotationMap.get(index);
 
-    const annotations = Array.from(annotationUIDsInSegment).map((uid) => {
-      const annotation = getAnnotation(uid);
+    // Todo: there is a bug right now where the annotationUIDsInSegment has both
+    // children and parent annotations, so we need to filter out the parent
```
lscoder, line 264: Should I make the changes in `contourSegmentationCompleted` to remove it?

packages/tools/src/stateManagement/segmentation/polySeg/Labelmap/convertContourToLabelmap.ts
```diff
@@ -256,14 +256,42 @@ function _getAnnotationMapFromSegmentation(
...
+
+    uids = uids.filter((uid) => {
+      const annotation = getAnnotation(uid) as ContourAnnotation;
+      return !uids.some((otherUid) => {
+        const parentAnnotation = getAnnotation(otherUid) as ContourAnnotation;
+        return parentAnnotation.childAnnotationUIDs?.includes(
+          annotation.annotationUID
+        );
+      });
+    });
```
lscoder, line 277: The child annotation also has its parent UID. Then you can rewrite this as:

```javascript
uids = uids.filter((uid) => 
  !(getAnnotation(uid) as Annotation).parentAnnotationUID
);
```

packages/core/src/utilities/getRandomSampleFromArray.ts
lscoder: I saw the size being limited by the caller in `projectTo2D.ts` (`Math.min(50, polyline.length))  but I think it should be here otherwise any other function that calls this method will have to do the same to avoid an infinity loop.

packages/tools/src/utilities/math/polyline/projectTo2D.ts
```diff
@@ -0,0 +1,48 @@
...
+  // We need to reduce one dimension to 2D, so basically
+  // we need to find the dimension index that is shared by all points
+  // Use the first three points, two is enough but three is more robust
+  let sharedDimensionIndex;
+
+  const testPoints = utilities.getRandomSampleFromArray(
+    polyline,
+    Math.min(50, polyline.length)
```
lscoder, line 19: `Math.min(50, polyline.length)` should be moved to `getRandomSampleFromArray`

packages/tools/src/utilities/math/polyline/projectTo2D.ts
```diff
@@ -0,0 +1,48 @@
...
+  );
+  for (let i = 0; i < 3; i++) {
```
lscoder, line 21: I didn't get why `getRandomSampleFromArray` is called with `size` set to 50 but you are using only 3. Why not get only 3 points instead of 50?

packages/tools/src/utilities/math/polyline/projectTo2D.ts
```diff
@@ -0,0 +1,48 @@
...
+    if (testPoints.every((point, index, array) => point[i] === array[0][i])) {
+      sharedDimensionIndex = i;
+      break;
+    }
```
lscoder, line 25: Is it possible to have precision error? Should it consider an `epsilon` threshold?

packages/tools/src/utilities/math/polyline/projectTo2D.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  // convert polyline list and point to 2D
+  const points2D = [] as Types.Point2[];
+
+  for (let i = 0; i < polyline.length; i++) {
+    points2D.push([
+      polyline[i][(sharedDimensionIndex + 1) % 3],
+      polyline[i][(sharedDimensionIndex + 2) % 3],
+    ]);
+  }
```
lscoder, line 42: `(sharedDimensionIndex + 1) % 3` and `(sharedDimensionIndex + 2) % 3` never change and can be stored in variables because `%` (mod) is a very expensive operation and that will degrade the performance considerably.

packages/tools/src/workers/polySegConverters.js
```diff
@@ -274,15 +294,29 @@ const polySegConverters = {
...
+        });
+
         // Run the pointInShapeCallback for the combined bounding box
         pointInShapeCallback(
           imageData,
           (pointLPS) => {
+            const point2D = [
+              pointLPS[(sharedDimensionIndex + 1) % 3],
+              pointLPS[(sharedDimensionIndex + 2) % 3],
+            ];
```
lscoder, line 312: Same here (`%` = -perf)

packages/tools/src/utilities/math/polyline/projectTo2D.ts
lscoder: This wouldn't work for oblique, right?

packages/tools/src/utilities/math/polyline/projectTo2D.ts
lscoder: If `sharedDimensionIndex` is equal to:
- 0 then [y, z]
- 1 then [z, x] (wouldn't get into an unexpected behavior in this case?)
- 2 then [x, y]

If the answer is yes, I would do:

```javascript
const [firstDim, secondDim] = [0, 1, 2].toSpliced(sharedDimensionIndex, 1)
```

packages/tools/src/stateManagement/segmentation/polySeg/Labelmap/convertContourToLabelmap.ts
sedghi: Yes later

packages/tools/src/utilities/math/polyline/projectTo2D.ts
sedghi: No it throws

packages/tools/src/workers/polySegConverters.js
```diff
@@ -177,15 +185,29 @@ const polySegConverters = {
...
```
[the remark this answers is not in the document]
sedghi: Good catch

packages/core/src/utilities/getRandomSampleFromArray.ts
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * Shuffles an array
+ * @param array - The array to shuffle.
+ */
+function shuffleArray<T>(array: T[]): void {
+  for (let i = array.length - 1; i > 0; i--) {
+    const j = Math.floor(Math.random() * (i + 1));
+    [array[i], array[j]] = [array[j], array[i]]; // Swap elements
+  }
+}
```
lscoder, line 30: :+1: 
This way it runs in `O(n)` (faster) while the other option would run in `O(n log(n))` (slower)

lscoder (approved):

PR #1099, opened 2024-02-15
fix(brush cursor): only show innter circle on dynamic threshold

### LINKED ISSUE #1086: [Bug] Dynamic threshold brush tool inner circle size
### Describe the Bug

Seems like we now have the inner circle shown for all the brush tools even if they are not threshold brush. 

### Steps to Reproduce

https://www.cornerstonejs.org/live-examples/labelmapsegmentationdynamicthreshold

### The current behavior

s

### The expected behavior

1. We should only show the inner circle for the dynamic threshold brush tool
2. It should be dynamically size with respect to its radius, right now it is fixed size

### OS

d

### Node version

d

### Browser

d

sedghi:
fixes #1086 

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  


packages/tools/src/tools/segmentation/BrushTool.ts
```diff
@@ -584,6 +584,33 @@ class BrushTool extends BaseTool {
       rightCursorInWorld,
     ];
 
+    const activeStrategy = this.configuration.activeStrategy;
+
+    // Note: i don't think this is the best way to implement this
+    // but don't think we have a better way to do it for now
+    if (activeStrategy.startsWith('THRESHOLD')) {
```
wayfarer3130, line 591: Instead of this, create a new active strategy callback that returns configuration values, can call that method and get the data from the response.  That way you can change the inner circle size based on whatever the strategy decides.

sedghi: Done

wayfarer3130 (approved):

Issue #1088, opened 2024-02-14
[Bug] window level tool for video


sedghi:
### Describe the Bug

https://www.cornerstonejs.org/live-examples/videocolor

It feels like the multiplier is too big and it shows very sudden changes in the brightness and contrast	

### Steps to Reproduce

d

### The current behavior

d

### The expected behavior

Most likely we should decrease that multiplier which should be easy

### OS

d

### Node version

d

### Browser

d

PR #1102, opened 2024-02-16
fix: windowLevel dynamic range from video viewport (#1088)

lscoder:
fixes https://github.com/cornerstonejs/cornerstone3D/issues/1088
### Context

WindowLevel tool was using a higher multiplier resulting in much faster changes to window width and window level making it harder to adjust.

### Changes & Results

Changed the function responsible for getting the range from the viewport which was returning an array instead of a single number breaking the calculation of the ratio (`NaN`).

### Testing

- Open [videoColor](https://www.cornerstonejs.org/live-examples/videocolor) example
- Click on the viewport to change the ww/wc

### DIFF
--- a/common/reviews/api/tools.api.md
+++ b/common/reviews/api/tools.api.md
@@ -6034,7 +6034,7 @@ export class WindowLevelTool extends BaseTool {
     // (undocumented)
     _getImageDynamicRangeFromMiddleSlice: (scalarData: any, dimensions: any) => number;
     // (undocumented)
-    _getImageDynamicRangeFromViewport(viewport: any): any;
+    _getImageDynamicRangeFromViewport(viewport: any): number;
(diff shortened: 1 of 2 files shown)


sedghi (approved):
Much better thanks. 

Update API please

reference the ticket to so that it gets closed  https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests

lscoder:
> Much better thanks.
> 
> Update API please
> 
> reference the ticket to so that it gets closed https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests

I'm manually linking my PRs to the bugs (see on right-side panel)

PR #1105, opened 2024-02-17
feat(volumeviewport): return volume specific colormap information

IbrahimCSAE:
### Context

We need to return the correct colormap information if there's more than 1 volume on the viewport

### Changes & Results

- Added getColormap function that attempts to identify the currently set colormap on the volume.

### Comments

i’m not sure if we should say “attempts” as from my testing this seems to return the correct opacity and colormap 100% of the time. but I’m not a vtk expert, so I can’t say for certain that this is 100% accurate.

### DIFF
--- a/packages/core/src/RenderingEngine/BaseVolumeViewport.ts
+++ b/packages/core/src/RenderingEngine/BaseVolumeViewport.ts
@@ -62,8 +62,6 @@ import type { vtkSlabCamera as vtkSlabCameraType } from './vtkClasses/vtkSlabCam
 import vtkSlabCamera from './vtkClasses/vtkSlabCamera';
 import transformWorldToIndex from '../utilities/transformWorldToIndex';
 import { getTransferFunctionNodes } from '../utilities/transferFunctionUtils';
-import isEqual from '../utilities/isEqual'
-
 /**
  * Abstract base class for volume viewports. VolumeViewports are used to render
  * 3D volumes from which various orientations can be viewed. Since VolumeViewports
--- a/packages/core/src/RenderingEngine/BaseVolumeViewport.ts
+++ b/packages/core/src/RenderingEngine/BaseVolumeViewport.ts
@@ -62,6 +62,7 @@ import type { vtkSlabCamera as vtkSlabCameraType } from './vtkClasses/vtkSlabCam
 import vtkSlabCamera from './vtkClasses/vtkSlabCamera';
 import transformWorldToIndex from '../utilities/transformWorldToIndex';
 import { getTransferFunctionNodes } from '../utilities/transferFunctionUtils';
+import isEqual from '../utilities/isEqual'
 
 /**
  * Abstract base class for volume viewports. VolumeViewports are used to render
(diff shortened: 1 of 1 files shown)


packages/core/src/RenderingEngine/BaseVolumeViewport.ts
```diff
@@ -282,6 +282,7 @@ abstract class BaseVolumeViewport extends Viewport implements IVolumeViewport {
     cfun.setMappingRange(range[0], range[1]);
     volumeActor.getProperty().setRGBTransferFunction(0, cfun);
 
+    // keeping this incase something is already using it (but this is wrong)
```
sedghi, line 285: please explain why this is wrong

packages/core/src/RenderingEngine/BaseVolumeViewport.ts
```diff
@@ -789,7 +790,7 @@ abstract class BaseVolumeViewport extends Viewport implements IVolumeViewport {
     }
 
     const {
-      colormap,
+      colormap: defaultColormap,
```
sedghi, line 793: why you call it default? is it like `current`?

packages/core/src/RenderingEngine/BaseVolumeViewport.ts
```diff
@@ -816,8 +817,10 @@ abstract class BaseVolumeViewport extends Viewport implements IVolumeViewport {
 
     const voiRange = voiRanges.length ? voiRanges[0].voiRange : null;
 
+    const colormap = this.getColormap(applicableVolumeActorInfo);
+
     return {
-      colormap: colormap,
+      colormap: volumeId ? colormap || defaultColormap : defaultColormap,
```
sedghi, line 823: please break this into multiple lines, I can't follow the logic here

packages/core/src/RenderingEngine/BaseVolumeViewport.ts
```diff
@@ -827,6 +830,61 @@ abstract class BaseVolumeViewport extends Viewport implements IVolumeViewport {
     };
   };
 
+
+  /**
+   *
```
sedghi, line 835: pleaes add function docs

packages/core/src/RenderingEngine/BaseVolumeViewport.ts
```diff
@@ -827,6 +830,61 @@ abstract class BaseVolumeViewport extends Viewport implements IVolumeViewport {
...
+                presetRGBPoints[i + 1] !== RGBPoints[i + 1] ||
+                presetRGBPoints[i + 2] !== RGBPoints[i + 2] ||
+                presetRGBPoints[i + 3] !== RGBPoints[i + 3]
+            ) {
+                return false;
+            }
+        }
+
+        return true;
+    });
```
sedghi, line 867: you should be Epsilon compare (you can use isEqual which can run on vectors too)

sedghi (commented):
see my comments please, thanks

sedghi (approved):

PR #1109, opened 2024-02-20
fix(livewire): polyline and handles (winding direction)

lscoder:


sedghi:
Failing ci
Please fix

packages/tools/src/tools/annotation/LivewireContourTool.ts
```diff
@@ -928,22 +924,29 @@ class LivewireContourTool extends ContourSegmentationBaseTool {
...
-    const { annotation, sliceToWorld } = this.editData;
+    const { annotation, sliceToWorld, worldToSlice, closed, newAnnotation } =
+      this.editData;
     let { pointArray: imagePoints } = livewirePath;
 
     if (imagePoints.length > 1) {
       imagePoints = [...imagePoints, imagePoints[0]];
     }
 
+    // Save the annotation in clockwise winding direction only after closing it
```
wayfarer3130, line 935: Thank you for the comment - I appreciate that because I wasn't sure how you were going to handle this one.

packages/tools/src/tools/base/ContourBaseTool.ts
```diff
@@ -219,6 +219,7 @@ abstract class ContourBaseTool extends AnnotationTool {
     },
     transforms: {
```
wayfarer3130, line 220: This this be on base annotation tool?  Not sure that belongs in this PR, but seems like having the transforms available separately is a useful addition.

packages/tools/src/utilities/math/polyline/getSignedArea.ts
```diff
@@ -5,9 +5,9 @@ import type { Types } from '@cornerstonejs/core';
  * https://www.youtube.com/watch?v=GpsKrAipXm8&t=1900s
  *
  * This functions has a runtime very close to `getArea` and it is recommended to
- * be called only if you need the area signal (eg: calculate polygon normal). If
- * you do not need the area signal you should always call `getArea`.
- *
+ * be called only if you need the area signal (eg: calculate polygon normal or
+ * winding direction). If you do not need the area signal you should always call
+ * `getArea`.
```
wayfarer3130, line 10: I'm not sure about the performance of the getSignedArea now, but if you want a bit more performance, then passing in a TypedArray<Types.Point2> and directly accessing the array data will be faster since you avoid creating so many tiny arrays.  Again, not part of this PR, but something to consider for the future.

wayfarer3130 (approved):

PR #1110, opened 2024-02-20
fix(freehandROI): moving mouse 1px back deletes the entire contour

lscoder:


packages/tools/src/utilities/math/polyline/addCanvasPointsToArray.ts
```diff
@@ -16,11 +16,16 @@ const addCanvasPointsToArray = (
   const enabledElement = getEnabledElement(element);
   const { viewport } = enabledElement;
 
+  if (!canvasPoints.length) {
+    canvasPoints.push(newCanvasPoint);
+    console.log('>>>>> !canvasPoints. :: RETURN');
```
sedghi, line 21: remove

sedghi (commented):
much better now thanks

sedghi (approved):

Issue #1094, opened 2024-02-14
[Bug] interpolate on contour holes as well


sedghi:
### Describe the Bug

d

### Steps to Reproduce

d

### The current behavior

d

### The expected behavior

d

### OS

d

### Node version

d

### Browser

d

PR #1112, opened 2024-02-21
feat: Add contour hole interpolation

wayfarer3130:
### Context

fixes https://github.com/cornerstonejs/cornerstone3D/issues/1094

Add contour hole interpolation.  The way this works is that the interpolation happens after the holes have been
created on the parent elements.  This then has the parentAnnotationUID already assigned, which is used to grab the parent interpolation UIDs.  These UIDs are then used positionally to assign new interpolation UIDs to the holes.  Then, that allows interpolating the holes directly.

### Testing

yarn example on videoContourSegmentation and on interpolationContourSegmentation
Create spline, freeform and livewire holes on the source interpolations, then see them on the intermediate results.
There are some rendering issues with the style on the intermediate ones when they are complete, but the parent is not (the child always takes the status of the parent interpolation.)

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  


packages/tools/examples/interpolationContourSegmentation/index.ts
```diff
@@ -304,10 +300,6 @@ function addBindings(toolGroupId) {
       {
         mouseButton: MouseBindings.Primary, // Left Click
       },
-      {
-        mouseButton: MouseBindings.Primary, // Shift + Left Click
-        modifierKey: KeyboardBindings.Shift,
-      },
     ],
```
sedghi, line 303: This will make the hole drawing not work, I guess Leo had some other PRs in hand to fix this so that we don't need to specify it here, right now when I want to draw a hole it does not work in examples @lscoder

sedghi:
I see some weird issues 

https://github.com/cornerstonejs/cornerstone3D/assets/7490180/4b682185-6bf4-4e28-8afd-fcb917abb5ef

PR #1117, opened 2024-02-23
fix(calibration tool): fix calibration type not being rendered on measurements 

IbrahimCSAE:
### Context

Fixes calibration types not rendering

### Changes & Results

add types

### DIFF
--- a/packages/tools/src/utilities/getCalibratedUnits.ts
+++ b/packages/tools/src/utilities/getCalibratedUnits.ts
@@ -173,7 +173,13 @@ const getCalibratedLengthUnitsAndScale = (image, handles) => {
     }
   } else if (calibration.scale) {
     scale = calibration.scale;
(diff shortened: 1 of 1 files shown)


packages/tools/src/utilities/getCalibratedUnits.ts
```diff
@@ -173,7 +173,13 @@ const getCalibratedLengthUnitsAndScale = (image, handles) => {
     }
   } else if (calibration.scale) {
     scale = calibration.scale;
-    calibrationType = calibration?.type;
+  }
+
+  // everything except REGION/Uncalibratted
```
IbrahimCSAE, line 178: REGION is excluded because it looks like there's handling above that adds it, and for uncalibrated, we don't display that one.

sedghi (approved):

PR #1128, opened 2024-02-28
feat(ROIThresholdsTools): Made the Rectangle and Circle ROIStartEndThresholds works on other planes

Celian-abd:
### Context
Updated version of the Rectangle/Circle ROIStartEndThreshold tools so they can work on other then acquisition plane.

### Changes & Results
- Update calculateCachedStats for other plane

### Testing
yarn run example RectangleROIStartEndThreshold or on CircleROIStartEndThreshold

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  


packages/tools/examples/CircleROIStartEndThreshold/index.ts
```diff
@@ -127,23 +124,24 @@ addButtonToToolbar({
       return;
     }
 
+    const viewport = annotation.metadata.enabledElement.viewport;
+
     // get the current slice Index
-    const sliceIndex = viewport[0].getCurrentImageIdIndex();
+    const sliceIndex = viewport.getCurrentImageIdIndex();
+    console.debug(annotation.data.startSlice)
```
sedghi, line 131: remove please

packages/tools/examples/RectangleROIStartEndThreshold/preset.js
```diff
@@ -0,0 +1,229 @@
+export default {
```
sedghi, line 1: can you import from some other file that we use this instead of duplicating?

packages/tools/src/utilities/planar/filterAnnotationsWithinPlan.ts
```diff
@@ -0,0 +1,77 @@
...
+ * given some `Annotations`, and the slice defined by the camera's normal
+ * direction and the spacing in the normal, filter the `Annotations` which
+ * is within the slice.
+ *
+ * @param annotations - Annotations
+ * @param camera - The camera
+ * @param spacingInNormalDirection - The spacing in the normal direction
+ * @returns The filtered `Annotations`.
+ */
+export default function filterAnnotationsWithinSamePlan(
```
sedghi, line 20: I'm not sure why we need this function, since annotationWithinSameSlice already checks for the same plane (`annotationsWithParallelNormals`)

packages/tools/src/tools/segmentation/CircleROIStartEndThresholdTool.ts
```diff
@@ -212,6 +219,23 @@ class CircleROIStartEndThresholdTool extends CircleROITool {
     return annotation;
   };
 
+  _arraysEqual = (a: Types.Point3, b: Types.Point3): boolean => {
```
sedghi, line 222: we have core.isEqual

packages/tools/src/tools/segmentation/RectangleROIStartEndThresholdTool.ts
```diff
@@ -561,29 +604,29 @@ class RectangleROIStartEndThresholdTool extends RectangleROITool {
...
+    const mprValues = CONSTANTS.MPR_CAMERA_VALUES
+    if (this._arraysEqual(viewPlaneNormal, mprValues.axial.viewPlaneNormal)) {
+      return Math.round(imageIdIndex[2]);
+    } else if (this._arraysEqual(viewPlaneNormal, mprValues.sagittal.viewPlaneNormal)) {
+      return Math.round(imageIdIndex[0]);
+    } else if (this._arraysEqual(viewPlaneNormal, mprValues.coronal.viewPlaneNormal)) {
+      return Math.round(imageIdIndex[1]);
+    } else {
+      return undefined;
     }
```
sedghi, line 629: I'm not sure if this is a correct logic, basically the volume has x,y,z , but it might be a sagittal acquistion volume, but still the z is the depth though sagittal. The logic is more complext than this, example of such volume can be found in this study https://viewer-dev.ohif.org/viewer?StudyInstanceUIDs=1.3.6.1.4.1.25403.345050719074.3824.20170125095438.5

packages/tools/src/utilities/planar/filterAnnotationsWithinPlan.ts
Celian-abd: `annotationWithinSameSlice` was returning the annotations that were on the same plane and slice aswell, but we wanted all the annotations for the same plane only.

packages/tools/src/tools/segmentation/RectangleROIStartEndThresholdTool.ts
Celian-abd: Yes I agree this may not be the good solution but I didn't find how to get the slice we project to get without checking in what plane we are. Maybe you can help us with that.

Celian-abd:
@sedghi  We've made some changes to the way we were doing the 3D ROI. Now there is no more sliceIndex but we are using instead the world coordinates of where the annotation is drawn and we compare it to the focalpoint of the camera.

Also we added the calculation of the basic stats with all the points contained in the 3D roi and the posibility to show the textbost like we have the for the other tools.

sedghi (commented):
Sorry for the delay. I'm back to contributing to open source much more now.

I see some weird behavior.

https://share.cleanshot.com/3pGRlKPZ

Celian-abd:
@sedghi I think you might have tested an older version.

Here is what I have : https://streamable.com/yx1qd4

salimkanoun:
Hi @sedghi, is there a chance you could review this PR ? We enhanced it, i think it is ready for review

sedghi:
Yes, we released version 3.8.1, so I will review this issue. Apologies for the delay.

sedghi (approved):
Great addition, thanks for the pull request, and apologies for the delay in merging it.

salimkanoun:
@sedghi thank you ! Next week we will make you a PR to update the command that set the start / stop slice in ohif as now the z limits are set by position and not to the viewport slice index

salimkanoun:
(PR in ohif not cornerstone)

Issue #985, opened 2024-01-05
[Bug] Image does not rendered in proper way (GPU rendering)


Max-Kharitonov:
### Describe the Bug

In case if we are using CPU rendering it works fine, but in case GPU this image inverted.
Please see attached anonimized file
[DICOM sample Issue with Invertion.zip](https://github.com/cornerstonejs/cornerstone3D/files/13840630/DICOM.sample.Issue.with.Invertion.zip)

### Steps to Reproduce

run GPU rendering
![image](https://github.com/cornerstonejs/cornerstone3D/assets/124146469/2b4f9647-7ca2-4762-a430-74d60b11bb11)

run CPU rendering
![image](https://github.com/cornerstonejs/cornerstone3D/assets/124146469/b4275442-32f5-4833-86c6-c7081124a7d7)

### The current behavior

Image inverted and not rendered in proper way

### The expected behavior

Image should be rendered in the correct way and the result should not depend on the rendering type.

### OS

Windows 10, Mac os

### Node version

v18.14.1

### Browser

Chrome 120.0.6099.131

sedghi:
old cornerstone was rendering it correctly in and OHIF v2 
![CleanShot 2024-01-10 at 09 07 48](https://github.com/cornerstonejs/cornerstone3D/assets/7490180/8495d4eb-ed98-4eb7-a596-6c253dbffb8f)

Max-Kharitonov:
@sedghi  any updates?

sedghi:
No not yet

tuongleit:
Hi @sedghi , Our customer is facing this issue, Would your team please help to check that or any clue we can address?

Max-Kharitonov:
@sedghi 
```
could this issue related to function toLowHighRange(
  windowWidth: number,
  windowCenter: number
): {
  lower: number;
  upper: number;
} {
  const delta = 0.001; // small value to prevent zero width
  const width = windowWidth > delta ? windowWidth : delta;
  const lower = windowCenter - 0.5 - (width - 1) / 2;
  let upper = windowCenter - 0.5 + (width - 1) / 2;

  // upper and lower should be non equal (upper should be bigger than lower)
  if (upper - lower < delta) {
    upper = lower + delta;
  }
  return { lower, upper };
}

export { toWindowLevel, toLowHighRange };
```
 Upper and lower should be non equal (upper should be bigger than lower).
and for this image we have 
Window Width: 1
Window Center: -2

[...]

PR #1135, opened 2024-03-01
pixel normalization

Max-Kharitonov:
### Describe the Bug

In case if we are using CPU rendering it works fine, but in case GPU this image inverted.
Please see attached anonimized file
[DICOM sample Issue with Invertion.zip](https://github.com/cornerstonejs/cornerstone3D/files/13840630/DICOM.sample.Issue.with.Invertion.zip)

### Steps to Reproduce

run GPU rendering
image

run CPU rendering
image

### The current behavior

Image inverted and not rendered in proper way

### The expected behavior

Image should be rendered in the correct way and the result should not depend on the rendering type.

### OS

Windows 10, Mac os

### Node version

v18.14.1

### Browser

Chrome 120.0.6099.131

### Changes & Results
Such images have a pixel range of [negative, negative]. Therefore, we are modifying the pixels per LUT to normal values [0, 255]. After these manipulations, everything renders fine.

### Testing
Try to open stackViewport with this series using GPU rendering:
[DICOM.sample.Issue.with.Invertion (1).zip](https://github.com/cornerstonejs/cornerstone3D/files/14971080/DICOM.sample.Issue.with.Invertion.1.zip)

### DIFF
--- a/packages/core/src/RenderingEngine/StackViewport.ts
+++ b/packages/core/src/RenderingEngine/StackViewport.ts
@@ -157,6 +157,7 @@ class StackViewport extends Viewport implements IStackViewport, IImagesLoader {
   private voiUpdatedWithSetProperties = false;
   private VOILUTFunction: VOILUTFunctionType;
   //
+  private outOfRange = false;
   private invert = false;
   // The initial invert of the image loaded as opposed to the invert status of the viewport itself (see above).
   private initialInvert = false;
@@ -1386,7 +1387,8 @@ class StackViewport extends Viewport implements IStackViewport, IImagesLoader {
 
     let voiRangeToUse = voiRange;
 
-    if (typeof voiRangeToUse === 'undefined') {
+    if (typeof voiRangeToUse === 'undefined' || this.outOfRange) {
+      this.outOfRange = false;
       const imageData = imageActor.getMapper().getInputData();
       const range = imageData.getPointData().getScalars().getRange();
       const maxVoiRange = { lower: range[0], upper: range[1] };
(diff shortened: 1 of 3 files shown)


sedghi (changes requested):
can you please provide context, what was the issue and how are you solving it with your PR? basically please fill in the template for the PR and provide test steps

Max-Kharitonov:
@sedghi
this PR is related to https://github.com/cornerstonejs/cornerstone3D/issues/985. Such images have a pixel range of [negative, negative]. Therefore, we are modifying the pixels per LUT to normal values [0, 255]. After these manipulations, everything renders fine.

You could find test series in original report: https://github.com/cornerstonejs/cornerstone3D/issues/985

sedghi (commented):
Thanks for your pull request. I apologize for the delay; I was away, but I'm now back to contributing to open source more actively.

I don't think your approach is correct. The minimum value can be negative, and there's no issue with that. The problem lies in the upper and lower voi ranges being the same, preventing `setVOI` from grabbing the correct range from the data

I'm willing to merge your pull request if you implement it like this.

CleanShot 2024-05-15 at 16 45 25@2x

Max-Kharitonov:
> Thanks for your pull request. I apologize for the delay; I was away, but I'm now back to contributing to open source more actively.
> 
> I don't think your approach is correct. The minimum value can be negative, and there's no issue with that. The problem lies in the upper and lower voi ranges being the same, preventing `setVOI` from grabbing the correct range from the data
> 
> I'm willing to merge your pull request if you implement it like this.
> 
> CleanShot 2024-05-15 at 16 45 25@2x

In case if min and max values negative?

s2ramana:
@sedghi any update on this?

sedghi:
I'm not sure if this is the right direction

PR #1137, opened 2024-03-04
fix: Change to names for statistics so that other implementations can be added

wayfarer3130:
### Context

The response for statistics is an array, but is always used as named/value pairs.  Having a single response type that allows either makes creating new statistics calculators much easier.

### Changes & Results

Use a named object from Calculator instead of an array, but include the array as one of the named objects.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  


packages/tools/src/utilities/math/basic/BasicStatsCalculator.ts
```diff
@@ -70,16 +70,28 @@ export default class BasicStatsCalculator extends Calculator {
...
-    return [
-      { name: 'max', value: singleArrayAsNumber(currentMax), unit: null },
-      { name: 'mean', value: singleArrayAsNumber(mean), unit: null },
-      { name: 'stdDev', value: singleArrayAsNumber(stdDev), unit: null },
-      {
+    const named: NamedStatistics = {
+      max: { name: 'max', value: singleArrayAsNumber(currentMax), unit: null },
+      mean: { name: 'mean', value: singleArrayAsNumber(mean), unit: null },
+      stdDev: {
+        name: 'stdDev',
```
sedghi, line 77: name: Standard Deviation? is the name like label?

sedghi (commented):
I'm not sure if `name` is like label, if so we should have better labels like `Standard Deviation With Sum Squared`, if they are not labels, i'm not sure why we need them

sedghi (approved):

sedghi:
FYI @salimkanoun @Celian-abd

PR #1208, opened 2024-04-15
Fix weird interaction between scroll() and setImageIdIndex() in StackViewport 

jobunk:
### Changes & Results

I had encountered an issue in the StackViewport class, where the user would scroll, afterwards use the public method "setImageIdIndex()" to jump to a specific image and then continue scrolling. The expected behavior in this situation would be to that the viewport continues scrolling from the image that the user just jumped to. However the actual behavior was that the scrolling would pick up back where the user had last scrolled and completely ignore the "setImageIdIndex()" that had happened.

This happens because "this.targetImageIdIndex", which is used during scrolling, was never updated. My workaround was to just use the "setStack()" method instead, since it actually updates "this.targetImageIdIndex" correctly. However I believe that that is not the intended use of the method, so here is my suggested fix. It adds an optional boolean that tells the viewport to also move the scrolling position to the given index. This boolean is optional, because the method is also called from within scroll(), in which case we don't want to adjust the targetImageIdIndex. 

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: Windows
- [] "Node version: v20.12.2
- [] "Browser: Firefox

### DIFF
--- a/common/reviews/api/core.api.md
+++ b/common/reviews/api/core.api.md
@@ -2094,7 +2094,7 @@ interface IStackViewport extends IViewport {
     // (undocumented)
     setDefaultProperties(ViewportProperties: StackViewportProperties, imageId?: string): void;
     // (undocumented)
-    setImageIdIndex(imageIdIndex: number): Promise<string>;
+    setImageIdIndex(imageIdIndex: number, overwriteScrollIndex?: boolean): Promise<string>;
     // (undocumented)
     setProperties({ voiRange, invert, interpolationType, rotation, colormap, }: StackViewportProperties, suppressEvents?: boolean): void;
     // (undocumented)
@@ -3138,7 +3138,7 @@ export class StackViewport extends Viewport implements IStackViewport, IImagesLo
     // (undocumented)
     setDefaultProperties(ViewportProperties: StackViewportProperties, imageId?: string): void;
     // (undocumented)
-    setImageIdIndex(imageIdIndex: number): Promise<string>;
+    setImageIdIndex(imageIdIndex: number, overwriteScrollIndex?: any): Promise<string>;
     // (undocumented)
     protected setInterpolationType: (interpolationType: InterpolationType) => void;
     // (undocumented)
--- a/packages/core/src/RenderingEngine/StackViewport.ts
+++ b/packages/core/src/RenderingEngine/StackViewport.ts
@@ -2447,7 +2447,10 @@ class StackViewport extends Viewport implements IStackViewport, IImagesLoader {
    * @param imageIdIndex - number represents imageId index
    * @param overwriteScrollIndex - can be used to also set the start position of scroll() to this index
    */
-  private async _setImageIdIndex(imageIdIndex: number, overwriteScrollIndex: boolean = false): Promise<string> {
+  private async _setImageIdIndex(
+    imageIdIndex: number,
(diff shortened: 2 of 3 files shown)


jobunk:
@sedghi Could you review this? Sorry that I can't provide more for the testing part. If something is unclear I'm happy to provide further explanation.

sedghi:
> If the scrolling picks up from the position that you jumped to, the error has been fixed.

I don't think that is the correct behavior. The scroll should not depend on the state (where it was jumped) and should only scroll in the requested direction.

It seems like you are injecting your app-level logic into the library, which is usually not a good idea.

I recommend implementing that logic in your app:
-  Detect the jump and keep track of it.
-  When you want to scroll, check the state of the app to see where it jumped.
-  Then set `ImageIdIndex` before you scroll from that location and then scroll.

Does this make sense?

Bunk96:
> > If the scrolling picks up from the position that you jumped to, the error has been fixed.
> 
> I don't think that is the correct behavior. The scroll should not depend on the state (where it was jumped) and should only scroll in the requested direction.
> 
> It seems like you are injecting your app-level logic into the library, which is usually not a good idea.
> 
> I recommend implementing that logic in your app:
> 
>     * Detect the jump and keep track of it.
> 
>     * When you want to scroll, check the state of the app to see where it jumped.
> 
>     * Then set `ImageIdIndex` before you scroll from that location and then scroll.
> 
> 
> Does this make sense?

Thank you for you response. What you describe in your implementation recommendation makes sense and is basically the behavior I expected from this module. However when I adjust "ImageIdIndex" by using the "setImageIdIndex()" method, the scrolling will not be aware of this since it uses the "targetImagingIdIndex" variable, which is never adjusted. It will basically jump back to the pre "setImageIdIndex()" position the next time it's called and then start scrolling. This issue is what I tried to solve in this PR. 

Please let me know if I misunderstood anything in your recommendation. In my opinion my solution is not very pretty, but it is the best I could come up with given that the "setImageIdIndex()" and "scroll()" functions are kind of intertwined.

jobunk:
> > If the scrolling picks up from the position that you jumped to, the error has been fixed.
> 
> I don't think that is the correct behavior. The scroll should not depend on the state (where it was jumped) and should only scroll in the requested direction.
> 
> It seems like you are injecting your app-level logic into the library, which is usually not a good idea.
> 
> I recommend implementing that logic in your app:
> 
>     * Detect the jump and keep track of it.
> 
>     * When you want to scroll, check the state of the app to see where it jumped.
> 
>     * Then set `ImageIdIndex` before you scroll from that location and then scroll.
> 
> 
> Does this make sense?

Thank you for you response. What you describe in your implementation recommendation makes sense and is basically the behavior I expected from this module. However when I adjust `ImageIdIndex` by using the `setImageIdIndex()` method, the scrolling will not be aware of this since it uses the `targetImagingIdIndex` variable, which is never adjusted. It will basically jump back to the pre `setImageIdIndex()` position the next time it's called and then start scrolling. This issue is what I tried to solve in this PR. 

Please let me know if I misunderstood anything in your recommendation. In my opinion my solution is not very pretty, but it is the best I could come up with given that the `setImageIdIndex()` and `scroll()` functions are kind of intertwined.

sedghi:
@jobunk can you please provide a minimum example to show case the issue?

Issue #1133, opened 2024-02-29
[Bug] 


gosvig123:
### Describe the Bug

The issue is that the image seem to load, but the canvas or image loaded is not vissible. This is only a problem when trying to load an image from my iphone SE, have tried different OS variations from 17.1 and up.
When trying to run the same code locally in the browser, or emulating the Iphone SE from the chrome dev tools the issue does not occur.

On the phone the loading indicate that the image has loaded, and the tool still works in the sense that windowing can be activated, and change the settings within the canvas that has no content.

### Steps to Reproduce

Try to load an image from an Iphone SE with OS 17.1 either using Ngrok or forward the tunnel to the device from the laptop running the.

### The current behavior

Everything works well, except when tested locally on an Iphone SE (have not tested other actual devices)
![image_123650291](https://github.com/cornerstonejs/cornerstone3D/assets/109414278/c3193483-86ab-418c-b662-2d297ae735e5)

### The expected behavior

For the Image to display inside the canvas after being loaded.

### OS

IOS 17.1

### Node version

18.19

### Browser

Chrome or Safari (tested both)

abustany:
This also happens with https://www.cornerstonejs.org/live-examples/stackbasic , which I suppose is using the latest cornerstone version (I'm on iOS 17.3.1). I have the same issue with cornerstone 1.58.4.

abustany:
https://www.cornerstonejs.org/live-examples/webloader works though, so maybe this is an issue with the dicom loader?

abustany:
From what I could gather so far:
- the DICOM loader seems to do its job fine and is not at fault
- forcing CPU rendering works around the issue (but is not a great solution)

so something in the GPU rendering pipeline is broken on iOS, going down the rabbit hole...

abustany:
I hacked `updateVTKImageDataWithCornerstoneImage` to remap single channel (numComponents=1) to three-channel images (numComponents=3), and the image displays on iOS correctly now. That doesn't really tell what the problem is, much less what the fix is, but at least I know what triggers the issue.

abustany:
I now have a reduced test case using only vtk.js:

```typescript
import '@kitware/vtk.js/Rendering/Profiles/Volume';
import vtkImageData from '@kitware/vtk.js/Common/DataModel/ImageData';
import vtkDataArray from '@kitware/vtk.js/Common/Core/DataArray';
[...]

Issue #1184, opened 2024-04-04
[Bug]  Volume viewports not working on ios


mauricio-cdr:
### Describe the Bug

Hello, how are you?

Volume viewports do not work on ios on any version of IOS 17.0 or higher

![image](https://github.com/cornerstonejs/cornerstone3D/assets/17281855/67edf16e-aad9-42e5-b968-f71bb2bb0595)

### Steps to Reproduce

1. Go to any example that includes a volume viewport. e.g. https://www.cornerstonejs.org/live-examples/volumeannotationtools .

### The current behavior

The images are not displayed and in the console you can see the following message.

![Captura de Pantalla 2024-04-04 a la(s) 14 30 42](https://github.com/cornerstonejs/cornerstone3D/assets/17281855/bf09551e-ff6a-4d5e-818d-2913f0f90c80)

![image](https://github.com/cornerstonejs/cornerstone3D/assets/17281855/67edf16e-aad9-42e5-b968-f71bb2bb0595)

### The expected behavior

Make it work the same way it works on IOS 16.x.x or less.

![Captura de Pantalla 2024-04-04 a la(s) 14 34 03](https://github.com/cornerstonejs/cornerstone3D/assets/17281855/e5a297de-638b-4d1e-b991-091d50c5b3e2)

### OS

IOS 17.0 , 17.1.4 , 17.2,  17.4.1

### Node version

18.13.0

### Browser

Chrome 90+

sedghi:
I created this PR which would be good enough for a lot of use cases (except the PT Scaled data)

https://github.com/cornerstonejs/cornerstone3D/pull/1212

PR #1212, opened 2024-04-17
fix(iOS): Implement texture handling for iOS issues in late 16 and 17

sedghi:
### Context

There were couple of bugs reported issues for cornerstone on iOS specifically late 16 and 17. 

As discussed in 

https://github.com/Kitware/vtk-js/issues/3027
https://github.com/cornerstonejs/cornerstone3D/issues/1133

The issue is iOS does not have support for `OES_texture_float_linear` (https://developer.mozilla.org/en-US/docs/Web/API/OES_texture_float_linear) but this feature was **wrongly** reported as true in the older iOS (https://bugs.webkit.org/show_bug.cgi?id=264404#c2) and from 16.5 ish iOS is throwing errors.

We have two other tools for not rendering floats

1. The 16 bit texture extension via https://developer.mozilla.org/en-US/docs/Web/API/EXT_texture_norm16 which has decent support https://web3dsurvey.com/webgl2/extensions/EXT_texture_norm16, and if exists we should use it 
2. `preferSizeOverAccuracy`: which originally was added for memory management, but under the hood is uses half float textures, and mostly works across the devices, though as the name suggest it has some inaccuracies in high bits 

So This PR will **detect** if the host device is mobile and set the 16 bit texture (if available) if not it will try to do preferSizeOverAccuracy

---

### Limitations
Although this PR addresses the rendering of the Stack and Volume Viewports, for scenarios like PT SUV scale, we require `Float32Array` as we can't render via 16-bit norms. Therefore, we detect this and simply render the PT in a non-scaled fashion. It would be ideal to find a proper solution to address this final issue.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  


IbrahimCSAE (approved):
Great fix, much needed

Congritulation:
Excuse me. I run my project in IPadOS 17.4.1 with cornerstone3D 1.70.10 and encounter with the same situation. The volumeViewport show nothing but a gray rectangle. so what i want to ask is [[Bug] Volume viewports not working on ios #1184](https://github.com/cornerstonejs/cornerstone3D/issues/1184) fixed and released before v1.70.10?

PR #1216, opened 2024-04-19
fix(wadouri): custom loaders should register their schema, and add new debounce listener to event target

sedghi:
### Context

We are reverting changes here https://github.com/cornerstonejs/cornerstone3D/pull/1154/files since it is causing troubles. 

Since it is wrong after further consideration with Bill Wallace, for your custom loader, you should register the scheme with it and use your custom scheme in your URLs.

CC: @jlopes90 

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  

### DIFF
--- a/packages/core/src/eventTarget.ts
+++ b/packages/core/src/eventTarget.ts
@@ -3,13 +3,16 @@
  */
 class CornerstoneEventTarget implements EventTarget {
   private listeners;
+  private debouncedListeners;
 
   constructor() {
     this.listeners = {};
+    this.debouncedListeners = {};
   }
 
   public reset() {
     this.listeners = {};
+    this.debouncedListeners = {};
   }
 
   public addEventListenerOnce(type, callback) {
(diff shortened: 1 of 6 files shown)


packages/core/src/eventTarget.ts
```diff
@@ -39,6 +42,49 @@ class CornerstoneEventTarget implements EventTarget {
     this.listeners[type].push(callback);
   }
 
+  public addEventListenerDebounced(type, callback, delay) {
```
wayfarer3130, line 45: Docs please - would like to know something about how it is debounced

wayfarer3130: Would be nice to see this used in at least one example please.

wayfarer3130 (approved):

shunia:
> ... for your custom loader, you should register the scheme with it and use your custom scheme in your URLs.

Please, share some detailed docs about how to implement a complete example on how to create a custom loader with just pure **ArrayBuffer**. 

It might be right to remove the wrong use cases to `wadouri.loadImage` with a custom loader since it's not designed for it, but just as [this comment](https://github.com/cornerstonejs/cornerstone3D/pull/990#issuecomment-1892221409) described, this is the only way I found that could easily use cornerstone with my dicom data from some custom sources like from a **Websocket**.

I don't want to say it's blocking ppl from using a custom loader but the reality here is that there's no real use cases or examples for us to learn from at all. 

It feels really helpless.

sedghi:
@shunia please see our docs

https://www.cornerstonejs.org/docs/how-to-guides/custom-image-loader

shunia:
> @shunia please see our docs
> 
> https://www.cornerstonejs.org/docs/how-to-guides/custom-image-loader

How to create the image object? Is there any utility function provided for convenience?

sgielen:
> @shunia please see our docs
> 
> https://www.cornerstonejs.org/docs/how-to-guides/custom-image-loader

Please bear with me if I'm wrong here -- but the docs, unfortunately, leave out the difficult part: they simply use `createImageObject` which isn't defined anywhere ("logic omitted" has the same energy as "I can prove this, but the margin of my scientific paper is too small for it"), and they don't mention metadata at all. Those are the two things we used WADOImageLoader for, which is now impossible. Could you please advise?

PR #1224, opened 2024-04-23
fix(stackContextPrefetch): fix disable function, reduce duplicated fetching of images

CarstenWalther:
### Context

This PR fixes the stackContextPrefetch.disable function that was always failing.
Additionally it makes sure that cached images are not fetched again.

### Changes & Results

stackContextPrefetch.disable: remove check of data that will never exist
stackContextPrefetch.prefetch: it is already checked, if images are already cached. Now it does not request cached images again, as it was intended

### Testing

I added a test for the disable function.

I tried to add a test for the duplicated fetching but found it difficult to test this asynchronous and open-ended behavior.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] "OS: Ubuntu 22.04
- [x] "Node version: v18.18.2
- [] "Browser:
  

### DIFF
--- a/packages/tools/src/utilities/stackPrefetch/stackContextPrefetch.ts
+++ b/packages/tools/src/utilities/stackPrefetch/stackContextPrefetch.ts
@@ -164,7 +164,7 @@ function prefetch(element) {
     imageLoadPoolManager.filterRequests(clearFromImageIds(stack));
   }
 
-  function doneCallback(imageId) {
+  function doneCallback(imageId: string) {
     const imageIdIndex = stack.imageIds.indexOf(imageId);
 
     removeFromList(imageIdIndex);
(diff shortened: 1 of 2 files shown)


packages/tools/src/utilities/stackPrefetch/stackContextPrefetch.ts
```diff
@@ -222,7 +222,7 @@ function prefetch(element) {
 
   const useNativeDataType = useNorm16Texture || preferSizeOverAccuracy;
 
-  indicesToRequestCopy.forEach((imageIdIndex) => {
+  stackPrefetch.indicesToRequest.forEach((imageIdIndex) => {
```
sedghi, line 225: Thanks for your pull request (PR). I tested your test code in the `main` branch, and it failed, as you mentioned. I also tried your proposed solution, and it passed - good work.

However, I reverted the change on line 255, and the tests still passed. What was the reason for modifying that line?

wayfarer3130: The indicesToRequestCopy is the original set of images, including things that are already in cache.  If you don't use the correct set, then you will end up re-fetching already cached images.  The line is correct as changed.

wayfarer3130: I'm thinking about how to test this, but without quite a bit of mocking, I don't see a good way of testing it.

wayfarer3130 (approved):
I think both changes are appropriate - the second one fixes the disable, while the first one fixes the extra fetch of already cached data.

PR #1229, opened 2024-04-26
Fix: rotate resolution decreasing on click

salimkanoun:
### Context
As discussed yesterday if Office hours, the click with the TrackballRotateTool was causing a permanent and additive resolution decrease

### Changes & Results

Reworked listener to do resolution decrease at drag callback.
However, this mean resolution is changed everytime a drag event occurs, it might have performance impact, as resolution decrease / reincrease could have computation cost.

### Testing

Go to Ortientation Maker Tool, click on the 3D viewport and drag, resolution doesn't decrease anymore

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  

### DIFF
--- a/packages/tools/src/tools/TrackballRotateTool.ts
+++ b/packages/tools/src/tools/TrackballRotateTool.ts
@@ -11,12 +11,15 @@ import { mat4, vec3 } from 'gl-matrix';
 import { EventTypes, PublicToolProps, ToolProps } from '../types';
 import { BaseTool } from './base';
 import { getToolGroup } from '../store/ToolGroupManager';
+import { IStackViewport, IVolumeViewport } from 'core/src/types';
+import vtkVolumeMapper from '@kitware/vtk.js/Rendering/Core/VolumeMapper';
 
 /**
  * Tool that rotates the camera in the plane defined by the viewPlaneNormal and the viewUp.
  */
 class TrackballRotateTool extends BaseTool {
   static toolName;
+  originalSampleDistance = null;
   touchDragCallback: (evt: EventTypes.InteractionEventType) => void;
   mouseDragCallback: (evt: EventTypes.InteractionEventType) => void;
   cleanUp: () => void;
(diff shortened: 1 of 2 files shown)


salimkanoun:
@sedghi if the user drag and drop a new series in an existing HP, is this rotation tool going to be reinstanciated ?

If not it will mean that the mapper is going to be set to the previous series value and not the new one....

salimkanoun:
And by the way I'm not sure about Performances,and qualify decrease / increase will be done at each drag event that are so many.

To handle more properly we will need to manage image quality reduction in drag start and drag end event

packages/tools/src/tools/TrackballRotateTool.ts
```diff
@@ -172,6 +170,7 @@ class TrackballRotateTool extends BaseTool {
     const { rotateIncrementDegrees } = this.configuration;
     const enabledElement = getEnabledElement(element);
     const { viewport } = enabledElement;
+    this._reduceSampleDistance(viewport);
```
sedghi, line 173: I'm not entirely certain about the reason for invoking the `reduceSampleDistance` function during each drag operation. This function executes multiple times, potentially tens of times, within a single interaction.

Ideally, it should start before the mouse event (preMouseDown) and reset after the mouseUp event.

salimkanoun:
@sedghi I tried but i'm unable to get the onMouseUp event, is passes throught anoter custom listener that is obscure to me.

I also tried to expose the dragstart / dragend event but didn't succeed as all mouse event seems to pass to a specific listener in cornerstone 

I think it would be better to use the native HTML even drag start / drag end, no chance to get these events in cornerstone ?

sedghi:
@IbrahimCSAE please see how to fix this properly

sedghi:
@salimkanoun Can you please provide reproducible steps so I can observe the incorrect behavior in the main branch?

salimkanoun:
https://github.com/cornerstonejs/cornerstone3D/assets/25229822/3079b9a3-4e7c-40ab-92bd-c53e6fe5be34

sedghi:
I wasn't able to push changes to this PR, so i created another PR, just FYI https://github.com/cornerstonejs/cornerstone3D/pull/1306

PR #1231, opened 2024-05-01
fix(video): video viewport annotation position

wayfarer3130:
### Context

Due to changes in the canvas positioning to fix pixel positioning, the transform in the video viewport got broken.  This PR adjusts the transform so that it works correctly now.

### Changes & Results

Updated the transform in VideoViewport to be correct for the updated canvas size/scaling.
Fixed the videoSegmentation example (had gotten overwritten with the wrong example).

### Testing
Run examples:
videoSegmentation
videoContourSegmentation
videoTools

pan/zoom

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  

### DIFF
--- a/packages/core/src/RenderingEngine/VideoViewport.ts
+++ b/packages/core/src/RenderingEngine/VideoViewport.ts
@@ -894,47 +894,52 @@ class VideoViewport extends Viewport implements IVideoViewport {
     const pan: Point2 = this.videoCamera.panWorld;
     const worldToCanvasRatio: number = this.getWorldToCanvasRatio();
 
-    const subCanvasPos: Point2 = [
+    const canvasPos: Point2 = [
       (worldPos[0] + pan[0]) * worldToCanvasRatio,
       (worldPos[1] + pan[1]) * worldToCanvasRatio,
     ];
-
-    const canvasPos: Point2 = [subCanvasPos[0], subCanvasPos[1]];
 
     return canvasPos;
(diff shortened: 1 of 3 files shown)


packages/core/src/RenderingEngine/VideoViewport.ts
```diff
@@ -965,17 +971,27 @@ class VideoViewport extends Viewport implements IVideoViewport {
     this.renderFrame();
   };
 
+  /**
+   * Creates a transform from video index coordinates to canvas coordinates.
+   */
   protected getTransform() {
     const panWorld: Point2 = this.videoCamera.panWorld;
+    const devicePixelRatio = window.devicePixelRatio || 1;
```
sedghi, line 979: does these changes mean the SM viewport also suffer from the DPR bug too?

wayfarer3130: Doesn't seem to, probably because we don't actually render the SM viewport itself.

packages/tools/examples/videoSegmentation/index.ts
wayfarer3130, line 1: The changes here are just straight reverting to the correct segmentation example.  Somehow it got replaced with videoCOntourSegmentation example.

sedghi (approved):

PR #1233, opened 2024-05-02
fix(rt): enhanced RT support, utilize SVGs for rendering.

IbrahimCSAE:
### Context

Add support for SVG rendering for RT

### Changes & Results

This PR changes the rendering process for RT contours, in the past we used to pass the `gemoteryIds` to our render function and create a VTK actor with the contour data.

Now, this has changed, we are still passing the `gemotryIds`, however, we take those `geomtryIds`, and call `cache.getGeomtry(geomtryId)` on each one, and use its data to create annotations that represent the contour, which gets stored internally and generate an `annotationsUIDsMap`

This helps us render the contours using SVG which is much more reliable than VTK.

From a user point of view, nothing is changing in the RT loading process, because you are still creating and caching geometries, we just handle them differently internally to generate SVGs instead of a VTK actor.

You will notice your  viewports only have 1 actor after loading an RT, which in the past would become 2 actors (1 for the scan,and 1 for RT).

### Fixes

Segment Select Tool In Both Stack and Volume Viewport:

- Fixed device pixel ratio rendering error
```yarn run example segmentSelect```

Contour Segmentation Representation for Volume Viewports:

- Removed 3D Volume Viewport
```yarn run example contourRendering```

Contour Rendering Configuration:

- Removed 3D Volume Viewport
- Fixed hide segment buttons
```yarn run example contourRenderingConfiguration```

Contour Segmentation Configuration:

- Fixed device pixel ratio error
```yarn run example contourSegmentationConfiguration```

### CODE REVIEW

File: packages/tools/examples/contourRendering/index.ts [hunk 1]
```diff
@@ -216,44 +149,72 @@ async function run() {
...
+  );
 
-    viewport3d.render();
+  // load the contour data
+  const promises = contour.contourSets.map(async (contourSet) => {
+    const { data, segmentIndex } = contourSet;
+    await Promise.all(
+      data.map(async (contour) => {
+        const { points } = contour;
+        const contourSegmentationAnnotation = {
```
IbrahimCSAE, line 176: create internally

packages/tools/examples/contourRenderingConfiguration/index.ts
```diff
@@ -341,21 +269,19 @@ async function run() {
...
 
   // Render the image
   renderingEngine.render();
+  segmentation.config.setToolGroupSpecificConfig(toolGroupId, {
+    renderInactiveSegmentations: true,
+    representations: {
+      CONTOUR: {
+        outlineWidthActive: 1,
+        renderFill: false,
+      },
```
sedghi, line 282: I think these should be default in contour rendering, can you move it internallly? so that we can remove the whole 

```
segmentation.config.setToolGroupSpecificConfig(toolGroupId, {
    renderInactiveSegmentations: true,
    representations: {
      CONTOUR: {
        outlineWidthActive: 1,
        renderFill: false,
      },
    },
  });
  ```

File: packages/tools/src/tools/displayTools/Contour/contourDisplay.ts [hunk 3]
```diff
@@ -95,89 +91,19 @@ async function render(
...
-  //   });
-  // } else if (contourData.annotationUIDsMap?.size) {
-  //   handleContourAnnotationSegmentation({
-  //     viewport,
-  //     representationConfig,
-  //     toolGroupConfig,
-  //     annotationUIDsMap: contourData.annotationUIDsMap,
-  //   });
-  // }
+  if (contourData?.geometryIds?.length) {
```
sedghi, line 98: let's keep the vtk path too, can we ?

sedghi (commented):
Looks great, please explain more in the PR what is happening (we are moving the contour segmentation representation to our SVG, and we are removing the support for 3D via SVG, if needed they should stick to geometryId with vtk stuff

packages/tools/src/tools/displayTools/Contour/contourDisplay.ts
IbrahimCSAE: there's only 1 path for both

IbrahimCSAE: review my messages and the PR description for this, we can do something else if its not ideal, I think its better to keep it like this

sedghi (approved):

Issue #1218, opened 2024-04-21
[Bug] Eraser Tool not working


simonbertoli:
### Describe the Bug

The Eraser Tool is not working 

### Steps to Reproduce

1. Install @cornerstonejs/tools^1.70.5
2. Add changes to package outline here https://github.com/cornerstonejs/cornerstone3D/issues/1203#issuecomment-2060433309 to avoid build error
3. Add EraserToo, LengthTool and PlanarFreehandROITool
4. Draw a Length Annotation, select the Eraser tool and select the Annotation 
5. Error appears in console log, screenshot below is from the Annotation Tools Stack at https://www.cornerstonejs.org/live-examples/stackannotationtools

![Screenshot 2024-04-22 at 8 40 39 AM](https://github.com/cornerstonejs/cornerstone3D/assets/122240522/5c6918c8-abfc-4e02-90c6-431c3363330e)

If we just use the Length tool and do not add the PlanarFreehandROITool, the eraser works as expected. Once both tools have been loaded the error occurs

### The current behavior

The annotation is not removed - getting the following error in the console

`Uncaught TypeError: interactableAnnotations is not iterable
    at AnnotationEraserTool._deleteNearbyAnnotations (AnnotationEraserTool.ts:69:32)
    at AnnotationEraserTool.preMouseDownCallback (AnnotationEraserTool.ts:21:17)`

### The expected behavior

The annotation is removed and the is no error in the console.log

### OS

MacOS 14.4.1

### Node version

20.12.1

### Browser

Brave 1.65.114

### UPDATE
It looks as is  [const interactableAnnotations =](https://github.com/cornerstonejs/cornerstone3D/blob/65604450273a4e7eee6203a83863e2b9674c4263/packages/tools/src/tools/AnnotationEraserTool.ts#L63C13-L63C36) loops through from the PlanarFreehandROITool but returns undefined as the annotation is from the Length tool, therefore causing the 'not iterable' error. If I add a freehand and a length annotation, the eraser tool works as expected.

PR #1237, opened 2024-05-04
fix: Eraser Tool Error with LengthTool and PlanarFreehandROITool Loaded

sedghi:
Related to #1218

Fixes the 'not iterable' error in the `AnnotationEraserTool` when using the Eraser Tool with both LengthTool and PlanarFreehandROITool loaded.

- Adds a conditional check to ensure `filterInteractableAnnotationsForElement` does not return `undefined` before attempting to iterate over its result. This prevents the 'not iterable' error by ensuring `interactableAnnotations` is always an array, even if it's empty.
- Modifies the `_deleteNearbyAnnotations` method to gracefully handle cases where `filterInteractableAnnotationsForElement` might return `undefined`, ensuring the Eraser Tool functions correctly regardless of the combination of tools loaded.

---

For more details, open the [Copilot Workspace session](https://copilot-workspace.githubnext.com/cornerstonejs/cornerstone3D/issues/1218?shareId=85160633-8878-43bd-9954-70e5959faf52).

### DIFF
--- a/packages/tools/src/tools/AnnotationEraserTool.ts
+++ b/packages/tools/src/tools/AnnotationEraserTool.ts
@@ -60,12 +60,11 @@ class AnnotationEraserTool extends BaseTool {
         continue;
       }
 
-      const interactableAnnotations = toolInstance.filterInteractableAnnotationsForElement
-        ? toolInstance.filterInteractableAnnotationsForElement(
-            element,
-            annotations
-          )
-        : [];
+      const interactableAnnotations =
+        toolInstance.filterInteractableAnnotationsForElement(
+          element,
+          annotations
+        ) || [];
 
       for (const annotation of interactableAnnotations) {
         if (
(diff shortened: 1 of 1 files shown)


simonbertoli (changes requested):
Action taken

- cloned branch
- ran `yarn run example stackannotationtools`
- added a Length Annotation
- selected the Eraser tool and clicked on Annotation 
- Received error
Screenshot 2024-05-09 at 2 06 06 PM

The ternary operator you have added will always return the output of the call to `filterInteractableAnnotationsForElement` because you are testing if the function exists, which it does, so it always returns the function output - which can be undefined

I suspect the code should like something like this
```
      const interactableAnnotations =
        toolInstance.filterInteractableAnnotationsForElement(
          element,
          annotations
        ) || [];
```

sedghi:
@IbrahimCSAE this was done by github workspace (just wanted to try it) and it wrong, can you fix it correctly? thanks

IbrahimCSAE (approved):
LGTM

PR #1246, opened 2024-05-13
docs(annotations): add/hydrate annotations examples

IbrahimCSAE:
### Context

Add an example that showcases how to add annotations programmatically using just points as input.

### Changes & Results

image


packages/tools/examples/dynamicallyAddAnnotations/index.ts
```diff
@@ -0,0 +1,406 @@
...
+            bottomRight: <Types.Point3>[0, 0, 0],
+          },
+        },
+      },
+      label: '',
+      cachedStats: {},
+    },
+    isLocked: true,
+    isVisible: true,
+  };
```
sedghi, line 260: Can we remove the unnecessary properties? For instance, do we need `label`? Do we need `cachedStats: {}`? Do we need `isLocked`? Do we need `isVisible`?

IbrahimCSAE: done

this is the bare min

```ts
  const annotation = {
    metadata: {
      toolName: 'Length',
      referencedImageId,
      viewUp,
      cameraPosition,
      FrameOfReferenceUID,
      viewPlaneNormal,
      cameraFocalPoint,
    },
    data: {
      handles: {
        points: [<Types.Point3>[...start], <Types.Point3>[...end]],
        textBox: {},
      },
      cachedStats: {},
    },
  };
  ```

File: packages/tools/examples/dynamicallyAddAnnotations/index.ts [hunk 2]
```diff
@@ -240,23 +238,10 @@ const addProgrammaticAnnotation = (
...
-          hasMoved: false,
-          worldPosition: <Types.Point3>[0, 0, 0],
-          worldBoundingBox: {
-            topLeft: <Types.Point3>[0, 0, 0],
-            topRight: <Types.Point3>[0, 0, 0],
-            bottomLeft: <Types.Point3>[0, 0, 0],
-            bottomRight: <Types.Point3>[0, 0, 0],
-          },
-        },
+        textBox: {},
```
sedghi, line 241: do we need textBox: {}? if so we should move it to internal, doesn't makes sense

packages/tools/examples/dynamicallyAddAnnotations/index.ts
IbrahimCSAE: textBox and cachedStats should be defined otherwise the tools tries to access undefined objects, rest is fine

packages/tools/examples/dynamicallyAddAnnotations/index.ts
IbrahimCSAE: we need it cuz it tries to access textBox.hasMoved

packages/tools/examples/dynamicallyAddAnnotations/index.ts
```diff
@@ -240,23 +238,10 @@ const addProgrammaticAnnotation = (
...
       },
-      label: '',
       cachedStats: {},
```
sedghi, line 243: Any necessary initialization (e.g., {}, []) should be implemented internally.

packages/tools/src/utilities/hydration.ts
```diff
@@ -0,0 +1,98 @@
...
+  utilities,
+  BaseVolumeViewport,
+  StackViewport,
+  VideoViewport,
+  cache,
+} from '@cornerstonejs/core';
+import { Annotation } from '../types';
+import { addAnnotation } from '../stateManagement';
+
+function hydration(
```
sedghi, line 12: please call annotationHydration or somethign related to annotaitons

packages/tools/src/utilities/hydration.ts
```diff
@@ -0,0 +1,98 @@
...
+  toolName: string,
+  worldPoints: Types.Point3[],
+  options?: {
+    FrameOfReferenceUID?: string;
+  }
+): Annotation {
+  const viewReference = viewport.getViewReference();
+  const { viewPlaneNormal, FrameOfReferenceUID } = viewReference;
+  const annotation = {
+    annotationUID: utilities.uuidv4(),
```
sedghi, line 23: sometimes it is desired to provider uid, have it in options

packages/tools/src/utilities/hydration.ts
```diff
@@ -0,0 +1,98 @@
...
+    referencedImageId = targetId.split('videoId:')[1];
+  } else {
+    const volumeId = utilities.getVolumeId(targetId);
+    const imageVolume = cache.getVolume(volumeId);
+
+    referencedImageId = utilities.getClosestImageId(
+      imageVolume,
+      worldPos,
+      viewPlaneNormal
+    );
```
sedghi, line 71: how come we are doing the getClosest only for volume?

sedghi (commented):
see my comments

sedghi (approved):

IT-Highmore:
@IbrahimCSAE hello, I use function `annotationHydration` to add "RectangleROI" annotations.  I use four points as input, But sometimes it works, and sometimes can't see. 

As the images show, there are three measurements. But it doesn't show in the third viewport.
I checked the relevant code and couldn't find the reason. I guess it's due to points error or other reasons？

image

IT-Highmore:
I know the reason why the annotations don't show, It is filtered in `filterAnnotationsWithinSlice`. Generally speaking, it is because of the point, Is there any way to produce effective points ? o(╥﹏╥)o

PR #1249, opened 2024-05-14
feat(RT): render single point segments in RT

IbrahimCSAE:
### Context

this PR allows use to render single point segments in RT, previously they would show up as nothing since 1 point was not enough to create the SVG.

### Changes & Results

Use the single point as a center, and draw a circle with a small radius around it, by generating points using the center.

Before:

image

After:
image


packages/tools/src/tools/annotation/planarFreehandROITool/renderMethods.ts
```diff
@@ -128,6 +128,48 @@ function renderClosedContour(
   const allContours = [canvasPolyline, ...childContours];
   const polylineUID = '1';
 
+  if (annotation.data.contour.isPoint) {
```
sedghi, line 131: have a configuration on the planar free hand that is default false, something like displayOnePointAsCrosshairs and if that is true, you detect it is one point and that config is set tot display crosshairs

sedghi (approved):

PR #1269, opened 2024-05-22
fix(rt): assign correct imageId for Stack

IbrahimCSAE:
### Context

Assign the correct imageId for annotations on stackviewport when loading an RT


sedghi (approved):

sedghi:
@IbrahimCSAE can you update the api?

PR #1283, opened 2024-05-27
feat(tools): New Tool - Height Tool

ArturRod:
### Context
A new height measurement tool is added, to be able to compare differences in height, for example, a hip.
To do this, the necessary classes and the SVG icon are added to make it identifiable. It is based on the length of the tool but only asks for the height.
image

### Changes & Results
I show a video of how it works.
https://github.com/cornerstonejs/cornerstone3D/assets/86238895/c2701431-4b94-44ce-b62a-b65768f82c2b

### Testing
It can be tested by running: yarn run example nifiWithTools, an automated test .js needs to be added.

Change in tools.md adding the height tool, but an image would need to be added along with the other annotations.

#### Tested Environment
- [] "OS: Windows 11
- [] "Node version: v20.13.1
- [] "Browser: Firefox 120.0


IbrahimCSAE (changes requested):
Please update the API and fix the merge conflict

IbrahimCSAE:
Its still failing the API, you can do yarn build:update-api to fix this

ArturRod:
Hi @IbrahimCSAE 
I think I found an error when calculating the height, currently it was done by subtracting the final position y - the initial position y. But I see that the other views give errors, it is not very clear to me if it is necessary to know which view it is in. I give an example comparing with the data of the length of the tool.
example
Is the formula correct? In principle, the tool consists of seeing the difference between the height of the hips.
Since I don't want to upload an incorrect formula.
imagen

ArturRod:
Hi @IbrahimCSAE
I have managed to solve the error mentioned above to calculate the height depending on whether the view is axial coronal or sagittal and it is now displayed correctly. I have taken into account that sagittal always X is 0, coronal always Y is 0 and AXIAL always Z is 0, you can do the calculations to see which one to use. 😊😊
imagen

IbrahimCSAE (approved):
I tested this locally and it seems to work great, @sedghi  can you please verify the implementation and if all the math checks out

sedghi (approved):
Nice addition thanks

PR #1315, opened 2024-06-12
tests: added e2e tests for splineContourSegmentationTools

lscoder:
- Added some tests for `examples/splineContourSegmentationTools`
- Added `test:e2e:ui` because it was taking me too much time to understand why a test was failing and after reading Playwright docs I found that `--ui` opens an UI which allow us to run tests and see the results step-by-step (screenshots) in a UI. It also allow us to "watch" a test which re-runs automatically when its is updated. That really make sense  to have in our project to help us save a lot of time.

https://github.com/cornerstonejs/cornerstone3D/assets/931820/828f98f4-e741-4d83-917e-02ce089a7494


lscoder:
@sedghi  Should we think about using LFS to store all screenshots? Otherwise the repo may grow very fast now as we are adding more e2e tests.

https://git-lfs.com/

sedghi (approved):
Thanks, we can look at lfs later

PR #1330, opened 2024-06-16
feat(synchronizer): frame view synchronizer

IbrahimCSAE:
### Context

Source of idea: https://santesoft.com/win/sante-dicom-viewer-lite/sante-dicom-viewer-lite.html

image

image

image

image


packages/tools/src/synchronizers/callbacks/frameViewSyncCallback.ts
```diff
@@ -0,0 +1,47 @@
+import { getRenderingEngine, Types } from '@cornerstonejs/core';
+
+import { Synchronizer } from '../../store';
+import { jumpToSlice } from '../../utilities';
+
+export default async function frameViewSyncCallback(
```
sedghi, line 6: I'm not sure if this belongs here, i rather have this inside OHIF

PR #1353, opened 2024-06-26
fix(video): Video play clip wasn't working

wayfarer3130:
### Context

Fix the video play clip so it starts/stops the video.
As part of that, added support for some more tags on image rendered event for video so that a scrollbar can be kept up to date.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  

### DIFF
--- a/packages/core/src/RenderingEngine/VideoViewport.ts
+++ b/packages/core/src/RenderingEngine/VideoViewport.ts
@@ -335,9 +335,9 @@ class VideoViewport extends Viewport implements IVideoViewport {
   public async play() {
     try {
       if (!this.isPlaying) {
+        this.isPlaying = true;
         // Play returns a promise that is true when playing completes.
         await this.videoElement.play();
-        this.isPlaying = true;
         this.renderWhilstPlaying();
       }
     } catch (e) {
@@ -346,10 +346,10 @@ class VideoViewport extends Viewport implements IVideoViewport {
     }
   }
 
-  public async pause() {
+  public pause() {
     try {
-      await this.videoElement.pause();
       this.isPlaying = false;
+      this.videoElement.pause();
     } catch (e) {
       // No-op - sometimes this happens on startup
     }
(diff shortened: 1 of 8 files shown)


packages/core/src/RenderingEngine/VideoViewport.ts
```diff
@@ -335,9 +335,9 @@ class VideoViewport extends Viewport implements IVideoViewport {
   public async play() {
     try {
       if (!this.isPlaying) {
+        this.isPlaying = true;
```
wayfarer3130, line 338: The timing of these was causing the cine play stuff to fail.  It had originally been done after the methods returned, but that ends up not actually fixing anything, so I have move them before so the cine works.

packages/core/src/RenderingEngine/VideoViewport.ts
```diff
@@ -808,7 +808,7 @@ class VideoViewport extends Viewport implements IVideoViewport {
   }
 
   public getSliceIndex() {
-    return this.getCurrentImageIdIndex();
+    return this.getCurrentImageIdIndex() / this.scrollSpeed;
```
wayfarer3130, line 811: The slice index needs to agree with the scroll speed/number of slices.

packages/core/src/RenderingEngine/VideoViewport.ts
```diff
@@ -1110,6 +1114,8 @@ class VideoViewport extends Viewport implements IVideoViewport {
       element: this.element,
       viewportId: this.id,
       viewport: this,
+      imageIndex: this.getCurrentImageIdIndex(),
```
wayfarer3130, line 1117: Need these for updating the scroll bar.  The integer frame index needs to be used so that every frame is accessible.

packages/tools/src/utilities/stackPrefetch/stackContextPrefetch.ts
wayfarer3130, line 1: Lots of little warning messages that were clogging things up trying to debug video playback.

sedghi (approved):
Thanks

PR #1356, opened 2024-06-26
examples(scale-overlay): update scale overlay example to contain 2 stack viewports

IbrahimCSAE:
### Context

Just adding an additional viewport to the example.


sedghi (approved):
Thanks

PR #1357, opened 2024-06-27
fix:(synchronizer): destroy remove Synchronizers from state.synChroni…

David2k13:
…zers

destroy should remove Synchronizers from state.synChronizers. I didn't see the code that removed Synchronizers

### Context

image

I didn't see the code that removed Synchronizers

I think the synchronizer should be deleted from the state when destroying.

### Changes & Results
```
 public destroy(): void {
  +  const index = state.synchronizers.findIndex((sync) => sync.id === this.id);

  +  if (index > -1) state.synchronizers.splice(index, 1);

    this._sourceViewports.forEach((s) => this.removeSource(s));
    this._targetViewports.forEach((t) => this.removeTarget(t));
  }
```

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  

### DIFF
--- a/packages/tools/src/store/SynchronizerManager/Synchronizer.ts
+++ b/packages/tools/src/store/SynchronizerManager/Synchronizer.ts
@@ -4,6 +4,7 @@ import {
   Enums,
   Types,
 } from '@cornerstonejs/core';
+import { state } from '../index';
 
 import { ISynchronizerEventHandler } from '../../types';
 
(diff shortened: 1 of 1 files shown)


IbrahimCSAE (changes requested):
I think this is already handled in 

https://github.com/cornerstonejs/cornerstone3D/blob/8eb59fdd500576bb371ba8987a1e8baec3371ac5/packages/tools/src/store/SynchronizerManager/destroySynchronizer.ts#L12

Adding this change would break that since we would be splicing twice now.

David2k13:
> I think this is already handled in

Thank you. I made a mistake

PR #1387, opened 2024-07-11
Fixed a bug with ReferenceLinesTool

Steven-DD:
In some medical imaging the "reference Lines Tool" does not get the intersection coordinates correctly with the "show Full Dimension" option turned on, causing the tool not to work properly.

I added a fallback action to it

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  

### DIFF
--- a/packages/tools/src/tools/ReferenceLinesTool.ts
+++ b/packages/tools/src/tools/ReferenceLinesTool.ts
@@ -246,11 +246,16 @@ class ReferenceLines extends AnnotationDisplayTool {
     const lineDash = this.getStyle('lineDash', styleSpecifier, annotation);
     const color = this.getStyle('color', styleSpecifier, annotation);
     const shadow = this.getStyle('shadow', styleSpecifier, annotation);
-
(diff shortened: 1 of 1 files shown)


sedghi:
Thanks for your PR. Could you provide a before-and-after image or share the specific data that is broken?

Steven-DD:
> Thanks for your PR. Could you provide a before-and-after image or share the specific data that is broken?

I'm sorry I can't show you the full cause of this problem, as it was discovered during the development of a commercial software, but I can give you some screenshots and describe what I think is going wrong.

I'm not sure what the underlying cause of the problem is, but in some of the images I've encountered, turning on the ‘showFullDimension’ option calculates an incorrect coordinate, as shown in the following image

image

This problem can be solved by not using the 'showFullDimension' option, so I'm not sure if adding a fallback mechanism is a sensible solution, but at least this would allow the line to show instead of nothing happening.

PR #1388, opened 2024-07-11
fix: Fix HTJ2K decoder leak

abustany:
### Context

Commit 85fd19396762f54c6806fdbebf0235139a67629a changed decodeHTJ2K.js to allocate a new HTJ2KDecoder with each call to decodeAsync. The old decoders somehow stay alive, and are not cleaned from the WASM memory. Each new HTJ2KDecoder allocates in turns buffers for the encoded and decoded pixel data while decoding.

### Changes & Results

The change to allocate a new decoder per frame argued that reusing the old decoder is "much slower", but I could not reproduce any slowdowns either in the browser or on Node. This commit therefore reverts the change, so that each call to decodeAsync reuses the same decoder instance.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] "OS: Linux 
- [x] "Node version: 20.11 
- [x] "Browser: Chrome 126, Firefox 128
  

### DIFF
--- a/packages/dicomImageLoader/src/shared/decoders/decodeHTJ2K.ts
+++ b/packages/dicomImageLoader/src/shared/decoders/decodeHTJ2K.ts
@@ -57,8 +57,7 @@ export function initialize(decodeConfig?: LoaderDecodeOptions): Promise<void> {
 // https://github.com/chafey/openjpegjs/blob/master/test/browser/index.html
 async function decodeAsync(compressedImageFrame: ByteArray, imageInfo) {
   await initialize();
-  // const decoder = local.decoder; // This is much slower for some reason
-  const decoder = new local.codec.HTJ2KDecoder();
+  const decoder = local.decoder;
 
   // get pointer to the source/encoded bit stream buffer in WASM memory
   // that can hold the encoded bitstream


packages/dicomImageLoader/src/shared/decoders/decodeHTJ2K.ts
```diff
@@ -57,8 +57,7 @@ export function initialize(decodeConfig?: LoaderDecodeOptions): Promise<void> {
 // https://github.com/chafey/openjpegjs/blob/master/test/browser/index.html
 async function decodeAsync(compressedImageFrame: ByteArray, imageInfo) {
   await initialize();
-  // const decoder = local.decoder; // This is much slower for some reason
-  const decoder = new local.codec.HTJ2KDecoder();
+  const decoder = local.decoder;
```
wayfarer3130, line 60: This was the original implementation, but it causes a bug where some images can't be decoded and the decoding gets slower and slower over time.
The problem is that some of the state in the decoder isn't being cleared correctly.  I would try updating the openjph library to latest, make sure the malloc library used is the same one we use for openjpeg, and rebuild that part of the system.  Then you can test to see if it works using the CS3D progressive loading tests.    Test the decompression time for the second decompression, and then for the last decompression time on the HTJ2K volume, as well as running the HTJ2K stack tests, with multiple runs through the stack.  The decompression time should stay roughly constant.

wayfarer3130, line 60: Should be 
const { decoder } = local;

wayfarer3130:
If you update this PR to include an openjph version change for the release after https://github.com/cornerstonejs/codecs/pull/46
then I will approve the PR and merge it.

PR #1419, opened 2024-08-08
performance(volume-rendering): Optimize image-based volume creation and caching

sedghi:
### Previous Approach

Previously, we had two separate caches: an image cache and a volume cache. The process involved putting images into the image cache from a stack viewport, and then converting these images to create a 3D scalar data volume when switching to a volume viewport. This approach led to performance issues and memory optimization challenges, as both images and scalar data were present in the CPU simultaneously, causing peak memory usage in applications.

### New Volume Creation Approach

To address these issues and improve syncing between stack and volume segmentations, a new approach for creating volumes has been introduced in this PR:
	•	Everything now loads as images
	•	Volume streaming is performed image by image
	•	Only images are cached in the image cache
	•	For volume rendering, data goes directly from image cache to GPU, bypassing CPU scalar data

This new method eliminates the need for scalar data in the CPU, reducing memory usage and improving performance.

### Voxel Manager for Tools

To handle tools that require pixel data in the CPU, a new voxel manager has been implemented:
	•	Acts as an intermediary between indexes and scalar data
	•	Provides mappers from IJK to indexes
	•	Retrieves information without creating scalar data
	•	Processes each image individually

### Handling Non-Image Volumes

For volumes without images (e.g., NIFTI volumes), the data is chopped and converted back to a stack format, making it compatible with the new image-based approach.

### Saving as Native format 

We've optimized the end loading pipeline's caching mechanism. Now, data is stored in its native format, rather than always caching as float32. This saves memory. When updating GPU textures, we convert data to the required format on-the-fly.

## Results

- Eliminated need for scalar data in CPU for volume rendering
- Improved memory usage and performance
- Streamlined data flow from image cache directly to GPU
- Implemented voxel manager for CPU-based tools
- Removed dependency on SharedArrayBuffer
- Reduced security constraints for image viewer
- Simplified web worker implementation (ArrayBuffer is now sufficient)
- Enhanced compatibility with various volume formats
- Improved syncing between stack and volume segmentations
- Optimized overall system architecture for image and volume handling
- Optimized memory usage by eliminating unnecessary data type conversions and reducing storage requirements for cached data.

### getPixelData() -> voxelManager.getScalarData()

If you relied on `image.getPixelData` in your code elsewhere, you can do the same thing via `image.voxelManager.getScalarData` now, since each image has one `voxelManager`

### getNumberOfComponents
If you previously used:

```javascript
const numberOfComponents = imageData
    .getPointData()
    .getScalars()
    .getNumberOfComponents();
```

You should now use:

```javascript
const { numberOfComponents } = imageData.get('numberOfComponents');
```

This change is necessary because we no longer have a scalar assigned to the image data, as there are no CPU scalars. Instead, we've passed the image data with metadata called 'numberOfComponents' that we can retrieve.

### deprecated useNativeDataType

You don't use the useNativeData type anymore because as mentioned above it was used to load the data into the destination buffer but we have moved away from that and we just convert on the fly.

### No SharedArrayBuffer

Remove shared array buffer dependencies, as they're no longer needed. Previously, it enabled sharing large buffers and 3D volume data among web workers for simultaneous image writing. Now, workers operate independently on individual images. For 3D textures, we upload directly to the GPU, using metadata to match images to texture parts.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  


packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -456,7 +448,9 @@ class StackViewport extends Viewport implements IStackViewport, IImagesLoader {
       spacing: vtkImageData.getSpacing(),
       origin: vtkImageData.getOrigin(),
       direction: vtkImageData.getDirection(),
-      scalarData: vtkImageData.getPointData().getScalars().getData(),
+      get scalarData() {
+        return this.csImage.voxelManager.getScalarData();
```
wayfarer3130, line 452: this.csImage?.voxelManager.getScalarData()
as csImage can be null if things aren't yet setup, and this shouldn't throw NPE

packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -552,7 +546,7 @@ class StackViewport extends Viewport implements IStackViewport, IImagesLoader {
       mapper.setPreferSizeOverAccuracy(true);
     }
 
-    if (imageData.getPointData().getNumberOfComponents() > 1) {
+    if (imageData.getPointData().getScalars().getNumberOfComponents() > 1) {
```
wayfarer3130, line 549: This doesn't agree with the issue text about how to get number of components - should it use the same format as the issue text?

packages/core/src/RenderingEngine/vtkClasses/vtkStreamingOpenGLTexture.js
```diff
@@ -239,6 +176,78 @@ function vtkStreamingOpenGLTexture(publicAPI, model) {
...
+  function updateTextureImagesUsingVoxelManager(updatedFrames) {
+    const volume = cache.getVolume(model.volumeId);
+    const imageIds = volume.imageIds;
+    for (let i = 0; i < updatedFrames.length; i++) {
+      if (updatedFrames[i]) {
+        // find the updated frames
+        const image = cache.getImage(imageIds[i]);
+
+        if (!image) {
+          continue;
```
wayfarer3130, line 206: Might drop a warn level message here, as this might indicate something got evicted before it could be rendered.

packages/core/src/RenderingEngine/vtkClasses/vtkStreamingOpenGLTexture.js
```diff
@@ -239,6 +176,78 @@ function vtkStreamingOpenGLTexture(publicAPI, model) {
...
+    for (let i = 0; i < imageIds.length; i++) {
+      model.updatedFrames[i] = true;
+    }
+  };
+
+  publicAPI.hasUpdatedFrames = () => {
+    return model.updatedFrames.some((frame) => frame);
+  };
+
+  function updateTextureImagesUsingVoxelManager(updatedFrames) {
```
wayfarer3130, line 197: This seems a lot cleaner now.  It becomes fairly obvious how to handle updating other data structures such as too large image sets given some of this behaviour.

packages/core/src/RenderingEngine/vtkClasses/vtkStreamingOpenGLTexture.js
wayfarer3130, line 1: I really like this change - I think it makes things quite a bit cleaner overall.

packages/core/src/RenderingEngine/vtkClasses/vtkStreamingOpenGLVolumeMapper.js
```diff
@@ -117,16 +131,44 @@ function vtkStreamingOpenGLVolumeMapper(publicAPI, model) {
...
+            publicAPI
+          );
+        }
+        model._scalarOpacityFunc = scalarOpacityFunc;
+      }
+    } else {
+      model.opacityTexture = opTex.oglObject;
     }
 
     // rebuild color tfun?
```
wayfarer3130, line 156: Is it possible to split some of these commented sections out into their own methods?  I prefer to have a large complex function be something like:

f() {
   doThis();
   doThat();
   doThisOtherTHing();
}

wayfarer3130:
@sedghi - could you check to see if data which is unsigned, but has a signed value after rescale still works, especially if it is 16 bit unsigned data using nearly the full range - eg CT data with intercept -1024 and slope of 1/16 roughly would need another internal data type after decompression if the rescale is to be applied.

packages/core/src/cache/cache.ts
```diff
@@ -168,19 +167,11 @@ class Cache implements ICache {
       volume.imageData.delete();
     }
 
-    // if we had views for the images of the volume, we need to restore them
-    // to avoid memory leaks
-    restoreImagesFromBuffer(volume);
```
wayfarer3130, line 173: It feels a lot better not to be restoring the images - that always felt kind of dangerous.

packages/core/src/cache/classes/ImageVolume.ts
```diff
@@ -181,19 +203,7 @@ export class ImageVolume implements IImageVolume {
...
-   * Return the scalar data for 3D volumes or the active scalar data
-   * (current time point) for 4D volumes
-   */
-  public getScalarData(): PixelDataTypedArray {
-    if (isTypedArray(this.scalarData)) {
-      return <PixelDataTypedArray>this.scalarData;
-    }
-
-    throw new Error('Unknown scalar data type');
+    return this.numTimePoints && this.numTimePoints > 1;
```
wayfarer3130, line 206: return this.numTimePoints > 1;
is sufficient, as numTimePoints not defined or null or !this.numTimePoints all result in false for that > 1.

packages/core/src/types/CPUFallbackEnabledElement.ts
```diff
@@ -36,6 +36,7 @@ interface CPUFallbackEnabledElement {
     imagePlaneModule?: ImagePlaneModule;
     imagePixelModule?: ImagePixelModule;
   };
+  voxelManager?: any;
```
wayfarer3130, line 39: Can this be typed?

packages/core/src/types/IViewport.ts
```diff
@@ -360,6 +363,18 @@ interface IViewport {
...
   getSliceIndex(): number;
+
+  /**
+   * Gets the slice view information
+   */
+  getSliceViewInfo(): {
+    width: number;
+    height: number;
+    sliceIndex: number;
+    slicePlane: number;
```
wayfarer3130, line 374: Could you document slicePlane please - it took me a moment to figure out what that was.  The rest seem self explanatory.

packages/core/src/types/VolumeProps.ts
```diff
@@ -31,10 +36,19 @@ interface VolumeProps {
   /** Image data representing the volume */
   imageData?: vtkImageData;
 
-  /** Scalar data representing the volume's intensity values */
-  scalarData: PixelDataTypedArray | Array<PixelDataTypedArray>;
+  /**
+   * The new volume model which solely relies on the separate image data
+   * and do not cache the volume data at all
+   */
+  voxelManager?: VoxelManager<number> | VoxelManager<RGB>;
```
wayfarer3130, line 43: Should the type here be exported as it's own type - so that if we add something in the future like RGBF with F being fluorescence, that we can use it directly?  I don't know if that seems likely, just concerned it might occur.

packages/core/src/utilities/VoxelManager.ts
```diff
@@ -26,19 +30,33 @@ export default class VoxelManager<T> {
...
-  // Provide direct access to the underlying data, if any
-  public scalarData: PixelDataTypedArray;
   public map: Map<number, T> | RLEVoxelMap<T>;
   public sourceVoxelManager: VoxelManager<T>;
-  public isInObject: (pointIPS, pointIJK) => boolean;
+  public isInObject: (pointLPS, pointIJK) => boolean;
   public readonly dimensions: Point3;
-  public numComps = 1;
+  public numberOfComponents = 1;
+  public getCompleteScalarDataArray?: () => ArrayLike<number>;
```
wayfarer3130, line 38: Nice to see that arraylike is getting used, that makes it much easier to understand things.

packages/core/src/utilities/VoxelManager.ts
```diff
@@ -167,46 +165,240 @@ export default class VoxelManager<T> {
...
+      const scanAxisStep = vec3.fromValues(
+        scanAxisNormal[0] * scanAxisSpacing,
+        scanAxisNormal[1] * scanAxisSpacing,
+        scanAxisNormal[2] * scanAxisSpacing
+      );
+
+      const currentPos = vec3.clone(worldPosStart) as Point3;
+
+      for (let k = kMin; k <= kMax; k++) {
+        const startPosJ = vec3.clone(currentPos);
```
wayfarer3130, line 247: There is an optimized version of this in the currently closed source code - it is quite noticeably faster as it pre-generates i,j,k arrays and adds the three numbers for each dimension.  That version actually works with sparse data voxel managers, whereas this one doesn't.  Maybe it is worth considering re-implementing that?  Lets leave it for now, but think about doing it later as it is really just a performance optimization, but the client for that closed source might be willing to contribute just that code back so it is easy for them to update in the future.

packages/core/src/utilities/VoxelManager.ts
```diff
@@ -250,24 +458,141 @@ export default class VoxelManager<T> {
...
+        }
+        break;
+      case 2: // XY plane
+        sliceSize = width * height;
+        sliceData = new SliceDataConstructor(sliceSize);
+        for (let i = 0; i < sliceSize; i++) {
+          this.setSliceDataValue(sliceData, i, this._get(startIndex + i));
+        }
+        break;
+      default:
```
wayfarer3130, line 548: If the slice plane was a normal, then it would be possible to implement this - maybe add a comment like
// Oblique plane - todo - implement as ortho normal vector.

packages/core/src/utilities/VoxelManager.ts
```diff
@@ -283,25 +608,159 @@ export default class VoxelManager<T> {
...
+    };
+
+    voxelManager._getScalarDataLength = () => {
+      const { pixelData } = getPixelInfo(0);
+      return pixelData.length * dimensions[2];
+    };
+
+    /**
+     * Retrieves the scalar data in a memory-inefficient manner.
+     * Useful for debugging, testing, or methods requiring all scalar data at once.
```
wayfarer3130, line 716: It is important to add a comment that this is READ ONLY - that is, changes to the buffer are never reflected in the underlying data unless individually merged back.

packages/core/test/stackViewport_cpu_render_test.js
```diff
@@ -55,7 +55,7 @@ function createViewport(renderingEngine, orientation, width, height) {
 }
 
 // For some reason the cpu rendering is not working properly in the CI
```
wayfarer3130, line 57: Should the comment go since I think the xdescribe => describe means it is working?

packages/core/test/utilities/VoxelManager.jest.js
wayfarer3130, line 1: Could you add tests for your new static creator methods?  There are also new methods like get the central slice that I think would be worth having unit tests.  This is a pretty core class to a lot of things now, and having better tests on it is worthwhile.

packages/core/test/volumeViewport_gpu_render_test.js
```diff
@@ -70,7 +70,7 @@ function createViewport(
   return element;
 }
 
-describe('Volume Viewport GPU -- ', () => {
+fdescribe('Volume Viewport GPU -- ', () => {
```
wayfarer3130, line 73: Is this change intended?

packages/nifti-volume-loader/src/cornerstoneNiftiImageLoader.ts
```diff
@@ -1,22 +1,213 @@
...
+): Promise<Types.IImage> {
+  let niftiBuffer = await fetchArrayBuffer({ url });
+  let niftiHeader = null;
+  let niftiImage = null;
+
+  if (NiftiReader.isCompressed(niftiBuffer)) {
+    niftiBuffer = NiftiReader.decompress(niftiBuffer);
+  }
+
+  if (NiftiReader.isNIFTI(niftiBuffer)) {
```
wayfarer3130, line 136: Should you do something if it isn't nifti - like warn or throw an error?

packages/tools/src/utilities/dynamicVolume/generateImageFromTimeData.ts
```diff
@@ -25,17 +25,16 @@ function generateImageFromTimeData(
...
 
-  // Gets scalar data for all time frames
-  const typedArrays = dynamicVolume.getScalarDataArrays();
+  const voxelManager = dynamicVolume.voxelManager;
 
-  const arrayLength = typedArrays[0].length;
+  const arrayLength = voxelManager.getScalarDataLength();
   const finalArray = new Float32Array(arrayLength);
 
   if (operation === Enums.DynamicOperatorType.SUM) {
```
wayfarer3130, line 33: In general having an object of type { [key:Enums.DynamicOperatorType]: function } and then calling the object[operation] allows for generically adding other operation types by just defining the function, and that also isolates the code for each object into the particular function, which reduces cyclomatic complexity of the method.  I know it was like this before, so I'm not really expecting you to change it.

packages/tools/src/utilities/dynamicVolume/getDataInTime.ts
```diff
@@ -187,12 +186,10 @@ function _getTimePointDataMask(frames, dynamicVolume, segmentationVolume) {
       count++;
     };
 
-    pointInShapeCallback(
-      dynamicVolume.imageData,
-      () => true,
-      averageCallback,
-      overlapIJKMinMax
-    );
+    dynamicVolume.voxelManager.forEach(averageCallback, {
```
wayfarer3130, line 189: THis is a much nicer syntax than the pointInShapeCallback syntax.

packages/tools/src/tools/annotation/ProbeTool.ts
```diff
@@ -550,37 +549,19 @@ class ProbeTool extends AnnotationTool {
...
-      const index = transformWorldToIndex(imageData, worldPos);
-
-      index[0] = Math.round(index[0]);
-      index[1] = Math.round(index[1]);
-      index[2] = Math.round(index[2]);
+      const ijk = transformWorldToIndex(imageData, worldPos);
 
-      const samplesPerPixel =
-        scalarData.length / dimensions[2] / dimensions[1] / dimensions[0];
+      ijk[0] = Math.round(ijk[0]);
```
wayfarer3130, line 557: vec3.round(ijk,ijk)

packages/tools/src/tools/annotation/ProbeTool.ts
```diff
@@ -550,37 +549,19 @@ class ProbeTool extends AnnotationTool {
...
-        let value =
-          samplesPerPixel > 2
-            ? [
-                scalarData[baseIndex],
-                scalarData[baseIndex + 1],
-                scalarData[baseIndex + 2],
-              ]
-            : scalarData[baseIndex];
+
+        let value = voxelManager.getAtIJK(ijk[0], ijk[1], ijk[2]);
```
wayfarer3130, line 564: getAtIJKPoint is more concise

wayfarer3130 (approved):
Lots of little comments, but nothing major so hitting approve - make the changes according to the comments and what looks worth doing, but I'm quite happy about this overall.

packages/core/src/RenderingEngine/vtkClasses/vtkStreamingOpenGLVolumeMapper.js
sedghi: Not really, since this is a "fork" of the vtk volume mapper, we should keep things similar to how it is done there

packages/core/src/types/VolumeProps.ts
sedghi: I would recommend doing it when it happens

packages/core/src/utilities/VoxelManager.ts
sedghi: Added a comment

PR #1420, opened 2024-08-11
fix(splineSegROI): fix for CCW splines

lscoder:
### Context

Drawing a CCW spline Segmentation ROI contour results in a weird behavior because the points are always sorted in CW direction.

### Changes & Results

Changed the spline Segmentation ROI to do not update the winding direction until the spline ROI is closed.

### Testing

Run `splineContourSegmentationTools` example and draw CW and CCW splines w/ or w/o a hole.


sedghi:
I'm gonna close this as i'm shipping it with my changes, thanks

PR #1449, opened 2024-08-30
feat(segmentation): Refactor segmentation and style handling

sedghi:
segmentation representation = segmentation + type + viewport 

segmentation style refactored for readability , it is now similar to annotation style


packages/core/src/RenderingEngine/BaseVolumeViewport.ts
```diff
@@ -1096,7 +1103,7 @@ abstract class BaseVolumeViewport extends Viewport {
       // same name, and we don't allow that) AND We cannot use only any uid, since
       // we rely on the volume in the cache for mapper. So we prefer actorUID if
       // it is defined, otherwise we use volumeId for the actor name.
-      const uid = actorUID || volumeId;
+      const uid = actorUID || uuidv4();
```
wayfarer3130, line 1106: Why is this one || and the next one ??

packages/core/src/RenderingEngine/BaseVolumeViewport.ts
```diff
@@ -1245,29 +1252,31 @@ abstract class BaseVolumeViewport extends Viewport {
...
+    if (!actorEntries?.length) {
       return;
     }
 
-    let volumeActor;
-
     if (volumeId) {
-      volumeActor = this.getActor(volumeId).actor as vtkVolume;
-    }
+      const actorEntry = actorEntries?.find(
```
wayfarer3130, line 1262: This doesn't need the ? any longer because you check it up above.

packages/core/src/RenderingEngine/BaseVolumeViewport.ts
```diff
@@ -1389,10 +1398,11 @@ abstract class BaseVolumeViewport extends Viewport {
       return;
     }
 
-    const { uid: defaultActorUID } = defaultActor;
-    volumeId = volumeId ?? defaultActorUID;
+    volumeId = volumeId ?? this.getVolumeId();
```
wayfarer3130, line 1401: I think this should probably be || - we really don't want to allow an empty string or null to be the Id, just seems like a bad idea.

packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -722,15 +723,15 @@ class StackViewport extends Viewport {
...
-    if (this.globalDefaultProperties == null) {
-      this.setDefaultProperties({
-        colormap,
-        voiRange,
-        VOILUTFunction,
-        invert,
-        interpolationType,
-      });
-    }
+    this.globalDefaultProperties = {
```
wayfarer3130, line 726: Can you explain why you use the global defaults here in preference to the specified property?  This isn't at all clear to me.

packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -1597,60 +1603,61 @@ class StackViewport extends Viewport {
 
   private matchImagesForOverlay(
```
wayfarer3130, line 1604: I would say to break this function up into smaller pieces that can be called/imported to detect if the match is present or not - the function is very hard to follow given how long it is, but since you are just changing a few things, I think leaving it for now, just noting that it probably needs at least a comment so that people don't have to read it to figure out what is going on.

packages/core/src/RenderingEngine/Viewport.ts
```diff
@@ -531,11 +531,11 @@ class Viewport {
 
   /**
    * Add a list of actors (actor entries) to the viewport
-   * @param resetCameraPanAndZoom - force reset pan and zoom of the camera,
+   * @param resetCamera - force reset pan and zoom of the camera,
    *        default value is false.
    * @param actors - An array of ActorEntry objects.
    */
-  public addActors(actors: ActorEntry[], resetCameraPanAndZoom = false): void {
+  public addActors(actors: ActorEntry[], resetCamera = false): void {
```
wayfarer3130, line 538: Could you pass an options bag here - I think it only needs the one option, but it would be really nice to have the option of providing more values here.

packages/core/src/RenderingEngine/VolumeViewport.ts
```diff
@@ -668,10 +680,11 @@ class VolumeViewport extends BaseVolumeViewport {
       this.updateClippingPlanesForActors(this.getCamera());
     }
 
-    const imageVolume = cache.getVolume(volumeActor.uid);
+    volumeId = volumeId ?? this.getVolumeId();
```
wayfarer3130, line 683: I think volumeId ||= this.getVolumeId() is probably better - it disallows null/empty string as the volume id (and also 0), which I think is best.  It also only re-assigns if the current value is empty, whereas your statement always re-assigns.

packages/core/src/RenderingEngine/VolumeViewport3D.ts
```diff
@@ -100,11 +100,12 @@ class VolumeViewport3D extends BaseVolumeViewport {
       this.updateClippingPlanesForActors(this.getCamera());
     }
 
-    const imageVolume = cache.getVolume(volumeActor.uid);
+    volumeId = volumeId ?? this.getVolumeId();
```
wayfarer3130, line 103: Ditto - volumeId ||= this.getVolumeId()

packages/core/src/RenderingEngine/WSIViewport.ts
```diff
@@ -233,29 +253,41 @@ class WSIViewport extends Viewport {
...
-      },
+
       hasPixelSpacing: this.hasPixelSpacing,
       calibration: this.calibration,
       preScale: {
         scaled: false,
       },
       scalarData: this.getScalarData(),
+      imageData,
+      voxelManager: {
```
wayfarer3130, line 264: I don't quite get this - it doesn't have the rest of hte voxel manager functions.  Can you at least comment what this is doing?

packages/core/src/RenderingEngine/helpers/setDefaultVolumeVOI.ts
```diff
@@ -88,26 +88,30 @@ function getVOIFromMetadata(imageVolume: IImageVolume): VOIRange {
...
-        windowCenter: Array.isArray(windowCenter)
-          ? windowCenter[0]
-          : windowCenter,
-      };
+      const width = Array.isArray(windowWidth) ? windowWidth[0] : windowWidth;
+      const center = Array.isArray(windowCenter)
+        ? windowCenter[0]
+        : windowCenter;
+
+      // Skip if both windowWidth and windowCenter are 0
```
wayfarer3130, line 96: Don't you only care if the width is 0?  The center shouldn't matter if the width is zero - I mean it is defined, everything below center is black, and everything above or equal is white.  But that actually works at 0, 0 for width/center for positive/negative values.

packages/core/src/loaders/configuration/interleavedRetrieve.ts
```diff
@@ -87,10 +87,5 @@ const interleavedRetrieveConfiguration: RetrieveStage[] = [
     requestType: RequestType.Thumbnail,
     retrieveType: 'multipleFinal',
   },
-  {
```
wayfarer3130, line 90: Can you leave this commented out - I need to fix things rather than removing it, because there are real situations where it needs to fallback to the default retrieval.

packages/core/src/utilities/VoxelManager.ts
```diff
@@ -468,6 +476,7 @@ export default class VoxelManager<T> {
    */
   public static addBounds(bounds: BoundsIJK, point: Point3) {
     if (!bounds) {
+      console.debug('🚀 ~ reset bounds');
```
wayfarer3130, line 479: If you are going to leave console statements around, could you introduce a logging library that can be redirected/replaced?  I'd love to use log.debug(....) and have log be a type specific to VoxelManager so we can actually figure out what is going on and turn on specific logging bits at runtime.

packages/core/src/utilities/pointInShapeCallback.ts
```diff
@@ -0,0 +1,179 @@
...
+ * true if the point is in the shape and false if it is not.
+ * @param callback - A function that will be called for
+ * every point in the shape.
+ * @param boundsIJK - The bounds of the volume in IJK coordinates.
+ */
+export function pointInShapeCallback(
+  imageData: vtkImageData | CPUImageData,
+  pointInShapeFn: ShapeFnCriteria,
+  callback?: PointInShapeCallback,
+  boundsIJK?: BoundsIJK
```
wayfarer3130, line 49: Can you change to an options argument here, that allows specifying boundsIJK, but also other types of ranges/sets of values for future calculations?  For example, knowing an orthogonal vector and a set of min/max basis points with a min/max generator function for each row allows an optimized sphere function to be calculated.  The pointInShapeFn should be part of hte options, and it should indicate whether or not to return the set of points.   There are just a lot of optimizations one can make for this which rely on having alternate options for how to determine the content.

packages/docs/docs/migration-guides/1.x-to-2.x.md
```diff
@@ -345,6 +345,14 @@ is now `VOLUME_VIEWPORT_SCROLL_OUT_OF_BOUNDS`
 This change was made to maintain consistency with the naming of other events in the library.
 </details>
 
+### STACK_VIEWPORT_NEW_STACK
+
+is now VIEWPORT_NEW_IMAGE_SET adn we will gradually bring all viewports to use this event instead
```
wayfarer3130, line 350: Thank you - this is a real improvement.

packages/tools/examples/PolySegWasmContourToSurface/index.ts
```diff
@@ -150,6 +152,21 @@ addButtonToToolbar({
   },
 });
 
+addLabelToToolbar({
```
wayfarer3130, line 155: Can you add a helper method for this bit of code?  It feels like it will be jsut common enough to be useful having.

packages/tools/examples/SculptorTool/index.ts
```diff
@@ -262,27 +251,16 @@ async function run() {
...
-
-  // Store the segmentation representation that was just created
-  segmentationRepresentationUID = segmentationRepresentationUIDs[0];
-
-  // Make the segmentation created as the active one
-  segmentation.activeSegmentation.setActiveSegmentationRepresentation(
-    toolGroupId,
-    segmentationRepresentationUID
-  );
+  await segmentation.addSegmentationRepresentations(viewportIds[0], [
```
wayfarer3130, line 254: I do like the new code - seems easier to manage.

packages/tools/examples/advancedMagnifyTool/index.ts
```diff
@@ -174,10 +174,12 @@ async function addSegmentationsToState(volumeId: string) {
   }
 
   // Create a segmentation of the same resolution as the source data
-  segmentationVolume =
-    await volumeLoader.createAndCacheDerivedSegmentationVolume(volumeId, {
+  segmentationVolume = await volumeLoader.createAndCacheDerivedLabelmapVolume(
```
wayfarer3130, line 177: What about allowing the second parameter to be either a string or the full definition?  It strikes me that most calls will just pass a string in, and it is much more obvious to me to be 
await volumeLoader.createAndCacheDerivedLabelmapVolume(volumeId,segmentationid)
for the 90% use case.

packages/tools/examples/splineContourSegmentationTools/index.ts
```diff
@@ -48,7 +48,7 @@ const { ViewportType } = Enums;
 const toolGroupId = 'STACK_TOOLGROUP_ID';
 
 const segmentationId = `SEGMENTATION_ID`;
-let segmentationRepresentationUID = '';
+const segmentationRepresentationUID = '';
```
wayfarer3130, line 51: This doesn't appear to still be used.

packages/tools/examples/stackLabelmapSegmentation/index.ts
```diff
@@ -112,11 +107,11 @@ const optionsValues = [
   ...brushValues,
   RectangleScissorsTool.toolName,
   CircleScissorsTool.toolName,
-  SphereScissorsTool.toolName,
   PaintFillTool.toolName,
 ];
 
 let viewport;
+const viewportId2 = 'STACK_VIEWPORT_2';
```
wayfarer3130, line 114: Could you create an import that has viewportId1='viewportId1' to viewportId4 or whatever you use, then just import them?   Seems like using a consistent set would be useful.  Can be done later, just that it is annoying to copy an example and find the ids no longer match up.

packages/tools/src/eventListeners/segmentation/imageChangeEventListener.ts
```diff
@@ -66,69 +71,41 @@ function _imageChangeEventListener(evt) {
...
-  //   }
-
-  //   return true;
-  // });
+  const labelmapActors = labelmapRepresentations
+    .map((representation) => {
+      const actorUID = getLabelmapActorUID(representation.segmentationId);
+      return actors.find((actor) => actor.uid === actorUID);
+    })
+    .filter((actor) => actor !== undefined);
```
wayfarer3130, line 98: !!actor is a bit shorter

packages/tools/src/stateManagement/segmentation/SegmentationStateManager.ts
```diff
@@ -62,6 +50,10 @@ export default class SegmentationStateManager {
     Map<string, string>
   >();
 
+  /**
+   * Creates an instance of SegmentationStateManager.
+   * @param {string} [uid] - Optional unique identifier for the manager.
+   */
   constructor(uid?: string) {
     if (!uid) {
       uid = csUtils.uuidv4();
```
wayfarer3130, line 59: How about uid ||= csUtils.uuidv4(); to replace the if part, or else assign a default value in the constructor itself.

packages/tools/src/stateManagement/segmentation/SegmentationStateManager.ts
```diff
@@ -72,24 +64,24 @@ export default class SegmentationStateManager {
 
   /**
    * Returns a copy of the current state of the segmentation.
-   * @returns A deep copy of the segmentation state.
+   * @returns {SegmentationState} A deep copy of the segmentation state.
    */
   getState(): SegmentationState {
     return this.state;
```
wayfarer3130, line 70: This is not a deep copy - can you update the comment?

packages/tools/src/stateManagement/segmentation/SegmentationStateManager.ts
```diff
@@ -151,56 +144,146 @@ export default class SegmentationStateManager {
...
+    renderingConfig: RenderingConfig
+  ): void {
+    const enabledElement = getEnabledElementByViewportId(viewportId);
+
+    if (!enabledElement) {
+      return;
+    }
+
+    if (!this.state.viewports[viewportId]) {
+      this.state.viewports[viewportId] = [];
```
wayfarer3130, line 166: Should the state updates be done in a way which doesn't modify the underlying state data?  That would mean you could use a frozen version of state safely for returning in the get state, and can in fact use an internal copy that is frozen and simply gets updated/re-frozen on changes.

packages/tools/src/stateManagement/segmentation/SegmentationStateManager.ts
```diff
@@ -277,26 +346,21 @@ export default class SegmentationStateManager {
...
-   * @param labelmapImageIds - The labelmap image IDs.
-   * @param updateCallback - A callback to update the reference map.
+   * @param {string} segmentationId - The ID of the segmentation representation.
+   * @param {Types.IStackViewport} stackViewport - The stack viewport.
+   * @param {string[]} labelmapImageIds - The labelmap image IDs.
+   * @param {Function} updateCallback - A callback to update the reference map.
+   * @returns {string | undefined} The labelmap imageId reference for the current imageId rendered on the viewport.
    */
   _updateLabelmapSegmentationReferences(
     segmentationId,
```
wayfarer3130, line 366: This may well not depend on it being a stack viewport any longer - is an IViewport sufficient?  
For the update callback, could you type that with function arguments?

packages/tools/src/stateManagement/segmentation/SegmentationStateManager.ts
```diff
@@ -497,225 +566,153 @@ export default class SegmentationStateManager {
...
-        active: false,
-        visible: true,
-        segmentsHidden: new Set(),
-      };
-    }
-
-    this.state.viewports[viewportId][segmentationRepresentationUID].active =
-      true;
-  }
+    const viewport = this.state.viewports[viewportId];
```
wayfarer3130, line 579: This naming feels odd - the return value is an array of something, which doesn't feel like a viewport

wayfarer3130: Sionce it seems to be segmentation representations on the viewport, could we call it this.state.segRepresentations?

packages/tools/src/stateManagement/segmentation/SegmentationStateManager.ts
```diff
@@ -497,225 +566,153 @@ export default class SegmentationStateManager {
...
-    if (!this.state.viewports?.[viewportId]) {
+    if (!viewport) {
       return;
     }
 
-    const activeSegRep = Object.entries(this.state.viewports[viewportId]).find(
-      ([, value]) => value.active
+    const viewportRendering = viewport.find(
+      (segRep) =>
+        segRep.segmentationId === segmentationId && segRep.type === type
```
wayfarer3130, line 587: Can one omit the type here, to remove the first representation for that segmentationId?  It seems like the type probably isn't needed much of the time.

packages/tools/src/stateManagement/segmentation/SegmentationStateManager.ts
```diff
@@ -497,225 +566,153 @@ export default class SegmentationStateManager {
...
+  setActiveSegmentation(viewportId: string, segmentationId: string): void {
+    const viewport = this.state.viewports[viewportId];
 
-    if (!segmentationRepresentation) {
+    if (!viewport) {
       return;
     }
 
-    return segmentationRepresentation.config;
+    Object.entries(viewport).forEach(([key, value]) => {
```
wayfarer3130, line 604: You don't use the key, so you could use Object.values(viewport).forEach(value => ...

packages/tools/src/stateManagement/segmentation/SegmentationStateManager.ts
```diff
@@ -497,225 +566,153 @@ export default class SegmentationStateManager {
...
    */
   getSegmentationRepresentationVisibility(
     viewportId: string,
-    segmentationRepresentationUID: string
+    segmentationId: string,
+    type: SegmentationRepresentations
   ): boolean {
-    const viewport = this.state.viewports[viewportId];
-    return viewport && viewport[segmentationRepresentationUID]?.visible;
+    const viewportRepresentations = this.getSegmentationRepresentations(
```
wayfarer3130, line 677: It seems to be a common pattern to get the segmentation representation specifically, eg the first one which matches.  Adding a funciton getSegmentationRepresentation(viewportId, segmentationid, type) return undefined or the representation seems like it would simplify this code.

packages/tools/src/stateManagement/segmentation/SegmentationStateManager.ts
```diff
@@ -497,225 +566,153 @@ export default class SegmentationStateManager {
...
    * @param viewportId - The ID of the viewport.
-   * @param segmentationRepresentationUID - The UID of the segmentation representation.
+   * @param segmentationId - The ID of the segmentation.
+   * @param type - The type of the segmentation representation.
    * @param visible - The visibility to set for the segmentation representation in the viewport.
    */
   setSegmentationRepresentationVisibility(
     viewportId: string,
-    segmentationRepresentationUID: string,
+    segmentationId: string,
```
wayfarer3130, line 697: Allow either of thse two to be not provided., as then one could set set invisble all contour types.

packages/tools/src/stateManagement/segmentation/SegmentationStateManager.ts
```diff
@@ -752,9 +749,9 @@ export default class SegmentationStateManager {
   }
 }
 
-async function computeVolumeSegmentationFromStack({
+async function internalComputeVolumeLabelmapFromStack({
   imageIds,
-  options,
+  options = {},
```
wayfarer3130, line 754: Make this optional rather than defining a default value, as line 766 below allows for optionality.  Also, the volumeId should be || not ??

packages/tools/src/stateManagement/segmentation/SegmentationStyle.ts
```diff
@@ -0,0 +1,378 @@
...
+  };
+  segmentations: {
+    [segmentationId: string]: {
+      [key in SegmentationRepresentations]?: {
+        allSegments?: RepresentationStyle;
+        perSegment?: { [key: number]: RepresentationStyle };
+      };
+    };
+  };
+  viewports: {
```
wayfarer3130, line 23: This should be called something like viewportsStyle

packages/tools/src/stateManagement/segmentation/addColorLUT.ts
```diff
@@ -1,13 +1,40 @@
...
+  let colorLUTToUse = colorLUT;
+  // make sure the colorLUT is always starts at [0, 0, 0, 0], and
+  // also has multiple entries, and if it is missing get it from
+  // constant CORNERSTONE_COLOR_LUT
+  // Append the "zero" (no label) color to the front of the LUT, if necessary.
+  if (!utilities.isEqual(colorLUTToUse[0], [0, 0, 0, 0])) {
+    console.warn(
+      'addColorLUT: [0, 0, 0, 0] color is not provided for the background color (segmentIndex =0), automatically adding it'
+    );
+    colorLUTToUse.unshift([0, 0, 0, 0]);
```
wayfarer3130, line 28: This ends up modifying the incoming lut data as a side affect - I'd rather it look like the pattern down below which copies the incoming array, eg
colorLUTToUse = [[0,0,0,0], ...colorLUTToUse]
Also, if there are methods to update hte color lut, then I'd rather always take a copy, and then modify that, eg start with
colorLUTToUse = [...colorLUT] as that generally prevents side affects.

packages/tools/src/stateManagement/segmentation/addSegmentationRepresentationsToViewport.ts
```diff
@@ -0,0 +1,153 @@
...
+ */
+export async function addSegmentationRepresentations(
+  viewportId: string,
+  segmentationInputArray: RepresentationPublicInput[]
+) {
+  const promises = segmentationInputArray.map((segmentationInput) => {
+    return internalAddSegmentationRepresentation(viewportId, segmentationInput);
+  });
+
+  Promise.all(promises);
```
wayfarer3130, line 16: This isn't an async function as written - just remove the async, and change the signature to:
export function addSegmentationRepresentations(...): Promise<any>

then on line 16, return Promise.all(promise)

You only need the async indicator when you use await within the function body itself.  Otherwise the function is simpler to understand when transpiled just returning a direct promise.

wayfarer3130: Even for a function without await, the async changes the return value and body when it transpiles, so please avoid the keyword unless you actually need it.

packages/tools/src/stateManagement/segmentation/addSegmentationRepresentationsToViewport.ts
```diff
@@ -0,0 +1,153 @@
...
+
+/**
+ * Adds multiple surface segmentations to multiple viewports.
+ *
+ * @param viewportInputMap - An object mapping viewport IDs to arrays of surface segmentation inputs.
+ * @returns A promise that resolves to an object mapping viewport IDs to arrays of segmentation representation UIDs.
+ */
+async function addSurfaceRepresentationToViewportMap(viewportInputMap: {
+  [viewportId: string]: RepresentationPublicInput[];
+}): Promise<{ [viewportId: string]: string[] }> {
```
wayfarer3130, line 132: This is a nice way to define the Promise result as it is clear what you return.

packages/tools/src/stateManagement/segmentation/config/segmentationColor.ts
```diff
@@ -1,89 +1,89 @@
...
+ * Sets the color LUT index for a segmentation in a specific viewport.
+ * @param viewportId - The ID of the viewport.
+ * @param segmentationId - The ID of the segmentation.
+ * @param colorLUTIndex - The index of the color LUT to set.
  */
 function setColorLUT(
-  segmentationRepresentationUID: string,
+  viewportId: string,
+  segmentationId: string,
   colorLUTIndex: number
```
wayfarer3130, line 32: Could we call this one colorLUTsIndex to indicate that it is the index into the color luts, not the index into a single color lut?  The overall function name then makes more sense to me, at it chooses which color lut to apply from the set of color luts.

packages/tools/src/stateManagement/segmentation/config/segmentationColor.ts
```diff
@@ -98,13 +98,15 @@ function getSegmentIndexColor(
...
   segmentIndex: number,
   color: Types.Color
 ): void {
   // Get the reference to the color in the colorLUT.
   const colorReference = getSegmentIndexColor(
-    segmentationRepresentationUID,
+    viewportId,
+    segmentationId,
     segmentIndex
   );
```
wayfarer3130, line 112: Since the color values are directly modified, when we add color luts, do we ensure we take a full copy so we don't later on modify it?

packages/tools/src/stateManagement/segmentation/getSegmentationRepresentation.ts
```diff
@@ -1,17 +1,62 @@
...
-    segmentationRepresentationUID
-  );
+
+  const representations =
+    segmentationStateManager.getSegmentationRepresentations(viewportId, {
+      segmentationId,
+      type,
+    });
+
+  return representations[0];
```
wayfarer3130, line 61: Could you add an assert that the representations.length is 0 or 1?  Otherwise it feels like it might ignore some, even though in the design it shouldn't.

packages/tools/src/stateManagement/segmentation/getViewportIdsWithSegmentation.ts
```diff
@@ -12,11 +12,13 @@ export function getViewportIdsWithSegmentation(
...
   const viewports = state.viewports;
 
-  return Object.keys(viewports).filter((viewportId) => {
-    const viewport = viewports[viewportId];
-    return Object.keys(viewport).some(
-      (segRepUID) =>
-        state.representations[segRepUID].segmentationId === segmentationId
-    );
-  });
+  const viewportIdsWithSegmentation = Object.entries(viewports)
```
wayfarer3130, line 15: Use Object.values if you aren't using the key.

packages/tools/src/stateManagement/segmentation/getViewportSegmentations.ts
```diff
@@ -0,0 +1,35 @@
...
+  const segmentationStateManager = defaultSegmentationStateManager;
+  const state = segmentationStateManager.getState();
+  const viewports = state.viewports;
+
+  const segmentationsForViewportId = viewports[viewportId];
+
+  const segmentations = segmentationsForViewportId.map(
+    (segmentationInViewport) =>
+      getSegmentation(
+        type && segmentationInViewport.type === type
```
wayfarer3130, line 24: If you are trying to allow excluding the type to mean any type, then this set of lines should be:
(!type || segmentationInViewport.type===type)

packages/tools/src/stateManagement/segmentation/getViewportSegmentations.ts
```diff
@@ -0,0 +1,35 @@
...
+import { defaultSegmentationStateManager } from './SegmentationStateManager';
+
+/**
+ * Retrieves the viewport IDs that have a specific segmentation.
+ * @param segmentationId - The ID of the segmentation.
+ * @returns An array of viewport IDs that have the specified segmentation.
+ */
+export function getViewportSegmentations(
+  viewportId: string,
+  type: SegmentationRepresentations
```
wayfarer3130, line 13: This should be type? to allow for any type.

packages/tools/src/stateManagement/segmentation/helpers/computeStackLabelmapFromVolume.ts
```diff
@@ -29,6 +29,10 @@ export async function convertVolumeToStackSegmentation({
 }): Promise<void> {
   const segmentation = getSegmentation(segmentationId);
 
+  if (!segmentation) {
+    return;
+  }
+
   const { volumeId } = segmentation.representationData
     .Labelmap as LabelmapSegmentationDataVolume;
   const segmentationVolume = cache.getVolume(volumeId) as Types.IImageVolume;
```
wayfarer3130, line 38: If you just return updateStackSegmentationState then you don't need the async declaration and the function becomes a bit simpler.

packages/tools/src/stateManagement/segmentation/setActiveSegmentation.ts
```diff
@@ -0,0 +1,15 @@
+import { defaultSegmentationStateManager } from './SegmentationStateManager';
+
+/**
+ * Retrieves the active segmentation representation for a given viewport.
```
wayfarer3130, line 4: Sets the active segmentation....

packages/tools/src/stateManagement/segmentation/setActiveSegmentation.ts
```diff
@@ -0,0 +1,15 @@
...
+ *
+ * @param viewportId - The ID of the viewport.
+ * @returns The active segmentation representation, or undefined if not found.
```
wayfarer3130, line 7: No return

packages/tools/src/stateManagement/segmentation/setActiveSegmentation.ts
```diff
@@ -0,0 +1,15 @@
...
+ */
+export function setActiveSegmentation(
+  viewportId: string,
+  segmentationId: string
```
wayfarer3130, line 11: Should this include the segmentation type?  Also, should it be possible to have 1 active of each type of segmentation?  That is, to have both the active labelmap and the active contour.

packages/tools/src/store/ToolGroupManager/ToolGroup.ts
```diff
@@ -8,7 +8,7 @@ import {
   getEnabledElementByIds,
   Settings,
 } from '@cornerstonejs/core';
-import type { Types } from '@cornerstonejs/core';
+import { type Types, utilities } from '@cornerstonejs/core';
```
wayfarer3130, line 11: Didn't know one could import types this way - that is cleaner than having separate imports.

packages/tools/src/store/ToolGroupManager/ToolGroup.ts
```diff
@@ -216,13 +216,21 @@ export default class ToolGroup {
...
       throw new Error('No rendering engines found.');
     }
-    if (renderingEngines.length > 1) {
-      throw new Error(
-        'Multiple rendering engines found. You must specify a renderingEngineId.'
-      );
-    }
 
-    const renderingEngineUIDToUse = renderingEngineId || renderingEngines[0].id;
+    // if there are multiple rendering engines, we need to search for the one that
```
wayfarer3130, line 220: Can you encapsulate this in a helper method - that just makes the overall function easier to understand.

packages/tools/src/store/ToolGroupManager/ToolGroup.ts
```diff
@@ -765,7 +773,7 @@ export default class ToolGroup {
       get(this._toolInstances[toolName].configuration, configurationPath) ||
       this._toolInstances[toolName].configuration;
 
-    return structuredClone(_configuration);
+    return utilities.deepClone(_configuration);
```
wayfarer3130, line 776: Why is this one deepClone and not structuredClone?

packages/core/src/RenderingEngine/WSIViewport.ts
sedghi: It is for the annotations to work, since all of them work on voxelManager and not on scalarData now, i added a comment

packages/core/src/utilities/VoxelManager.ts
sedghi: It was a mistake, removed it, but logging lib is needed

packages/tools/src/tools/base/ContourSegmentationBaseTool.ts
```diff
@@ -79,17 +75,15 @@ abstract class ContourSegmentationBaseTool extends ContourBaseTool {
       return contourAnnotation;
     }
 
-    const activeSeg = getActiveSegmentationRepresentation(viewport.id);
+    const activeSeg = getActiveSegmentation(viewport.id);
```
wayfarer3130, line 78: Here is where it would be nice to just include the type, and then if !activeSeg, it simply will throw the single error.

packages/tools/examples/advancedMagnifyTool/index.ts
sedghi: The library evolves over time, and we may eventually need options for various reasons I'm pretty sure. I'm confident we'll regret not including them.

packages/tools/src/tools/displayTools/Labelmap/addLabelmapToElement.ts
```diff
@@ -36,15 +47,40 @@ async function addLabelmapToElement(
   const immediateRender = false;
   const suppressEvents = true;
 
-  if (isVolumeSegmentation(labelMapData, viewport)) {
+  if (viewport instanceof BaseVolumeViewport) {
```
wayfarer3130, line 50: This bit feels like it should be in a helper method  'ensureVolumeHasVolumeId'   I think that would reduce the complexity of the method.

packages/tools/examples/stackLabelmapSegmentation/index.ts
sedghi: Yes, later sounds good

packages/tools/src/tools/displayTools/Surface/removeSurfaceFromElement.ts
```diff
@@ -1,36 +1,35 @@
...
+  segmentationId: string
 ): void {
   const enabledElement = getEnabledElement(element);
 
   const { viewport } = enabledElement;
 
   const actorEntries = (viewport as Types.IVolumeViewport).getActors();
 
-  // remove actors whose id has the same prefix as the segmentationRepresentationUID
+  const actorUID = `${segmentationId}-${SegmentationRepresentations.Surface}`;
```
wayfarer3130, line 24: Could you please create a utility function which generates the actorUID given the segmentationId and type?  I know it is trivial, but if it is externalized, then it can be updated in the future without changing all the spots you have embedded that knowledge - that is, at the code level you have encapsulated the volatility of hte actorUID name.

packages/tools/src/stateManagement/segmentation/SegmentationStateManager.ts
sedghi: I like it, like React, done

packages/tools/src/stateManagement/segmentation/SegmentationStateManager.ts
sedghi: I'll call it `viewportSegmentationRepresentations`

packages/tools/src/stateManagement/segmentation/SegmentationStateManager.ts
sedghi: hmm maybe, i'll keep it here, but not in the pulbic method, this is state manager

packages/tools/src/tools/segmentation/strategies/fillCircle.ts
```diff
@@ -67,12 +66,13 @@ const initializeCircle = {
       segmentationImageData.getDimensions()
     );
 
-    segmentationVoxelManager.boundsIJK = boundsIJK;
-    imageVoxelManager.isInObject = createPointInEllipse({
+    operationData.isInObject = createPointInEllipse({
```
wayfarer3130, line 69: Here is a spot where getting the object iterator as a complete options object which includes currently the is in object and the bounds ijk would allow us in the future to define it using more efficient maps, and as long as it is just an instance of the right type, then the type can be extended in the future to allow for improved definitions.  I believe this should be part of CS3D 2 as it fixes the types now, and then allows for better future extension.

packages/tools/src/tools/StackScrollTool.ts
```diff
@@ -22,32 +31,26 @@ class StackScrollTool extends BaseTool {
         invert: false,
         debounceIfNotLoaded: true,
         loop: false,
+        // this tool can also be used to rotate the volume for instance for MIP
```
wayfarer3130, line 34: Can you bind rotate to the drag callback too?  It looks like that doesn't work, and drag/mouse move only scrolls.

packages/tools/test/segmentationState_test.js
```diff
@@ -204,19 +203,19 @@ describe('Segmentation State:', () => {
       expect(globalConfig.representations).toBeDefined();
       expect(globalConfig.representations[Labelmap]).toBeDefined();
 
-      const representationConfig = segUtils.getDefaultRepresentationConfig({
-        type: Labelmap,
-      });
+      // const representationConfig = segUtils.getDefaultRepresentationConfig({
```
wayfarer3130, line 206: Can you remove this code now?

wayfarer3130 (changes requested):
A few bug fixes as noted, plus some naming changes a few small improvements are needed, but overall this is a much cleaner design than the CS3D 1 version.

packages/tools/examples/advancedMagnifyTool/index.ts
wayfarer3130: I'm not saying to now allow the full definition, just that it is easy enough to support a string version of it for the 90% use case of only specifying segmentation id.  In fact one already needs other parameters as an option, just that making it easy to create it is convenient and is easier to read in the code.

packages/tools/src/stateManagement/segmentation/SegmentationStateManager.ts
wayfarer3130: Thank you, that is a good naming.

packages/tools/src/stateManagement/segmentation/SegmentationStateManager.ts
wayfarer3130: I was just meaning make the find condition be:
segRep.segmentationId === segmentationId && (segRep.type === type || !type)

packages/tools/src/stateManagement/segmentation/config/segmentationColor.ts
sedghi: Yes during add we do

packages/tools/src/stateManagement/segmentation/setActiveSegmentation.ts
sedghi: No, in the concept of activeness, only one segmentation (not representation) is active at a time. We can't have the contour of one segmentation active while the labelmap of another segmentation is active.

packages/tools/src/store/ToolGroupManager/ToolGroup.ts
sedghi: Since `structuredClone` cannot copy functions or classes, our custom `deepClone` function uses `structuredClone` for objects while also handling functions and classes.

packages/tools/src/tools/segmentation/strategies/fillCircle.ts
sedghi: We have the options for the forEach, i added it to the pointInShape too

packages/core/src/utilities/pointInShapeCallback.ts
```diff
@@ -35,19 +54,21 @@ export type ShapeFnCriteria = (pointLPS: vec3, pointIJK: vec3) => boolean;
...
- * @param dimensions - The dimensions of the image.
- * @param pointInShapeFn - A function that takes a point in LPS space and returns
- * true if the point is in the shape and false if it is not.
- * @param callback - A function that will be called for
- * every point in the shape.
- * @param boundsIJK - The bounds of the volume in IJK coordinates.
+ * @param options - Configuration options for the shape callback.
+ * @returns An array of points in the shape if returnPoints is true, otherwise undefined.
  */
 export function pointInShapeCallback(
```
wayfarer3130, line 60: Thank you - this feels like a better interface.

wayfarer3130 (approved):
Thank you for all these changes - I think they are really looking pretty good now

PR #1450, opened 2024-09-02
save 2nd point for arrow annotation

Joeycho:
Fortunately, TID300 at dcmjs, supports 2nd point in Point class. 

https://github.com/dcmjs-org/dcmjs/blob/master/src/utilities/TID300/Point.js

To support ArrowAnnotate at OHiF, OHIF code should be also modified (`extensions/cornerstone-dicom-sr/src/tools/DICOMSRDIsplayTool.ts`)


Joeycho:
@sedghi

sedghi (approved):

PR #1459, opened 2024-09-07
fix(ContourSegmentationBaseTool ): Fixing ContourSegmentationBaseTool filter validation

andrebot:
### Context

Cornerstone is thrown into an error state after ContourSegmentationBaseTool fails to execute `_getSegmentationRepresentation ` causing some unwanted side effects.

### Changes & Results

Fixed the validation to better fail safe the method by validating an undefined return.

before:
with error

After:
fixed

#### Tested Environment

- [] "OS: Windows 11
- [] "Node version: 18.20.4"
- [] "Browser:  Edge 128.0.2739.63 (Official build) (64-bit)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,12 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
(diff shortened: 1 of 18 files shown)


IbrahimCSAE (approved):

sedghi:
@andrebot is this not needed anymore? can you comment why you closed it?

sedghi:
sure

Issue #1455, opened 2024-09-04
[Bug] Annotations appear on incorrect images when using the wadoURI fileManager


plt-joey:
### Describe the Bug

When loading images with `cornerstoneDICOMImageLoader.wadouri.fileManager.add` they are given an id with the format `dicom:<image number>`. Annotations created on images loaded this way will also show up on the images whose image number matches the last digits of the image number they were created on.

For example: If you were to load 200 images, then create an annotation on the image with the ID `dicom:126` it will also be displayed when rendering the images `dicom:26` or `dicom:6`.

### Steps to Reproduce

I have created a minimal repro of this issue here (instructions in the README): https://github.com/plt-joey/cornerstone-issue-repro/tree/master

### The current behavior

The repro displays two viewports with the `CircleROITool` active, the left viewport is displaying `dicom:0`, the right viewport is displaying `dicom:10`. If you create an annotation in the right viewport it will also be displayed in the left one, but notably the reverse is not true, creating annotations in the left viewport only display in the left viewport.

This is not limited to displaying multiple viewports, the annotation will also incorrectly show up if you have a single viewport on `dicom:10`, create an annotation, then switch it to `dicom:0`

### The expected behavior

The expected behavior is that annotations should only show up on the image they was created on.

### OS

Windows 10

### Node version

18.12.1

### Browser

Chrome (128.0.6613.114) & Firefox (130.0)

plt-joey:
Created https://github.com/cornerstonejs/cornerstone3D/pull/1460 to address this

PR #1460, opened 2024-09-10
fix: Annotations showing up on incorrect images

plt-joey:
### Context

Fixes issue described here: https://github.com/cornerstonejs/cornerstone3D/issues/1455

### Changes & Results

Modified the image URI check in the `isReferenceViewable` function in `StackViewport` from an `endsWith` match to an exact match. This stops annotations on images with IDs like `dicom:126` (which is the format used by the wadoURI filemanager) from displaying on the images `dicom:26` and `dicom:6` as well.

### Testing

The repro provided in the linked issue demonstrates the bug, replacing the UMD build of `@cornerstonejs/core` with one containing this change will fix the bug.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] "OS: Windows 11"
- [x] "Node version: 18.12.1"
- [x] "Browser: Chrome 128.0.6613.120"
  

### DIFF
--- a/packages/core/src/RenderingEngine/StackViewport.ts
+++ b/packages/core/src/RenderingEngine/StackViewport.ts
@@ -3042,13 +3042,11 @@ class StackViewport extends Viewport {
       }
     }
 
-    let { imageURI } = options;
+    const { imageURI } = options;
 
-    if (!imageURI) {
-      // Remove the dataLoader scheme since that can change
-      imageURI = imageIdToURI(referencedImageId);
-      return true;
-    } else if (imageIdToURI(referencedImageId) === imageURI) {
+    // Direct URI match check
+    const currentImageURI = imageIdToURI(referencedImageId);
+    if (!imageURI || currentImageURI === imageURI) {
       return true;
     }
 
--- a/packages/core/src/RenderingEngine/StackViewport.ts
+++ b/packages/core/src/RenderingEngine/StackViewport.ts
@@ -3049,11 +3049,13 @@ class StackViewport extends Viewport {
       imageURI = imageIdToURI(currentImageId);
     }
 
-    const referencedImageURI = imageIdToURI(referencedImageId);
+    const referenceColonIndex = referencedImageId.indexOf(':');
+    const referencedImageURI = referencedImageId.substring(
+      referenceColonIndex + 1
+    );
 
-    const endsWith = referencedImageId?.endsWith(imageURI);
-    if (endsWith) {
-      return endsWith;
+    if (referencedImageURI === imageURI) {
+      return true;
     }
 
     // if camera focal point is provided, we can use that as a point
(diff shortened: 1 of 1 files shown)


packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -2927,9 +2927,13 @@ class StackViewport extends Viewport implements IStackViewport, IImagesLoader {
       imageURI = imageId.substring(colonIndex + 1);
     }
 
-    const endsWith = referencedImageId?.endsWith(imageURI);
-    if (endsWith) {
-      return endsWith;
+    const referenceColonIndex = referencedImageId.indexOf(':');
```
sedghi, line 2930: can you use this one packages/core/src/utilities/imageIdToURI.ts

plt-joey: Sure can, updated this check as well as the check above it (and fixed some redundancy there)

packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -3046,14 +3046,10 @@ class StackViewport extends Viewport {
...
-      imageURI = imageIdToURI(currentImageId);
-    }
-
-    const referencedImageURI = imageIdToURI(referencedImageId);
-
-    const endsWith = referencedImageId?.endsWith(imageURI);
-    if (endsWith) {
-      return endsWith;
+      imageURI = imageIdToURI(referencedImageId);
+      return true;
```
sedghi, line 3050: why return true here? I don't understand

plt-joey: just removes a redundant check

before when it looked like:
```
if (!imageURI) {
    imageURI = imageIdToURI(referencedImageId);
}

if (imageIdToURI(referencedImageId) === imageURI) {
    return true;
}
```

if the first `if (!imageURI)` was true then the second would always be true, so simplifying it to
```
if (!imageURI) {
    imageURI = imageIdToURI(referencedImageId);
    return true;
} else if (imageIdToURI(referencedImageId) === imageURI) {
    return true;
}
```

saves a call to `imageIdToURI(referencedImageId)` and an extra equivalency check

sedghi: can you do this instead 

```js
const { imageURI } = options;
  
  // Direct URI match check
  const currentImageURI = imageIdToURI(referencedImageId);
  if (!imageURI || currentImageURI === imageURI) {
    return true;
  }
```

plt-joey: updated to use that instead

sedghi (approved):
Thanks for your PR and patience

plt-joey:
@sedghi Is there anything I need to do to fix this pipeline and get this merged in? This bug is blocking us.

sedghi:
oh sorry

PR #1465, opened 2024-09-12
feat: stack scroll out of bounds event

m00n620:
Triggered `STACK_SCROLL_OUT_OF_BOUNDS` event in the `scroll` function of the stack viewport. Updated `stackManipulationTools` example to demonstrate the events trigger. 

https://github.com/OHIF/OHIF-Office-Hours/blob/main/notes/2024-08-08.md - followed the guides in the answer here.

### DIFF
--- a/packages/core/src/RenderingEngine/StackViewport.ts
+++ b/packages/core/src/RenderingEngine/StackViewport.ts
@@ -84,7 +84,6 @@ import {
   PixelDataTypedArray,
 } from '../types';
 import {
-  StackScrollOutOfBoundsEventDetail,
   StackViewportNewStackEventDetail,
   StackViewportScrollEventDetail,
   VoiModifiedEventDetail,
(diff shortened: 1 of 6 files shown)


packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -2601,6 +2602,19 @@ class StackViewport extends Viewport implements IStackViewport, IImagesLoader {
...
+    if (
+      imageIdIndex + delta > this.getImageIds().length - 1 ||
+      imageIdIndex + delta < 0
+    ) {
+      const eventData: StackScrollOutOfBoundsEventDetail = {
+        imageIdIndex,
+        direction: delta,
+      };
+      triggerEvent(this.element, Events.STACK_SCROLL_OUT_OF_BOUNDS, eventData);
+    }
```
sedghi, line 2616: I'm not sure if this makes sense, it is talking about a tool event inside the viewport, i think this should not happen here, but rather in the scroll() utilitiy

sedghi (changes requested):

sedghi (approved):

PR #1471, opened 2024-09-18
fix: volume progressive not working in beta

wayfarer3130:
#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  


wayfarer3130:
@sedghi  - I don't have this quite working.  I think the issue is that somewhere the wrong id gets used for deleting a frame.  However, it is qutie close.  If you just undo the removal of hte interleaved stages it will work for the basic retrieve, just not byte ranges.

packages/core/src/cache/cache.ts
```diff
@@ -443,7 +445,10 @@ class Cache {
       );
     }
 
+    // Starts with an existing cached image and extend it with information
+    // about being loaded.
     const cachedImage: ICachedImage = {
+      ...alreadyCached,
```
sedghi, line 451: How do we get here, since we are throwing error above if alreadyCached

packages/core/src/cache/cache.ts
```diff
@@ -1049,9 +1053,13 @@ class Cache {
    * Returns the image associated with the imageId
    *
    * @param imageId - image ID
+   * @param minQuality - the minimum image quality to fetch
    * @returns Image
    */
-  public getImage = (imageId: string): IImage | undefined => {
+  public getImage = (
+    imageId: string,
+    minQuality = ImageQualityStatus.FAR_REPLICATE
```
sedghi, line 1061: let's make it options object instead of positional argument

packages/core/src/cache/cache.ts
```diff
@@ -1061,13 +1069,60 @@ class Cache {
...
+      cachedImage.imageLoadObject = null;
+      this.incrementImageCacheSize(-cachedImage.sizeInBytes);
+      cachedImage.sizeInBytes = 0;
+      cachedImage.image = partialImage || cachedImage.image;
+    } else {
+      cachedImage.image = partialImage || cachedImage.image;
+    }
+  }
+
+  /** Gets the current image quality for the given image id */
```
sedghi, line 1118: can we get better docs for params and return thanks

sedghi (commented):
Looks good see my minor comments thanks

sedghi (approved):

PR #1486, opened 2024-10-01
fix(workflow): use a fixed ubuntu version for the runner

IbrahimCSAE:
### Context

Updating the workflow to use a stable version of ubuntu.


sedghi (approved):
Thanks a lot

PR #1495, opened 2024-10-10
final/stuff beta

sedghi:
- **Core: Update RequestType enum and geometry loader**
- **Segmentation Tools: Update getSegmentationActor to return ActorEntry instead of VolumeActor or ImageActor**
- **Segmentation: Update actor retrieval and filtering logic**
- **Surface: Improve Surface Validation and Rendering**
- **Volume Loader: Refactor volume creation and caching logic**
- **CornerstoneJS: Implement Jump to World and Slice Functionality**
- **Core: Improve surface rendering and remove segmentation representations**

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  


packages/core/src/cache/classes/Contour.ts
```diff
@@ -9,64 +9,137 @@ interface ContourProps {
...
+
+  /** Size of the contour data in bytes */
   readonly sizeInBytes: number;
-  points: Point3[];
-  color: Point3;
-  type: ContourType;
-  segmentIndex: number;
 
+  /** Array of 3D points defining the contour */
+  private _points: Point3[];
```
wayfarer3130, line 26: Should you use a PointsManager - the efficiency of the contour access is better because you can just get references to particular points, and the size is just the native data type times the number of values total?  Not sure this is worth doing right now, but I thought I'd mention it.

packages/core/src/cache/classes/Surface.ts
```diff
@@ -1,61 +1,146 @@
...
  */
 export class Surface {
   readonly id: string;
   readonly sizeInBytes: number;
   readonly frameOfReferenceUID: string;
-  private color: RGB = [200, 0, 0]; // default color
-  private points: number[];
-  private polys: number[];
+  private _color: RGB = [200, 0, 0]; // default color
+  private _points: number[];
```
wayfarer3130, line 14: The PointsManager manages the total points array/re-allocates as necessary and allows direct access as a Point3 reference value.

packages/core/src/enums/RequestType.ts
```diff
@@ -3,13 +3,13 @@
  */
 enum RequestType {
   /** Highest priority for loading*/
-  INTERACTION = 'INTERACTION',
+  Interaction = 'interaction',
```
wayfarer3130, line 6: Why are you changing the case here but not for Events?

packages/core/src/loaders/geometryLoader.ts
```diff
@@ -6,31 +6,156 @@ import type {
...
+    return geometryLoadObject.promise;
+  }
+
+  geometryLoadObject = loadGeometryFromGeometryLoader(geometryId, options);
+
+  return geometryLoadObject.promise;
+}
+
+/**
+ * Loads a geometry given a geometryId and optional options and returns a promise which will resolve to
```
wayfarer3130, line 110: This function and the loadGeometric from the outside seem to do the same sort of thing - can you add notes as to why this one is called createAndCache?  Maybe it should be loadAndCache?

packages/core/src/loaders/geometryLoader.ts
```diff
@@ -6,31 +6,156 @@ import type {
...
+
+/**
+ * Creates and caches a geometry synchronously
+ *
+ * @param geometryId - A Cornerstone Geometry Object's geometryId
+ * @param options - Options to be passed to the Geometry Loader
+ *
+ * @returns IGeometry
+ */
+export function createAndCacheLocalGeometry(
```
wayfarer3130, line 149: This one makes sense to call it createAndCacheGeometry, if the previous one is loadAndCacheGeometry then it makes more sense.

packages/core/src/loaders/geometryLoader.ts
```diff
@@ -6,31 +6,156 @@ import type {
...
+ * The geometry loader that is used will be
+ * determined by the geometry loader scheme matching against the geometryId.
+ *
+ * @param geometryId - A Cornerstone Geometry Object's geometryId
+ * @param options - Options to be passed to the Geometry Loader.
+ *
+ * @returns An Object which can be used to act after a geometry is loaded or loading fails
+ *
  */
+function loadGeometryFromGeometryLoader(
```
wayfarer3130, line 40: Consider encapsulating all the functions into a static class, and then export the class - it ties things together better, and allows you to encapsulate things like the geometryLoaders values.  Maybe not worth doing right now, given the timing.

packages/core/src/types/IVolumeInput.ts
```diff
@@ -31,6 +31,8 @@ interface IVolumeInput {
   blendMode?: BlendModes;
   /** Slab thickness of the volume - by default it is 0.05*/
   slabThickness?: number;
+  /** other metadata that is needed for the volume */
```
wayfarer3130, line 34: Please change the IVolumeInput to a type - that way you don't need to declare the key function, you can simply create an extended type based on it, and use that.

packages/core/src/utilities/scroll.ts
```diff
@@ -45,21 +49,17 @@ export default function scroll(
 
     if (
       imageIdIndex + delta >
-        (viewport as Types.IStackViewport).getImageIds().length - 1 ||
+        (viewport as IStackViewport).getImageIds().length - 1 ||
```
wayfarer3130, line 52: This is the line here that required adding getImageIds to WSIViewport.

packages/tools/examples/PolySegWasmSurfaceToContour/index.ts
```diff
@@ -189,23 +185,7 @@ async function run() {
...
-      });
-
-      const segmentIndex = index + 1;
-      acc.set(segmentIndex, geometryId);
-
-      return acc;
-    },
-    new Map()
-  );
+  const geometryIds = await createAndCacheGeometriesFromSurfaces();
```
wayfarer3130, line 188: Much cleaner like this.

packages/core/src/utilities/clip.js
wayfarer3130, line 1: Could you rename to clip.ts since you are renaming it anyways?

wayfarer3130 (approved):
There are a few comments, but generally this all looks like worthwhile cleanup/preparation for releasing CS3D 2.0

packages/core/src/cache/classes/Contour.ts
sedghi: Great mention, i added a todo

packages/core/src/loaders/geometryLoader.ts
sedghi: We don't have a loader yet for geometry. In the future, we can write loaders for STL and other relevant data types. For now, this is points and polys, so I think create and cache is relevant. I updated the doc.

sedghi: Ok done

PR #1603, opened 2024-11-18
Fix: bug causing slice images to intermingle across NIfTI volumes during simultaneous loading

stada526:
### Context

In a case where multiple NIfTI files are being fetched simultaneously, I encountered an issue where slice images from different volumes became intermingled when the volumes were loaded. Upon investigation, it was determined that the cause lies in the global access of `niftiScalarData` within `cornerstoneNiftiImageLoader.ts`.

### Changes & Results

I replaced `niftiScalarData` with `niftiScalarDataMap`. `niftiScalarDataMap` is a HashMap where the key is the URL and the value is a newly defined `NiftiDataFetchState`. `NiftiDataFetchState` consists of the fetch status and the fetched scalar data. This change prevents conflicts in scalar data updates across different URLs.

Additionally, I believe the scalar data stored in `niftiScalarDataMap` should be cleared after all processes for a specific URL are completed. One possible approach to achieve this is to store the number of processed slices for each URL and, once this number matches the total number of slices, execute `map.delete(url)`.
However, I am unsure how to obtain the total number of slices. I reviewed `imagePixelModule` and `imagePlaneModule`, but they did not seem to contain this information.

### Testing

In my environment where multiple NIfTI files are being fetched simultaneously, this change worked, but I'm sorry I'm not sure how to verify it using the assets in the repository.
At least, you can confirm that this change doesn't affect the examples named "niftiBasic" and "niftiWithTools".

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] "OS: maxOS 15.1"
- [x] "Node version: Node version: v21.7.1"
- [x] "Browser: Chrome 130.0.6723.117"


sedghi (approved):
I like it thanks!

Issue #1604, opened 2024-11-18
[Bug] `colorLUIOrIndex` in `RepresentationPublicInput` doesn't appear to work properly.


stada526:
### Describe the Bug

The `colorLUIOrIndex` property in `RepresentationPublicInput` doesn't appear to work properly.

### Steps to Reproduce

In `labelmapSegmentColorChange/index.ts`, specify the `colorLUTOrIndex` as shown below.
 (By the way, this file hasn't been fully updated to align with version2.)

```ts
  await segmentation.addSegmentationRepresentations(viewportId1, [
    {
      segmentationId: segmentationId1,
      type: csToolsEnums.SegmentationRepresentations.Labelmap,
      config: {
        colorLUTOrIndex: [
          [
            [0, 0, 0, 0],
            [250, 152, 180, 255],
            [125, 250, 20, 255],
          ],
        ],
      },
    },
  ]);
```

### The current behavior

No matter which color look up table is provided to `addSegmentationRepresentations`, the initial colors of the segments don't change.

### The expected behavior

The initial colors of the segments should change according to the `colorLUIOrIndex` configuration.

### OS

maxOS 15.1

### Node version

v21.7.1

### Browser

Chrome 130.0.6723.117

stada526:
@sedghi 
Thank you for addressing this issue! I appreciate it 🙏
However, there still seems to be an issue with `colorLUIOrIndex` property in `RepresentationPublicInput`.

In `labelmapSegmentColorChange/index.ts`, specify the colorLUTOrIndex as shown below.
```ts
  await segmentation.addSegmentationRepresentations(viewportId1, [
    {
      segmentationId: segmentationId1,
      type: csToolsEnums.SegmentationRepresentations.Labelmap,
      config: {
        colorLUTOrIndex: [
          [0, 0, 0, 0],
          [0, 0, 255, 255],
          [255, 255, 0, 255],
        ],
      },
    },
  ]);

  await segmentation.addSegmentationRepresentations(viewportId2, [
    {
      segmentationId: segmentationId1,
      type: csToolsEnums.SegmentationRepresentations.Labelmap,
      config: {
        colorLUTOrIndex: [
          [0, 0, 0, 0],
          [0, 255, 0, 255],
          [255, 0, 255, 255],
        ],
      },
    },
  ]);
```

In this case, the first color lookup table appears to work properly, but the second one doesn't.
In version 1 of Cornerstone.js, the both work properly.
[...]

PR #1611, opened 2024-11-22
fix(colorlut): use the correct colorlut index if there's multiple segmentations

IbrahimCSAE:
### Context

The segmentation rendering was defaulting to index 0 all the time for the color lut

CleanShot 2024-11-22 at 11 44 40

also fixes : https://github.com/cornerstonejs/cornerstone3D/issues/1604


sedghi (approved):
Thanks

PR #1669, opened 2024-12-04
Fix logic that determines if nifti file is compressed

faridyagubbayli:
### Context

The logic which decides if remote nifti file is compressed or not has a flaw. It checks the end of the URL string for `.gz` extension. When url includes parameters/headers, `.gz` extension appears in the middle of the string. 

For example, current logic will work for a URL like: `https://<some_address>/<filename>.nii.gz` but fail for a url like `https://<some_address>/<filename>.nii.gz?<some_param>=<some_value>`. 

### Changes & Results

This PR updates the logic to check for inclusion of the `.gz` extension anywhere in the string.

### Testing

Tested in a limited local environment and works as expected.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  

### DIFF
--- a/packages/nifti-volume-loader/src/createNiftiImageIdsAndCacheMetadata.ts
+++ b/packages/nifti-volume-loader/src/createNiftiImageIdsAndCacheMetadata.ts
@@ -25,7 +25,7 @@ export async function fetchArrayBuffer({
   onHeader,
   loadFullVolume = false,
 }) {
-  const isCompressed = url.endsWith('.gz');
+  const isCompressed = url.includes('.gz');
   let receivedData = new Uint8Array(0);
   let niftiHeader = null;
   const sliceInfo = null;
--- a/packages/nifti-volume-loader/src/createNiftiImageIdsAndCacheMetadata.ts
+++ b/packages/nifti-volume-loader/src/createNiftiImageIdsAndCacheMetadata.ts
@@ -25,7 +25,8 @@ export async function fetchArrayBuffer({
   onHeader,
   loadFullVolume = false,
 }) {
-  const isCompressed = url.includes('.gz');
+  const _url = new URL(url);
+  const isCompressed = _url.pathname.endsWith('.gz');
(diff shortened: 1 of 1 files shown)


packages/nifti-volume-loader/src/createNiftiImageIdsAndCacheMetadata.ts
```diff
@@ -25,7 +25,7 @@ export async function fetchArrayBuffer({
   onHeader,
   loadFullVolume = false,
 }) {
-  const isCompressed = url.endsWith('.gz');
+  const isCompressed = url.includes('.gz');
```
sedghi, line 28: but then how are you cleaning the url for the fetch?

faridyagubbayli: I don't need to clean the url. In my case, I use a signed URL and additional parameters are needed for authentication.

daker: Maybe use `URL` api instead

```js
const url = "https://toto.blob.core.windows.net/test5/myfoo1.gz?sv=2019-10-10&ss=bfqt&srt=sco&sp=rwdlacupx&se=2020-07-17T10:09:48Z&st=2020-07-17T02:09:48Z&spr=https&sig=xxxxxxxxx";
const _url = new URL(url);
const isCompressed = _url.pathname.endsWith('.gz');
console.log(isCompressed);


const url = "https://myserver.com/sample.gz";
const _url = new URL(url);
const isCompressed = _url.pathname.endsWith('.gz');
console.log(isCompressed);
```

faridyagubbayli: @daker good suggestion thanks! Updated the code.

PR #1670, opened 2024-12-04
feat(webgpu): growcut base

lscoder:
### Context

Add GrowCut implementation running in the gpu (webgpu, compute shader)

--- 

Inspired by https://github.com/OHIF/ohif-step/blob/master/src/GrowCutGenerator.js


sedghi (approved):
This is really great work, thanks

PR #1710, opened 2024-12-14
fix: passing metadata to getTextLines callback in annotation tools

salimkanoun:
### Context

In nuclear medicine, some quantifications may have different normalization such as SUV lean body mass or glycaemia corrected SUV.

To implement specific quantification we previously add callback definition to override statistics computation or result rendering (stat calculator and getTextLines parameters)

Our current problem is that we want to change the rendered value SUV => SUL, to do this we need to get PatientWeight, PatientSize and PatientSex in the Metadata store.

The probleme we are facing is that the getTextLines callaback is not giving enough context about the measurement especially to retrieve the volumeId/displaySet/SeriesInstanceUID that are needed to link the measurement rendering from the Metadata Store.

So in this PR i add the metadata in the getTextLines of Circle and Rectangle start stop, it should solve our problem but now the issues is that getTextLines has different definition from a tool to another,  for instance : 

In 2D annotation we have : defaultGetTextLines(data, targetId)
in UltrasoundDirectional we have : defaultGetTextLines(data, targetId, configuration)
And now for 3D annotation we will have : defaultGetTextLines(data, _metadata)

I'm a bit afraid seeing different callback definition from a tool to another, that can be complicated to document and maintain in long term, i wonder if we shouldn't send the whole annotation object as one single param so the getTextLines call back can pickup watever needed, but this will become a breaking change 

### Changes & Results

Add more context to getTextLines params

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  

### DIFF
--- a/packages/tools/src/tools/segmentation/CircleROIStartEndThresholdTool.ts
+++ b/packages/tools/src/tools/segmentation/CircleROIStartEndThresholdTool.ts
@@ -881,9 +881,9 @@ class CircleROIStartEndThresholdTool extends CircleROITool {
  * target volume enclosed by the rectangle.
  *
  * @param data - The annotation tool-specific data.
- * @param targetId - The volumeId of the volume to display the stats for.
+ * @param metadata - The metadata annotation.
  */
-function defaultGetTextLines(data): string[] {
+function defaultGetTextLines(data, _metadata): string[] {
   const cachedVolumeStats = data.cachedStats.statistics;
 
   const { area, mean, max, stdDev, areaUnit, modalityUnit } = cachedVolumeStats;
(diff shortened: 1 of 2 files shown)


packages/tools/src/tools/segmentation/CircleROIStartEndThresholdTool.ts
```diff
@@ -881,9 +881,9 @@ class CircleROIStartEndThresholdTool extends CircleROITool {
  * target volume enclosed by the rectangle.
  *
  * @param data - The annotation tool-specific data.
- * @param targetId - The volumeId of the volume to display the stats for.
+ * @param metadata - The metadata annotation.
  */
-function defaultGetTextLines(data): string[] {
```
sedghi, line 886: can we have it as defaultGetTextLines(data, {} ), i'm pretty sure in future something else need to get injected too

salimkanoun:
here it is, yes this object is more extensible in the future, 

What to do with 2D annotation which have : 
defaultGetTextLines(data, targetId)
And ultrasounds direction tools has 
defaultGetTextLines(data, targetId, configuration):

Do we change something for 2D roi ? (but can be a breaking change)

sedghi (approved):

PR #1717, opened 2024-12-18
feat(voi): add linear exact voi lut function

sedghi:
related https://github.com/OHIF/Viewers/issues/4298
related https://github.com/OHIF/Viewers/issues/4538

This pull request includes an important update to the `toLowHighRange` function in the `packages/core/src/utilities/windowLevel.ts` file. The change adds a conditional check to handle cases where `windowWidth` or `windowCenter` are less than 1, providing a fallback calculation for the `lower` and `upper` values.

Key change:

* [`packages/core/src/utilities/windowLevel.ts`](diffhunk://#diff-6f4cb97bb6e19327efc7d92bc58b84bc28cef3d473fe4547e83ca6dd46bdcb04L44-R55): Added a conditional check in the `toLowHighRange` function to handle cases where `windowWidth` or `windowCenter` are less than 1, and provided a fallback calculation for `lower` and `upper` values.

### DIFF
--- a/packages/core/src/RenderingEngine/helpers/setDefaultVolumeVOI.ts
+++ b/packages/core/src/RenderingEngine/helpers/setDefaultVolumeVOI.ts
@@ -85,8 +85,8 @@ function getVOIFromMetadata(imageVolume: IImageVolume): VOIRange | undefined {
     const imageIdIndex = Math.floor(imageIds.length / 2);
     const imageId = imageIds[imageIdIndex];
     const voiLutModule = metaData.get('voiLutModule', imageId);
-    voi.voiLUTFunction = voiLutModule.voiLUTFunction;
-    if (voiLutModule?.windowWidth && voiLutModule.windowCenter) {
+    if (voiLutModule && voiLutModule.windowWidth && voiLutModule.windowCenter) {
+      voi.voiLUTFunction = voiLutModule.voiLUTFunction;
       const { windowWidth, windowCenter } = voiLutModule;
       const width = Array.isArray(windowWidth) ? windowWidth[0] : windowWidth;
       const center = Array.isArray(windowCenter)
--- a/packages/core/src/RenderingEngine/helpers/setDefaultVolumeVOI.ts
+++ b/packages/core/src/RenderingEngine/helpers/setDefaultVolumeVOI.ts
@@ -81,10 +81,11 @@ function handlePreScaledVolume(imageVolume: IImageVolume, voi: VOIRange) {
 function getVOIFromMetadata(imageVolume: IImageVolume): VOIRange | undefined {
   const { imageIds, metadata } = imageVolume;
   let voi;
-  if (imageIds.length) {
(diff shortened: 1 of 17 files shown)


packages/core/src/utilities/windowLevel.ts
```diff
@@ -41,10 +41,18 @@ function toLowHighRange(
   lower: number;
   upper: number;
 } {
-  const lower = windowCenter - 0.5 - (windowWidth - 1) / 2;
-  const upper = windowCenter - 0.5 + (windowWidth - 1) / 2;
+  if (windowWidth < 1 || windowCenter < 1) {
+    // use fallback
```
wayfarer3130, line 45: Please add a console.warn('Non DICOM standard compliant width/center", windowWidth, windowCenter)

packages/core/src/utilities/windowLevel.ts
```diff
@@ -41,10 +41,18 @@ function toLowHighRange(
   lower: number;
   upper: number;
 } {
-  const lower = windowCenter - 0.5 - (windowWidth - 1) / 2;
-  const upper = windowCenter - 0.5 + (windowWidth - 1) / 2;
+  if (windowWidth < 1 || windowCenter < 1) {
```
wayfarer3130, line 44: Only check windowWidth, the windowCenter should be allowed to be bigger.

packages/core/src/utilities/windowLevel.ts
```diff
@@ -41,10 +41,18 @@ function toLowHighRange(
...
+    // use fallback
+    return {
+      lower: windowCenter - windowWidth / 2,
+      upper: windowCenter + windowWidth / 2,
+    };
+  }
 
-  return { lower, upper };
+  return {
```
wayfarer3130, line 52: Please add a comment:
// This definition is from the DICOM standard PS3.3

wayfarer3130 (approved):

packages/core/src/RenderingEngine/helpers/setDefaultVolumeVOI.ts
```diff
@@ -81,11 +81,12 @@ function handlePreScaledVolume(imageVolume: IImageVolume, voi: VOIRange) {
...
   const { imageIds, metadata } = imageVolume;
   let voi;
-  if (imageIds.length) {
+  if (imageIds?.length) {
     const imageIdIndex = Math.floor(imageIds.length / 2);
     const imageId = imageIds[imageIdIndex];
     const voiLutModule = metaData.get('voiLutModule', imageId);
-    if (voiLutModule?.windowWidth && voiLutModule.windowCenter) {
+    if (voiLutModule && voiLutModule.windowWidth && voiLutModule.windowCenter) {
+      voi.voiLUTFunction = voiLutModule.voiLUTFunction;
```
IbrahimCSAE, line 89: this is not always defined, it's crashing tests in OHIF

PR #1732, opened 2024-12-26
fix(ArrowAnnotate): use svg marker to draw the arrow

### LINKED ISSUE #1671: [Bug] Arrow annotate tool ui's become broken when increase the line width
### Describe the Bug

When I set annotations line width to 10 than I'm getting following result. 

image

### Steps to Reproduce

1. Set ArrowAnnotationTool lineWidth style to 10.
2. Draw an arrow.

### The current behavior

When I draw arrow the arrow line overflows outside of arrow.

### The expected behavior

The line should not overflow.

### OS

Windows 10

### Node version

20

### Browser

Chrome 131.0.6778.108

daker:
### Context
fix #1671

### Changes & Results

Using `marker` html def instead of two separate arrows for better control over arrow

| size 1 |  size 2 |  size 3 |  size 4 |
|---|---|---|---|
|||||

By default, ArrowAnnotateTool uses the previous version, but you can opt for the new one by using the following code: 
```js
toolGroup.addTool(ArrowAnnotateTool.toolName, {
  arrowHeadStyle: 'standard',
});
```
To change the marker head size, you can use the ToolStyle API. For example, you can do something like this:
```js
const newStyles = {
  global: {
    ...cornerstoneTools.annotation.config.style.getDefaultToolStyles().global,
    markerSize: 20,
  },
};

cornerstoneTools.annotation.config.style.setDefaultToolStyles(newStyles);
```

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  

### DIFF
--- a/packages/tools/src/drawingSvg/drawLine.ts
+++ b/packages/tools/src/drawingSvg/drawLine.ts
@@ -11,15 +11,7 @@ export default function drawLine(
   lineUID: string,
   start: Types.Point2,
   end: Types.Point2,
-  options: {
-    color?: string;
-    width?: number;
-    lineWidth?: number;
-    lineDash?: string;
-    markerStartId?: string;
-    markerEndId?: string;
-    shadow?: boolean;
-  },
+  options = {},
   dataId = ''
 ): void {
   // if length is NaN return
@@ -35,7 +27,15 @@ export default function drawLine(
     markerStartId = null,
     markerEndId = null,
     shadow = false,
-  } = options;
+  } = options as {
+    color?: string;
+    width?: string;
+    lineWidth?: string;
+    lineDash?: string;
+    markerStartId?: string;
+    markerEndId?: string;
+    shadow?: boolean;
+  };
 
   // for supporting both lineWidth and width options
   const strokeWidth = lineWidth || width;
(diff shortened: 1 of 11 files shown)


packages/tools/src/drawingSvg/drawLine.ts
```diff
@@ -11,34 +11,50 @@ export default function drawLine(
...
-      width: '2',
-      lineWidth: undefined,
-      lineDash: undefined,
-      shadow: undefined,
-    },
-    options
-  );
+  const {
+    color = 'rgb(0, 255, 0)',
+    width = 10,
```
sedghi, line 32: Why are we changing the `width` here? Could this break people's tools since we're changing it from 2 to 10?

packages/tools/src/drawingSvg/drawLine.ts
```diff
@@ -11,34 +11,50 @@ export default function drawLine(
...
-    : '';
+  const layerId = svgDrawingHelper.svgLayerElement.id;
+  const dropShadowStyle = shadow ? `filter:url(#shadow-${layerId});` : '';
+  const arrow = svgDrawingHelper.svgLayerElement.querySelector(
+    `#${markerEndId}-${layerId}`
+  );
+
+  if (arrow) {
+    arrow.setAttribute('fill', color);
+    const newstrokeWidth = strokeWidth >= 4 ? 3 : 6;
```
sedghi, line 54: Why is `strokeWidth >= 4 ? 3 : 6;` used here? It seems arbitrary.

daker: if the stroke is more than 4 the arrow becomes bigger hence the reduction to 3 otherwise it's 6 but i think i need to find a better way for this

sedghi: I guess put it in the configuration

sedghi:
@daker, thanks for starting this fix. 
I made a couple of changes. 

1) One marker definition isn't enough since we have different colors based on the state (like selected or locked), and that was causing some issues when the color wasn't changing. So, I refactored it to have multiple marker definitions.
2) I'm making this a configuration option on the tool, because I don't want people's tests to fail. For now, let's keep it under a configuration, and in Cornerstone 3.x, we'll default to it.

Thanks again. Take a look at the changes and let me know if you're okay with them, then I'll merge.

daker:
@sedghi all good (y)

sedghi (approved):

PR #1743, opened 2025-01-09
feat: Add range support for key image tool

wayfarer3130:
### Context

Key images can be used to identify several different things within a series.  They might identify specific points as a Fiducial, or they might be a series label or a range, or a selection of several images within  the same series.  This change adds the ability to select various parts of a series within an annotation.

Also, the performance of the isReferenceViewable was very poor and the results were inconsistent.  This PR addresses the performance by adding a map of image id to location and caching additional test locations where needed.

The API is nearly the same - there are a couple of small internal changes in the view references, but unless there are previously stored view references with ranges, those should not affect this PR.

The range references are not supported yet for volume viewports, but the existing view references are supported.

### Testing

Run the stackRange and videoRange examples
Note the consistsency between range value settings between both viewport types.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  


utils/ExampleRunner/example-info.json
```diff
@@ -175,6 +175,10 @@
         "name": "Stack Annotation Tools",
         "description": "Demonstrates usage of various annotation tools (Probe, Rectangle ROI, Elliptical ROI, Bidirectional measurements) on a Stack Viewport."
       },
+      "stackRange": {
```
sedghi, line 178: Can we make it a habit to append to the end for easier access?

packages/tools/src/enums/ChangeTypes.ts
```diff
@@ -35,6 +35,10 @@ enum ChangeTypes {
    * Occurs when an annotation is changed do to an undo or redo.
    */
   History = 'History',
+  /**
+   * Send when the metadata selector changes
```
sedghi, line 39: is this about a changeType for annotation? can we get more docs please, Occurs when annotation's metadata changs? what is a metadata selector?

packages/core/src/types/IViewport.ts
```diff
@@ -18,7 +18,15 @@ export interface ViewReferenceSpecifier {
    * and cannot be shared across different view types such as stacks and
    * volumes, or two viewports showing different orientations or slab thicknesses.
    */
-  sliceIndex?: number | [number, number];
+  sliceIndex?: number;
+  /**
+   * The end index - this requires sliceIndex to be specified.
+   */
+  sliceRangeEnd?: number;
```
sedghi, line 25: does it makes sense to have

```
sliceIndex?: number -> for single index
sliceRange?: [number, number] -> for range
```

packages/core/src/types/IViewport.ts
```diff
@@ -122,20 +138,26 @@ export interface ViewReference {
...
    * ends or can be co-incident but separate types of images.
    */
-  sliceIndex?: number | [number, number];
+  sliceIndex?: number;
+  /**
+   * An end of a slice range.  This is used to indicate the end of the slices
+   * where the sliceIndex is the first point.  This is a positive value if defined,
+   * so can be tested for `sliceRangeEnd`.
+   */
+  sliceRangeEnd?: number;
```
sedghi, line 160: same

packages/core/src/types/IViewport.ts
```diff
@@ -97,9 +105,17 @@ export interface ViewReference {
...
+   *
+   * For range/selection, this must be the starting range referenced image id
    */
   referencedImageId?: string;
 
+  /**
+   * An internal URI version of hte referencedImageId, used for performance
+   * while checking the referenced image id.
+   */
+  referencedImageUri?: string;
```
sedghi, line 117: we have `FrameOfReferenceUID` above, so is this `referencedImageURI` ? 

I don't see anywhere else in the code we have `referencedImageUri` but i see 16 occurences of `referencedImageURI`

packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -1826,6 +1828,11 @@ class StackViewport extends Viewport {
     this._throwIfDestroyed();
 
     this.imageIds = imageIds;
+    this.imageIdsMap.clear();
+    imageIds.forEach((imageId, index) => {
+      this.imageIdsMap.set(imageId, index);
+      this.imageIdsMap.set(imageIdToURI(imageId), index);
```
sedghi, line 1834: so this seems like it is not just imageIds, but also imageURIs too, so I guess some other variable naming, `imageKeyToIndexMap` ?

packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -3145,35 +3153,27 @@ class StackViewport extends Viewport {
    * Assumes that the slice index is correct for this viewport
    */
   public setViewReference(viewRef: ViewReference): void {
-    if (!viewRef) {
+    if (!viewRef?.referencedImageId) {
```
sedghi, line 3156: so seems like this is not (and was not) working with just `sliceIndex` can we handle that case too so that we are able to say `setViewReference({sliceIndex})`

packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -3035,48 +3042,49 @@ class StackViewport extends Viewport {
...
-
-    if (options.withNavigation && typeof sliceIndex === 'number') {
-      testIndex = sliceIndex;
-      currentImageId = this.imageIds[testIndex];
-    }
-
-    if (!currentImageId) {
+    const testIndex = this.getCurrentImageIdIndex();
+    const currentImageId = this.imageIds[testIndex];
+    if (!currentImageId || !viewRef) {
```
sedghi, line 3047: How there is a case where there is no `currentImageId`? I don't see how `this.imageIds[this.getCurrentImageIdIndex]` become undefined

packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -3035,48 +3042,49 @@ class StackViewport extends Viewport {
...
       return false;
     }
+    const {
```
sedghi, line 3050: We should start writing tests for the `isReferenceViewable` feature. Consider adding some for this PR; if not, we should do it in the future.

packages/core/src/utilities/frameRangeUtils.ts
```diff
@@ -0,0 +1,56 @@
+import { triggerEvent, eventTarget } from '@cornerstonejs/core';
+
+export type FramesRange = [number, number] | number;
+
+/**
+ * This class handles the annotation frame range values for multiframes.
+ * Mostly used for the Video viewport, it allows references to
+ * a range of frame values.
+ */
+export default class FrameRangeUtils {
```
sedghi, line 10: can we name the file at least to FrameRangeUtils so that it is clear it is a class?

packages/core/src/utilities/frameRangeUtils.ts
```diff
@@ -0,0 +1,56 @@
...
+  protected static frameRangeExtractor =
+    /(\/frames\/|[&?]frameNumber=)([^/&?]*)/i;
+
+  protected static imageIdToFrames(imageId: string): FramesRange {
```
sedghi, line 14: please add docs for each method

packages/tools/src/utilities/annotationFrameRange.ts
```diff
@@ -47,32 +48,88 @@ export default class AnnotationFrameRange {
     );
   }
 
+  public static setStartRange(
```
sedghi, line 51: There seems to be duplicated logic in this class, since lots is implemented in frameRangeUtils now

packages/core/src/utilities/frameRangeUtils.ts
```diff
@@ -0,0 +1,56 @@
...
+      return null;
+    }
+    const range = match[2].split('-').map((it) => Number(it));
+    if (range.length === 1) {
+      return range[0];
+    }
+    return range as FramesRange;
+  }
+
+  public static multiframeImageId(imageId: string, frameNumber = 1) {
```
sedghi, line 26: multiframeImaegIdToFrames? do you need this extra? can't you make the `imageIdToFrames` handle multiframe case too?

packages/tools/src/utilities/annotationFrameRange.ts
```diff
@@ -47,32 +48,88 @@ export default class AnnotationFrameRange {
...
+    viewport,
+    annotation,
+    startRange = viewport.getCurrentImageIdIndex()
+  ) {
+    this.setRange(viewport, annotation, startRange);
+  }
```
sedghi, line 57: I don't like how we are making range and frames to be two separate construct in these two utils. Can we just pick one? i prefere range (like your new method namings), we can add fallbacks just for not having breaking change and call them deprecated

```
// Assuming the class gets re-named to 'AnnotationRange' and there we also have `RangeUtils`

AnnotationRange.imageIdToRange(imageId: string);
AnnotationRange.rangeToString(range: FrameRange);
AnnotationRange.rangeToImageId(imageId: string, range: FrameRange); 
AnnotationRange.setRange(viewport, annotation, newRange);
AnnotationRange.setStart(viewport, annotation, startRange);
AnnotationRange.setEnd(viewport, annotation, endRange); 
AnnotationRange.getRange(annotationUID); 
AnnotationRange.setViewportRange(viewport, specifier);

```

I don't know what is .setSingle

Notice the getRange is via UID not the whole annotation, similar to other locking, selection etc APIs

packages/core/src/types/IViewport.ts
wayfarer3130: No, I'm trying to get away from sliceIndex at all since it really isn't deterministic between different display sets contain the same set of images.  I would have liked to get rid of it in this PR, but that was more of an API change than I was willing to contemplate here, whereas the simple change to have two different values instead of number | [number,number] is much more internally consistent with the current API, even though it is a small change.

packages/core/src/types/IViewport.ts
wayfarer3130: I'd rather have a separate numberr here - this one will be left as indices as it actually refers to specific slices in a consistent fashion, but even with that, the use of sliceIndex is consistent between single and ranges, so I think I'd like to leave it as a separate value.

packages/core/src/types/IViewport.ts
wayfarer3130: Updated the docs to explain it - there is no relationship to FrameOfReferenceUID

packages/core/src/RenderingEngine/StackViewport.ts
wayfarer3130: The viewport is uninitialized is primarily what I was concerned with, but if you are ok with it throwing, then we could ignore it.

packages/core/src/types/IViewport.ts
sedghi: we don't have `referencedImageUri` at all in the codebase and have 16 `referencedImageURI` already, that is my point

packages/core/src/RenderingEngine/StackViewport.ts
sedghi: it's fine leave it

packages/core/src/types/IViewport.ts
sedghi: As discussed we will move off `number` and do a referencedEndId or something

packages/tools/examples/videoRange/index.ts
```diff
@@ -191,36 +174,31 @@ addButtonToToolbar({
...
-        : [rangeSelection, viewport.getNumberOfSlices()];
-      range[1] = frame;
-      range[0] = Math.min(frame, range[0]);
-      annotationFrameRange.setFrameRange(
-        annotation,
-        range as [number, number],
-        baseEventDetail
-      );
-      viewport.setFrameRange(range);
+      AnnotationMultiSlice.setEndRange(viewport, annotation);
```
sedghi, line 177: Much cleaner thanks

packages/core/src/types/IViewport.ts
```diff
@@ -11,14 +11,24 @@ import type Viewport from '../RenderingEngine/Viewport';
...
    */
-  sliceIndex?: number | [number, number];
+  sliceIndex?: number;
+  /**
+   * The end index - this requires sliceIndex to be specified.
+   * This will result in a view reference containing a range, specified
+   * by the starting and ending images or by a thickness for a volume.
+   */
+  sliceRangeEnd?: number;
+
```
sedghi, line 28: How about `rangeEndSliceIndex` and below instead of `endRangeImageURI` we replace with `rangeEndImageURI` (I assume there should be a `rangeEndImageId` too?)

sedghi (commented):
just a minor renaming proposal

sedghi:
can you rename endRangeImageURI to `rangeEndImageURI` as well?

PR #1744, opened 2025-01-10
[OHI-1544] fix(annotations): make ArrowAnnotate, CobbAngle and Angle check for visibilty status before rendering

### LINKED ISSUE #1741: [Bug] ArrowAnnotationTool and AngleTool Annotations are always visible. Even after using setAnnotationVisibility to make them disappear.
### Describe the Bug

I am using different tool and i use setAnnotationVisibility to control if their are displayed or not.
For most tools, it work without problems.

But for "ArrowAnnotateTool" and "AngleTool", they are always visible.
Even after using setAnnotationVisibility.
With "annotation.visibility.isAnnotationVisible", i can see that the visibility has been changed correctly, but the display stay the same.

Thank you.

### Steps to Reproduce

- Create a project that use "AngleTool" or "ArrowAnnotateTool".
- Change the visiblity of an "AngleTool" or "ArrowAnnotateTool" annotation after creation with "setAnnotationVisibility "

### The current behavior

The annotation doesn't disapear after setting the visibility at false.

### The expected behavior

The annotation should disapear after setting the visibility at false.

### OS

Windows 11

### Node version

22.11.0

### Browser

Chrome

IbrahimCSAE:
### Context

A few tools were missing the implementation for visibility checking before rendering, so when they were set to hidden they were still being rendered anyway.

This has been fixed, and I added most of the measurement tools in the annotation visibility example for easier testing in the future.

Fixes OHIF-1544
Fixes #1741

### Demo

CleanShot 2025-01-10 at 00 31 28

### DIFF
--- a/packages/tools/examples/annotationVisibility/index.ts
+++ b/packages/tools/examples/annotationVisibility/index.ts
@@ -9,10 +9,20 @@ import {
   createImageIdsAndCacheMetaData,
   setTitleAndDescription,
   addButtonToToolbar,
+  addDropdownToToolbar,
 } from '../../../../utils/demo/helpers';
 import {
   LengthTool,
+  BidirectionalTool,
+  ArrowAnnotateTool,
+  EllipticalROITool,
   RectangleROITool,
+  CircleROITool,
+  PlanarFreehandROITool,
+  SplineROITool,
+  LivewireContourTool,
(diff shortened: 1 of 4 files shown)


sedghi (approved):
Thanks

Issue #1490, opened 2024-10-03
[Feature Request] Expose more utility functions


LadislavBohm:
### What feature or change would you like to see made?

I'm trying to implement my own annotation tool that is based on LengthTool. When looking at it's source code it uses some very useful utility functions that are not exposed publicly. For example these functions I'm not able to import and use in my own project:

1. triggerAnnotationCompleted
2. triggerAnnotationModified

Is it by design or is it ok to export those as well?

### Why should we prioritize this feature?

It helps people correctly implement annotation tools. It ensures same behavior as already existing tools.

sedghi:
You can create a PR and I will merge it

PR #1748, opened 2025-01-14
WIP proof of concept for simplifying tool imports

jmhmd:
### Context

When creating a new tool, it helps to start with one of the existing tools and modifying it. However, there are tons of imports which reference internal files in the cstools repo. Most, but not all, of these are exposed in the package exports. Even the ones that are exposed, are not always in the same path as what is imported by existing tools, so the process of resolving the imports is very tedious.

### Changes & Results

This PR proposes to change how existing tools import dependencies, such that nearly all imports come from the root index.ts file. This makes it so that to clone a tool, all the dev has to do is change the import path from the root index.ts file to `@cornerstonejs/tools`, and everything else should work. This process would also ensure that all tool dependencies are exposed in the package index.

I have modified three tools in this way, and if core devs approve, can modify more/the rest.

One point to discuss: I doubt whether this change would significantly change bundle size, but am not completely sure if the ability of bundlers to tree-shake would be affected. Open to other proposals.

Would address issue #1490

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  


jmhmd:
Closing this, taking a less aggressive approach.

PR #1752, opened 2025-01-14
fix(sortImageIdsAndGetSpacing): sliceThickness and spacingBetweenSlices in case of nonstrict spacing

sedghi:
Not sure why the logic was like this, it didn't makes sense
zSpacing also should not be negative

### DIFF
--- a/packages/core/src/utilities/sortImageIdsAndGetSpacing.ts
+++ b/packages/core/src/utilities/sortImageIdsAndGetSpacing.ts
@@ -157,14 +157,14 @@ export default function sortImageIdsAndGetSpacing(
   // If possible, we use the sliceThickness, but we warn about this dicom file
   // weirdness. If sliceThickness is not available, we set to 1 just to render
   if (zSpacing === 0 && !strictZSpacingForVolumeViewport) {
-    if (!sliceThickness && spacingBetweenSlices) {
+    if (spacingBetweenSlices) {
       console.log('Could not calculate zSpacing. Using spacingBetweenSlices');
-      zSpacing = Math.abs(spacingBetweenSlices);
+      zSpacing = spacingBetweenSlices;
     } else if (sliceThickness) {
       console.log(
         'Could not calculate zSpacing and no spacingBetweenSlices. Using sliceThickness'
       );
-      zSpacing = Math.abs(sliceThickness);
+      zSpacing = sliceThickness;
     } else {
       console.log(
         'Could not calculate zSpacing. The VolumeViewport visualization is compromised. Setting zSpacing to 1 to render'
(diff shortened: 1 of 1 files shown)


sedghi:
FYI @salimkanoun

salimkanoun:
It could be negative, see https://dicom.innolitics.com/ciods/ct-image/image-plane/00180088
"and a negative spacing indicates slices are stacked in front of the first slice. See Image Orientation (0020,0037) in the [NM Detector Module](http://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.8.4.11.html#sect_C.8.4.11)."

If you use absolute value you will reintroduced flipped NM images issue

packages/core/src/utilities/sortImageIdsAndGetSpacing.ts
```diff
@@ -157,14 +157,14 @@ export default function sortImageIdsAndGetSpacing(
   // If possible, we use the sliceThickness, but we warn about this dicom file
   // weirdness. If sliceThickness is not available, we set to 1 just to render
   if (zSpacing === 0 && !strictZSpacingForVolumeViewport) {
-    if (sliceThickness && spacingBetweenSlices) {
+    if (!sliceThickness && spacingBetweenSlices) {
```
wayfarer3130, line 160: Why do you care about slice thickness at all - just order it starting at the preferred item, then moving to the least preferred.  So maybe
if( spacingBetweenSclies ) {
...
else if ( sliceThickness)


That is, ignore the sliceThickness in the first test, OR put the slice thickness version first.

wayfarer3130 (commented):
See my comment inline.

pieper:
More discussion (much more) about issues with negative SpacingBetweenSlices that popped up in ITK and Slicer not so long ago:

https://github.com/Slicer/Slicer/pull/7987

https://github.com/InsightSoftwareConsortium/ITK/issues/4794

There are some nonstandard CT instances in the world (at least some in IDC that appear to have come from real modalities) where the spacing value is negative even though the standard says it should only happen for NM.

sedghi:
> It could be negative, see [dicom.innolitics.com/ciods/ct-image/image-plane/00180088](https://dicom.innolitics.com/ciods/ct-image/image-plane/00180088) "and a negative spacing indicates slices are stacked in front of the first slice. See Image Orientation (0020,0037) in the [NM Detector Module](http://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.8.4.11.html#sect_C.8.4.11)."
> 
> If you use absolute value you will reintroduce flipped NM images issue

It can be negative for spacing between slices, but our construct (zSpacing, vtk spacing) cannot be negative as far as i know

sedghi:
@salimkanoun Actually i reverted the Math.abs and still works so most likely vtk.js Match abs internally

packages/core/src/utilities/sortImageIdsAndGetSpacing.ts
sedghi: Done

salimkanoun:
Sounds good!

wayfarer3130 (approved):

PR #1774, opened 2025-01-23
Modify ROI tools and arrow annotate tool to make rendering the text-box (or label) configurable

jmhmd:
### Context
Multiple tools by default render a text box, i.e. ROIs and arrow annotate. There are use cases for being able to configure these tools to not render the text (many pacs default to a plain arrow with no label), also ROIs without stats are useful for marking abnormalities when the stats are not relevant.

### Changes & Results
This PR makes the text box rendering configurable, and adds examples in the stack annotation tools example. The examples extend the default tools to allow both configurations to coexist.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  

### DIFF
--- a/packages/tools/src/tools/annotation/ArrowAnnotateTool.ts
+++ b/packages/tools/src/tools/annotation/ArrowAnnotateTool.ts
@@ -67,6 +67,8 @@ class ArrowAnnotateTool extends AnnotationTool {
       supportedInteractionTypes: ['Mouse', 'Touch'],
       configuration: {
         shadow: true,
+        // Whether to ask for and render text box (label) on the annotation
+        renderTextBox: true,
         getTextCallback,
         changeTextCallback,
         preventHandleOutsideImage: false,
@@ -164,7 +166,7 @@ class ArrowAnnotateTool extends AnnotationTool {
     const { arrowFirst } = this.configuration;
     const FrameOfReferenceUID = viewport.getFrameOfReferenceUID();
 
-    const annotation = {
+    const annotation: ArrowAnnotation = {
       highlighted: true,
       invalidated: true,
       metadata: {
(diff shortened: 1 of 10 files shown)


packages/tools/src/tools/annotation/ArrowAnnotateTool.ts
```diff
@@ -181,20 +183,21 @@ class ArrowAnnotateTool extends AnnotationTool {
...
-              topRight: <Types.Point3>[0, 0, 0],
-              bottomLeft: <Types.Point3>[0, 0, 0],
-              bottomRight: <Types.Point3>[0, 0, 0],
-            },
-          },
         },
-        label: '',
       },
     };
+    if (!this.configuration.renderTextBox) {
```
sedghi, line 189: This should really be initialized, because the app config can switch renderTextBox on or off, and if it's not initialized, it will throw an error since the data is missing.

packages/tools/src/tools/annotation/ArrowAnnotateTool.ts
```diff
@@ -801,54 +812,56 @@ class ArrowAnnotateTool extends AnnotationTool {
...
-          worldBoundingBox: {
-            topLeft: <Types.Point3>[0, 0, 0],
-            topRight: <Types.Point3>[0, 0, 0],
-            bottomLeft: <Types.Point3>[0, 0, 0],
-            bottomRight: <Types.Point3>[0, 0, 0],
-          },
-        };
-        continue;
-      }
+      if (this.configuration.renderTextBox) {
```
sedghi, line 815: if (!this.configuration.renderTextBox){ return renderStatus} 

instead of indenting the code

packages/tools/src/tools/annotation/CircleROITool.ts
```diff
@@ -196,25 +198,27 @@ class CircleROITool extends AnnotationTool {
...
           points: [[...worldPos], [...worldPos]] as [
             Types.Point3, // center
             Types.Point3 // end
           ],
           activeHandleIndex: null,
         },
-        cachedStats: {},
       },
     };
+    if (this.configuration.renderTextBox) {
```
sedghi, line 209: same comment

packages/tools/src/tools/annotation/CircleROITool.ts
```diff
@@ -795,60 +801,62 @@ class CircleROITool extends AnnotationTool {
...
-          worldBoundingBox: {
-            topLeft: <Types.Point3>[0, 0, 0],
-            topRight: <Types.Point3>[0, 0, 0],
-            bottomLeft: <Types.Point3>[0, 0, 0],
-            bottomRight: <Types.Point3>[0, 0, 0],
-          },
-        };
-        continue;
-      }
+      if (this.configuration.renderTextBox) {
```
sedghi, line 804: same

sedghi (commented):
I'm not sure about this PR. The issue is that render and calculate are two separate things, and we're packing them into one boolean inside the configuration. 

Having said that, rendering also sounds like a presentation state, so one instance of a tool might want to decide whether to render a text box or not. I'm really wondering whether we should put that option in the `ToolStyle` class, similar to color and other properties.

Now the question is, how do we handle just rendering a rectangle, for example? I kind of wish we had a better compositional aspect, rather than inheritance, so we could say we want rectangle SVG drawing, but not stats, and we want to render a text box, and so on. But we don't have that now. 

For now, I guess building a new tool and just keeping what you want is the way to go.

jmhmd:
That makes sense, I can certainly make these custom tools, just trying to minimize the amount of custom code I have to maintain :) I can probably still just extend the tools as in this example and replace the `renderAnnotation` method to not calculate/draw the text box. Thanks for reviewing!

PR #1781, opened 2025-01-24
[OHI-1588] fix(flip): return flip as part of view presentation

IbrahimCSAE:
### Context

Flip property was being lost on new image renders, now it's returned as part of the view presentation so it's always maintained.

CleanShot 2025-01-24 at 12 29 31

Fixes: https://github.com/OHIF/Viewers/issues/4729

### DIFF
--- a/packages/core/src/RenderingEngine/Viewport.ts
+++ b/packages/core/src/RenderingEngine/Viewport.ts
@@ -1863,8 +1863,8 @@ class Viewport {
       zoom = this.getZoom(),
       pan,
       rotation,
-      flipHorizontal,
-      flipVertical,
+      flipHorizontal = this.flipHorizontal,
+      flipVertical = this.flipVertical,
     } = viewPres;
     if (displayArea !== this.getDisplayArea()) {
       this.setDisplayArea(displayArea);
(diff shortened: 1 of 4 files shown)


packages/core/src/RenderingEngine/Viewport.ts
```diff
@@ -1809,11 +1809,14 @@ class Viewport {
       displayArea: true,
       zoom: true,
       pan: true,
+      flipHorizontal: true,
```
wayfarer3130, line 1812: Please represent this as rotation + flip horizontal to make the values canonical

IbrahimCSAE: We will do this in another PR Bill

fgarcia760:
I found another bug. While I was focused on the buttons, I overlooked the hotkeys. The issue now appears when using the 'Flip Vertical' hotkey function.

Bug Details:
Incorrect Laterality Marker Orientation: When flipping vertically, the laterality marker becomes incorrect and changes orientation.
Unexpected Scroll Behavior: If you flip vertically and then flip horizontally, the series exhibits a strange behavior—it reverses direction as you scroll.

PR #1782, opened 2025-01-24
fix(StackViewport): Fix patient position calculation in StackViewport CPU mode.

sgielen:
### Context

I believe there is an error in the ImagePositionPatient calculations in the StackViewport CPU mode. The current code multiplies X coordinates (columns) with row cosines and vice versa. Because of this, in StackViewport CPU mode on non-axial volumes, my patient coordinate positioning seems to be off.

Unfortunately my cornerstone3D build fails, so I haven't been able to verify this fix.

### Changes & Results

Fix the output of the `canvasToWorldCPU`.

### Testing

My cornerstone3D build fails on my Mac and I don't have another machine handy to check my fixes, so this will be more of a theoretical fix.

I have a volume containing a coronal view. For this volume, the ImageOrientationPatient is `[0, 0, -1, 1, 0, 0]` i.e. rows are increasing towards the feet of the patient, and columns are increasing towards the left of the patient.

I've set up an image `onClick` which shows the canvas locations and world coordinates calculated by the Stack viewport in CPU mode on this volume. If I click one location on the volume, and then move a bit to the right and click again, then the X position of the canvas is slightly higher. Since higher column (X) values are increasing towards the left, I would expect the world coordinate to be slightly increasing towards the left (a higher [0] value), too.

However, because of this bug, the stack viewport accidentally takes the row cosine instead of the column cosine, so the world coordinate is instead slightly increasing towards the feet (a lower [2] value).

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] "OS: macOS"
- [x] "Node version: v23.5.0"
- [x] "Browser: Chrome 131"

### DIFF
--- a/packages/core/src/RenderingEngine/StackViewport.ts
+++ b/packages/core/src/RenderingEngine/StackViewport.ts
@@ -2861,8 +2861,8 @@ class StackViewport extends Viewport {
     const jVector = direction.slice(3, 6) as Point3;
 
     // Calculate the world coordinate of the pixel
-    vec3.scaleAndAdd(worldPos, origin, iVector, px * spacing[0]);
-    vec3.scaleAndAdd(worldPos, worldPos, jVector, py * spacing[1]);
+    vec3.scaleAndAdd(worldPos, origin, iVector, py * spacing[0]);
+    vec3.scaleAndAdd(worldPos, worldPos, jVector, px * spacing[1]);
 
     return worldPos;
   };


sedghi:
Can you share the anonymized data for me to test couple of things please?

sgielen:
> Can you share the anonymized data for me to test couple of things please?

Apparently, this rubber-duck is what I needed to figure out my issue.

In examining the data before sending it to you, I found out that I mixed up my interpretation of the ImageOrientationPatient tag. My data was incorrect, not the StackViewport implementation.

> Image Orientation (Patient) (0020,0037) specifies the direction cosines of the first row and the first column with respect to the patient. These Attributes shall be provide as a pair. Row value for the x, y, and z axes respectively followed by the Column value for the x, y, and z axes respectively.

I had interpreted this as "the first three values describe how ImagePositionPatient increases over rows (Y values), and the last three values describe how the ImagePositionPatient increases over columns (X values)". Hence, this issue. But it seems to be the reverse: the first three values describe how ImagePositionPatient increases **over the first row**, i.e. over the columns in the first row, i.e. over X values. I've fixed my data accordingly and now the patient positioning matches up with what I expect.

Apologies for wasting your time! Closing this PR.

sedghi:
It's fine, no PR is a bad PR

PR #1789, opened 2025-01-30
fix: inheritted configuration for tool

wayfarer3130:
### Context

Calling static methods from instance versions needs to make sure the correct static instance gets called.  This PR just calls the right version of the static method.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  


packages/tools/src/tools/annotation/ProbeTool.ts
```diff
@@ -92,7 +92,7 @@ const { transformWorldToIndex } = csUtils;
  */
 
 class ProbeTool extends AnnotationTool {
-  static toolName;
+  static toolName = 'Probe';
```
sedghi, line 95: this will cause some issues in some evns i remember, can we keep it outside like before?

wayfarer3130: It used to cause issues because of the build, but in 2.x it is fine.

sedghi: I mean either change all or none, i rather have one PR to change all, than spread them into separate PRs

wayfarer3130: Changed all.

sedghi (approved):

PR #1815, opened 2025-02-05
feat(adapters): overlapping segs with labelmap images

pedrokohler:
### Context

In order to address https://github.com/OHIF/Viewers/issues/3496 we're taking a two steps approach:

1 - chopping the volume data buffer array into planar data buffer arrays
2 - adopt some strategy to duplicate these planar data buffer arrays when we get overlapping segs

This PR aims to cover the second step.

### Changes & Results

Before - loss of information:
image

After - overlapping segs maintained:
image

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  


sedghi:
I don't think we're on the right track here. The adapter should return `imageIds`, and the example/consumer/ohif should create a **single** segmentation from those `imageIds`. We might need to tweak the internal segmentation addition and hydration process to make this work, but it's pretty likely to be okay. For now, let's put the volume one on the back burner - we'll need more optimizations to create sub-volumes for each `imageId` group, and that's a bit more complicated since we'd need to do it only where they're actually located. So make sure the example work with having one segmentation, the segmentation is consisted of multiple imageIds

pedrokohler:
> I don't think we're on the right track here. The adapter should return `imageIds`, and the example/consumer/ohif should create a **single** segmentation from those `imageIds`. We might need to tweak the internal segmentation addition and hydration process to make this work, but it's pretty likely to be okay. For now, let's put the volume one on the back burner - we'll need more optimizations to create sub-volumes for each `imageId` group, and that's a bit more complicated since we'd need to do it only where they're actually located. So make sure the example work with having one segmentation, the segmentation is consisted of multiple imageIds

I was really hoping you would point to specific parts of the code and ask for specific changes instead of abstract ones. I'm not following what you're asking me to do.

packages/adapters/examples/segmentationStack/index.ts
```diff
@@ -104,7 +104,7 @@ const state = {
     toolGroup: null,
     toolGroupId: "MY_TOOL_GROUP_ID",
     viewportIds: ["CT_AXIAL"],
-    segmentationId: "LOAD_SEG_ID:" + cornerstone.utilities.uuidv4(),
+    segmentationIds: [],
```
sedghi, line 107: here we should have the segmentationId, there is one segmentation for this case and it happens to be overlapping, so don't change this to an array

sedghi:
So basically the idea is to have this work. 

```js
const { labelMapImages } = 
        await Cornerstone3D.Segmentation.createFromDICOMSegBuffer(
            referenceImageIds,
            arrayBuffer,
            {
                metadataProvider: metaData,
                skipOverlapping: false
            }
        );
```

Then we do 

```js
csToolsSegmentation.addSegmentations([
        {
            segmentationId,
            representation: {
                type: cornerstoneTools.Enums.SegmentationRepresentations
                    .Labelmap,
                data: {
                    imageIds: labelMapImages.flat()
                }
            }
        }
    ]);
```

The key files that need to be changed are 

https://github.com/cornerstonejs/cornerstone3D/blob/main/packages/tools/src/eventListeners/segmentation/imageChangeEventListener.ts

Here

const derivedImageId = getCurrentLabelmapImageIdForViewport(
        viewportId,
        representation.segmentationId
      );

as you see it only get one labelmapimageId for viewport => but we have changed that so that it can have multiple for one segmentationId (overlapping)

and then later on in that file 

```js
viewport.addImages([
        {
          imageId: derivedImageId,
          representationUID: `${segmentationId}-${SegmentationRepresentations.Labelmap}`,
          callback: ({ imageActor }) => {
            imageActor.getMapper().setInputData(imageData);
          },
        },
      ]);
```

you see we only add one derivedImageId to the viewport (since we only had support for one slice -> one imageId -> per labelmap) but now we want to support multiple 

so we need to loop over derivedImageIds for that slice and then add them one by one which should work 

Hope this helps

fedorov:
@sedghi is there a timeline for feedback on this? We are blocked by this PR.

pedrokohler:
@sedghi as asked on slack, here is the table of comparison for the skipOverlapping flag (master vs this branch):

image

PR #1820, opened 2025-02-10
feat(seg): Improve Labelmap Statistics, Interpolation, and Threshold Configuration

sedghi:
This PR will: 

## Labelmap Statistics
- makes it work on stack viewports too
- moves the calculation to web worker to not block ui
- moves it to a utility function instead of a brush strategy (we should be able to calculate stats on anything not just brush instances)

## Sphere Brush 
- was broken on stack viewports now it is fixed

## Interpolation
- makes it work on stack viewports too
- moves the calculation to web worker to not block ui
- moves it to a utility function instead of a brush strategy (we should be able to calculate stats on anything not just brush instances)

## THRESHOLD -> Strategy specific threshold 

Previously, Brush tools using thresholding might have used a generic THRESHOLD property within their strategySpecificConfiguration. This PR introduces a shift towards more explicit and strategy-specific threshold configurations. Instead of a single THRESHOLD property (which is NOT a strategy), you'll now configure thresholds under properties named after the specific strategy itself (e.g., THRESHOLD_INSIDE_CIRCLE, THRESHOLD_INSIDE_SPHERE, THRESHOLD_INSIDE_SPHERE_WITH_ISLAND_REMOVAL). 

It is now the app-level concern to make sure threshold is synced between different tool strategies if desired

```js
{
    configuration: {
        activeStrategy: 'THRESHOLD_INSIDE_CIRCLE', 
        strategySpecificConfiguration: {
            THRESHOLD: { // Old generic THRESHOLD property
                threshold: [50, 200], // Example threshold range
            },
        },
    },
});
```

- If your activeStrategy is 'THRESHOLD_INSIDE_CIRCLE', rename THRESHOLD to THRESHOLD_INSIDE_CIRCLE.
- If your activeStrategy is 'THRESHOLD_INSIDE_SPHERE', rename THRESHOLD to THRESHOLD_INSIDE_SPHERE.
- If your activeStrategy is 'THRESHOLD_INSIDE_SPHERE_WITH_ISLAND_REMOVAL', rename THRESHOLD to THRESHOLD_INSIDE_SPHERE_WITH_ISLAND_REMOVAL.

```js
After (Example - Migrated Configuration):

{
     configuration: {
         activeStrategy: 'THRESHOLD_INSIDE_CIRCLE',
        strategySpecificConfiguration: {
            useCenterSegmentIndex: true,
            THRESHOLD_INSIDE_CIRCLE: { // Renamed to strategy-specific property
                threshold: [50, 200], // Threshold settings remain the same inside
            },
        },
    },
});

### CODE REVIEW

**Comment by sedghi** [comment]
@wayfarer3130, could you check out `peerImport` versus `import`? I'm having trouble getting the `labelmapInterpolation` example to work without `await import`. For some reason, running `peerImport` inside a web worker just isn't working.

File: packages/tools/src/workers/computeWorker.ts [hunk 1]
```diff
@@ -0,0 +1,132 @@
...
+  },
+
+  interpolateLabelmap: async (args) => {
+    const { segmentationInfo, configuration } = args;
+    const { scalarData, dimensions, spacing, origin, direction } =
+      segmentationInfo;
+
+    let itkModule;
+    try {
+      itkModule = await import('@itk-wasm/morphological-contour-interpolation');
```
wayfarer3130, line 62: This will cause webpack/ohif to do a hard link to itk-wasm, breaking the fact that it should be able to load it dynamically.

sedghi: Done

packages/core/src/cache/cache.ts
```diff
@@ -33,13 +33,45 @@ class Cache {
...
   private _imageCacheSize = 0;
   private _maxCacheSize = 3 * ONE_GB;
   private _geometryCacheSize = 0;
 
+  /**
+   * Generates a deterministic volume ID from a list of image IDs
+   * @param imageIds - Array of image IDs
+   * @returns A deterministic volume ID
+   */
+  public generateVolumeId(imageIds: string[]): string {
```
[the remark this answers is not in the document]
wayfarer3130: Also, does it need to be order invariant?  Should the entries be sorted first?

packages/tools/examples/labelmapInterpolation/index.ts
```diff
@@ -162,20 +162,10 @@ addDropdownToToolbar({
...
-addButtonToToolbar({
-  title: 'Run Overlapping Interpolation',
-  onClick: () => {
-    const toolGroup = ToolGroupManager.getToolGroup(toolGroupId);
-    const activeName = toolGroup.getActivePrimaryMouseButtonTool();
-    const brush = toolGroup.getToolInstance(activeName);
-    brush.interpolate?.(element1, { extendedConfig: false });
+    segmentationUtils.interpolateLabelmap({
+      segmentationId,
+      segmentIndex: 1,
```
wayfarer3130, line 167: interpolation seems to fail if segmentIndex is not the same as what is generated

packages/tools/examples/stackLabelmapSegmentation/index.ts
wayfarer3130, line 1: I believe you are missing the dynamic threshold getting added to the tool group - it says not added to tool group when I try to select it.  
Also, could you leave the entire stack there so there is a bit more data to test with for the CT, so that volumes can be looked at?

packages/core/src/cache/cache.ts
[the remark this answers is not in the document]
sedghi: Seems like crypto apis rely on secure context, so i guess we can't

utils/demo/helpers/initDemo.js
```diff
@@ -49,15 +49,19 @@ export async function peerImport(moduleId) {
   }
 
   if (moduleId === '@icr/polyseg-wasm') {
-    return import('@icr/polyseg-wasm');
+    return import(
+      /* webpackChunkName: "icr-polyseg-wasm" */ '@icr/polyseg-wasm'
```
wayfarer3130, line 53: Can you update with the vite ignore and the separate variable name?

sedghi: These are inside our demo which we don't build and don't ship

packages/labelmap-interpolation/src/workers/interpolationWorker.js
```diff
@@ -0,0 +1,172 @@
+import { expose } from 'comlink';
+import vtkImageData from '@kitware/vtk.js/Common/DataModel/ImageData';
+import vtkDataArray from '@kitware/vtk.js/Common/Core/DataArray';
+
+/**
+ * Dynamically imports ITK WASM modules needed for labelmap interpolation
```
wayfarer3130, line 6: Could you add a comment that this doesn't currently allow the dynamic import?

sedghi:
closing in favor of https://github.com/cornerstonejs/cornerstone3D/pull/1844

PR #1821, opened 2025-02-12
Updating proper values for ChangeType parameter in AnnotationModifiedEventDetail.

nithin-trenser:
### Context

 
Updated ChangeType parameter value in AnnotationModifiedEventDetail for LengthTool
### Changes & Results
The ```ChangeType```parameter in ```AnnotationModifiedEventDetail``` now returns ```StatsUpdated``` following the execution of the ```_calculateCachedStats``` function, it returns ```HandlesUpdated``` when the handle position is modified.

The ```initialSetup``` change type can be fired from addNewAnnotation(), we currently left it out. Please let us know if you think it should be there as cs3d already has a annotation created event for it. Should we remove the change type InitialSetup ?

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  

### DIFF
--- a/packages/tools/src/tools/annotation/PlanarFreehandROITool.ts
+++ b/packages/tools/src/tools/annotation/PlanarFreehandROITool.ts
@@ -685,7 +685,7 @@ class PlanarFreehandROITool extends ContourSegmentationBaseTool {
       const { data } = annotation;
       if (
         !data.cachedStats[targetId] ||
-        data.cachedStats[targetId].areaUnit == null
+        data.cachedStats[targetId].areaUnit === null
       ) {
         data.cachedStats[targetId] = {
           Modality: null,
--- a/packages/tools/src/tools/annotation/PlanarFreehandROITool.ts
+++ b/packages/tools/src/tools/annotation/PlanarFreehandROITool.ts
@@ -684,10 +684,7 @@ class PlanarFreehandROITool extends ContourSegmentationBaseTool {
     if (!this.commonData?.movingTextBox) {
       const { data } = annotation;
       const { closed } = data.contour;
-      if (
-        !data.cachedStats[targetId] ||
-        (closed && !data.cachedStats[targetId]?.areaUnit)
-      ) {
+      if (!data.cachedStats[targetId] || !data.cachedStats[targetId]?.unit) {
         data.cachedStats[targetId] = {
           Modality: null,
           area: null,
@@ -695,6 +692,7 @@ class PlanarFreehandROITool extends ContourSegmentationBaseTool {
           mean: null,
           stdDev: null,
           areaUnit: null,
+          unit: null,
         };
 
         this._calculateCachedStats(
--- a/packages/tools/src/tools/annotation/PlanarFreehandROITool.ts
+++ b/packages/tools/src/tools/annotation/PlanarFreehandROITool.ts
@@ -683,8 +683,7 @@ class PlanarFreehandROITool extends ContourSegmentationBaseTool {
 
     if (!this.commonData?.movingTextBox) {
       const { data } = annotation;
-      const { closed } = data.contour;
-      if (!data.cachedStats[targetId] || !data.cachedStats[targetId]?.unit) {
+      if (!data.cachedStats[targetId]?.unit) {
         data.cachedStats[targetId] = {
           Modality: null,
           area: null,
--- a/packages/tools/src/tools/annotation/PlanarFreehandROITool.ts
+++ b/packages/tools/src/tools/annotation/PlanarFreehandROITool.ts
@@ -683,9 +683,10 @@ class PlanarFreehandROITool extends ContourSegmentationBaseTool {
 
     if (!this.commonData?.movingTextBox) {
       const { data } = annotation;
+      const { closed } = data.contour;
       if (
         !data.cachedStats[targetId] ||
-        data.cachedStats[targetId].areaUnit === null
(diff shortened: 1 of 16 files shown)


packages/tools/src/utilities/setAnnotationLabel.ts
```diff
@@ -0,0 +1,12 @@
...
+import { ChangeTypes } from '../enums';
+
+export default function setAnnotationLabel(
+  annotation: Annotation,
+  element: HTMLDivElement,
+  updatedLabel: string
+) {
+  annotation.data.label = updatedLabel;
+  triggerAnnotationModified(annotation, element, ChangeTypes.LabelChange);
+}
```
sedghi, line 12: we don't have a label on Annotation in the Types as i see, maybe we add it in the types as well in AnnotationTypes.ts

nithin-trenser: Label added.

packages/tools/src/tools/annotation/PlanarFreehandROITool.ts
```diff
@@ -685,7 +685,7 @@ class PlanarFreehandROITool extends ContourSegmentationBaseTool {
       const { data } = annotation;
       if (
         !data.cachedStats[targetId] ||
```
wayfarer3130, line 687: How about !data.cachedStates[targetId]?.areaUnit
The areaUnit should be a non empty string, even for unknown units as that will be pixels

packages/tools/src/types/AnnotationTypes.ts
```diff
@@ -81,6 +81,8 @@ type Annotation = {
     [key: string]: unknown;
     /** Cached Annotation statistics which is specific to the tool */
     cachedStats?: Record<string, unknown>;
+    /** Label of an annotation */
+    label?: string;
```
wayfarer3130, line 85: This is inconsistent with the label in LengthTool - label belongs to the data element.   Note the difference between CS3D and OHIF - in CS3D the data is in the data object, while in OHIF it is at the top level.

packages/tools/src/utilities/setAnnotationLabel.ts
wayfarer3130: This is the correct level here, but the declaration is at the upper level.

nithin-trenser: The declaration is defined inside the data element in Annotation types.

packages/tools/src/tools/annotation/PlanarFreehandROITool.ts
nithin-trenser: It seems there is an existing issue that calculateCachedStats is always called for each mouse movement of open freenad annotation. This is due to [the areaUnit for open freehand ROIs](https://github.com/nithin-trenser/cornerstone3D/blob/feat/updating-changetype-values/packages/tools/src/tools/annotation/PlanarFreehandROITool.ts#L993) in cachedStates is always ```undefined```.

As this [condition](https://github.com/cornerstonejs/cornerstone3D/blob/main/packages/tools/src/tools/annotation/PlanarFreehandROITool.ts#L688) is always true for Open freehand annotations, resulting in the unintended activation of an ```annotationModified``` event, we are restricting the second condition for closed freehand only.

nithin-trenser:
> The initial creation should leave the label changed event out

At present, the label change event is not triggered upon the creation of an annotation. We will trigger this event from the OHIF side when an annotation label is added or modified.

sedghi (approved):
I'm fine, let's see what @wayfarer3130 thinks

packages/tools/src/tools/annotation/PlanarFreehandROITool.ts
wayfarer3130: There should be an equivalent check for open contours, and I'd really rather have a single check that works for both.

packages/tools/src/types/AnnotationTypes.ts
wayfarer3130: Sorry, was seeing the nesting wrong.  Could you move the handles to it's own type - I think that is sufficiently long to be worth having by itself.  That would resolve the issue with having it hard to see the nesting.

packages/tools/src/types/AnnotationTypes.ts
```diff
@@ -106,6 +84,32 @@ type AnnotationState = {
   [key: string]: GroupSpecificAnnotations;
 };
 
+type handles = {
```
wayfarer3130, line 87: Needs to be
export type Handles = {
to match the default casing.

packages/tools/src/types/AnnotationTypes.ts
wayfarer3130: Thanks, I like that a lot better.  I never like having to try to manually figure out the indentation to figure out what is what.

wayfarer3130 (commented):
Just rename the handles type to Handles to agree with the general casing (and export it too), and shorten the if condition and I will approve/merge.

nithin-trenser:
@wayfarer3130 could you please check again and approve/merge?

sedghi:
can you rebase, we can merge it

nithin-trenser:
@sedghi rebased.

sedghi (approved):
Thanks

PR #1828, opened 2025-02-14
initial commit for fixing video segmentations

minimal-scouser:
### Context

Video segmentations are broken in cornerstone3d. Well sort of. They are working but they are not properly tied to a frame. Rather they persist across frames which is not ideal.

They used to work in 3d 1.0

### Changes & Results

I sense that the mappings of derivedImageId and the currentImageId are not proper, and I confirmed that for every imageId, the mapped derivedImageId is same, which is why these labelmaps are being carried forward.

working demo after changes

https://github.com/user-attachments/assets/4af91b53-07ce-479e-bf3a-8369208f2e2b

### Testing

run the videosegmentation example

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  


packages/tools/examples/videoSegmentation/index.ts
```diff
@@ -224,11 +224,14 @@ async function run() {
...
   const allImageIds = viewport.getImageIds();
   const firstImage = allImageIds[0];
-  const segImages = await imageLoader.createAndCacheDerivedImages(
-    [firstImage],
+  const segImages = await imageLoader.createAndCacheDerivedLabelmapImages(
+    allImageIds,
     {
       skipCreateBuffer: true,
       onCacheAdd: csUtils.VoxelManager.addInstanceToImage,
+      getDerivedImageId: (referenceImageId, index) => {
```
wayfarer3130, line 232: Please use the multiframe utilities to update the frame reference.
Also, there is no need to use a random uid, you can use the original uid with derived in front of it, but updating the frame index so that every time you generate the same uids.
Finally, this derivation should probably go into the core code so that the handling is consistent and can be done from things like OHIF.

wayfarer3130: @sedghi  - should we add a vewport call 'getDerivedImageIds'   It could take a range request to allow getting a sub-range of it.

minimal-scouser: https://github.com/cornerstonejs/cornerstone3D/blob/0b67f91a41eb069e0ef212d4be7acdcb4c68c542/packages/core/src/loaders/imageLoader.ts#L356-L368

line 360 is creating a derivedImageId using a random uuid anyway if we don't implement `getDerivedImageId`. so i am little confused.

here's a trace:
createAndCacheDerivedLabelmapImages
createAndCacheDerivedImages
line 360

hene i assumed i was not doing anything drastically different reg the creation of a derivedImageId. i thought `getDerivedImageId` lets the user create a derviedImageId in a format they want, which is why i used it to append frames to the id.

minimal-scouser: @wayfarer3130

wayfarer3130: Could you at least move the create derived images into a specialized method in imageLoaders.ts that takes a type argument that can be video/stack/volume/wsi and returns the right type of response?  It should take the same type of argument as the rendering engine create does.  That way the same code can be re-used more generally, and we can update it to use non-random uids if we choose, or continue as is.

minimal-scouser: that makes more sense. on it.

PR #1829, opened 2025-02-16
fix(rendering): fix image overflow issues by restoring data type check, and fix window level to be maintained if it was modified by the user and image data can't be reused

IbrahimCSAE:
### Context

Some MR images were overflowing and not rendering correctly because we were no longer checking the data type, this has been restored. A fix for the window level has been added to also maintain it if the image data can't be re-used


danielschwartz4 (approved):

sedghi (approved):
Thanks

PR #1833, opened 2025-02-18
fix for changing opacity based on segment index

### LINKED ISSUE #1832: [Feature Request] Support per-segment active/inactive styles within a segmentation

Currently, active and inactive styles can only be set at the segmentation level, meaning all segments within the same segmentation share the same active/inactive styling. I would like to request a feature that allows each segment within a segmentation to have its own distinct active and inactive styles.

- Improved Visual Clarity: Users can better distinguish active and inactive segments within the same segmentation.
- Increased Flexibility: Allows per-segment customization instead of enforcing a single active/inactive style at the segmentation level.

minimal-scouser:
### Context

Fixes #1832 

While rendering, segment styles such as opacity, fill etc do not consider active segment based on its index.

### Changes & Results

Active segmentationId along with the segmentIndex should determine whether a segment is active or not.

https://github.com/user-attachments/assets/5c0491e6-0d0b-4d94-8b7e-8882d0d348f8

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  

### DIFF
--- a/packages/tools/src/utilities/segmentation/getSVGStyleForSegment.ts
+++ b/packages/tools/src/utilities/segmentation/getSVGStyleForSegment.ts
@@ -63,7 +63,11 @@ export function getSVGStyleForSegment({
   );
 
   const activeSegmentation = getActiveSegmentation(viewportId);
-  const isActive = activeSegmentation?.segmentationId === segmentationId;
+  const isSegmentAtIndexActive =
+    activeSegmentation.segments[segmentIndex].active;
+  const isActive =
+    activeSegmentation?.segmentationId === segmentationId &&
+    isSegmentAtIndexActive;
 
   // Merge the configurations from different levels based on its precedence
   const style = segmentationStyle.getStyle({


sedghi (approved):

stada526:
Thank you for addressing the issue I raised!

sedghi:
I think I messed up by merging this. I thought the opacity was at the segmentation level, but it seems to be segment-specific. Could you please create a pull request to revert this change? My reasoning is that there's only one layer of opacity change at the API level, which is between segments, not within them. If someone needs a different opacity for any active segment, they can use the `perSegment` configuration.

This PR also cannot get merged, https://github.com/cornerstonejs/cornerstone3D/pull/1836 as it is adding another layer of opacity configuration at api level. There is already API for perSegment configuration

minimal-scouser:
doing it

minimal-scouser:
per segment styles are not respected

the original intent itself is not aligned to support perSegment styles
https://github.com/cornerstonejs/cornerstone3D/blob/06ebd68df854a774e7e2fa5063d52902704935a8/packages/tools/src/stateManagement/segmentation/SegmentationStyle.ts#L21-L22

look at `updatedStyles` - it is basically the styles set at segmentation level
https://github.com/cornerstonejs/cornerstone3D/blob/06ebd68df854a774e7e2fa5063d52902704935a8/packages/tools/src/stateManagement/segmentation/SegmentationStyle.ts#L98-L117

and here we are assigning the high level styles to segment level style
https://github.com/cornerstonejs/cornerstone3D/blob/06ebd68df854a774e7e2fa5063d52902704935a8/packages/tools/src/stateManagement/segmentation/SegmentationStyle.ts#L146-L155

sedghi:
I think they are working at least in OHIF, since this is the effect we have in OHIF, if you look at SegmentationService in OHIF you will see it 

https://viewer-dev.ohif.org/viewer?StudyInstanceUIDs=1.3.6.1.4.1.32722.99.99.62087908186665265759322018723889952421

https://github.com/user-attachments/assets/3c08f743-a8e7-4484-b4f8-55cf284fa207

see how per segment we apply different thickness

minimal-scouser:
true. it works for visibility. even in the example.

but it does not consider fillAlphaInactive, fillOutlineInactive - especially at segment level. 

try this,

go to https://viewer-dev.ohif.org/segmentation?StudyInstanceUIDs=2.16.840.1.114362.1.11972228.22789312658.616067305.306.2

create two segments (under same segmentation)

run
```javascript
cornerstoneTools.segmentation.config.style.getStyle({segmentationId: "Segmentation 1", type: "Contour"})
```

and this is what you will get
```javascript
{
    "renderOutline": true,
    "outlineWidthAutoGenerated": 3,
    "outlineWidth": 1,
    "outlineWidthInactive": 1,
    "outlineOpacity": 1,
    "outlineOpacityInactive": 0.85,
    "outlineDashAutoGenerated": "5,3",
    "activeSegmentOutlineWidthDelta": 0,
    "renderFill": false,
    "fillAlpha": 0.5, <---------
    "fillAlphaInactive": 0.3, <----------
    "fillAlphaAutoGenerated": 0.3
}
```

notice how `fillAlpha` and `fillAlphaInactive` have different values yet on the viewport, there is no difference.

PR #1846, opened 2025-02-24
feat(surface-segmentation): add visibilty configuration for surface segmentation

IbrahimCSAE:
### Context

Adds visibility configuration to surface segmentations

Fixes: https://github.com/OHIF/Viewers/issues/4796
Fixes: https://linear.app/ohif/issue/OHI-1667


packages/tools/src/stateManagement/segmentation/helpers/getSegmentationActor.ts
```diff
@@ -76,11 +77,12 @@ export function getSurfaceActorEntry(
   segmentationId: string,
   segmentIndex?: number | string
 ) {
-  return getActorEntry(viewportId, segmentationId, (actor) =>
-    // @ts-expect-error
-    actor.representationUID?.startsWith(
```
IbrahimCSAE, line 81: this gave me so much pain... 4 hours to figure out what was going on

this is not good because it will match Surface-1 with Surface 10, 11 , 12, 13, 14, 15 etc, or Surface-2 with 21, 22, 23.

such a horrible bug because it was saying there's an actor for a segment when there isn't one

packages/tools/src/tools/displayTools/Surface/addOrUpdateSurfaceToElement.ts
```diff
@@ -25,6 +25,16 @@ function addOrUpdateSurfaceToElement(
 
   const surfaceActor = surfaceActorEntry?.actor as Types.Actor;
 
+  const isVisible = surface.visible;
+
+  if (!isVisible) {
```
sedghi, line 30: if (!isVisible & surfaceActor)

packages/tools/src/tools/displayTools/Surface/addOrUpdateSurfaceToElement.ts
```diff
@@ -25,6 +25,16 @@ function addOrUpdateSurfaceToElement(
...
+    if (surfaceActor) {
+      viewport.removeActors([surfaceActorEntry.uid]);
```
sedghi, line 32: Try using `actor.setVisibility(true)` at the actor level instead of `remove`; there should be a `setVisibility` API there.

sedghi (approved):
Thanks

PR #1848, opened 2025-02-24
fix: add combined size pool

wayfarer3130:
### Context

This PR lays the groundwork for adding display fetch controls that delay initial fetching of thumbnails and prefetches until the primary viewports are completed.

Default combined cap raised 10 → 50. The initial maxConcurrentRequests default of 10 was too low for a general default — it would globally throttle the shared imageLoadPoolManager (which currently sets Interaction/Thumbnail/Prefetch to 1000 each) down to 10 concurrent HTTP requests. Bumped the default to 50 as a more reasonable starting point.
RequestType.Metadata producer is intentionally not wired here. The new metadata queue is added in this PR but nothing enqueues RequestType.Metadata yet. Tagging metadata fetches with this type will be wired up later on the OHIF side.
setMaxConcurrentRequests caller is intentionally not wired here. No consumer calls it in this PR, so the 50 default governs. OHIF will call setMaxConcurrentRequests to tune the cap for its own use case in a follow-up.

### Changes & Results

Adds a prefetch queue for metadata
Adds an overall maximum size prefetch queue that applies to the metadata through prefetch queues, with the higher priority queues being done first.  The intent of this is in an application such as OHIF to allow all the metadata queries to be sent off initially, and then let those continue to come in, while the initial images are not fetched initially.  Then, the viewport specific images would be fetched and then thumbnails and finally prefetch images.

### Testing

Link into OHIF, and display a study with thumbnails and prefetch.  Add a log message on prefetch queue.
Reduce the delay in study browser to 125 ms or something like that, and all the thumbnails should be before the prefetch.
Should also add the metadata prefetch to the queue at level metadata, and add a viewport fetch for first image at low priority as soon as the display sets are created.  That will demonstrate slowly fetching viewport/thumbnail images mixed in with metadata requests.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  

## Summary

* **New Features**
  * Added `Metadata` as a highest-priority request type.
  * Introduced a combined concurrency cap across metadata, interaction, thumbnail, and prefetch (default: 50).
  * Added reporting for outstanding interaction, thumbnail, and prefetch requests (metadata excluded).
* **Bug Fixes**
  * Improved request scheduling to prevent interaction requests from being starved and ensure priority/fairness when the pool is full.
  * Compute requests continue running independently of the shared limit.
* **Documentation**
  * Updated request pool manager concepts and added a 5.x migration guide for the combined limit behavior.
* **Tests**
  * Added Jest coverage for the new prioritization, limits, and outstanding-count rules.


packages/core/src/requestPool/requestPoolManager.ts
```diff
@@ -66,6 +66,12 @@ type RequestPool = {
  * You don't need to directly use the imageRetrievalPoolManager to load images
  * since the imageLoadPoolManager will automatically use it for retrieval. However,
  * maximum number of concurrent requests can be set by calling `setMaxConcurrentRequests`.
+ *
+ * ### Requests between types
+ * There are different types of requests, and each type starves the next request
+ * type, under the assumption that they are needed in order for fetching.
+ * The total number of requests is set by the setMaxCOncurrentRequests, defaulting to 10
```
jbocce: Typo setMaxCo*ncurrentRequests.

packages/core/src/requestPool/requestPoolManager.ts
```diff
@@ -280,19 +310,48 @@ class RequestPoolManager {
...
+        (this.numRequests[RequestType.Interaction] === 0 && 1) ||
+        0
     );
+    available = this.maxConcurrentRequests - this.outstandingRequests;
+    // Allow an extra request for thumbnail if there are no outstanding
+    // interaction or thumbnail requests.
     const hasRemainingThumbnailRequests = this.sendRequests(
-      RequestType.Thumbnail
+      RequestType.Thumbnail,
+      (available > 0 && available) ||
```
jbocce, line 341: Can't this be simplied to (available > 0)?

packages/core/src/requestPool/requestPoolManager.ts
```diff
@@ -280,19 +310,48 @@ class RequestPoolManager {
...
+    const hasRemainingMetadataRequests = this.sendRequests(
+      RequestType.Metadata,
+      this.maxConcurrentRequests - this.outstandingRequests
+    );
+    let available = this.maxConcurrentRequests - this.outstandingRequests;
+    // Allow 1 interaction request always
     const hasRemainingInteractionRequests = this.sendRequests(
-      RequestType.Interaction
+      RequestType.Interaction,
+      (available > 0 && available) ||
```
jbocce, line 332: Can't this be simplified to (available > 0)?

jbocce (dismissed):
Couple of things to look at. Thanks.

packages/core/src/requestPool/requestPoolManager.ts
wayfarer3130: No, it would change the logic.  Added a comment explaining it.

jbocce:
@claude review

packages/core/src/requestPool/requestPoolManager.ts
wayfarer3130: No - added a comment to explain that.

jbocce:
FYI... the OHIF downstream validation is failing.

wayfarer3130:
Based on feedback offline, it seems like having different queues for different things that all are basically http sounds like a bad idea.  Going to redo this in a way that combines the queues

PR #1849, opened 2025-02-24
Feat/new cs3d brush threshold

sedghi:
writing...

TLDR:

Just externalizing itk-wasm and polyseg-wasm to satisfy Webpack and Vite.
some seg threshold tool reworking


packages/core/src/enums/Events.ts
wayfarer3130, line 1: Could you merge stack and volume new image/new dataset events?  That way they can be used for video, sm etc without needing to have different listeners.  That was missed in the previous 2.0 release and had been something we were going to do.  A separate PR is fine for that, just into beta.

wayfarer3130 (approved):
The changes I suggested probably need to go in a different PR so that we have a set of changes which together make up beta for CS3D 3.0
The rest of hte changes look good - I particularly like just having a separate module for some of the large external libraries being used.

packages/core/src/enums/Events.ts
sedghi: Yes i will create another PR

Issue #1861, opened 2025-02-26
[Bug]


pnewhook:
### Describe the Bug

Flipping MRI/CT stacks vertically **and** horizontally in a stack viewport loses flip after scrolling. We've found this issue in our application, and we believe it's reproducible in the Stack API example.

### Steps to Reproduce

1. Visit https://www.cornerstonejs.org/live-examples/stackapi
2. Flip both horizontally and vertically
3. Click on Next Image
4. The next image is no longer flipped horizontally

### The current behavior

Individual images in the stack lose the horizontal flip after scrolling. The vertical flip flag is now false, and the image is instead rotated 180 degrees.

### The expected behavior

All images in a stack should be flipped horizontally and vertically. If only a horizontal or vertical flip is selected, the flip is visible on the full stack when scrolling.

### OS

macOS 14.7.2 (23H311)

### Node version

NA

### Browser

Chrome 133.0.6943.127

PR #1868, opened 2025-03-01
fix(flip): seperate flip calls

IbrahimCSAE:
### Context

Fixes issues where flip wasnt being maintained correctly by seperating the flip calls

https://github.com/user-attachments/assets/14a1de61-ce9d-4676-b091-461392cc8754

Fixes: https://github.com/cornerstonejs/cornerstone3D/issues/1861
Fixes: https://github.com/OHIF/Viewers/issues/4729
Fixes: https://linear.app/ohif/issue/OHI-1588


sedghi (changes requested):
can you add an e2e test for this one

packages/core/src/RenderingEngine/Viewport.ts
minimal-scouser, line 1: can flip not take null as a value?

sedghi (approved):
Thanks

PR #1887, opened 2025-03-10
fix: Only copy patient/study attributes rather than all

wayfarer3130:
### Context

When creating an SR, the previous code used ...instance on an example instance in the study to create the new patient/study headers.  However, that added headers like "Rows" and "Columns" which are completely bogus.  This PR copies just selected attributes.  It might not be quite a full set, but includes all the major ones.

Fixes https://github.com/OHIF/Viewers/issues/4804#issuecomment-2701056580

### Changes & Results

Add a copyStudyTags helper method to create a new study tags object.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  

### DIFF
--- a/packages/adapters/src/adapters/Cornerstone3D/MeasurementReport.ts
+++ b/packages/adapters/src/adapters/Cornerstone3D/MeasurementReport.ts
@@ -1,7 +1,7 @@
 import { normalizers, data, utilities, derivations } from "dcmjs";
 
 import CORNERSTONE_3D_TAG from "./cornerstone3DTag";
-import { toArray, codeMeaningEquals } from "../helpers";
+import { toArray, codeMeaningEquals, copyStudyTags } from "../helpers";
 import Cornerstone3DCodingScheme from "./CodingScheme";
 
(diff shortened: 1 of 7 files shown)


packages/adapters/src/adapters/Cornerstone3D/MeasurementReport.ts
```diff
@@ -206,11 +206,7 @@ export default class MeasurementReport {
 
         const _meta = MeasurementReport.generateDatasetMeta();
 
-        const derivationSourceDataset = {
-            ...instance,
-            _meta: _meta,
-            _vrMap: _vrMap
-        };
+        const derivationSourceDataset = copyStudyTags(instance);
```
wayfarer3130, line 209: Use the new copyStudyTags which encapsulates which tags are needed.  The previous code was creating invalid DICOM because it would copy things like Rows and Columns, and wouldn't work on the multiframe inheritted objects.

sedghi (approved):

sedghi (approved):

PR #1891, opened 2025-03-11
fix(tests): reduce timeouts / retries / workers

IbrahimCSAE:
### Context

Reducing the test timeout so a test doesn't run for an unexpected duration like 12 minutes. A test that can't pass in 2 minutes is a test we should consider not supported by our CI machine specs for the time being. We can still add the test and use it locally, however we add a `test.skip` before commiting until we upgrade our CI machines in the future.

Reduce workers as CI machines can't handle the load of multiple workers, workers are now 1

Disable retries, since they double/triple the duration of a single test.


sedghi (approved):
Thanks

Issue #1874, opened 2025-03-04
[Bug] `getStatistics` utility: `min.pointIJK` is null


nagychris:
### Describe the Bug

I am trying to calculate a boundingBox using the values returned by the `getStatistics` utility function for segmentations.

However, I noticed that the function does not return any value for the `min.pointIJK` value, even though the `max.pointIJK` is set correctly. 

![Image](https://github.com/user-attachments/assets/2ede8186-faf1-448a-8308-d406c2bff4e8)

### Steps to Reproduce

1. Go to demo: https://www.cornerstonejs.org/live-examples/labelmapstatistics
2. Open dev tools and go to `BasicStatsCalculator.ts` file
3. Set a breakpoint at line 162 (`return named`)
4. Draw a segment in the viewport with the brush tool
5. inspect the `named` object and its properties -> `min.pointIJK` is `null`

### The current behavior

 `min.pointIJK` is `null` 

### The expected behavior

 `min.pointIJK` is set to the minimum IJK values of the segment, similar to how `max.pointIJK` is set to the maximum values.

### OS

Ubuntu 24

### Node version

22.7.0

### Browser

Chrome latest

nagychris:
Closed via #1892

PR #1892, opened 2025-03-11
fix(BasicStatsCalculator): remove redundant Math.min call

nagychris:
### Context

Fixes the following bug, that involves a missing value for `min.pointIJK` in the result of `getStatistics` utility:
https://github.com/cornerstonejs/cornerstone3D/issues/1874

### Changes & Results

Removed the call to `Math.min()` which is redundant (the following two lines to the exact same), and prevents the following code to execute, because the minimum value is already inside the `stats.min[idx]`, so the comparison will always evaluate to false (`value < state.min[idx]`). Therefore, the fix makes sure that `stats.min.pointIJK` and `stats.min.pointLPS` values will always be set accordingly when the min value updates. 
This is basically the same logic as we do for the `stats.max` values.

### Testing

1. Visit https://www.cornerstonejs.org/live-examples/labelmapstatistics 
2. debug `BasicStatsCalculator.ts` with the changes -> the min value is updated correctly

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] "OS: Ubuntu 24
- [x] "Node version: v22.7.0
- [x] "Browser: Latest Chrome

### DIFF
--- a/packages/tools/src/utilities/math/basic/BasicStatsCalculator.ts
+++ b/packages/tools/src/utilities/math/basic/BasicStatsCalculator.ts
@@ -70,7 +70,6 @@ function basicStatsCallback(
     const delta2 = value - state.runMean[idx];
     state.m2[idx] += delta * delta2;
 
-    state.min[idx] = Math.min(state.min[idx], value);
     if (value < state.min[idx]) {
       state.min[idx] = value;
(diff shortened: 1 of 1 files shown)


sedghi (approved):
Thanks!

PR #1897, opened 2025-03-12
Fix undefined window error in logger

faridyagubbayli:
### Context

Logger accesses `window` object which can be `undefined` if used in wasm worker. For example, check the [Volume Labelmap to Surface](https://www.cornerstonejs.org/live-examples/polysegwasmvolumelabelmaptosurface) example - segment some part and try to convert the labelmap to surface segmentation, you'll get the error below:

`Uncaught (in promise) ReferenceError: window is not defined`

### Changes & Results

Added check for undefined window. If window is undefined, then skip setting the log level.

### Testing

Tested by running the example mentioned above and making sure it runs correctly.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  

### DIFF
--- a/packages/core/src/utilities/logger.ts
+++ b/packages/core/src/utilities/logger.ts
@@ -8,7 +8,10 @@ type WindowLog = {
   log: unknown;
 };
 
-(window as unknown as WindowLog).log = loglevel;
+if (typeof window !== 'undefined') {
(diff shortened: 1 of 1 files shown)


sedghi (approved):
Good catch

PR #1902, opened 2025-03-14
Fixes loading error when using wado-uri to load a volume

jmhmd:
### Context

In the case of loading a new volume with WADO-URI, we may not have loaded
metadata for all imageIds, as they are streaming in. The
getViewReference() call uses getClosestImageId() in its call stack, which
will error in that scenario as it tries to loop over all imageId
metadata. So only call getViewReference if necessary.

### Changes & Results

Removes unnecessary function call that errors in specific scenarios.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  

### DIFF
--- a/packages/core/src/RenderingEngine/Viewport.ts
+++ b/packages/core/src/RenderingEngine/Viewport.ts
@@ -576,14 +576,19 @@ class Viewport {
       this.addActor(actor);
     });
 
-    const prevViewPresentation = this.getViewPresentation();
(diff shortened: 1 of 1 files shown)


sedghi (approved):

sedghi:
can you push something new (dummy) to this PR so that ci runs

PR #1909, opened 2025-03-18
fix: avoid NaN Z spacing if possible

slak44:
### Context

The code that calculates Z spacing:
```typescript
    spacing =
      Math.abs(
        distanceImagePairs[numImages - 1].distance -
          distanceImagePairs[0].distance
      ) /
      (numImages - 1);
```
causes a division by 0 when `numImages === 1`.

The NaN ends up in some vtk shader and throws a syntax error.

### Changes & Results

Reuse existing fallback for spacing === 0


sedghi (approved):
Makes sense

PR #1923, opened 2025-03-20
feat(RenderingEngine): Make reset camera behavior configurable

### LINKED ISSUE #1922: [Feature Request] Make reset camera behavior configurable

Before v2.0, it was necessary to manually control the camera reset in the client code. After the v2.0 release, the camera resets itself on resize and orientation. It's very useful and helps a lot, but there are some cases when client application needs custom behavior. For example, if client application need to make viewport full screen, in previos version client application just change size of viewport and call `renderEngine.resize` and get result with full screen viewport with same zoom and pan as before resizing. In v2.0 result will be different, it will reset camera so user has to configure zoom and pan manually again.

I'd really appreciate it if client application could configure reset on resize and set orientation behavior.

Because it will save a lot of user's time not to configure camera manually every time user need to resize viewports.

KlimMixer:
### Context

Closes #1922

### Changes & Results

Add configuration for `resetCameraForResizez` so client application can configure behavior how `RenderEngine.resize` should reset camera.
Add configuration for `setOrientation` so client application can decide whether to reset camera on orientation change or not.
All default behaviors are kept the same as v2.0 default behaviors, so clients that don't need them won't be affected.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] "OS: Windows 11" 
- [x] "Node version: v22.14.0" 
- [x] "Browser: Chrome 134.0.6998.89"
  

### DIFF
--- a/packages/core/src/RenderingEngine/BaseVolumeViewport.ts
+++ b/packages/core/src/RenderingEngine/BaseVolumeViewport.ts
@@ -1223,10 +1223,12 @@ abstract class BaseVolumeViewport extends Viewport {
    *
    * @param orientation - The orientation to set the camera to.
    * @param immediate - Whether the `Viewport` should be rendered as soon as the camera is set.
+   * @param resetCamera - Whether the 'Camera' should be reset.
    */
   public setOrientation(
     _orientation: OrientationAxis | OrientationVectors,
-    _immediate = true
+    _immediate = true,
+    _resetCamera = true
   ): void {
     console.warn('Method "setOrientation" needs implementation');
   }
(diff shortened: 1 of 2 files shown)


sedghi:
I'm not really on board with this change.

It works fine in OHIF, so how we handle double-clicking is likely an app-level concern.

In OHIF, we're probably getting the `viewPresentation` via `viewport.getViewPresentation`, and then applying it after a double-click to preserve it.

We should not mix the resize with camera stuff

https://github.com/user-attachments/assets/278aa1f2-367d-433c-8998-8b8199527f2f

KlimMixer:
Yes, it makes sense to handle camera settings at the application level. However, there are already mix of reset camera call at viewport resize. And because of that, I thought it might be useful to make it configurable.

PR #1928, opened 2025-03-21
feat(segmentation): Simplify preview handling and remove unused code

sedghi:
This pull request introduces significant enhancements and new features to the `ONNXSegmentationController` class and related files. The most important changes include the addition of auto-segmentation mode, updates to the viewport orientation, and various improvements in segmentation handling.

Enhancements and new features:

* [`packages/ai/src/ONNXSegmentationController.ts`](diffhunk://#diff-c3076d8d8d5836e6283cd84b72c6aacce71555a7a4b2a1a0873dd043c0b8827dL220-R222): Added auto-segmentation mode with properties and methods to enable/disable it, set the number of random points, and handle segmentation in non-acquisition planes. [[1]](diffhunk://#diff-c3076d8d8d5836e6283cd84b72c6aacce71555a7a4b2a1a0873dd043c0b8827dL220-R222) [[2]](diffhunk://#diff-c3076d8d8d5836e6283cd84b72c6aacce71555a7a4b2a1a0873dd043c0b8827dR251-R256) [[3]](diffhunk://#diff-c3076d8d8d5836e6283cd84b72c6aacce71555a7a4b2a1a0873dd043c0b8827dR288-R289) [[4]](diffhunk://#diff-c3076d8d8d5836e6283cd84b72c6aacce71555a7a4b2a1a0873dd043c0b8827dR299-R301) [[5]](diffhunk://#diff-c3076d8d8d5836e6283cd84b72c6aacce71555a7a4b2a1a0873dd043c0b8827dR318-R357) [[6]](diffhunk://#diff-c3076d8d8d5836e6283cd84b72c6aacce71555a7a4b2a1a0873dd043c0b8827dR657-R761) [[7]](diffhunk://#diff-c3076d8d8d5836e6283cd84b72c6aacce71555a7a4b2a1a0873dd043c0b8827dR1078-R1103) [[8]](diffhunk://#diff-c3076d8d8d5836e6283cd84b72c6aacce71555a7a4b2a1a0873dd043c0b8827dL931-R1122) [[9]](diffhunk://#diff-c3076d8d8d5836e6283cd84b72c6aacce71555a7a4b2a1a0873dd043c0b8827dL1098-R1284) [[10]](diffhunk://#diff-c3076d8d8d5836e6283cd84b72c6aacce71555a7a4b2a1a0873dd043c0b8827dR1295-R1301) [[11]](diffhunk://#diff-c3076d8d8d5836e6283cd84b72c6aacce71555a7a4b2a1a0873dd043c0b8827dR1310-R1312)

* [`packages/ai/examples/autoSegmentAI/index.ts`](diffhunk://#diff-e3cc7566f3c2db6fc4e653816f71215ce738b5b4b91c0c08aff7143ed096d457R1-R265): Implemented a complete example for the new auto-segmentation feature, including UI elements for brush size, toggle enable, and clear actions, as well as the setup for segmentation models and tools.

Viewport updates:

* [`packages/ai/examples/SAMClientSide/index.ts`](diffhunk://#diff-e5ca9c7162c66cefa9485db2ea0be810d5370aa28b04af1db35e25bdd328f3f8L256-R256): Changed the default orientation for orthographic viewports from `SAGITTAL` to `AXIAL`.

Cache improvements:

* [`packages/core/src/cache/cache.ts`](diffhunk://#diff-8786ee7fb52d86a98d90b77085582043df8fbb5cfece302d5efd17a0354e2b6cR39-R40): Added a reverse lookup map for referenced image IDs to image IDs to improve segmentation handling.

Import updates:

* [`packages/ai/src/ONNXSegmentationController.ts`](diffhunk://#diff-c3076d8d8d5836e6283cd84b72c6aacce71555a7a4b2a1a0873dd043c0b8827dR7): Added `cache` to the imports from `@cornerstonejs/core`.


packages/core/src/utilities/historyMemo/index.ts
```diff
@@ -64,6 +64,7 @@ export class HistoryMemo {
    */
   public undo(items = 1) {
     while (items > 0 && this.undoAvailable > 0) {
+      console.log('running udno');
```
wayfarer3130, line 67: Can you start using the log framework?  Having debug level logs for this sort of thing can be quite useful to be able to turn on to debug things.

packages/core/src/cache/cache.ts
```diff
@@ -1080,6 +1080,19 @@ class Cache {
     this._geometryCacheSize -= decrement;
   };
 
+  /**
+   * Find the image that has the referenced ImageId
+   */
+  public getImageByReferencedImageId = (
```
wayfarer3130, line 1086: Probably you should maintain a map for referenced image id to cached image.  Otherwise this will get really slow.

packages/core/src/utilities/historyMemo/index.ts
```diff
@@ -73,6 +73,12 @@ export class HistoryMemo {
     }
   }
 
+  public undoIf(condition: (item: Memo) => boolean) {
+    if (condition(this.ring[this.position])) {
```
wayfarer3130, line 77: You need to check if there is anything at the given position - number of elements in the undo ring should be at least 1.

packages/core/src/utilities/historyMemo/index.ts
```diff
@@ -73,6 +73,12 @@ export class HistoryMemo {
     }
   }
 
+  public undoIf(condition: (item: Memo) => boolean) {
```
wayfarer3130, line 76: Please return true if the undo was done, and false otherwise.

packages/tools/src/enums/Events.ts
```diff
@@ -179,6 +179,14 @@ enum Events {
    */
   SEGMENTATION_DATA_MODIFIED = 'CORNERSTONE_TOOLS_SEGMENTATION_DATA_MODIFIED',
 
+  /**
+   * Triggers on the eventTarget when a labelmap undo operation is performed.
+   * Make use of {@link EventTypes.LabelmapUndoEventType | Labelmap Undo Event Type}
+   * for typing your event listeners for this labelmap undo event, and see what
+   * event detail is included in {@link EventTypes.LabelmapUndoEventDetail | Labelmap Undo Event Detail}
+   */
+  LABELMAP_UNDO = 'CORNERSTONE_TOOLS_LABELMAP_UNDO',
```
wayfarer3130, line 188: Can we call it HISTORY_UNREDO or something like that so it fires on general history undo/redo?
Should there be an event for things getting added to the history?

packages/tools/src/tools/segmentation/BrushTool.ts
```diff
@@ -153,8 +157,19 @@ class BrushTool extends LabelmapBaseTool {
...
+    // Add event listener for labelmap undo
+    this._labelmapUndoHandler = this._labelmapUndoHandler.bind(this);
+    eventTarget.addEventListener(
+      Events.LABELMAP_UNDO,
+      this._labelmapUndoHandler
+    );
   }
 
+  private _labelmapUndoHandler = () => {
+    this._previewData.isDrag = true;
```
wayfarer3130, line 170: It should compute this based on the event data - so that if the isDrag had been true, it will be again, and if it was false previously it will stay false.  That way it works for undo and redo regardless of whether the before/after values were with drag or not.

wayfarer3130:
There is a bug in undo/redo which marks everything redone as preview, even if a subsequent redo is done.
Try this - in toolHistory create/accept 3 segmentation areas
Undo 3 times
Redo 3 times
You will see that all 3 redone segmentations are highlighted
Also, delete annotation on toolHistory seems broken - it should remove what is currently in view, and the ability to redo it should exist.

packages/ai/src/ONNXSegmentationController.ts
```diff
@@ -601,6 +644,74 @@ export default class ONNXSegmentationController {
...
+              const worldCoords = utilities.imageToWorldCoords(
+                nextLabelmapImage.imageId,
+                [pointIJK[0], pointIJK[1]]
+              );
+              pointLists.push(worldCoords);
+            }
+          });
+
+          // Pick random points from the pointLists
+          this.randomPoints = pointLists
```
wayfarer3130, line 708: This seems a little too random - it can end up picking points really close to the edge.  It also feels like it should calculate some number of possible points based on how many are available.  Finally, adding negative points outside a reasonable distance would be useful as it seems pretty aggressive about selecting points.

packages/tools/src/tools/segmentation/LabelmapBaseTool.ts
```diff
@@ -407,10 +460,8 @@ export default class LabelmapBaseTool extends BaseTool {
       this.getOperationData(element),
       StrategyCallbacks.AcceptPreview
     );
-    this._previewData.isDrag = false;
     this._previewData.preview = null;
-    // Store the edit memo too
-    this.doneEditMemo();
+    this._previewData.isDrag = false;
```
wayfarer3130, line 464: This should reset the isDrag to whatever it was before this - which could still be isDrag depending on how often memos are stored.  That is definitely on mouse up - you can test the wrong isDrag setting by checking the un/re do functionality for selecting a region, moue up, selecting more region, mouse up, removing a region, mouse up, then accept.

packages/tools/src/utilities/segmentation/createLabelmapMemo.ts
```diff
@@ -47,6 +44,16 @@ export function restoreMemo(isUndo?: boolean) {
   });
   const slices = useVoxelManager.getArrayOfModifiedSlices();
   triggerSegmentationDataModified(this.segmentationId, slices);
+
+  // Trigger the LABELMAP_UNDO event
+  eventTarget.dispatchEvent(
```
wayfarer3130, line 49: Can we add that to the history class instead, and include the memo itself in the detail rather than trying to make it segmentation specific?  It should also say whether it is undo or redo.

wayfarer3130 (changes requested):
A couple of fixes still for 
remembering what state the preview/regular view is in
Ability to remove existing segmentation by starting outside the segmentation view
Sending undo/redo notifications for all undo/redo events

packages/tools/src/tools/segmentation/LabelmapBaseTool.ts
```diff
@@ -122,8 +129,50 @@ export default class LabelmapBaseTool extends BaseTool {
...
+    this.centerSegmentIndexInfo = {
+      segmentIndex: null,
+      hasSegmentIndex: false,
+      hasPreviewIndex: false,
+      changedIndices: [],
+    };
+
+    // Add event listener for history undo
+    this._historyUndoHandler = this._historyUndoHandler.bind(this);
+    eventTarget.addEventListener(Events.HISTORY_UNDO, this._historyUndoHandler);
```
wayfarer3130, line 149: You need to listen to both undo and redo, or only have a single history event.

packages/core/src/utilities/historyMemo/index.ts
wayfarer3130, line 1: Why isn't this utility firing the undo notification?  It seems it should happen on the historyMemo for all undo/redo operations.

packages/tools/src/utilities/segmentation/createLabelmapMemo.ts
```diff
@@ -47,6 +46,18 @@ export function restoreMemo(isUndo?: boolean) {
   });
   const slices = useVoxelManager.getArrayOfModifiedSlices();
   triggerSegmentationDataModified(this.segmentationId, slices);
+
+  // Trigger the HISTORY_UNDO event
+  eventTarget.dispatchEvent(
+    new CustomEvent(Events.HISTORY_UNDO, {
```
wayfarer3130, line 52: This should probably fire generically on the history undo/redo after the restoreMemo call.  Otherwise it can't be used for listening to other types of undo/redo changes.  It should also include the memo itself rather than including the segmentationId so that it is a generic event type.

wayfarer3130: If you want to leave it here for now, moving it later, at least add the memo object itself rather than the segmentationId so that it has the right object design to work.

wayfarer3130:
The remove preview/segmentation by starting outside but overlapped onto an existing segmentation isn't working.
If you redo after accepting it is always in the preview colour AND it messes up the final segmentation indexes.
Steps:
1. Create a preview section in index 1
2. Create a preview section in indexes 2,3
3. Undo 3 times
4. Redo 3 times

Expected result is that all the redos will be in the final segment index that they had been in

wayfarer3130 (approved):
Just the one small bug to complete for undo while previewing.

sedghi:
Fixed that bug + 2 more bugs

PR #1940, opened 2025-03-26
fix CircleROI tool for wsi

mbellehumeur:
### Context

CircleROI and other ROI tools were not generating stats for wsi viewport.  

### Changes & Results

This PR fixes the CircleROI, RectangleROI, EllipseROI and FreehandROI tool for whole slide (wsi) viewports by skipping statistics calculation that require the voxel manager. 
Key changes include:

Changing wsiViewport so that getScalarData returns an empty CanvasScalarData object instead of null.

Wrapping the voxel manager-based points collection of each tool  in a condition that bypasses the process when the  scalarData is empty.

### DIFF
--- a/packages/core/src/RenderingEngine/WSIViewport.ts
+++ b/packages/core/src/RenderingEngine/WSIViewport.ts
@@ -231,8 +231,6 @@ class WSIViewport extends Viewport {
 
   public getScalarData() {
     const scalarData = null as PixelDataTypedArray;
-    //scalarData.length = 0;
-    //scarlarData = [];
     return scalarData;
   }
 
--- a/packages/core/src/RenderingEngine/WSIViewport.ts
+++ b/packages/core/src/RenderingEngine/WSIViewport.ts
@@ -10,6 +10,7 @@ import type {
   ViewportInput,
   BoundsIJK,
   CPUImageData,
+  PixelDataTypedArray,
 } from '../types';
 import uuidv4 from '../utilities/uuidv4';
 import * as metaData from '../metaData';
@@ -228,8 +229,11 @@ class WSIViewport extends Viewport {
     this.setProperties({});
   }
 
-  protected getScalarData() {
-    return null;
+  public getScalarData() {
+    const scalarData = null as PixelDataTypedArray;
+    //scalarData.length = 0;
+    //scarlarData = [];
+    return scalarData;
   }
 
   public getImageData(): CPUIImageData {
@@ -272,7 +276,7 @@ class WSIViewport extends Viewport {
       preScale: {
         scaled: false,
       },
-      scalarData: this.getScalarData(),
+      scalarData: null,
       imageData,
       // It is for the annotations to work, since all of them work on voxelManager and not on scalarData now
       voxelManager: {
--- a/packages/core/src/RenderingEngine/WSIViewport.ts
+++ b/packages/core/src/RenderingEngine/WSIViewport.ts
@@ -233,15 +233,7 @@ class WSIViewport extends Viewport {
   public resetProperties() {
     this.setProperties({});
   }
-  /*
-  protected getScalarData() {
-    // Return an empty CanvasScalarData object
-    const emptyData = new Uint8ClampedArray() as CanvasScalarData;
-    emptyData.getRange = () => [0, 255];
-    emptyData.frameNumber = -1;
-    return emptyData;
-  }
- */
+
   protected _getScalarData() {
     // Return an empty CanvasScalarData object
     const emptyData = new Uint8ClampedArray() as CanvasScalarData;
--- a/packages/core/src/RenderingEngine/WSIViewport.ts
+++ b/packages/core/src/RenderingEngine/WSIViewport.ts
@@ -615,12 +615,18 @@ class WSIViewport extends Viewport {
   protected canvasToIndex = (canvasPos: Point2): Point2 => {
     const transform = this.getTransform();
     transform.invert();
-    return transform.transformPoint(canvasPos);
+    return transform.transformPoint(
+      canvasPos.map((it) => it * devicePixelRatio) as Point2
+    );
+    // return transform.transformPoint(canvasPos);
   };
 
   protected indexToCanvas = (indexPos: Point2): Point2 => {
     const transform = this.getTransform();
-    return transform.transformPoint(indexPos);
+    return transform
+      .transformPoint(indexPos)
+      .map((it) => it / devicePixelRatio) as Point2;
+    // return transform.transformPoint(indexPos);
   };
 
   /** This can be implemented later when multi-slice WSI is supported */
--- a/packages/core/src/RenderingEngine/WSIViewport.ts
+++ b/packages/core/src/RenderingEngine/WSIViewport.ts
@@ -293,6 +293,7 @@ class WSIViewport extends Viewport {
       scalarData: this._getScalarData(),
       imageData,
       // It is for the annotations to work, since all of them work on voxelManager and not on scalarData now
+      /*
       voxelManager: {
         forEach: (
           callback: (args: {
@@ -316,6 +317,7 @@ class WSIViewport extends Viewport {
           });
         },
       },
+      */
     };
 
     // @ts-expect-error we need to fully migrate the voxelManager to the new system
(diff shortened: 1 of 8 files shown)


packages/tools/src/tools/annotation/CircleROITool.ts
```diff
@@ -919,7 +919,10 @@ class CircleROITool extends AnnotationTool {
       // Check if one of the indexes are inside the volume, this then gives us
       // Some area to do stats over.
 
-      if (this._isInsideVolume(pos1Index, pos2Index, dimensions)) {
+      if (
+        this._isInsideVolume(pos1Index, pos2Index, dimensions) ||
```
wayfarer3130, line 923: Can you make the isInsideVolume work for WSI as well?  I'd rather not have a viewport.type==='whileSlide' anywhere inside the tools.

wayfarer3130: It might be necessary to improve the WSIViewport with better mapping functions, or perhaps the transformWorldToIndex mapping functions.

packages/tools/src/tools/annotation/CircleROITool.ts
```diff
@@ -982,17 +985,19 @@ class CircleROITool extends AnnotationTool {
...
-            isInObject: (pointLPS) =>
-              pointInEllipse(ellipseObj, pointLPS, { fast: true }),
-            boundsIJK,
-            imageData,
-            returnPoints: this.configuration.storePointData,
-          }
-        );
-
+        let pointsInShape;
+        if (viewport.type !== 'wholeSlide') {
```
wayfarer3130, line 989: What about 
if (voxelManager) {
...
}
although I'm not quite sure how this calculates or doesn't calculate the statistics for perimeter and area.  Maybe those are calculated in another way?

mbellehumeur: wsiviweport returns null for imageData.getScalar()

packages/tools/src/tools/annotation/CircleROITool.ts
mbellehumeur: yes, wsiviewport worldToCanvas returns negative values.

wayfarer3130 (approved):

wayfarer3130 (approved):

wayfarer3130:
Can you ensure the tests pas[s - I think those are real issues

packages/tools/src/tools/annotation/CircleROITool.ts
wayfarer3130: That impacts too much right now - @pedrokohler is working on a fix for that separately, so lets leave this as is for now.

packages/core/src/RenderingEngine/WSIViewport.ts
```diff
@@ -244,7 +254,7 @@ class WSIViewport extends Viewport {
       getDirection: () => metadata.direction,
       getDimensions: () => metadata.dimensions,
       getRange: () => [0, 255],
-      getScalarData: () => this.getScalarData(),
```
wayfarer3130, line 247: Can you add a viewReference to the getter as an optional?  I know we don't pass it in yet, but we need to start thinking about it as the getScalarData for whole slide imaging won't be able to return the entire scalar data.

packages/core/src/RenderingEngine/WSIViewport.ts
```diff
@@ -228,8 +234,12 @@ class WSIViewport extends Viewport {
     this.setProperties({});
   }
 
-  protected getScalarData() {
-    return null;
+  protected _getScalarData() {
+    // Return an empty CanvasScalarData object
+    const emptyData = new Uint8ClampedArray() as CanvasScalarData;
+    emptyData.getRange = () => [0, 255];
+    emptyData.frameNumber = -1;
```
wayfarer3130, line 241: Can you add an emptyData.region that is a rectangle of the enclosing region, and a scale parameter?  The region should be zero sized for now, and the scale can be 1, but again I'd like to start thinking about the future.

packages/tools/src/tools/annotation/CircleROITool.ts
```diff
@@ -1028,6 +1030,14 @@ class CircleROITool extends AnnotationTool {
   };
 
   _isInsideVolume = (index1, index2, dimensions) => {
+    // for wsiViewport
+    if (index1[1] < 0) {
```
wayfarer3130, line 1034: Can you put this fix into the WSIViewport itself?

packages/core/src/RenderingEngine/WSIViewport.ts
```diff
@@ -615,12 +627,16 @@ class WSIViewport extends Viewport {
   protected canvasToIndex = (canvasPos: Point2): Point2 => {
     const transform = this.getTransform();
     transform.invert();
-    return transform.transformPoint(canvasPos);
+    return transform.transformPoint(
```
wayfarer3130, line 630: Can you use Math.abs on the result here - I think that will get us close to the right value, or on initial setup, figure out which attribute is negative and store a negative multiplier for it.

packages/core/src/RenderingEngine/WSIViewport.ts
```diff
@@ -615,12 +627,16 @@ class WSIViewport extends Viewport {
...
+      canvasPos.map((it) => it * devicePixelRatio) as Point2
+    );
   };
 
   protected indexToCanvas = (indexPos: Point2): Point2 => {
     const transform = this.getTransform();
-    return transform.transformPoint(indexPos);
+    return transform
```
wayfarer3130, line 637: Will need to convert back to negative values, so if you on initial setup store an index of which values are negative and multiple by that vector you should get the right sign.

packages/tools/src/tools/annotation/EllipticalROITool.ts
```diff
@@ -1248,7 +1250,7 @@ function defaultGetTextLines(data, targetId): string[] {
     textLines.push(`Mean: ${csUtils.roundNumber(mean)} ${modalityUnit}`);
   }
 
-  if (max) {
+  if (max && isFinite(max)) {
```
wayfarer3130, line 1253: Please use the new AnnotationTool.isNumber here

packages/tools/src/tools/annotation/PlanarFreehandROITool.ts
```diff
@@ -1085,7 +1086,7 @@ function defaultGetTextLines(data, targetId): string[] {
     textLines.push(`Mean: ${csUtils.roundNumber(mean)} ${modalityUnit}`);
   }
 
-  if (Number.isFinite(max)) {
+  if (Number.isFinite(max) && !isNaN(max)) {
```
wayfarer3130, line 1089: Again, please call AnnotationTool.isNumber here

wayfarer3130 (commented):
Couple of small remaining changes, hopefully merge this today.

packages/tools/src/tools/base/AnnotationTool.ts
```diff
@@ -732,6 +732,17 @@ abstract class AnnotationTool extends AnnotationDisplayTool {
       viewport,
     };
   }
+
+  /**
+   * @param n - array of numbers or a simple number
+   * @returns True if n or the first element of n is finite and not NaN
+   */
+  public static isNumber(n: number[] | number): boolean {
```
sedghi, line 740: there is already a isNumber in packages/core/src/utilities/isEqual.ts , adapt that for your use

packages/tools/src/tools/annotation/RectangleROITool.ts
```diff
@@ -968,6 +968,14 @@ class RectangleROITool extends AnnotationTool {
   };
 
   _isInsideVolume = (index1, index2, dimensions) => {
+    // for wsiViewport
+    if (index1[1] < 0) {
```
sedghi, line 972: This doesn't look good. Could you please check if the viewport type is WSI before proceeding? I imagine this might disrupt other types of viewports.

sedghi (commented):
please see my comments

mbellehumeur:
Replaced with https://github.com/cornerstonejs/cornerstone3D/pull/2107.

PR #1952, opened 2025-03-30
refactor(Tools): Cleanup code for drawingSvg, make 'data-id' attribut…

siclesia:
…e consistant

### Context

When adding svg annotation, there is a custom attribute 'data-id' intended for automation test.
the feature was not consistently implemented ( most of the time missing ) and is somehow redundant with internal unique id.

This PR remove custom value for data-id and uses the internal uid.

### Changes & Results

only impacts 'data-id' attribute in SVG elements.

### Testing

use stack annotation tools /  https://www.cornerstonejs.org/live-examples/stackannotationtools

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] "OS: Ubuntu 24
- [x] "Node version: 22
- [x] "Browser: firefox

### DIFF
--- a/packages/tools/src/drawingSvg/_draw.ts
+++ b/packages/tools/src/drawingSvg/_draw.ts
@@ -1,0 +1,34 @@
+import type { SVGDrawingHelper } from '../types';
+import setAttributesIfNecessary from './setAttributesIfNecessary';
+import setNewAttributesIfValid from './setNewAttributesIfValid';
+
+function _draw(
+  what: 'circle' | 'line' | 'ellipse' | 'rect' | 'path' | 'polyline' | 'g',
+  svgDrawingHelper: SVGDrawingHelper,
+  annotationUID: string,
+  drawingId: string,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  attributes: Record<string, any>
+): {
+  element: SVGGElement;
+  isNew: boolean;
+} {
+  // variable for the namespace
+  const svgns = 'http://www.w3.org/2000/svg';
+  const dataId = `${what}-${annotationUID}-${drawingId}`;
+  const existingElement = svgDrawingHelper.getSvgNode(dataId);
+
+  if (existingElement) {
+    setAttributesIfNecessary(attributes, existingElement);
+    svgDrawingHelper.setNodeTouched(dataId);
+    return { element: existingElement, isNew: false };
+  } else {
+    const newElement = document.createElementNS(svgns, what);
+    newElement.setAttribute('data-id', dataId);
+    setNewAttributesIfValid(attributes, newElement);
+    svgDrawingHelper.appendNode(newElement, dataId);
+    return { element: newElement, isNew: true };
+  }
+}
+
+export default _draw;
--- a/packages/tools/src/drawingSvg/drawTextBox.ts
+++ b/packages/tools/src/drawingSvg/drawTextBox.ts
@@ -1,8 +1,7 @@
 import type { Types } from '@cornerstonejs/core';
 import type { SVGDrawingHelper } from '../types';
-
-import _getHash from './_getHash';
 import setAttributesIfNecessary from './setAttributesIfNecessary';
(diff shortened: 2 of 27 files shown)


packages/tools/src/drawingSvg/_draw.ts
```diff
@@ -0,0 +1,34 @@
...
+  drawingId: string,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  attributes: Record<string, any>
+): {
+  element: SVGGElement;
+  isNew: boolean;
+} {
+  // variable for the namespace
+  const svgns = 'http://www.w3.org/2000/svg';
+  const dataId = `${what}-${annotationUID}-${drawingId}`;
```
sedghi, line 18: Could we please keep the order as `${annotationUID}::${drawingElementType}::${nodeUID}`? I'm just not sure what others might be using it for, and I'd rather be safe.

packages/tools/src/drawingSvg/_draw.ts
```diff
@@ -0,0 +1,34 @@
...
+import setNewAttributesIfValid from './setNewAttributesIfValid';
+
+function _draw(
+  what: 'circle' | 'line' | 'ellipse' | 'rect' | 'path' | 'polyline' | 'g',
+  svgDrawingHelper: SVGDrawingHelper,
+  annotationUID: string,
+  drawingId: string,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  attributes: Record<string, any>
+): {
```
sedghi, line 12: Can you use an object instead of positional parameters?

packages/tools/src/drawingSvg/setAttributesIfNecessary.ts
```diff
@@ -1,13 +1,26 @@
...
-      svgNode.setAttribute(key, newValue);
-    }
-  });
+/**
+ *
+ * @param attributes
+ * @param svgNode
+ * @returns true if attributes were modified
+ */
+export function setAttributesIfNecessary(attributes, svgNode): boolean {
```
sedghi, line 7: I'm not sure where the returned boolean is used. Why does this function return a boolean?

packages/tools/src/drawingSvg/drawTextBox.ts
```diff
@@ -57,80 +56,55 @@ function _drawTextGroup(
 ): SVGRect {
   const { padding, color, fontFamily, fontSize, background } = options;
 
-  let textGroupBoundingBox;
   const [x, y] = [position[0] + padding, position[1] + padding];
```
sedghi, line 59: Why is this file being modified so much more often than the others? I thought this pull request was just a refactoring.

sedghi (commented):
please see my comments

packages/tools/src/drawingSvg/_draw.ts
siclesia: yes, you are right.

( note: there is a risk of breaking change. since data-id logic was not consistent, some user may be impacted by change in value ).

packages/tools/src/drawingSvg/_draw.ts
siclesia: over my dead body ;-). this is typically over engineering in my opinion.

( we are talking of 5 positional arguments that fit into a single line of text in most cases )

packages/tools/src/drawingSvg/setAttributesIfNecessary.ts
siclesia: you are right.

this is over-engineering on my side.

packages/tools/src/drawingSvg/drawTextBox.ts
siclesia: this is because there is extra logic after svg element set-up ( call to _draw method ) in order to set-up every line of text.

the refactor change the 2 branch code ( one for creating, one for updating ) into a single branch.

this is pure refactor. no functional change.

Issue #1972, opened 2025-04-07
[Feature Request] Customise path to wasm files when initialising `dicom-image-loader`


jmannau:
### What feature or change would you like to see made?

This is more of an RFC.

Angular doesn't work well with loading wasm files using `new URL('@cornerstonejs/codec-penjpeg/decodewasm',  import.meta.url);`. There are open issues in both Angular CLI project and Cornerstone projects regarding this.

For Angular, it is common to copy assets like this using `angular.json` configuration in to the output path (https://angular.dev/reference/configs/workspace-config#assets-configuration).  This is commonly used for loading web font and other static assets.

In order to workaround the issue loading WASM in the web worker automatically, I propose adding an option to the `dicom-image-loader` `init` method that will allow setting a path to the required wasm files.

ie:

```js
cornerstoneDICOMImageLoader.init({
  decodeConfig: {
    wasmPaths: {
      "codec-charls": "./codec-charls.wasm",
      "codec-openjph": "./codec-openjph.wasm",
      "codec-openjpeg": "./codec-openjpeg.wasm",
      "codec-libjpeg-turbo-8bit": "./codec-libjpeg-turbo-8bit.wasm",
    },
  },
});
```

These paths will be passed to the web worker and can be passed to the relevant decoder.  If present, the wasm can be loaded from the specified path. If the path is not present, it will fallback to the current implementation.

This could be a non breaking change.

I am happy to create a PR if you believe this is worthwhile.

### Why should we prioritize this feature?

This will allow Angular users to load the wasm correctly.

mhmmd7777:
Hi @jmannau 
I am using AnalogJS and Vite in Angular. The WASM files are located in the assets folder, but Cornerstone still does not use them and instead decodes images using decodeImageFrameWorker.js. Can you help me?

wayfarer3130:
Closed based on the CS3D change in https://github.com/cornerstonejs/cornerstone3D/pull/2826

PR #1982, opened 2025-04-08
feat(dicom-image-loader): Add ability to customise path to web worker and codec wasm files

jmannau:
### Context

Currently, the dicom-image-loader relies on a bundler like Webpack or Vite to be correctly configured and identity the web worker and wasm files created.  This doesn't work well for Angular projects where customising the build tools is significantly more difficult.

This PR addresses https://github.com/cornerstonejs/cornerstone3D/issues/1972

This PR adds the ability for users to override the default `webWorkerFn` where the worker is loaded and customise the location of the OpenJPH, OpenJPEG, Charls and LibTurboJpeg wasm files that are used in the decoder. 

These can be set when the dicomImageLoader is initialised.

for instance:

```typescript
import { init } from '@cornerstonejs/dicom-image-loader';

init({
  maxWebWorkers: 1,
  webWorkerUrl: () => {
    return new Worker(
      // use the Angular build system to create a work worker where the `decodeImageFrameWorker.js` is manually included
      new URL('../cornerstone-image-decoder.worker.ts', import.meta.url)
    );
  },
  decodeConfig: {
    // Use `angular.json` to copy the required wasm files to the output `assets` directory. 
    // Set the path to where the library should load the wasm files from.
    wasmUrlCodecCharls: './assets/charlswasm_decode.wasm',
    wasmUrlCodecLibjpegTurbo8bit: './assets/libjpegturbowasm_decode.wasm',
    wasmUrlCodecOpenJpeg: './assets/openjpegwasm_decode.wasm',
    wasmUrlCodecOpenJph: './assets/openjphjs.wasm',
  },
});
```

If these aren't passed, the library will use the default loaders.

### Testing

I have created a demo angular project that uses a dev build of this library to demonstrate that it works.  You can find the demo angular project here - https://github.com/jmannau/angular-c3d

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:


jmannau:
A couple of things that I would like your help with.  I would like to verify that the wasm files decode appropriate files.  Do you have sample files that we can include in the angular project to verify the wasm loads correctly?

I would also like to write some tests for these changes to.

jmannau:
I have added tests in the angular repo to test the custom wasm paths for the 4 wasm decoders.  These tests are here https://github.com/jmannau/angular-c3d/blob/main/src/tests/dicom-image-loader.spec.ts

jmannau:
Just checking to see if you have any comments/thoughts on this PR?

sedghi:
I need to test this in many different settings, but I haven't had time yet, unfortunately. I need to test all the Vite, Webpack, etc. configurations to see if they can handle this.

daemon29:
Hi @jmannau, I follow your solution, and it works, thanks a lot. I know that there are a lot of people having the same problem, you saved us all.

packages/dicomImageLoader/src/decodeImageFrameWorker.js
```diff
@@ -331,8 +384,14 @@ function createDestinationImage(
   };
 }
 
-/** Scales the image frame, updating the frame in place with a new scaled
- * version of it (in place modification)
+/**
+ *
+ * @description Scales the image frame, updating the frame in place with a new
+ * scaled version of it (in place modification)
+ * @param {import("@cornerstonejs/core").Types.IImageFrame} imageFrame
```
wayfarer3130: Can you rename the file to decodeImageFrameWorker.ts so you can directly declare at least some of the relevant types?

packages/dicomImageLoader/src/types/LoaderDecodeOptions.ts
```diff
@@ -1,3 +1,55 @@
...
+   * static assets directory.  Then set the the correct paths to where
+   * cornerstone should load the files.
+   *
+   * This should point to the export `@cornerstonejs/codec-charls/decodewasm`
+   * that resolves to `@cornerstonejs/codec-charls/dist/charlswasm_decode.wasm`.
+   * Copy this file to your static assets directory and set the path to it.
+   *
+   * @example `wasmUrlCodecCharls: '/static/charlswasm_decode.wasm'`
+   */
+  wasmUrlCodecCharls?: string;
```
wayfarer3130: I'd rather treat these all the same and have it be a standard object, eg:
charls: {
   wasmPath?: string;
   modulePath?: string;
}

That way there is some chance of being able to centralize the loading code, and perhaps even re-use it for other things.

packages/docs/docs/concepts/streaming-image-volume/streaming.md
wayfarer3130, line 1: Please exclude unnecessary reformats as they make the review significantly harder

jmannau: fixed and updated

packages/dicomImageLoader/src/decodeImageFrameWorker.js
jmannau: done and updated

packages/dicomImageLoader/src/types/LoaderDecodeOptions.ts
jmannau: I'm not sure I fully understand.  Is your preference that the configuration is 
```ts
  charls?: {
    wasmPath: string
  },
  openJph?: {
    wasmPath: string
  },
  openJpeg?: {
    wasmPath: string
  },
  libJpegTurbo?: {
    wasmPath: string
  }
```

I'm not sure what the `modulePath?:string` would be used for.

wayfarer3130: The modulePath would be used to also dynamically load the typescript file for charls etc, but that isn't needed right now as those are recreated from the codec library assuming a fixed interface.  Ideally they would have used the standard exported version, but that isn't something you need to resolve.

Please update the LoaderDecodeOptions with a new interface containing the wasmPath, and mapping from the transfer syntax uid, eg
LoaderDecodePath {
   [tsuid:string]: TsuidSpecificOptions,
}

where TsuidSpecificOptions is an interface with just wasmPath in it as an optional.

packages/dicomImageLoader/src/shared/decoders/decodeHTJ2K.ts
```diff
@@ -34,7 +37,17 @@ function calculateSizeAtDecompositionLevel(
...
+/**
+ *
+ * @param decodeConfig
+ * @param wasmUrlCodecOpenJph Optional path to load the OpenJPH WASM codec from.
+ * If not given, it will default to using the default `openjphWasm`
+ * @returns
+ */
+export function initialize(
+  decodeConfig?: LoaderDecodeOptions,
+  wasmUrlCodecOpenJph?: string
```
wayfarer3130: Please pass in the generic TsuidSpecificOptions

packages/dicomImageLoader/src/decodeImageFrameWorker.ts
```diff
@@ -353,16 +378,16 @@ function scaleImageFrame(imageFrame, targetBuffer, TypedArrayConstructor) {
...
-  transferSyntax,
-  pixelData,
-  decodeConfig,
-  options,
-  callbackFn
-) {
+  imageFrame: CoreTypes.IImageFrame,
+  transferSyntax: string,
+  pixelData: ByteArray,
+  decodeConfig: LoaderDecodeOptions,
```
wayfarer3130, line 384: You can read the TsuidConfigOptions here for the specific transferSyntaxUID, then pass it in generally below, eg
const tsuidConfigOptions = decodeConfig?.[transferSyntax]
Then other options can be passed in the future.

Thessi:
@jmannau @wayfarer3130 what's the status here? Our Angular builds have been broken for a few months at this point. Maybe someone else could take this over?

jmannau:
Have been flat out - let me see what I can do now.  I have just merged main

packages/dicomImageLoader/src/types/LoaderDecodeOptions.ts
jmannau: How do you think we handle cases like HTJ2K that have multiple transfer syntaxes that use the same decoder? 

ie:
```
    case '3.2.840.10008.1.2.4.96':
    case '1.2.840.10008.1.2.4.201':
    case '1.2.840.10008.1.2.4.202':
    case '1.2.840.10008.1.2.4.203':
      // HTJ2K
      opts = {
        ...imageFrame,
      };

      decodePromise = decodeHTJ2K(
        pixelData,
        opts,
        decodeConfig?.wasmUrlCodecOpenJph
      );
      break;
```

Do you foresee that the `LoaderDecoderOptions` would look like:

```ts
const options: LoaderDecodeOptions = {
  '3.2.840.10008.1.2.4.96': {
    wasmPath: "/path/to/openjphjs.wasm'"
  },
  '1.2.840.10008.1.2.4.201': {
    wasmPath: "/path/to/openjphjs.wasm'"
  },
  '1.2.840.10008.1.2.4.202': {
    wasmPath: "/path/to/openjphjs.wasm'"
  },
  '1.2.840.10008.1.2.4.203': {
    wasmPath: "/path/to/openjphjs.wasm'"
  }
}
```

and similar for JPEG2000 with TSUIDs of `1.2.840.10008.1.2.4.80` and `1.2.840.10008.1.2.4.81`

I propose that the `LoaderDecodeOptions` should specify the options based on the decoder type (jpegBaseline, Jpeg2000, HTJ2K, JPEG-LS) and then `decodeImageFrameWorker` would use the appropriate configuration based in the decoder it needs.

chuber50:
Any updates on this? Looks like just one last hurdle to take

packages/dicomImageLoader/src/types/LoaderDecodeOptions.ts
jmannau: Just checking in to see if I understood your comment correctly or if you had any thoughts on my proposal

sedghi: Bill will start hopefully next week with significant more capacity on OHIF side. This will be the first thing he will work on. Sorry for the delay

jmannau: 🙏 thanks for the update.

wayfarer3130:
@jmannau  - for this PR, is the idea that the libraries should be located in a general location relative to a root path, OR is the idea that each library will be in a different location?  The first one would allow setting a generic path like is done in initDemo

window.PUBLIC_URL ||= '/';
// Use a relative library path that includes the component name
window.PUBLIC_LIB_URL ||= './${component}/';

so that the library URL would automatically take the PUBLIC_URL and combine it with the PUBLIC_LIB_URL based on the component name, and then would just append the default WASM libraries.
If that is sufficient, then I'd prefer to resolve the issue by including the PUBLIC_URL and PUBLIC_LIB_URL, and updating your sample application to just use those libraries.   

If what you really need is the ability to specify a full URL, with specific renames and changing host names, then I'm going to propose a structure like:

```
LoaderDecodeOptions: {
   sharedOptions: {
       publicUrl: '/subpath',
       libUrl: './${component}/";
       wasmUrl: 'http://cdn.com/${component}/',
   },
   jpegls: {
      parent: 'sharedOptions',
      wasmUrl: 'https://cdn.com/jpegls/jpegls.wasm',
      libUrl 'https://cdn.com/jpegls/lib.js',
   },
   '1.2.840....81': {
        parent: 'jpegls',
        encodeOptions: { lossy encode options for JPEG LS }
        decodeOptions: { ... }
   }
```

That way very specific overrides can be specified, but the general ones are all that is needed for the normal use case.

Can you let me know if this works for you?

jmannau:
Hi @wayfarer3130 the purpose is to allow a user to customise where the `.wasm` files should be loaded from.  This could be a cdn (different host(s)) or a location relative to the other app assets.  

The specific use case I am trying to solve for is including these assets in an Angular project.  The angular build tool doesn't automatically extract the `.wasm` files required by the loader.  It is possible to include these files as static assets in the angular build process, allowing them to be incorporated in the output assets (for example see https://github.com/jmannau/angular-c3d/blob/5a9a8a45f737905f4e57edf5bf221f5a525c1f02/angular.json#L23-L42)

The `PUBLIC_LIB_URL` would work - but this wouldn't work with a cdn.  In this case I will refactor the PR to follow your proposed `LoaderDecodeOptions` interface.  

In this PR I will update to support only loading the `wasmUrl`.

Cheers,

James

wayfarer3130:
Let me make a few changes to the API here to support your use case, and then if you could update the angular example on the CS3D github repo so that it works with the updates, then I think we can continue fixing this.  It will take me a few days to update the codec interface so that we don't introduce API incompatibilities in the future.

jmannau:
@wayfarer3130 did you have any luck with the api changes?

mhmmd7777:
Any updates?

wayfarer3130:
@jmannau - this change allows customizing the load path using the standard peerImport configuration.  wasm files are clearly peer files, so I think that makes sense to keep this sort of load time customization in one place.
@sedghi - I updated this to match other parts of the code for consistency, and am now ok with it.  It might need a small change on the OHIF side, but not a significant one, just to support a fallback loader.

I've added documentation on how to create custom paths - it can also be done generically by updating the peerImport to return named paths.

wayfarer3130:
I have created PRs on the other projects showing how to edit the paths directly in the build files.  I think that makes mroe sense than adding complexity to the codec loader portion itself, so I am closing this PR in favour of the PRs:
https://github.com/cornerstonejs/angular-cornerstone3d/pull/4
https://github.com/cornerstonejs/vue-cornerstone3d/pull/3
https://github.com/cornerstonejs/cornerstone3D/pull/2632
https://github.com/cornerstonejs/vite-react-cornerstone3d/pull/5

sedghi:
@jmannau Could you please check and let us know if those PRs work for you, 
@wayfarer3130 we probably should add a section on our documentation and talk about how to customize (pointing to these repos)

jmannau:
Hi @sedghi @wayfarer3130 Looks great.  We've tested this branch internally and looks good.  

Just wondering what the release plan is for this? This PR has been closed, but not merged. Will it be merged through a different PR?

sedghi:
@jmannau seems like Bill found a way without touching cs3d code to do custom path, i was wondering if you could look at the following PRs, if it works, we just document and don't change the cs3d source code

https://github.com/cornerstonejs/angular-cornerstone3d/pull/4
https://github.com/cornerstonejs/vue-cornerstone3d/pull/3
https://github.com/cornerstonejs/cornerstone3D/pull/2632
https://github.com/cornerstonejs/vite-react-cornerstone3d/pull/5

wayfarer3130:
The angular builder requires a patch to the javascript generated files, which is quite ugly, so going to try to just have a minimal change using the two argument peerImport function just for loading the wasm data/path url

wayfarer3130:
I've added a much simpler mechanism that adds a single configuration value.
This PR has some nice pieces in it, but it needs better consideration overall for how to set the configuration consistently and separately for each decode versus initialization - those aren't the same things, and the initialization has a lot of commonalities between different versions that should get factored out.

PR #1994, opened 2025-04-14
feat(Circle/RectangleStartEndThresholdTool): Added debounce mode for Start/End tools

Celian-abd:
### Context
The CircleStartEndThresholdTool and RectangleStartEndThresholdTools were having performance issues when drawing big ROIS.
We modified when and how we decide to calculate stats.

### Changes & Results
- Added throttleTimeout to configuration
- Renamed calculatePointsInsideVolume to computeStatsDuringEditing
- Added debounce for calculating stats

Now, by default the tools will use the throttle to calculate cached stats every time we move the cursor while drawing but we added a way to config the tools to calculate the stats when the cursor stops moving using a debounce. This allow to draw big ROIS without having performance issues and the tools freezing due to the cached stats calculation.

### Testing

Can be tested in circleROIStartEndThresholdWithSegmentation or rectangleROIStartEndThresholdWithSegmentation by modifying the tool config:

```
  toolGroup.addTool(CircleROIStartEndThresholdTool.toolName, {
    /* Define if the stats are calculated while drawing the annotation or at the end */
    computeStatsDuringEditing: true => false,
    showTextBox: true,
    storePointData: true,
    /* Set a custom wait time */
    throttleTimeout: 100 => custom time,
  });
```

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  


packages/tools/examples/circleROIStartEndThresholdWithSegmentation/index.ts
```diff
@@ -236,8 +235,12 @@ async function run() {
 
   // Segmentation Tools
   toolGroup.addTool(CircleROIStartEndThresholdTool.toolName, {
-    calculatePointsInsideVolume: true,
+    /* Define if the stats are calculated while drawing the annotation or at the end */
+    computeStatsDuringEditing: true,
```
sedghi, line 239: can we make this not a breaking change, seems like you are removing a public configuraiton

salimkanoun: In fact we figured that calculatePointsInsideVolume was useless, as it was equivalent to showTextBox, if you need to display stats you need to calculatePointInsideVolume.

sedghi (commented):
I still don't understand why we're removing the public API `calculatePointsInsideVolume`. Can we avoid this? I'm not comfortable breaking this API just yet.

salimkanoun:
Ok, @Celian-abd could you let this settings to avoid a breaking change

Celian-abd:
Yes, I'll change it

salimkanoun:
It is possible to get this one merged ? We would need this PR for our further developpements

sedghi (approved):

salimkanoun:
thanks !

Issue #1393, opened 2024-07-15
[Bug] StackViewport.setImageIdIndex should set this.targetImageIdIndex too


abustany:
### Describe the Bug

Calling `StackViewport.setImageIdIndex` on a stack viewport does update `this.currentImageIdIndex`, but does not update `this.targetImageIdIndex`. `StackViewport.scroll`, on the other hand, using `this.targetImageIdIndex` as a base. As a result, if you scroll with the scroll tool to eg. image 10, then call `setImageIdIndex(1)`, and then use the scroll tool again, the viewport will not scroll to image 2, but to image 11.

### Steps to Reproduce

1. Create a stack viewport
2. Use the mouse scroll tool to scroll to a given image
3. Call `setImageIdIndex` on a different image
4. Scroll again

### The current behavior

Scroll continues from the last image scrolled, instead of continuing from the image set using `setImageIdIndex`

### The expected behavior

Scroll should be continuous even if `setImageIdIndex` is called.

### OS

Linux

### Node version

20

### Browser

Firefox 128

abustany:
I added the following hack for now, which fixes the issue:
```
viewport.setImageIdIndex(imageIndex); // will update this.currentImageIdIndex but not this.targetImageIdIndex
(viewport as unknown as {targetImageIdIndex: number}).targetImageIdIndex = imageIndex; // workaround for #1393
```

cisacpalma:
I have the same issue

sedghi:
Can you check the latest OHIF viewer, which now includes Cornerstone 3D 2.0?

Here are the migration guides:

Try OHIF locally: [https://viewer-dev.ohif.org/localbasic](https://viewer-dev.ohif.org/localbasic)<br>Try Cornerstone3D 2.0 locally: [https://www.cornerstonejs.org/live-examples/local.html](https://www.cornerstonejs.org/live-examples/local.html)

[https://docs.ohif.org/migration-guide/3p8-to-3p9/](https://docs.ohif.org/migration-guide/3p8-to-3p9/)

sedghi:
I ran into this bug today and fixed it. Sorry if I didn't understand the issue at first.

PR #2039, opened 2025-04-29
fix(colorbar): enable bottom placement

sedghi:
fixes https://github.com/cornerstonejs/cornerstone3D/issues/1393 

This pull request introduces significant improvements to the colorbar functionality in the `colorbar` example, including the ability to dynamically change the colorbar's position and enhanced rendering for horizontal tick positions. The most important changes are grouped below by theme.

### New Feature: Dynamic Colorbar Positioning
* Added a dropdown to the toolbar for selecting the colorbar's position (`right`, `left`, or `bottom`) and implemented the `updateColorbarPosition` function to update the layout and recreate the colorbar dynamically. (`[[1]](diffhunk://#diff-9a97d79d1d7cffc750affcc4f4b745bb37638b25dc16c04510abe6218b1caed6R105-R119)`, `[[2]](diffhunk://#diff-9a97d79d1d7cffc750affcc4f4b745bb37638b25dc16c04510abe6218b1caed6R136-R232)`)
* Replaced the static initialization of the colorbar in the `run` function with a call to `updateColorbarPosition`, defaulting the position to `right`. (`[packages/tools/examples/colorbar/index.tsL237-R351](diffhunk://#diff-9a97d79d1d7cffc750affcc4f4b745bb37638b25dc16c04510abe6218b1caed6L237-R351)`)

### UI Enhancements
* Updated the grid layout logic to support horizontal (`bottom`) and vertical (`left`, `right`) colorbar positions. (`[packages/tools/examples/colorbar/index.tsR136-R232](diffhunk://#diff-9a97d79d1d7cffc750affcc4f4b745bb37638b25dc16c04510abe6218b1caed6R136-R232)`)
* Added a label to the colormap selection dropdown for improved user experience. (`[packages/tools/examples/colorbar/index.tsR91](diffhunk://#diff-9a97d79d1d7cffc750affcc4f4b745bb37638b25dc16c04510abe6218b1caed6R91)`)

### Rendering Improvements
* Implemented the `_getTopTickInfo` and `_getBottomTickInfo` methods in the `ColorbarTicks` class to support horizontal tick rendering. (`[packages/tools/src/utilities/voi/colorbar/ColorbarTicks.tsL422-R440](diffhunk://#diff-74ed9cb2912caa5fd8f670f29140601ee04b9c917496b3f21a22a8c78d20dfb7L422-R440)`)
* Adjusted text alignment and baseline in the `ColorbarTicks` rendering logic to handle horizontal colorbars correctly. (`[packages/tools/src/utilities/voi/colorbar/ColorbarTicks.tsL449-R464](diffhunk://#diff-74ed9cb2912caa5fd8f670f29140601ee04b9c917496b3f21a22a8c78d20dfb7L449-R464)`)

### DIFF
--- a/packages/tools/src/tools/displayTools/Contour/contourDisplay.ts
+++ b/packages/tools/src/tools/displayTools/Contour/contourDisplay.ts
@@ -191,13 +191,29 @@ function _checkContourGeometryMatchViewport(
   geometryIds: string[],
   viewportNormal: Types.Point3
 ): boolean {
-  const geometry = cache.getGeometry(geometryIds[0]);
+  // Find a geometry with at least 3 points in its first contour
+  let validGeometry = null;
+  let geometryData = null;
 
-  if (!geometry) {
-    return false;
+  for (const geometryId of geometryIds) {
+    const geometry = cache.getGeometry(geometryId);
+
+    if (!geometry) {
+      continue;
+    }
+
+    const data = geometry.data as Types.IContourSet;
+
+    if (data.contours?.[0]?.points?.length >= 3) {
+      validGeometry = geometry;
+      geometryData = data;
+      break;
+    }
   }
 
-  const geometryData = geometry.data as Types.IContourSet;
+  if (!validGeometry || !geometryData) {
+    return false;
+  }
 
   const contours = geometryData.contours;
   const points = contours[0].points;


packages/tools/examples/colorbar/index.ts
```diff
@@ -101,6 +102,21 @@ addDropdownToToolbar({
...
+  labelText: 'Colorbar Position: ',
+  options: {
+    values: ['right', 'bottom'],
+    defaultValue: 'right',
+  },
+  style: {
+    maxWidth: '120px',
+  },
+  onSelectedValueChange: (position) => {
+    updateColorbarPosition(position as string);
```
wayfarer3130, line 116: This should go through a command function so that it can be tied to a keyboard or other action such as onLoad in a hanging protocol.

packages/tools/examples/colorbar/index.ts
wayfarer3130, line 1: Can you use a zustand store to specify the colorbar position so it gets remembered?  That way it just renders in the default position, and gets re-assigned correctly when re-entering.  It also means it can be specified from a commands module and also from a mode startup.  Finally, it is a place to start storing other values that might be needed with the colorbar.

wayfarer3130 (changes requested):
Could you do this in zustand so it gets remembered/assigned correctly?

packages/tools/examples/colorbar/index.ts
wayfarer3130: Another option might be to use the presentation objects to define how the screen appears - although they should also be using zustand.

packages/tools/examples/colorbar/index.ts
sedghi: This is just an example in cornerstone3D, not OHIF. Its purpose is to demonstrate the minimum code needed to test different functionalities in `cs3d`. I don't want to create a HP engine for cs3d

packages/tools/examples/colorbar/index.ts
sedghi: There's really no need for any kind of store or state management in cs3d. The examples should be the bare minimum JavaScript code, just enough to showcase the different features in cs3d.

PR #2040, opened 2025-04-29
feat(polymorphic-segmentation): Add polyseg worker init options

tye-shanklin:
### Context
I have a specific use case where I need to keep the polyseg worker thread alive so that subsequent calls remain fast (I call the conversion methods directly instead of through Cornerstone), but the default PolySeg Worker registration is configured to delete the thread after 2 seconds.

I am updating the init function to accept optional configuration parameters so that I can configure the worker thread lifecycle without impacting existing workflow expectations.

### Changes & Results
Updated the polySeg.init() function to accept optional parameters. Not all options are required, and I deep merge the user-provided options with the default options so that we still have defaults, but user input takes precedence.

### Testing
I tested this by modifying the Segmentation workflow to add a Contour segmentation to a stack viewport and a button to convert the Contour to a labelmap and console log the conversion's image IDs.

The function I used in the OHIF viewers repo to test the conversion, invoked from the button click mentioned above:

```ts
const calculateLabelMap = async () => {
    const activeSegmentation =
      segmentation.defaultSegmentationStateManager.getActiveSegmentation('default');
    if (!activeSegmentation) {
      return;
    }
    const viewport = getEnabledElementByViewportId('default')?.viewport;
    if (!viewport) {
      return;
    }

    const start = performance.now();
    const results = (await computeLabelmapData(activeSegmentation.segmentationId, {
      viewport,
    })) as LabelmapSegmentationDataStack;
    const end = performance.now();
    console.log(results);
    console.log(`runtime: ${end - start}`);
  };
```

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: Windows 11
- [] "Node version: 23.5.0
- [] "Browser: Edge 135.0.3179.73


sedghi (approved):
Makes sense thanks

sedghi:
can you pull main, i fixed the tests

PR #2054, opened 2025-05-07
Feat(coordinate mapping)

rodrigobasilio2022:
### Changes & Results
Fixed coordinate conversions between slide coordinate system, image/pixel/index coordinates and canvas coordinates. 

image

Given the points 1, 2 and 3

| # | Point (World Coordinates) | Expected Canvas Result | CS3D Canvas Result | Difference |
|---|---------------------------|------------------------|-------------------|------------|
| 1 | [5, 25, 0]                | [0, 74.02]            | [0, 778]          | [0, +703.98] |
| 2 | [12.5, 32.5, 0]           | [250, 250]            | [250, 602]        | [0, +352] |
| 3 | [20, 40, 0]               | [500, 426.02]         | [500, 426.02]     | [0, 0] |

Also we implement the indexToWorld and worldToIndex functions

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  


packages/core/src/RenderingEngine/WSIViewport.ts
```diff
@@ -247,13 +272,10 @@ class WSIViewport extends Viewport {
...
       worldToIndex: (point: Point3) => {
-        const canvasPoint = this.worldToCanvas(point);
-        const pixelCoord = this.canvasToIndex(canvasPoint);
-        return [pixelCoord[0], pixelCoord[1], 0] as Point3;
+        return this.worldToIndex(point);
       },
       indexToWorld: (point: Point3) => {
-        const canvasPoint = this.indexToCanvas([point[0], point[1]]);
-        return this.canvasToWorld(canvasPoint);
+        return this.indexToWorld(point);
```
wayfarer3130, line 278: This is providing a Point3 coordinate to a point2 requiring function - you might change the signatures to allow either type here, or just see if you can cast it.  You might need to cast it to unknown first.

wayfarer3130: index coordinates are actually Point3 as they include a frame# in the z coordinate, which is not relevant for this case

packages/core/src/RenderingEngine/WSIViewport.ts
wayfarer3130, line 1: If you run bun example wsi and zoom in/pan, you will see that left/right pan is ok, but up/down pan is reversed.

packages/core/src/RenderingEngine/WSIViewport.ts
```diff
@@ -428,6 +453,48 @@ class WSIViewport extends Viewport {
     this.refreshRenderValues();
   };
 
+  /**
+   * Converts a slide coordinate to a image coordinate using WSI utils functions
+   * @param point
+   * @returns
+   */
+  public worldToIndexWSI(point: Point3): Point2 {
+    const affine = this.viewer[Symbol.for('affine')];
```
wayfarer3130, line 462: Can you do this as a constant?

packages/core/src/RenderingEngine/WSIViewport.ts
```diff
@@ -428,6 +453,48 @@ class WSIViewport extends Viewport {
...
+    const pixelCoords = WSIUtilFunctions?.applyInverseTransform({
```
wayfarer3130, line 463: You don't need the ? here as it will throw anyways if not defined.  You could add a shortcut return if it isn't defined.
Applies to all the spots you use it.

packages/core/src/RenderingEngine/WSIViewport.ts
wayfarer3130: I mean the affine symbol.  Also, could you create a bug report for dicom-microscopy-viewer to make affine not use the symbol setup?  That would be worthwhile generally to improve interaction with different front ends.

wayfarer3130 (changes requested):
Couple of small changes and then I can merge.

sedghi (commented):
Please update the description to explain what's currently wrong or what this unlocks.

rodrigobasilio2022:
@wayfarer3130 I think you approve on the worng place...

PR #2057, opened 2025-05-09
fix(metadata): add get pixelSpacing from ImagerPixelSpacing (0018,1164)

### LINKED ISSUE #2042: [Bug] metaDataProvider only looks at PixelSpacing which gives incorrect measurements
### Describe the Bug

`metaDataProvider` (for both WADO-RS and WADO-URI) only looks at PixelSpacing (0028,0030), but there are other metadata associated with the pixel spacing.

https://github.com/cornerstonejs/cornerstone3D/blob/f7d656d8b175c3ff00b145cdd5ceedf910afeda1/packages/dicomImageLoader/src/imageLoader/wadors/metaData/metaDataProvider.ts#L148

Cornerstone (incorrect):

Image

OHIF (correct):

Image

(anonymised image can be provided if you want).

### Steps to Reproduce

1. Using `dicomImageLoader` (WADO-URI or WADO-RS), load an image into the viewer which has `ImagerPixelSpacing` (0018,1164) or other tags which affect pixel spacing.
2. Annotate a length.
3. The length value displayed is incorrect.

### The current behavior

Incorrect length displayed, different to OHIF and in the context of the scan shown above obviously wrong.

I worked around it for now by writing a higher priority provider for image plane module in my app, which calls `getPixelSpacingInformation` but it is quite messy and should be fixed upstream.

### The expected behavior

Correct length displayed.

### OS

Linux mint 21

### Node version

22.5.1

### Browser

All

kubanm3:
### Context

Fixes #2042.

### Changes & Results

Takes pixelSpacing metadata from imagerPixelSpacing tag - 0018,1164 if not found earlier.

### Testing

Load dicom without Pixel Spacing tag - 0028,0030, in current version of application measurement would give wrong values. With change it will have correct value. Dicom to test is provided in issue #2042. 

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] "OS: Windows  11
- [x] "Node version:  v20.18.0
- [x] "Browser:
   Brave v1.78.97

### DIFF
--- a/packages/dicomImageLoader/src/imageLoader/wadouri/metaData/extractPositioningFromDataset.ts
+++ b/packages/dicomImageLoader/src/imageLoader/wadouri/metaData/extractPositioningFromDataset.ts
@@ -151,6 +151,12 @@ function extractSpacingFromDataset(dataSet) {
     );
   }
 
+  // If pixelSpacing not valid to this point, trying to get the spacing
+  // from the Imager Pixel Spacing
+  if (!pixelSpacing && dataSet.elements.x00181164) {
+    pixelSpacing = getNumberValues(dataSet, 'x00181164', 2);
+  }
+
   return pixelSpacing;
 }
 


packages/dicomImageLoader/src/imageLoader/wadouri/metaData/extractPositioningFromDataset.ts
```diff
@@ -151,6 +151,12 @@ function extractSpacingFromDataset(dataSet) {
     );
   }
 
+  // If pixelSpacing not valid to this point, trying to get the spacing
+  // from the Imager Pixel Spacing
+  if (!pixelSpacing && dataSet.elements.x00181164) {
```
sedghi, line 156: @wayfarer3130 Could you please check this? Also, should we do this for WADO-RS as well?

wayfarer3130: This is quite dangerous to do - it can result in surgery in the wrong location as the imager pixel spacing values are different from the spacing in the body.  You need to get the ERMF values and apply them, and if not available, then indicate this is imager spacing, not body spacing.  
ERMF = Estimated Radiographic Magnification Factor
A single tag containing the value.
Also, can be computed via distance detector to emitter, detector to body, body to emitter.  I don't have the DICOM tags for those immediately at hand, but any of of them can be used to calculate the last value since:
detector to emitter = detector to body + body to emitter
and 
ERMF = detector to emitter / body to emitter

kubanm3: I understand thats it's more complicated, but the example I provided gives completly wrong measurement with current solution. Where other dicom viewers get it right. So I belive this is important to get working.

PR #2061, opened 2025-05-12
test(rectangleROIThreshold+statistics): Modify rectangleStartEndROIThreshold example for threshold value + add test for testing rectangleROIThreshold with statistics result comparison with MIM

dxlin:
Modify rectangleStartEndROIThreshold example to:
->use PT image
->allow input for threshold value
->relative to max ROI
-added addInputToToolbar for examples
-added test to check accuracy of routine relative to MIM software on same image set NOTE:
1. Doesn't include end slice (test index for endSlice higher than should be by 1 to account)
2. Edge slice peakValue discrepancies with MIM (ignoring for now that check)

Should the example be separated?
cachedStats.statistics worked in example but did not populate in time for playwright test, a function based on OHIF's implementation of getting max voxel value was used -> should this be made into a utility function?

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  

### DIFF
--- a/packages/tools/examples/rectangleROIStartEndThresholdWithSegmentation/index.ts
+++ b/packages/tools/examples/rectangleROIStartEndThresholdWithSegmentation/index.ts
@@ -11,7 +11,9 @@ import {
   createImageIdsAndCacheMetaData,
   setTitleAndDescription,
   addButtonToToolbar,
-  setCtTransferFunctionForVolumeActor,
+  addCheckboxToToolbar,
+  addInputToToolbar,
+  setPetTransferFunctionForVolumeActor,
   getLocalUrl,
 } from '../../../../utils/demo/helpers';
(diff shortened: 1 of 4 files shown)


sedghi:
Thanks! is it expected that they fail?

dxlin:
The current code should pass, the endSlice in the testcases defined are incremented by 1 to account for endSlice not being included (I've added a comment where the endSlice are incremented).

Let me know if you'd like the offset to be removed so it can be used to test against a fix for the endSlice not being included (or if that is intended behaviour, I can remove the comments).

dxlin:
Sorry, made a mistake with unit conversion and noticed a race condition for the debounce. These should be fixed now.

Issue #403, opened 2023-01-30
Passive tool enters draw mode when selected


Edward-Son:
Hello,

I am using the `PlanarFreehandRoiTool` to display contours on an image. The tool is set to passive, since I only want to display and not edit. When I click the annotation, I listen to the `ANNOTATION_SELECTION_CHANGE` event to do an action. However, the tool is then launched into drawing mode, even though it is set to passive. 

Inside `mouseDown.js`  there is 
`const annotationToolsWithMoveableHandles = filterToolsWithMoveableHandles(element, annotationToolsWithAnnotations, canvasCoords, 'mouse');`
which returns the planar tool. However, if the tool is passive, then the planar tool should not be returned here? Is this intended, or should this method filter passive tools from having moveable handles?

Another option would be to check for passive mode inside of `PlanarFreehandROITool.js` `handleSelectedCallback()` before `activateOpenContourEndEdit()`. 

Let me know what you think, and I can open an MR if needed.

sedghi:
Yes that makes sense, can you open a PR?

PR #2063, opened 2025-05-12
Fix(tools) don't allow passive tools to edit annotations

h1zqeel:
### Context
Fixes the Issue reported here: https://github.com/cornerstonejs/cornerstone3D/issues/403

### Changes & Results

Before:
When adding contours with any tool or using contour segmentations, there was no mechanism to prevent users from modifying them via mouse interactions.

After:
Passive tools are now excluded from filterMoveableAnnotationTools; if a tool is in passive mode, it will no longer be considered moveable and cannot be modified using mouse events.

### Testing

1. Run the example: `contourRendering`.
2. Navigate to `http://localhost:3000`.
3. **Expected Behavior (Default Tool Mode):**

   * Since the tool mode is not explicitly set to *Active*, it defaults to *Passive*.
   * As a result, contour annotations should **not** be moveable via mouse interactions.
4. **Activate Tool:**

   * Set the `PlanarFreehandROITool` to *Active* mode.
   * Now, the user **should** be able to move the contour annotations as before.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- Mac OS 15.1
- Node v20.19.0
- Edge Version 135.0.3179.98

### DIFF
--- a/packages/tools/src/store/filterMoveableAnnotationTools.ts
+++ b/packages/tools/src/store/filterMoveableAnnotationTools.ts
@@ -35,12 +35,11 @@ export default function filterMoveableAnnotationTools(
   const moveableAnnotationTools = [];
 
   ToolAndAnnotations.forEach(({ tool, annotations }) => {
+    if (tool.mode === ToolModes.Passive) {
+      return;
+    }
     for (const annotation of annotations) {
-      if (
-        annotation.isLocked ||
-        !annotation.isVisible ||
-        tool.mode === ToolModes.Passive
-      ) {
+      if (annotation.isLocked || !annotation.isVisible) {
         continue;
       }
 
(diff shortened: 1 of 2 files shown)


h1zqeel:
@sedghi can this be looked into, this bug has been open for 2 years

sedghi:
Isnt disabled state exactly for this use case? Why we are making passive behaviour change?

hizqeel-javed-seetreat:
> Isnt disabled state exactly for this use case? Why we are making passive behaviour change?

The disabled state also removes the contours from the viewport which makes them hidden. If i am understanding correctly, as mentioned in the original issue Passive tools should not be able to perform actions on mouseEvents.

h1zqeel:
> Isnt disabled state exactly for this use case? Why we are making passive behaviour change?

@sedghi  
 
If we set this tool to the disabled state, it would also remove the segmented contours from the Viewport and not render them.

My understanding is that Passive tools should not make adjustments on MouseEvents, isnt that the case?

sedghi:
There are certainly usecases for having passive, so i am not comfortable to change that behaviour. We should fixed them in contour and segmentation instead

sedghi:
Sorry i meant enabled, we should set them as enbaled not disabled, that will allow rendering but not interaction

PR #2067, opened 2025-05-13
feat(version): add version property

IbrahimCSAE:
### Context

Add the ability to check the current package version dynamically.

You can do

```ts
import { version } from '@cornerstonejs/core'

console.debug(version)
```

or

```ts
import * as cornerstoneTools from '@cornerstonejs/tools'

console.debug(cornerstoneTools.version)
```


sedghi (changes requested):
This would make it always one version behind, since we don't know what is the version being merged with fix/feat

PR #2085, opened 2025-05-19
fix(hydrate): bidirectional was not hydrating properly

IbrahimCSAE:
### context

fix bidirectional hydrate

### DIFF
--- a/packages/tools/src/tools/annotation/BidirectionalTool.ts
+++ b/packages/tools/src/tools/annotation/BidirectionalTool.ts
@@ -240,7 +240,7 @@ class BidirectionalTool extends AnnotationTool {
     } = this.hydrateBase<BidirectionalTool>(
       BidirectionalTool,
       enabledElement,
-      [],
+      axis[0],
       options
     );
 


IbrahimCSAE:
I will merge this as I need this change for some work I'm doing, if there's any issues let me know and I can revert, but I'm pretty sure this is a bug.

sedghi:
Please don't merge without a review, @IbrahimCSAE. Bill and I are actively monitoring all the repositories.

packages/tools/src/tools/annotation/BidirectionalTool.ts
```diff
@@ -240,7 +240,7 @@ class BidirectionalTool extends AnnotationTool {
     } = this.hydrateBase<BidirectionalTool>(
       BidirectionalTool,
       enabledElement,
-      [],
+      axis[0],
```
wayfarer3130, line 243: axis is an array of array of Point3, so axis[0] is an array of Point, so this is correct.  Will approve retroactively.

wayfarer3130 (commented):
Reviewed this retroactively.

PR #2087, opened 2025-05-20
feat(WSIViewport): add support for window level tool

mbellehumeur:
### Context

ticket: 9038784383

### Changes & Results

Added window level support to WSIViewport in the same way that it is done in videoViewport using an feFilter on the canvas style property.
The public 'setVOI' method is the one used by the window level tool.

### Testing

The window level tool changes the display in both the main canvas and the small 'map' canvas in the bottom left corner. 

#### PR

#### Code

  etc.)


packages/core/src/RenderingEngine/WSIViewport.ts
```diff
@@ -217,17 +230,63 @@ class WSIViewport extends Viewport {
...
+    this.voiRange = voiRange;
+    this.setColorTransform();
+  }
+
+  public setAverageWhite(averageWhite: [number, number, number]) {
+    this.averageWhite = averageWhite;
+    this.setColorTransform();
+  }
+
+  protected setColorTransform() {
```
wayfarer3130, line 259: Can you move the setColorTransform call to Viewport, and just describe it as a filter on the canvas?  That way both Video and WSI Viewports use the same code.  The setColorTransform can take the voiRange/averageWhite as arguments.

packages/core/src/RenderingEngine/WSIViewport.ts
```diff
@@ -217,17 +230,63 @@ class WSIViewport extends Viewport {
...
+    return {
+      voiRange: { ...this.voiRange },
+    };
   };
 
   public resetProperties() {
     this.setProperties({});
   }
 
+  public setVOI(voiRange: VOIRange): void {
```
sedghi, line 248: please add docs above each publich function

packages/core/src/RenderingEngine/WSIViewport.ts
sedghi: Stack/Volume should not be able to call them, i think we need another base class

wayfarer3130: It doesn't matter if stack/volume calls them, it would apply as a secondary window level, and there are cases that might be worth allowing such as having a JPEG original image.  Applying window level using a transform for that case avoids certain types of issues around caching data.  However, as a protected method or even  a static method one could just declare it as being a helper method in the docs.  However, you might be right, adding a new base class for non-vtk based stacks might be worthwhile as there are other helper methods to consider.

packages/core/src/RenderingEngine/VideoViewport.ts
```diff
@@ -553,6 +553,10 @@ class VideoViewport extends Viewport {
       loop: false,
       muted: true,
```
sedghi, line 554: Other viewports, such as StackViewport and VolumeViewport, implement a `voiRange` in their `setProperties`. If we're resetting properties, we can use `this.setProperties({voiRange})` to align with those other implementations.

wayfarer3130 (approved):
Can you just add the comment requested and then let me know and I will merge?

sedghi:
this is failing please fix 

cross-env NODE_OPTIONS=--max_old_space_size=32896 bun x lerna run format-check
lerna notice cli v8.2.1
lerna info ci enabled

 Lerna (powered by Nx)   Running target format-check for [8](https://github.com/cornerstonejs/cornerstone3D/actions/runs/15301879477/job/43044314161?pr=2087#step:7:9) projects:

- @cornerstonejs/adapters
- @cornerstonejs/ai
- @cornerstonejs/core
- @cornerstonejs/dicom-image-loader
- @cornerstonejs/labelmap-interpolation
- @cornerstonejs/nifti-volume-loader
- @cornerstonejs/polymorphic-segmentation
- @cornerstonejs/tools



✅ > @cornerstonejs/nifti-volume-loader:format-check
✅ > @cornerstonejs/dicom-image-loader:format-check
❌ > @cornerstonejs/core:format-check
✅ > @cornerstonejs/ai:format-check
✅ > @cornerstonejs/adapters:format-check
✅ > @cornerstonejs/tools:format-check


 Lerna (powered by Nx)   Running target format-check for 8 projects failed

Tasks not run because their dependencies failed or --nx-bail=true:

- @cornerstonejs/labelmap-interpolation:format-check
- @cornerstonejs/polymorphic-segmentation:format-check

Failed tasks:

- @cornerstonejs/core:format-check

error: script "format-check" exited with code 1

yarn run format-check

PR #2099, opened 2025-05-27
fix: Distinguishing between scoord3d and scoord (3d/2d) measurements and parsing them

pedrokohler:
This pull request introduces several updates to the `packages/adapters` module, focusing on streamlining measurement data handling, improving code maintainability, and updating dependencies. The changes primarily involve refactoring `Angle`, `ArrowAnnotate`, `Bidirectional`, and `BaseAdapter3D` classes, introducing helper functions for coordinate conversions, and upgrading the `dcmjs` dependency.

### Refactoring and Code Simplification:
* **Refactored `Angle`, `ArrowAnnotate`, and `Bidirectional` classes**: Replaced separate methods for handling SCOORD and SCOORD3D groups with a unified approach using helper functions (`toScoord` and `toScoords`) to convert coordinates. This simplifies the logic and reduces redundant code. (`[[1]](diffhunk://#diff-c021a8a3d3fa1d553c2ee8a6a680e55c78a6a01ee1dd05e43623eb02c05b0248L21-R82)`, `[[2]](diffhunk://#diff-38eff3e959e9491943e62fda38c20866266dc252b7bad6e9e3d58d2b8e247fd1L23-R41)`, `[[3]](diffhunk://#diff-5b2a4f6299dd3281e42246cc8de88fdcc0a8edb1edc20cbd49d41e1c7bb8f678L21-L33)`)
* **Removed redundant methods in `BaseAdapter3D`**: Consolidated methods for TID300 representation arguments into a single implementation, utilizing `toScoords` for coordinate conversion. (`[packages/adapters/src/adapters/Cornerstone3D/BaseAdapter3D.tsL145-L195](diffhunk://#diff-7980ee0e8f8cb7710663f5ddd518f88d4809605a41d27bba1e0087358a97bed3L145-L195)`)

### Dependency Update:
* **Updated `dcmjs` dependency**: Upgraded from version `0.42.0` to `0.43.1` in `package.json` to ensure compatibility and access to newer features. (`[packages/adapters/package.jsonL88-R88](diffhunk://#diff-e7b11bd4833791e7f2ee0c6ead2346c5228ae519a20b799fe54384ee8239adedL88-R88)`)

### Testing Configuration:
* **Added `testMatch` pattern in Jest configuration**: Updated `jest.config.js` to include a specific test match pattern for easier test discovery. (`[packages/adapters/jest.config.jsR8](diffhunk://#diff-5614049a099c7010222b5acb5423c9a3345fd7a97ebbc2a43cf03eeeff8de9ecR8)`)

### Helper Functions Integration:
* **Introduced `toScoord` and `toScoords` helper functions**: Added utility functions for coordinate conversion, improving consistency and reusability across multiple adapter classes. (`[[1]](diffhunk://#diff-c021a8a3d3fa1d553c2ee8a6a680e55c78a6a01ee1dd05e43623eb02c05b0248R4)`, `[[2]](diffhunk://#diff-38eff3e959e9491943e62fda38c20866266dc252b7bad6e9e3d58d2b8e247fd1R5)`, `[[3]](diffhunk://#diff-7980ee0e8f8cb7710663f5ddd518f88d4809605a41d27bba1e0087358a97bed3R6)`, `[[4]](diffhunk://#diff-5b2a4f6299dd3281e42246cc8de88fdcc0a8edb1edc20cbd49d41e1c7bb8f678L2-R3)`)


wayfarer3130:
@pedrokohler  - can you mark this as ready for review?  I still have to finish fixing the unit tests, but it is working fairly well now overall.  Might be one or two more tweaks.  I'm going to ask @sedghi  for a review as I've made a lot of changes to it.

pedrokohler:
@wayfarer3130 done.

packages/adapters/src/adapters/helpers/scoordToWorld.ts
```diff
@@ -0,0 +1,29 @@
+import type { Types } from "@cornerstonejs/core";
+
+export function scoordToWorld(
+    { isMeasurement3d, referencedImageId, imageToWorldCoords },
+    scoord
+): Types.Point3[] {
+    const worldCoords = [];
+    if (isMeasurement3d) {
```
sedghi: i see `is3DMeasurement` elsewhere, pick one

packages/adapters/src/adapters/helpers/scoordToWorld.ts
```diff
@@ -0,0 +1,29 @@
+import type { Types } from "@cornerstonejs/core";
+
+export function scoordToWorld(
```
sedghi, line 10: doc please

packages/adapters/src/adapters/helpers/toPoint3.ts
```diff
@@ -0,0 +1,22 @@
+import type { Types } from "@cornerstonejs/core";
+
+export function toPoint3(flatPoints: number[]): Types.Point3[] {
```
sedghi, line 9: doc please

packages/adapters/src/adapters/helpers/toScoordType.ts
```diff
@@ -0,0 +1,16 @@
+import type { Types } from "@cornerstonejs/core";
+
+export function toScoord(
+    { worldToImageCoords, is3DMeasurement, referencedImageId },
+    point
+) {
+    if (is3DMeasurement) {
+        return { x: point[0], y: point[1], z: point[2] };
+    }
+    const point2 = worldToImageCoords(referencedImageId, point);
```
sedghi: why are we passing worldToImageCoords here? this can get imported from cornerstone since it is a utility

packages/adapters/src/adapters/helpers/scoordToWorld.ts
```diff
@@ -0,0 +1,29 @@
...
+                GraphicData[i],
+                GraphicData[i + 1],
+                GraphicData[i + 2]
+            ];
+            worldCoords.push(point);
+        }
+    } else {
+        const { GraphicData } = scoord;
+        for (let i = 0; i < GraphicData.length; i += 2) {
+            const point = imageToWorldCoords(referencedImageId, [
```
sedghi, line 28: isn't this utility already part of Cornerstone Core? Why are we passing these items as parameters?

packages/core/src/utilities/updateReferencedPlane.ts
```diff
@@ -0,0 +1,54 @@
+import type { Point3, ViewReference } from '../types';
+import { isEqual } from '../utilities/isEqual';
+import { vec3 } from 'gl-matrix';
+
+/** Updates the referencedPlane inside the view reference */
```
sedghi: appreciate much longer doc, expand on the logic in the doc, what happens if there point list is 1 and what happens if it is more than one

packages/core/src/utilities/updateReferencedPlane.ts
```diff
@@ -0,0 +1,54 @@
...
+export function updateReferencedPlane(
```
sedghi: updateViewReferencedPlane?

packages/core/src/utilities/updateReferencedPlane.ts
```diff
@@ -0,0 +1,54 @@
...
+    FrameOfReferenceUID: reference.FrameOfReferenceUID,
+    point: points[0],
+    inPlaneVector1: null,
+    inPlaneVector2: null,
+  };
+  const { referencedPlane } = reference;
+
+  if (points.length === 1) {
+    referencedPlane.inPlaneVector1 = null;
+    referencedPlane.inPlaneVector2 = null;
```
sedghi: can you return the plane here, so the we can remove indentation from the else 

if (points.length === 1) {
    plane.inPlaneVector1 = null;
    plane.inPlaneVector2 = null;
    return plane;
}

// else without indentation
referencedPlane.inPlaneVector1 = <Point3>(
      vec3.sub(vec3.create(), points[0], points[Math.floor(points.length / 2)])
    );

packages/core/src/utilities/updateReferencedPlane.ts
```diff
@@ -0,0 +1,54 @@
...
+  } else {
+    referencedPlane.inPlaneVector1 = <Point3>(
+      vec3.sub(vec3.create(), points[0], points[Math.floor(points.length / 2)])
+    );
+    vec3.normalize(
+      referencedPlane.inPlaneVector1,
+      referencedPlane.inPlaneVector1
+    );
```
sedghi: can you do the other way around. 

v1 = vec3.sub(...)
vec3.normalize(v1,v1)
referenced.inplane1= v1

packages/core/src/utilities/updateReferencedPlane.ts
```diff
@@ -0,0 +1,54 @@
...
+    if (n > 2) {
+      for (let i = Math.floor(n / 3); i < n; i++) {
+        const testVector = vec3.sub(vec3.create(), points[i], points[0]);
+        const length = vec3.length(testVector);
+        if (isEqual(length, 0)) {
+          continue;
+        }
+        if (
+          vec3.dot(testVector, referencedPlane.inPlaneVector1) <
+          length * 0.95
```
sedghi: make EPSION a CONST on top

packages/core/src/types/IViewport.ts
```diff
@@ -110,17 +110,76 @@ export type ReferencedImageRange = ViewReference & {
...
+   * The referenced plane is visible depth wise if the point minus the focal point of hte
+   * current view is orthogonal to the view plane normal of the viewport.
+   * This is sufficient for a single-point identifier.
+   * For a line, the inPlaneVector must be orthogonal to the view plane normal of the viewport.
+   * For a planar annotation, both inPlaneVectors must be orthogonal to the view plane normal of the viewport.
+   *
+   * This extension from the least specific to the most specific types of views allows
+   * determining whether a view can be seen in a variety of conditions.  It does not
+   * allow recovering the original view reference.
+   */
```
sedghi, line 183: is it one plane or more than one plane? the ts type says otherwise

packages/core/src/types/IViewport.ts
```diff
@@ -110,17 +110,76 @@ export type ReferencedImageRange = ViewReference & {
   referencedImageId: string;
 };
 
+/**
+ *  A referenced plane specifies a planar coordinate system
+ * The most degenerate case has a single point, and includes all planes
+ * which pass through that point.  For this case, if a view plane normal
+ * is required to display the point, the current view plane normal of the
+ * viewport is recommended, or the acquisition plane if there is no view plane
+ * normal.
```
sedghi: what is the case that we don't have view plane normal? there is always a vpn no?

packages/core/src/types/IViewport.ts
```diff
@@ -110,17 +110,76 @@ export type ReferencedImageRange = ViewReference & {
...
+ *
+ * A specific plane reference further restricts the colinear planes to those
+ * which also include point+inPlaneVector2
+ * The view plane normal to display this point is required to be the cross product
+ * of the two inPlaneVector values.
+ */
+export type ReferencedPlane = {
+  FrameOfReferenceUID: string;
+
+  /** A single point within the reference plane is required for all references */
```
sedghi: if you do this /** absd the doc generation has problem. It should be on new lines

packages/core/src/types/IViewport.ts
```diff
@@ -110,17 +110,76 @@ export type ReferencedImageRange = ViewReference & {
...
+  point: Point3;
+
+  /** An inPlaneVector1 is required for all colinear referenced planes.
+   * Shall not be undefined if inPlaneVector2 is defined.
+   */
+  inPlaneVector1?: Point3;
```
sedghi, line 144: vec3? instead of Point3?

packages/core/src/types/IViewport.ts
```diff
@@ -110,17 +110,76 @@ export type ReferencedImageRange = ViewReference & {
...
+   * (but NOT the focal point), and up to two coplanar vectors.
+   *
+   * The referenced plane is visible depth wise if the point minus the focal point of hte
+   * current view is orthogonal to the view plane normal of the viewport.
+   * This is sufficient for a single-point identifier.
+   * For a line, the inPlaneVector must be orthogonal to the view plane normal of the viewport.
+   * For a planar annotation, both inPlaneVectors must be orthogonal to the view plane normal of the viewport.
+   *
+   * This extension from the least specific to the most specific types of views allows
+   * determining whether a view can be seen in a variety of conditions.  It does not
```
sedghi, line 181: let's simplify the doc, and remove the logic from it. 

```md
/**
 * Encapsulates the spatial reference used to decide whether an annotation
 * should appear in the active imaging slice.
 *
 * Contents
 * ────────
 * • **point** – establishes the depth at which the annotation resides.  
 * • **up to two in-plane orientation vectors** – refine the description from
 *   a depth marker (no vectors) to a line-based marker (one vector) or to a
 *   full planar marker (two non-parallel vectors).
 *
 * Behaviour in a viewer
 * ─────────────────────
 * 1. **Point only** The annotation is rendered whenever the current slice
 *    intersects that depth.  
 * 2. **Point + one vector** Adds an orientation cue; the annotation becomes
 *    visible only when the slice both intersects the depth and aligns with
 *    the specified direction  
 * 3. **Point + two vectors** Specifies one unique slice; the annotation is
 *    displayed exclusively when the viewer presents that exact plane.
 */
 ```

 edit as necessary

File: packages/adapters/src/adapters/Cornerstone3D/MeasurementReport.ts [hunk 16]
```diff
@@ -286,22 +325,31 @@ export default class MeasurementReport {
...
                         toolName: toolType,
                         FrameOfReferenceUID:
-                            SCOORD3DGroup.ReferencedFrameOfReferenceUID,
-                        label: ""
+                            SCOORD3DGroup.ReferencedFrameOfReferenceUID
                     }
                 }
             }
         };
+        console.warn("SCOORD3DGroup=", SCOORD3DGroup);
```
sedghi: remove

packages/adapters/src/adapters/Cornerstone3D/MeasurementReport.ts
```diff
@@ -371,15 +420,18 @@ export default class MeasurementReport {
             defaultState.description = defaultState.finding.CodeMeaning;
         }
 
-        defaultState.annotation.metadata.label =
+        defaultState.annotation.data.label =
             MeasurementReport.getCornerstoneLabelFromDefaultState(defaultState);
 
         return {
             defaultState,
+            state: defaultState,
```
sedghi: Why are we adding a new state which is equal to default state when a default state is already being returned?

sedghi (changes requested):
see my comments

packages/adapters/src/adapters/helpers/toScoordType.ts
wayfarer3130: Because the API has been like that for a long time, since the code was in dcmjs and it wasn't available.  I'd like to leave the API as is until we are ready to update the CS3D version and then update it all at once

packages/adapters/src/adapters/helpers/scoordToWorld.ts
wayfarer3130: Yes, but it has been like that a long while, this is just centralized code.  I'd like to leave it like that until we are ready to update the CS3D version.

packages/core/src/utilities/updateReferencedPlane.ts
wayfarer3130: I think updateReferencedPlane is sufficient, but if you think otherwise we can update it.

packages/core/src/types/IViewport.ts
wayfarer3130: Updated to explain that there are some degenerate cases.

packages/core/src/types/IViewport.ts
wayfarer3130: Restoring 3d point or length annotations from DICOM SR don't have a view plane normal

packages/core/src/types/IViewport.ts
wayfarer3130: Everything else we declare is Point3, and vec3 is a Point3 in everything except declaration so I'd prefer to go with Point3.

packages/adapters/src/adapters/Cornerstone3D/MeasurementReport.ts
wayfarer3130: Added comments - basically there are enough uses of it out there that haven't switched over to use state yet.

packages/core/src/RenderingEngine/Viewport.ts
```diff
@@ -1823,16 +1825,66 @@ class Viewport {
...
-      FrameOfReferenceUID: this.getFrameOfReferenceUID(),
+      FrameOfReferenceUID,
       cameraFocalPoint,
       viewPlaneNormal,
       viewUp,
       sliceIndex: viewRefSpecifier?.sliceIndex ?? this.getSliceIndex(),
+      /** The referenced plane is the canonical specifier for whether
+       * this view reference is visible or not.
+       */
+      planeRestriction: {
```
wayfarer3130, line 1838: @sedghi - I've renamed this to planeRestriction as it is setup to restrict what planes can be shown.  Does that sound better to you?  I'm thinking we can use it more generally in the future with different types of restrictions.

packages/tools/src/tools/CrosshairsTool.ts
wayfarer3130, line 1: Some of these files have type declaration fixes, needed because I made some of the IViewport more comprehensive.

packages/core/src/types/IViewport.ts
wayfarer3130: Resolved this by renaming and changing the docs.

packages/core/src/types/IViewport.ts
wayfarer3130: Resolved by updating the name/docs a fair bit here.

packages/tools/src/types/AnnotationTypes.ts
```diff
@@ -1,6 +1,42 @@
...
+   * The viewUp for the view position
+   */
+  viewUp?: Types.Point3;
+
+  /** The color of the segmentation - really this is annotation data */
+  segmentColor?;
+};
+
+export type AnnotationData = {
+  /** Annotation handles that are grabbable for manipulation */
```
sedghi: I'm not sure if the apidocs understand /** same line */ if it is same line use //

sedghi (approved):

Issue #2049, opened 2025-05-05
[Bug] Wrong rendering from dicom 1.2.840.10008.1.2.1 (Explicit VR Little Endian) on PALETTE COLOR


AnnaKuduk:
### Describe the Bug

On the last version is wrong rendering from dicom file (i attach this zipped file).  On GPU Mode - bad colors, on CPU Mode - black image. I used for now old version (1.83.1 - im sure in CPU mode) a few month ago was working correctly. I check too on examples sites :  https://www.cornerstonejs.org/live-examples/local and https://www.cornerstonejs.org/live-examples/localcpu - with the same effect - for me means - no error in my code (changes to new version)

[1.2.840.10008.1.2.1-ExplicitVRLittleEndian-PaletteColor.zip](https://github.com/user-attachments/files/20040801/1.2.840.10008.1.2.1-ExplicitVRLittleEndian-PaletteColor.zip)

### Steps to Reproduce

1. Open the site https://www.cornerstonejs.org/live-examples/local
2. Unzip and put attachment file 

### The current behavior

![Image](https://github.com/user-attachments/assets/2e0a1fa7-dbce-48e0-9154-077d73b36313)
![Image](https://github.com/user-attachments/assets/9673941e-87eb-4e7d-99eb-2db379426130)

### The expected behavior

On MicroDicom

![Image](https://github.com/user-attachments/assets/8c32c53b-a5e7-4c1b-b68d-50435cc66e5b)

### OS

Windows 10/11

### Node version

v20.11.0

### Browser

Chrome : 135.0.7049.115, Edge : 136.0.3240.50

daker:
@mbellehumeur maybe close the issue i think you already fixed this

mbellehumeur:
I don't have permissions to close issues.  Not sure who to ask.

daker:
@sedghi can you close those issues (https://github.com/cornerstonejs/cornerstone3D/issues/2049 , https://github.com/cornerstonejs/cornerstone3D/issues/1857) they are already fixed.

Issue #1857, opened 2025-02-26
[Bug] PALETTE COLOR Dicom not displaying properly


qaqaqaqaowh:
### Describe the Bug

Image not applying LUT correctly

### Steps to Reproduce

1. Upload [this file](https://github.com/user-attachments/files/18979148/dicom_viewer_0020.1.zip) to [https://www.cornerstonejs.org/live-examples/local.html]()

### The current behavior

Currently image is being displayed as Grayscale

![Image](https://github.com/user-attachments/assets/913bb19f-26cc-4f60-aae4-eec8fda4351f)

### The expected behavior

This is the same image viewed from another viewer

![Image](https://github.com/user-attachments/assets/a68228c6-25a2-4d42-8f69-aaa40b098f37)

### OS

Windows 11

### Node version

18.19.1

### Browser

Opera GX 116.0.5366.148, Chromium 131.0.6778.266

diaztula:
I can confirm this bug. In version 1.x the image with color palette opens correctly.

daker:
@qaqaqaqaowh i think it's fixed, can you close the issue?

PR #2113, opened 2025-06-02
fix(rendering): palette color 

mbellehumeur:
### Context
Issue 1857 and 2049.

https://github.com/cornerstonejs/cornerstone3D/issues/2049
https://github.com/cornerstonejs/cornerstone3D/issues/1857

Images of type PALETTE_COLOR did not display correctly.

### Changes & Results

1. Turned on color conversion in the iSColorConversionRequired function for PALETTE.

2.  Set 'numberOfComponents' to be 3 or 4 (RGB/RGBA) for palette color instead of equal to samplesPerPixel.

### Testing

Use the attached DICOM file for example local and localCPU.  It should display in colour like this.
Run the examples  dicomImageLoaderWADOURI and ultrasoundColors.

image


packages/dicomImageLoader/src/imageLoader/colorSpaceConverters/convertPALETTECOLOR.ts
```diff
@@ -45,51 +45,22 @@ export default function (
...
-        colorBuffer[bufferIndex++] = rDataCleaned[value];
-        colorBuffer[bufferIndex++] = gDataCleaned[value];
-        colorBuffer[bufferIndex++] = bDataCleaned[value];
-        colorBuffer[bufferIndex++] = 255;
-      }
-
-      return;
-    }
-
+  const rData = imageFrame.redPaletteColorLookupTableData;
```
sedghi: Why not promise all? since OHIF has a metadata provider that create promises, i'm pretty sure this will break OHIF

sedghi:
Please check if this solves this issue in OHIF, link this PR to OHIF and try 

https://github.com/OHIF/Viewers/issues/4642

How to link 

https://www.cornerstonejs.org/docs/contribute/linking

mbellehumeur:
> Please check if this solves this issue in OHIF, link this PR to OHIF and try
> 
> [OHIF/Viewers#4642](https://github.com/OHIF/Viewers/issues/4642)
> 
> How to link
> 
> https://www.cornerstonejs.org/docs/contribute/linking

It works in OHIF for wado but not for dicomfile (loading from the viewer which I was using to test...).  This will require a PR in OHIF I think.
image

sedghi (approved):
Great fix! I've added back the tests we skipped.

packages/dicomImageLoader/src/imageLoader/isColorConversionRequired.ts
```diff
@@ -27,8 +27,7 @@ export default function isColorConversionRequired(imageFrame) {
   }
 
   if (photometricInterpretation === 'PALETTE COLOR') {
-    // since it is based on a look up table we don't need to convert
-    return false;
+    return true;
```
wayfarer3130, line 30: This means we don't use any native LUT support @sedghi  - is there native LUT support for palette colour images?

wayfarer3130 (approved):
Even if we have native LUT support, jsut fixing it for now with applying the LUT manually means we fix a number of images, so I think we should proceed with this as is.

wayfarer3130:
@sedghi  - is there any way to have the playwright tests publish the failing images so that we can just replace images with the new one if needed, or at least see if it is a real failure or not?

sedghi:
@wayfarer3130 @mbellehumeur yes, pass the `--update-snapshots` to that exact test and it will update it

packages/dicomImageLoader/src/imageLoader/isColorConversionRequired.ts
sedghi: I don't think so

PR #2128, opened 2025-06-06
fix(Circle/RectangleROIStartEndThresholdTools) : fix projection points calculation when the start point is after end point

Celian-abd:
### Context
With the Circle/RectangleROIStartEndThresholdTools if you only modify **the start point** to put it after **the current end point**, the projection point calculation was wrong and the points we're not right.

### Changes & Results
- Change how we calculate the projection points (now we always goes from the **smallest** to the **largest coordinate**)

### Testing
Can be tested with the rectangleROIStartEndThresholdWithSegmentation or circleROIStartEndThresholdWithSegmentation by only setting the start point after the current end point.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  


salimkanoun:
Probablement interne à OHIF
https://github.com/OHIF/Viewers/issues/5126

sedghi:
what is the issue now? I'm not following

salimkanoun:
@sedghi this PR fixes a bug that happend when the user set the start slice bellow the End slice.
Celian fixed the computation of point no matter if the start slice is over or under the end slice

salimkanoun:
If your talkin about the refrence about the other decaching bug it is just a mistake in my part, i removed by previous post in this thread (simply wrote my text in the wrong place, the two issues are not related)

rodrigobasilio2022:
Looks good to me.

PR #2129, opened 2025-06-07
feat: add logic to change background color in setOptions method

MrWerton:


packages/core/src/RenderingEngine/Viewport.ts
```diff
@@ -301,6 +301,16 @@ class Viewport {
...
       this.setDisplayArea(this.options?.displayArea);
     }
+
+    if (options.background) {
+      const renderingEngine = this.getRenderingEngine();
+
+      const offscreenRenderer =
+        renderingEngine.offscreenMultiRenderWindow.getRenderer(this.id);
+
+      offscreenRenderer.setBackground(options.background);
```
sedghi, line 311: is this a real api? i'm not sure

MrWerton: Yes. Here are the main details:

- The background is set during the renderer creation via the call:  
  [`offscreenMultiRenderWindow.addRenderer(...)`](https://github.com/cornerstonejs/cornerstone3D/blob/main/packages/core/src/RenderingEngine/RenderingEngine.ts#L773C7-L773C17)

- The class `offscreenMultiRenderWindow` has the method `getRenderer(this.id)`, which returns the corresponding viewport, an instance of `vtkViewport`.

- The `vtkViewport` type, defined in VTK.js, provides the method:  
  [`setBackground(background: number[])`](https://github.com/Kitware/vtk-js/blob/master/Sources/Rendering/Core/Viewport/index.d.ts#L147)

rodrigobasilio2022: The background color defined in a viewportInput configuration like that
```javascript
  const viewportInput = {
    viewportId,
    type: ViewportType.STACK,
    element,
    defaultOptions: {
      background: [0.8, 0, 0.8] as Types.Point3,
    },
  };
```
is configured here

https://github.com/cornerstonejs/cornerstone3D/blob/main/packages/core/src/RenderingEngine/RenderingEngine.ts#L177-L180

I think its best to folow the same structure to not deviate

PR #2138, opened 2025-06-12
Fix suppress camera bug

rodrigobasilio2022:
### Context

This PR address a bug that occurs when one have multiple calls in the call stack for the methods resetCameraNoEvent and setCameraNoEvent. The first time one of these methods are called they set suppressCameraModifiedEvents flag to false breaking calls stacks where one didnt want to fire events until the flag is set to false by the programmers.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  

### DIFF
--- a/packages/core/src/RenderingEngine/Viewport.ts
+++ b/packages/core/src/RenderingEngine/Viewport.ts
@@ -718,7 +718,7 @@ class Viewport {
    * Reset the camera to the default viewport camera without firing events
    */
   protected resetCameraNoEvent(): void {
-    const savedValue = this._suppressCameraModifiedEvents;
+    const savedValue = this._suppressCameraModifiedEvents; // save the value pf the flag to restore it later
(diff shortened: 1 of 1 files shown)


packages/core/src/RenderingEngine/Viewport.ts
```diff
@@ -718,19 +718,21 @@ class Viewport {
    * Reset the camera to the default viewport camera without firing events
    */
   protected resetCameraNoEvent(): void {
+    const savedValue = this._suppressCameraModifiedEvents;
```
sedghi: can you please add a comment of what is this supposed to do thanks

wayfarer3130 (approved):

wayfarer3130:
Thanks Rodrigo - I'd been thinking about making the same change for a while now, just hadn't had it bubble up high enough.

PR #2139, opened 2025-06-12
feat: volume cropping 

mbellehumeur:
### Context
ticket: 9248395925

This pull request adds two tools:  VolumeCroppingTool and VolumeCroppingControlTool.

The VolumeCroppingTool acts on a VRT viewport and provides the ability to crop the volume using control spheres/handles along the x, y and z axis and the corners of the cropping volume.
The display of the control spheres can be enabled/disabled through the setHandlesVisible/ getHandlesVisible methods. 
The VolumeCroppingControlTool acts on 1 to 3 orthogonal viewports and adds reference lines that control and reacts to changes in the cropping volume.

The tools communicate through new cornerstone events:  VOLUMECROPPING_TOOL_CHANGED and  VOLUMECROPPINGCONTROL_TOOL_CHANGED.

Configurable items of the VolumeCroppingTool are the following:
        showCornerSpheres: true,
        showHandles: true,
        sphereColors: {
          x: [1.0, 1.0, 0.0], //  Yellow for X
          y: [0.0, 1.0, 0.0], // Green for Y
          z: [1.0, 0.0, 0.0], // Red for Z
          corners: [0.0, 0.0, 1.0], // Blue for corners
        },
        sphereRadius: 10,

A new 'VolumeCroppingTool' example demontrates the two tools working together and a button to toggle the display of the control spheres:

image

### Changes & Results

New files are VolumeCropping, VolumeCroppingControl and index.js for the example.

The Pan, TrackballRotate and Zoom 3D tools were modified to send events to the volume cropping tool in order to adjust  the clipping planes .  The new events are  PAN_TOOL_CHANGED, ZOOM_TOOL_CHANGED,TRACKBALLROTATE_TOOL_CHANGE. 

### Testing

Run the example and use the spheres to modify the cropping.  The volume should change the control lines in the orthogonal views as well.
Use the control lines in the orthogonal views to change the cropping volume.
Use the 'Toggle 3D handles' button to hide/show the control spheres.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  


packages/tools/src/tools/PanTool.ts
```diff
@@ -58,6 +127,8 @@ class PanTool extends BaseTool {
       focalPoint: updatedFocalPoint,
       position: updatedPosition,
     });
+    // Update clipping planes after pan
+    this._updateClippingPlanes(enabledElement.viewport);
```
igoroctaviano: This should belong to the cropping tool / domain, not the PanTool. Maybe you can use mouse drag events ```packages/tools/src/enums/Events.ts```?

packages/tools/src/tools/PanTool.ts
```diff
@@ -25,6 +26,74 @@ class PanTool extends BaseTool {
...
+      matrix[5],
+      matrix[6],
+      matrix[8],
+      matrix[9],
+      matrix[10],
+    ];
+
+    // --- Get the VolumeCroppingTool instance for this viewport ---
+    const toolGroup = getToolGroup(this.toolGroupId);
+    const croppingTool = toolGroup?.getToolInstance?.('VolumeCroppingTool');
```
igoroctaviano: It's better to import the tool and use its static name prop e.g. SomeTool.toolName.

packages/tools/src/tools/PanTool.ts
```diff
@@ -25,6 +26,74 @@ class PanTool extends BaseTool {
   mouseDragCallback(evt: EventTypes.InteractionEventType) {
     this._dragCallback(evt);
   }
+  _transformNormal(normal: Types.Point3, mat: number[]): Types.Point3 {
```
igoroctaviano: Maybe you can use something from gl-matrix (vec3) to do this? Rather than manually multiplying out the matrix. If there's still something specific here you want this moved to ```packages/tools/src/utilities/geometricSurfaceUtils.ts``` so it becomes reusable.

packages/tools/src/tools/PanTool.ts
```diff
@@ -25,6 +26,74 @@ class PanTool extends BaseTool {
...
+          matrix[5] * origin[1] +
+          matrix[9] * origin[2] +
+          matrix[13],
+        matrix[2] * origin[0] +
+          matrix[6] * origin[1] +
+          matrix[10] * origin[2] +
+          matrix[14],
+      ];
+      // Transform normal (rotation only)
+      const n: Types.Point3 = this._transformNormal(normal, rot);
```
igoroctaviano: Maybe after transforming the normal, you may want to normalize it to ensure its a unit vector?: ```const n: Types.Point3 = vec3.normalize([], this._transformNormal(normal, rot));```

packages/tools/src/tools/PanTool.ts
```diff
@@ -25,6 +26,74 @@ class PanTool extends BaseTool {
...
+        typeof plane.getOrigin === 'function'
+          ? plane.getOrigin()
+          : plane.origin;
+      const normal =
+        typeof plane.getNormal === 'function'
+          ? plane.getNormal()
+          : plane.normal;
+
+      // Transform origin (full 4x4)
+      const o: Types.Point3 = [
```
igoroctaviano: Something that might not apply here but I got it from code analysis: normals should be transformed by the inverse transpose of the rotation matrix if the matrix includes non-uniform scaling or shearing. If your transforms are always pure rotation (or uniform scaling), this is fine. If not, you may get incorrect results for the normal direction. If you want to be robust, you should compute the inverse transpose of the 3x3 rotation part and use that to transform the normal.

packages/tools/src/tools/TrackballRotateTool.ts
```diff
@@ -152,6 +154,76 @@ class TrackballRotateTool extends BaseTool {
...
+          matrix[4] * origin[1] +
+          matrix[8] * origin[2] +
+          matrix[12],
+        matrix[1] * origin[0] +
+          matrix[5] * origin[1] +
+          matrix[9] * origin[2] +
+          matrix[13],
+        matrix[2] * origin[0] +
+          matrix[6] * origin[1] +
+          matrix[10] * origin[2] +
```
igoroctaviano: Same here and any other tool. Code from one tool cannot live in another. This should be under the volume cropping tool domain. Use existent events to do this, or create new ones if not existent.

packages/tools/src/tools/TrackballRotateTool.ts
```diff
@@ -191,7 +266,10 @@ class TrackballRotateTool extends BaseTool {
     const { rotateIncrementDegrees } = this.configuration;
     const enabledElement = getEnabledElement(element);
     const { viewport } = enabledElement;
-
+    if (viewport.type !== Enums.ViewportType.VOLUME_3D) {
```
igoroctaviano: This tool should only be active for volume 3d. Is this check necessary?

packages/tools/src/tools/ZoomTool.ts
```diff
@@ -46,6 +48,74 @@ class ZoomTool extends BaseTool {
     }
     this.mouseDragCallback = this._dragCallback.bind(this);
   }
+  // Helper to transform a normal by a 3x3 matrix
+  _transformNormal(normal: Types.Point3, mat: number[]): Types.Point3 {
```
igoroctaviano: Same here!

packages/tools/src/tools/ZoomTool.ts
```diff
@@ -46,6 +48,74 @@ class ZoomTool extends BaseTool {
...
+    return [
+      mat[0] * normal[0] + mat[3] * normal[1] + mat[6] * normal[2],
+      mat[1] * normal[0] + mat[4] * normal[1] + mat[7] * normal[2],
+      mat[2] * normal[0] + mat[5] * normal[1] + mat[8] * normal[2],
+    ];
+  }
+
+  _updateClippingPlanes(viewport) {
```
igoroctaviano: Also, you have the same code across different tools; we should always extract it into a utility and reuse it. But I think these things should not live in Zoom, rotate, etc, only in the volume cropping tool. If existing events are not enough, we could think about adding events for these tools and indirectly using them from the volume cropping domain.

packages/tools/src/tools/VolumeCroppingTool.ts
```diff
@@ -0,0 +1,1534 @@
...
+  viewport,
+  point1,
+  point2,
+  radius = 0.5,
+  color: [number, number, number] = [0.5, 0.5, 0.5],
+  uid = ''
+) {
+  const cylinderSource = vtkCylinderSource.newInstance();
+  // Compute direction and length
+  const direction = [
```
igoroctaviano: I think you need to normalize direction vector, no? something like:
```
const dirNorm = Math.sqrt(direction[0]**2 + direction[1]**2 + direction[2]**2);
const directionNormalized = [
  direction[0] / dirNorm,
  direction[1] / dirNorm,
  direction[2] / dirNorm,
];
cylinderSource.setDirection(...directionNormalized);
```

packages/tools/src/tools/VolumeCroppingTool.ts
```diff
@@ -0,0 +1,1534 @@
...
+  XMAX_YMAX_ZMIN: 12,
+  XMAX_YMAX_ZMAX: 13,
+};
+const POINTINDEX = {
+  X: 0,
+  Y: 1,
+  Z: 2,
+};
+
+function addCylinderBetweenPoints(
```
igoroctaviano: If point1 and point2 are the same, length will be zero, and the direction will be [0,0,0].
This will likely cause errors or invisible geometry. Maybe you need to add a check to avoid creating a cylinder if the points are the same.

packages/tools/src/tools/VolumeCroppingTool.ts
```diff
@@ -0,0 +1,1534 @@
...
+    const spacing = imageData.getSpacing(); // [xSpacing, ySpacing, zSpacing]
+    const cropFactor = 0.2;
+    const xMin = origin[0] + cropFactor * (dimensions[0] - 1) * spacing[0];
+    const xMax =
+      origin[0] + (1 - cropFactor) * (dimensions[0] - 1) * spacing[0];
+    const yMin = origin[1] + cropFactor * (dimensions[1] - 1) * spacing[1];
+    const yMax =
+      origin[1] + (1 - cropFactor) * (dimensions[1] - 1) * spacing[1];
+    const zMin = origin[2] + cropFactor * (dimensions[2] - 1) * spacing[2];
+    const zMax = origin[2] + 0.8 * (dimensions[2] - 1) * spacing[2];
```
igoroctaviano: What 0.8 means?

packages/tools/src/tools/VolumeCroppingTool.ts
```diff
@@ -0,0 +1,1534 @@
...
+    sphereActor.setPickable(true);
+    viewport.addActor({ actor: sphereActor, uid: uid });
+  }
+
+  _initialize3DViewports = (viewportsInfo): void => {
+    const [viewport3D] = viewportsInfo;
+    const renderingEngine = getRenderingEngine(viewport3D.renderingEngineId);
+    const viewport = renderingEngine.getViewport(viewport3D.viewportId);
+    const volumeActors = viewport.getActors();
+    const imageData = volumeActors[0].actor.getMapper().getInputData();
```
igoroctaviano: Always check for volume actors to avoid runtime errros.

packages/tools/src/tools/VolumeCroppingTool.ts
igoroctaviano: You use zMax = origin[2] + 0.8 * (dimensions[2] - 1) * spacing[2];
This is inconsistent with the other axes, which use (1 - cropFactor).
If this is intentional, that's fine, but if not, it should be: ```    const zMax = origin[2] + (1 - cropFactor) * (dimensions[2] - 1) * spacing[2];```

packages/tools/src/tools/VolumeCroppingTool.ts
```diff
@@ -0,0 +1,1534 @@
...
+            uid
+          );
+          this.edgeCylinders[uid] = { actor, source, key1, key2 };
+        }
+      });
+    }
+
+    const defaultActor = viewport.getDefaultActor();
+    const actor = defaultActor.actor as vtkActor | vtkVolume;
+    if (actor && (actor.isA?.('vtkActor') || actor.isA?.('vtkVolume'))) {
```
igoroctaviano: Do you need these vtkactor and volume checks?

packages/tools/src/tools/VolumeCroppingTool.ts
```diff
@@ -0,0 +1,1534 @@
...
+        const otherXSphere = this.sphereStates.find(
+          (s, i) => s.axis === 'x' && i !== 0
+        );
+        const newXCenter =
+          (otherXSphere.point[0] + planeXmin.getOrigin()[0]) / 2;
+        this.sphereStates.forEach((state, idx) => {
+          if (
+            !state.isCorner &&
+            state.axis !== 'x' &&
+            !evt.detail.viewportOrientation.includes('sagittal') // sagittal is y axis in yellow
```
igoroctaviano: Use viewport orientation enums from CS3D insteasd of hardcoded values.

packages/tools/src/tools/VolumeCroppingTool.ts
```diff
@@ -0,0 +1,1534 @@
...
+        ['XMAX_YMIN_ZMAX', 'XMAX_YMAX_ZMAX'],
+        // Z edges
+        ['XMIN_YMIN_ZMIN', 'XMIN_YMIN_ZMAX'],
+        ['XMIN_YMAX_ZMIN', 'XMIN_YMAX_ZMAX'],
+        ['XMAX_YMIN_ZMIN', 'XMAX_YMIN_ZMAX'],
+        ['XMAX_YMAX_ZMIN', 'XMAX_YMAX_ZMAX'],
+      ];
+
+      edgeCornerPairs.forEach(([key1, key2], i) => {
+        const state1 = this.sphereStates.find(
```
igoroctaviano: At this point sphereStates has been initialized already?

packages/tools/src/tools/VolumeCroppingTool.ts
```diff
@@ -0,0 +1,1534 @@
...
+      clippingPlanes[PLANEINDEX.YMAX].setOrigin(planeYmax.getOrigin());
+      clippingPlanes[PLANEINDEX.ZMAX].setOrigin(planeZmax.getOrigin());
+    }
+    if (
+      this.configuration.showHandles &&
+      this.configuration.showCornerSpheres
+    ) {
+      this._updateCornerSpheres(viewport);
+    }
+    // this._updateCornerSpheres(viewport);
```
igoroctaviano: Remove comment

packages/tools/src/tools/VolumeCroppingTool.ts
```diff
@@ -0,0 +1,1534 @@
...
+        );
+        const newZCenter =
+          (otherZSphere.point[POINTINDEX.Z] +
+            planeZmax.getOrigin()[POINTINDEX.Z]) /
+          2;
+        this.sphereStates.forEach((state, idx) => {
+          if (
+            !state.isCorner &&
+            state.axis !== 'z' &&
+            !evt.detail.viewportOrientation.includes('axial')
```
igoroctaviano: Use enums!

packages/tools/src/tools/VolumeCroppingTool.ts
```diff
@@ -0,0 +1,1534 @@
...
+      });
+      const planeYmax = vtkPlane.newInstance({
+        origin: [0, toolMax[1], 0],
+        normal: [0, -1, 0],
+      });
+      const planeZmax = vtkPlane.newInstance({
+        origin: [0, 0, toolMax[2]],
+        normal: [0, 0, -1],
+      });
+      // viewport.setOriginalClippingPlane(PLANEINDEX.XMAX, planeXmax.getOrigin());
```
igoroctaviano: Remove comments

packages/tools/src/tools/VolumeCroppingTool.ts
```diff
@@ -0,0 +1,1534 @@
...
+          }
+        });
+
+        // y
+        this.sphereStates[SPHEREINDEX.YMIN].point[1] = planeYmin.getOrigin()[1];
+        this.sphereStates[SPHEREINDEX.YMIN].sphereSource.setCenter(
+          this.sphereStates[SPHEREINDEX.YMIN].point
+        );
+        this.sphereStates[SPHEREINDEX.YMIN].sphereSource.modified();
+        const otherYSphere = this.sphereStates.find(
```
igoroctaviano: You assume otherXSphere, otherYSphere, and otherZSphere always exist. If not found, this will throw an error.

packages/tools/src/tools/VolumeCroppingTool.ts
```diff
@@ -0,0 +1,1534 @@
...
+        });
+      }
+      const volumeActor = viewport.getDefaultActor()?.actor;
+      if (!volumeActor) {
+        console.warn('No volume actor found');
+        return;
+      }
+      const mapper = volumeActor.getMapper() as vtkVolumeMapper;
+      const clippingPlanes = mapper.getClippingPlanes();
+      clippingPlanes[PLANEINDEX.XMIN].setOrigin(planeXmin.getOrigin());
```
igoroctaviano: You assume clippingPlanes[PLANEINDEX.XMIN] etc. exist. If the array is not initialized or is too short, this will throw.

packages/tools/src/tools/VolumeCroppingTool.ts
```diff
@@ -0,0 +1,1534 @@
...
+        this.sphereStates.forEach((state, idx) => {
+          if (
+            !state.isCorner &&
+            state.axis !== 'x' &&
+            !evt.detail.viewportOrientation.includes('sagittal')
+          ) {
+            state.point[POINTINDEX.X] = newXCenter;
+            state.sphereSource.setCenter(state.point);
+            state.sphereActor.getProperty().setColor(state.color);
+            state.sphereSource.modified();
```
igoroctaviano: Do you need modified after setCenter?

packages/tools/src/tools/VolumeCroppingTool.ts
```diff
@@ -0,0 +1,1534 @@
...
+    const viewport = renderingEngine.getViewport(viewport3D.viewportId);
+    const mouseCanvas = [evt.offsetX, evt.offsetY];
+    // Find the sphere under the mouse
+    for (let i = 0; i < this.sphereStates.length; ++i) {
+      const sphereCanvas = viewport.worldToCanvas(this.sphereStates[i].point);
+      const dist = Math.sqrt(
+        Math.pow(mouseCanvas[0] - sphereCanvas[0], 2) +
+          Math.pow(mouseCanvas[1] - sphereCanvas[1], 2)
+      );
+      if (dist < 20) {
```
igoroctaviano: What 20 means? maybe name it

packages/tools/src/tools/VolumeCroppingTool.ts
```diff
@@ -0,0 +1,1534 @@
...
+        // 20 pixels threshold
+        this.draggingSphereIndex = i;
+        element.style.cursor = 'grabbing';
+        return;
+      }
+    }
+    this.draggingSphereIndex = null;
+  };
+
+  _onMouseMoveSphere = (evt) => {
```
igoroctaviano: This is a big function, if possible extract code into separate named functions or utils so the tool is not too big and hard to understand.

packages/tools/src/tools/VolumeCroppingTool.ts
```diff
@@ -0,0 +1,1534 @@
...
+          draggingSphereIndex: this.draggingSphereIndex,
+        });
+      }
+    }
+  };
+
+  _updateCornerSpheres(viewport) {
+    // Get current face sphere positions
+    const xMin =
+      this.sphereStates.find((s) => s.axis === 'x' && s.point[0] <= s.point[1])
```
igoroctaviano: You use s.point[0] <= s.point[1] to determine min/max for X, and similar for Y/Z.
This assumes that for a face sphere, the min/max axis value is always less/greater than the other axis value, which is not generally true. If the face spheres are not always aligned this way, you may get the wrong min/max.

You can use the known indices or uids for your face spheres (e.g., SPHEREINDEX.XMIN, SPHEREINDEX.XMAX, etc.) to directly get the min/max values e.g. 

```
const xMin = this.sphereStates[SPHEREINDEX.XMIN].point[0];
const xMax = this.sphereStates[SPHEREINDEX.XMAX].point[0];
const yMin = this.sphereStates[SPHEREINDEX.YMIN].point[1];
const yMax = this.sphereStates[SPHEREINDEX.YMAX].point[1];
const zMin = this.sphereStates[SPHEREINDEX.ZMIN].point[2];
const zMax = this.sphereStates[SPHEREINDEX.ZMAX].point[2];
```

packages/tools/src/tools/VolumeCroppingTool.ts
```diff
@@ -0,0 +1,1534 @@
...
+          direction[0] ** 2 + direction[1] ** 2 + direction[2] ** 2
+        );
+        const center = [
+          (point1[0] + point2[0]) / 2,
+          (point1[1] + point2[1]) / 2,
+          (point1[2] + point2[2]) / 2,
+        ];
+        source.setCenter(center[0], center[1], center[2]);
+        source.setHeight(length);
+        source.setDirection(direction[0], direction[1], direction[2]);
```
igoroctaviano: Maybe you should normalize the direction vector before calling setDirection, as VTK expects a unit vector for direction?

packages/tools/src/tools/VolumeCroppingControlTool.ts
```diff
@@ -0,0 +1,1771 @@
...
+    if (viewportsInfo && viewportsInfo.length > 0) {
+      const { viewportId, renderingEngineId } = viewportsInfo[0];
+      const enabledElement = getEnabledElementByIds(
+        viewportId,
+        renderingEngineId
+      );
+      const renderingEngine = getRenderingEngine(renderingEngineId);
+      const viewport = renderingEngine.getViewport(viewportId);
+      const volumeActors = viewport.getActors();
+      const imageData = volumeActors[0].actor.getMapper().getInputData();
```
igoroctaviano, line 264: Always check for volume actors to avoid runtime errors.

packages/tools/src/tools/VolumeCroppingControlTool.ts
```diff
@@ -0,0 +1,1771 @@
...
+      const { element } = viewport;
+
+      element.addEventListener(
+        Enums.Events.VOLUME_VIEWPORT_NEW_VOLUME,
+        this._onNewVolume
+      );
+    });
+  }
+
+  _areViewportIdArraysEqual = (viewportIdArrayOne, viewportIdArrayTwo) => {
```
igoroctaviano: The function always returns true unless the lengths differ or the last item is not found.
If any item is not found, it returns false from inside the forEach, but this does not break out of the function, so the function will still return true at the end. Maybe you can use a for loop or every instead of forEach for early return.

packages/tools/src/tools/VolumeCroppingControlTool.ts
```diff
@@ -0,0 +1,1771 @@
...
+    const { element } = eventDetail;
+
+    this.editData.annotation.data.handles.activeOperation = null;
+    this.editData.annotation.data.activeViewportIds = [];
+
+    this._deactivateModify(element);
+
+    resetElementCursor(element);
+
+    this.editData = null;
```
igoroctaviano, line 1773: This is being used?

packages/tools/src/tools/VolumeCroppingControlTool.ts
```diff
@@ -0,0 +1,1771 @@
...
+    // filter the annotations to only keep that are for this toolGroup
+    const toolGroupAnnotations = annotations.filter((annotation) => {
+      const { data } = annotation;
+      return viewportIds.includes(data.viewportId);
+    });
+
+    return toolGroupAnnotations;
+  };
+
+  _onSphereMoved = (evt) => {
```
igoroctaviano, line 1550: The logic for updating min/max/corner spheres is a bit hard to follow and could be error-prone if the indices change. Consider using named constants or enums for clarity.

igoroctaviano (changes requested):
Great work here Martin! 🎖️  There's a decent chunk to revise, but nothing majorly off — just some cleanup and alignment needed.

packages/tools/src/tools/VolumeCroppingTool.ts
mbellehumeur: Removed

packages/tools/src/tools/ZoomTool.ts
```diff
@@ -46,6 +48,74 @@ class ZoomTool extends BaseTool {
...
+          matrix[6] * origin[1] +
+          matrix[10] * origin[2] +
+          matrix[14],
+      ];
+      // Transform normal (rotation only)
+      const n: Types.Point3 = this._transformNormal(normal, rot);
+      const planeInstance = vtkPlane.newInstance({ origin: o, normal: n });
+      mapper.addClippingPlane(planeInstance);
+    });
+  }
```
sedghi: so why are we touching clpping plane in zoom? is it only for 3d viewport?

packages/tools/src/tools/PanTool.ts
```diff
@@ -25,6 +26,74 @@ class PanTool extends BaseTool {
...
+      ];
+      // Transform normal (rotation only)
+      const n: Types.Point3 = this._transformNormal(normal, rot);
+      const planeInstance = vtkPlane.newInstance({ origin: o, normal: n });
+      mapper.addClippingPlane(planeInstance);
+    });
+  }
 
   _dragCallback(evt: EventTypes.InteractionEventType) {
     const { element, deltaPoints } = evt.detail;
```
sedghi: so why are we touching clpping plane in pan? is it only for 3d viewport?

packages/tools/src/tools/VolumeCroppingTool.ts
```diff
@@ -0,0 +1,1622 @@
...
+    );
+    eventTarget.addEventListener(Events.ZOOM_TOOL_CHANGED, (evt) => {
+      this._updateClippingPlanes(evt.detail.viewport);
+    });
+    eventTarget.addEventListener(Events.PAN_TOOL_CHANGED, (evt) => {
+      this._updateClippingPlanes(evt.detail.viewport);
+    });
+    eventTarget.addEventListener(Events.TRACKBALLROTATE_TOOL_CHANGED, (evt) => {
+      this._updateClippingPlanes(evt.detail.viewport);
+    });
```
sedghi: The `zoom_changed`, `pan_changed`, and `trackball_changed` events essentially just send the viewport to `updateClippingPlanes`. I'm wondering why we don't simply subscribe to the camera event and run it when changed.

sedghi (changes requested):
so when i removed the coronal viewport this happened

CleanShot 2025-07-11 at 10 30 15@2x

we should not tie this tool to specific layout in the screenshot above, the axial and sagittal already have enough information to project the cropper on their planes, but they don't show anything

packages/tools/src/tools/VolumeCroppingTool.ts
mbellehumeur: Please check, I tried adding CAMERA_MODIFIED events to pan, zoom, trackball tools as I was not receiving any. 
There is some conflict between 
import { Events } from '../enums';
and 
import {Enums,} from '@cornerstonejs/core';
const { Events } = Enums;
So I hardcoded the event name in a couple of places.  Perhaps you can guide me here.

packages/tools/src/tools/VolumeCroppingControlTool.ts
mbellehumeur: Thanks.  removed .That's a remnant of the crosshair tool (which I started with) and is not used.

packages/tools/src/tools/VolumeCroppingControlTool.ts
mbellehumeur: yes, the mouse cursor disappears if removed.

packages/tools/src/tools/VolumeCroppingTool.ts
sedghi: maybe it fires on the element?

packages/tools/src/tools/PanTool.ts
mbellehumeur: updated in VolumeCroppingTool

packages/tools/src/tools/VolumeCroppingTool.ts
mbellehumeur: I tried that first.  I try again tomorrow.

packages/tools/src/tools/VolumeCroppingTool.ts
mbellehumeur: Added the check.

packages/tools/src/tools/VolumeCroppingTool.ts
mbellehumeur: I think it's a bug in Viewport.ts
image

packages/tools/src/tools/VolumeCroppingTool.ts
mbellehumeur: Now in configuration under
grabSpherePixelDistance: 20, //pixels threshold for closeness to the sphere being grabbed

packages/tools/src/tools/PanTool.ts
mbellehumeur: Pan, zoom and rotate changes have been replaces by camera-modified event.

mbellehumeur:
> I don't understand the changes to zoom and pan and trackball rotate tool. There is certainly other ways to do that

The changes to zoom and pan and trackball rotate tool have been removed.

sedghi:
@mbellehumeur let me know when you fix this issue so that i can review again https://github.com/cornerstonejs/cornerstone3D/pull/2139#pullrequestreview-3010550044

mbellehumeur:
> @mbellehumeur let me know when you fix this issue so that i can review again [#2139 (review)](https://github.com/cornerstonejs/cornerstone3D/pull/2139#pullrequestreview-3010550044)

I unfortunately cannot spend more time on the 2 viewport  feature in the current project.

packages/tools/src/tools/VolumeCroppingTool.ts
mbellehumeur: added a check

packages/tools/src/tools/VolumeCroppingTool.ts
mbellehumeur: refactored

packages/tools/src/tools/VolumeCroppingTool.ts
mbellehumeur: reworked

packages/tools/src/tools/VolumeCroppingTool.ts
mbellehumeur: Yes, this is done by addSphere just above.

sedghi (changes requested):
Right now, the implementation is tightly coupled to having all three views (axial, sagittal, coronal) present. If I remove, for example, the coronal view, it fails on the other two — even though those remaining views already have sufficient data to render the crop cube projection.

CleanShot 2025-07-17 at 10 21 55@2x


This is very similar to the issue we ran into with the Crosshairs implementation back in 2020, where it was also tied to the presence of all three views. That approach turned out to be problematic, and I’d like to avoid repeating that mistake here.

The volume cropper should be viewport-agnostic — it should work even if only a single planar view is present.

You already have the cube state at render time. At that point, you can check the viewport type (stack or volume), and if it’s planar, just project the cube onto the view plane normal and render it using SVG (or whatever method you're using). This should also work fine for stack viewports since those are just planes too.

Let me know if you want to discuss implementation details — happy to chat!

mbellehumeur:
> Right now, the implementation is tightly coupled to having all three views (axial, sagittal, coronal) present. If I remove, for example, the coronal view, it fails on the other two — even though those remaining views already have sufficient data to render the crop cube projection.
> 
> CleanShot 2025-07-17 at 10 21 55@2x
> This is very similar to the issue we ran into with the Crosshairs implementation back in 2020, where it was also tied to the presence of all three views. That approach turned out to be problematic, and I’d like to avoid repeating that mistake here.
> 
> The volume cropper should be viewport-agnostic — it should work even if only a single planar view is present.
> 
> You already have the cube state at render time. At that point, you can check the viewport type (stack or volume), and if it’s planar, just project the cube onto the view plane normal and render it using SVG (or whatever method you're using). This should also work fine for stack viewports since those are just planes too.
> 
> Let me know if you want to discuss implementation details — happy to chat!

image

image

packages/tools/examples/volumeCroppingTool/index.ts
```diff
@@ -0,0 +1,420 @@
...
+  BaseVolumeViewport,
+  Types,
+  VolumeViewport3D,
+} from '@cornerstonejs/core';
+import {
+  RenderingEngine,
+  Enums,
+  setVolumesForViewports,
+  volumeLoader,
+  getRenderingEngine,
```
sedghi: There are many unused variables in this file, so please remove them. For example, `BaseVolumeViewport` and `getRenderingEngine` are unused, and my IDE shows them as faded, which is how I identified them.

packages/tools/src/tools/VolumeCroppingTool.ts
```diff
@@ -0,0 +1,1432 @@
...
+  getEnabledElement,
+  utilities as csUtils,
+  Enums,
+  triggerEvent,
+  eventTarget,
+} from '@cornerstonejs/core';
+
+import { getToolGroup } from '../store/ToolGroupManager';
+
+import { state } from '../store/state';
```
sedghi: same here, state is unused, InteractionTypes below etc

packages/tools/src/tools/VolumeCroppingTool.ts
```diff
@@ -0,0 +1,1432 @@
...
+/**
+ * VolumeCroppingTool is a tool that provides clipping planes to crop a volume
+ */
+class VolumeCroppingTool extends BaseTool {
+  // Store 2D edge lines between corner spheres
+  edgeLines: {
+    [uid: string]: {
+      actor: vtkActor;
+      source: vtkPolyData;
+      key1: string;
```
sedghi: please comment above what is key1 and key2, or rename better

packages/tools/src/tools/VolumeCroppingTool.ts
```diff
@@ -0,0 +1,1432 @@
...
+      key2: string;
+    };
+  } = {};
+
+  // Helper to add a 3D line between two points using vtkActor
+  addLine3DBetweenPoints(
```
sedghi: please move this function (method) after the properties definitions

sedghi: seems like it is actually before constructor which is wrong

packages/tools/src/tools/VolumeCroppingTool.ts
```diff
@@ -0,0 +1,1432 @@
...
+        sphereColors: {
+          x: [1.0, 1.0, 0.0], //  Yellow for X
+          y: [0.0, 1.0, 0.0], // Green for Y
+          z: [1.0, 0.0, 0.0], // Red for Z
+          corners: [0.0, 0.0, 1.0], // Blue for corners
+        },
+        sphereRadius: 8,
+        grabSpherePixelDistance: 20, //pixels threshold for closeness to the sphere being grabbed
+        rotateIncrementDegrees: 2,
+        rotateSampleDistanceFactor: 2, // Factor to increase sample distance (lower resolution) when rotating
```
sedghi: please add proper documentation above each of these, you can look at other files such as packages/tools/src/tools/annotation/SplineROITool.ts to see how we comment

packages/tools/src/tools/VolumeCroppingTool.ts
```diff
@@ -0,0 +1,1432 @@
...
+  }
+
+  setHandlesVisible(visible: boolean) {
+    this.configuration.showHandles = visible;
+    // Before showing, update sphere positions to match clipping planes
+    if (visible) {
+      const viewportsInfo = this._getViewportsInfo();
+      const [viewport3D] = viewportsInfo;
+      const renderingEngine = getRenderingEngine(viewport3D.renderingEngineId);
+      const viewport = renderingEngine.getViewport(viewport3D.viewportId);
```
sedghi: viewport is unused

sedghi: seems like this logic can get extracted out, as it is used after if (visible)

packages/tools/src/tools/VolumeCroppingTool.ts
```diff
@@ -0,0 +1,1432 @@
...
+    this._triggerToolChangedEvent(sphereState);
+
+    return true;
+  };
+
+  // Helper method to get viewport and world coordinates
+  _getViewportAndWorldCoords = (evt) => {
+    const [viewport3D] = this._getViewportsInfo();
+    const renderingEngine = getRenderingEngine(viewport3D.renderingEngineId);
+    const viewport = renderingEngine.getViewport(viewport3D.viewportId);
```
sedghi: this is shared a lot in the code, maybe refactor to a private method _getViewport or something

packages/tools/src/tools/VolumeCroppingTool.ts
```diff
@@ -0,0 +1,1432 @@
...
+      : evt.detail;
+    const enabledElement = getEnabledElement(element);
+    this._updateClippingPlanes(enabledElement.viewport);
+    enabledElement.viewport.render();
+  };
+
+  _onNewVolume = () => {
+    const viewportsInfo = this._getViewportsInfo();
+    this._initialize3DViewports(viewportsInfo);
+  };
```
sedghi: there is a mix of order between public and private methods, can we have all public methods first then _private ones

packages/tools/src/tools/VolumeCroppingTool.ts
```diff
@@ -0,0 +1,1432 @@
...
+      const { element } = viewport;
+
+      element.addEventListener(
+        Enums.Events.VOLUME_VIEWPORT_NEW_VOLUME,
+        this._onNewVolume
+      );
+    });
+  }
+
+  preMouseDownCallback = (evt: EventTypes.InteractionEventType) => {
```
sedghi: move way up

packages/tools/src/tools/VolumeCroppingTool.ts
```diff
@@ -0,0 +1,1432 @@
...
+    mat4.identity(transform);
+    mat4.rotate(transform, transform, angle, axis);
+    vec3.transformMat4(newViewUp, viewUp, transform);
+
+    viewport.setCamera({
+      position: newPosition,
+      viewUp: newViewUp,
+      focalPoint: newFocalPoint,
+    });
+  };
```
sedghi: why this tool rotates the camera? hmmmm i don't follow

packages/tools/src/tools/VolumeCroppingControlTool.ts
```diff
@@ -0,0 +1,1657 @@
...
+          opacity: 0.8,
+        },
+        lineColors: {
+          AXIAL: [1.0, 0.0, 0.0], //  Red for axial
+          CORONAL: [0.0, 1.0, 0.0], // Green for coronal
+          SAGITTAL: [1.0, 1.0, 0.0], // Yellow for sagittal
+          UNKNOWN: [0.0, 0.0, 1.0], // Blue for unknown
+        },
+        lineWidth: 1.5,
+        lineWidthActive: 2.5,
```
sedghi, line 229: documentation for each on top

packages/tools/src/tools/VolumeCroppingControlTool.ts
```diff
@@ -0,0 +1,1937 @@
...
+
+      vtkMath.add(camera.focalPoint, projectedDelta, newFocalPoint);
+      vtkMath.add(camera.position, projectedDelta, newPosition);
+
+      viewport.setCamera({
+        focalPoint: newFocalPoint,
+        position: newPosition,
+      });
+      viewport.render();
+    }
```
igoroctaviano, line 1892: In the next iteration, I think it would be a good idea to extract some utility functions from this tool in the future to help keep it more maintainable. Hitting 2,000 lines of code is starting to feel a bit overwhelming.

igoroctaviano (approved):
LGTM, great work @mbellehumeur

packages/tools/src/tools/VolumeCroppingControlTool.ts
mbellehumeur: Thanks. Agreed, we started from the crosshairs tool at 2700 lines… Perhaps we could add single/dual orthographic viewports to crosshairs and reduce both tools line counts and with a common utility file.  Maybe merge the cropping tools also.

PR #2155, opened 2025-06-23
fix(acquisition): fix bug in acquisition mode

rodrigobasilio2022:
### Context

This PR fixes a regression introduced by the recent reformatting PR, which inadvertently broke support for the ACQUISITION orientation mode.

Issue:
The reformatting implementation does not support ACQUISITION orientation, but the reformatting PR had rearranged conditional branches in a way that overlooked the specific check for acquisition mode, resulting in unexpected behavior.

Fix:
The condition order has been updated to explicitly prioritize the ACQUISITION check. This ensures that if the user selects acquisition orientation, the system correctly routes and processes it before falling through to other conditions.

Impact:
Users can now reliably select and use the ACQUISITION orientation mode again.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  

### DIFF
--- a/packages/core/src/RenderingEngine/VolumeViewport.ts
+++ b/packages/core/src/RenderingEngine/VolumeViewport.ts
@@ -85,7 +85,10 @@ class VolumeViewport extends BaseVolumeViewport {
       );
     }
 
-    if (
+    if (this._useAcquisitionPlaneForViewPlane) {
+      this._setViewPlaneToAcquisitionPlane(firstImageVolume);
+      this._useAcquisitionPlaneForViewPlane = false;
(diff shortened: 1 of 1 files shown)


wayfarer3130:
Can you update your description with a link to the bug, and a description of how to test using the standard examples dataset (I can modify an example, just want to know which one to look at)

anaisabel1909:
You can change the mprReformat example, changing the orientation of the last viewports to AQUISITION. Before the fix, you notice an image different from the first axial image. After the fix you will see the same image

PR #2159, opened 2025-06-25
Draft - feat(renderingEngine): multi-offscreen canvases

IbrahimCSAE:
### Offscreen Canvas Cropping and Misaligned Viewports Due to Canvas/Browser Limits

**Background**
Cornerstone3D uses a single, large offscreen canvas for all viewports. This canvas grows horizontally as more viewports are added. When rendering, each viewport pulls its image data from a region of this shared canvas. This strategy works well, until it doesn’t.

**The Problem**
Chrome (and most modern browsers) enforce a maximum WebGL canvas size of **16,384 × 16,384**. Once the combined width of all viewports exceeds that (e.g., on ultra-wide screens, high-DPI settings, or a high number of side-by-side viewports), the offscreen canvas silently overflows. Viewports that should render beyond this threshold start getting cropped.

Since Cornerstone3D copies from the offscreen canvas to the on-screen viewports using fixed source coordinates, this cropping causes serious misalignment.

**Visual Evidence**
Before hitting the limit (expected behavior):

image1
image2
image3
image4

---

#### **Approach 1: One RenderingEngine Per Viewport**

This would give each viewport its own offscreen canvas. Technically possible, but practically a dead end:

* **Volume Viewports must share a RenderingEngine**. Breaking this would break volume rendering.
* Cornerstone3D and OHIF assume a **single shared engine**—tools like Crosshairs, Reference Lines, and Sync rely on this.
* Dynamic creation/teardown of RenderingEngines causes instability in OHIF due to how `cornerstoneViewportService` and React’s lifecycle interact, many black viewports occur.
* WebGL context limits (usually **16 per tab**) mean this approach caps us at 16 viewports max, regardless of layout or canvas size.

#### **Approach 2: Multi-Offscreen Strategy (In this PR)**

In this approach, multiple offscreen canvases are created (2–3 recommended). Viewports are evenly distributed across them. Important constraints and tradeoffs:

* **Volume Viewports must all share the same offscreen canvas**. Volumes cannot span canvases, same as they can't span RenderingEngines.
* Stack viewports are spread across remaining offscreen canvases to balance usage.
* Works well with **2 or 3 offscreens**. Beyond that, performance starts to degrade.
* This approach avoids the WebGL canvas size cap and also stays within the context limit.

---

### Behavior in Other Viewers

We also tested how similar viewers handle this:

| Viewer    | Strategy                     | Fails >16 Viewports |
| --------- | ---------------------------- | ------------------- |
| MedDream  | 1 WebGL context per viewport | ✅ Yes               |
| VolView   | 1 WebGL context per viewport | ✅ Yes               |

Demo: MedDream fails once the 16-context limit is hit

https://github.com/user-attachments/assets/7808ef82-da24-42cb-aa60-3d14c0383433

---

### How to Reproduce the Bug in OHIF

To reliably hit the canvas size limit:

1. Fullscreen the viewer on a wide/high-res display, if you don't have a high res one don't worry, you can still do it.
2. Zoom out Chrome to **25%**
3. In DevTools Console, run:

   ```js
   window.devicePixelRatio = 2; // or 3, 4...
   ```

   Then trigger a re-render by resizing the browser or calling `renderingEngine.resize()`
4. Eventually, viewports will appear misaligned or blank — they’re being cropped from the offscreen canvas.

---

### Possible Solution

* **viewport packing logic** to make more efficient use of offscreen space.

---

### How to run this PR in OHIF

Link cornerstone3D locally, and in `cornerstoneViewportService`, change the `new RenderingEngine(id)` call to be `new RenderingEngine(id, 2)` or number of offscreen canvases desired

### DIFF
--- a/packages/core/src/RenderingEngine/BaseVolumeViewport.ts
+++ b/packages/core/src/RenderingEngine/BaseVolumeViewport.ts
@@ -1629,7 +1629,7 @@ abstract class BaseVolumeViewport extends Viewport {
     const renderer = this.getRenderer();
     const displayCoords = this.getVtkDisplayCoords(canvasPos);
     const offscreenMultiRenderWindow =
-      this.getRenderingEngine().offscreenMultiRenderWindow;
+      this.getRenderingEngine().getOffScreenMultiRenderWindow(this.id);
     const openGLRenderWindow =
       offscreenMultiRenderWindow.getOpenGLRenderWindow();
     const worldCoord = openGLRenderWindow.displayToWorld(
@@ -1658,7 +1658,7 @@ abstract class BaseVolumeViewport extends Viewport {
       canvasPos[1] * devicePixelRatio,
     ];
     const offscreenMultiRenderWindow =
-      this.getRenderingEngine().offscreenMultiRenderWindow;
+      this.getRenderingEngine().getOffScreenMultiRenderWindow(this.id);
     const openGLRenderWindow =
       offscreenMultiRenderWindow.getOpenGLRenderWindow();
     const size = openGLRenderWindow.getSize();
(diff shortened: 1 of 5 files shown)


IbrahimCSAE:
I don’t think the multi offscreen is fast enough, I think this leaves canvas packing as the only option.

sedghi:
I've tried packing it is too slow too unfortunately

IbrahimCSAE:
Well, not sure if there’s a solution at this point :/

PR #2160, opened 2025-06-26
feat: enable historyMemo action grouping

pedrokohler:
### Context

Currently, if you have a batch of actions, such as clearing multiple annotations at once, cornerstone is not capable of track that and undo/redo these batched actions.

This PR introduces an api of start/end recording, which will enable consumer apps to group actions that are related. 

### Changes & Results

Example of Before/After with OHIF
Before:

https://github.com/user-attachments/assets/a8bc2197-04bf-4199-8ef9-0cf46f7c1fd2

After:

https://github.com/user-attachments/assets/a0f44eec-ee75-405c-b060-a43f99daee0f

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  

### DIFF
--- a/packages/core/src/utilities/historyMemo/index.ts
+++ b/packages/core/src/utilities/historyMemo/index.ts
@@ -89,7 +89,7 @@ export class HistoryMemo {
     while (items > 0 && this.undoAvailable > 0) {
       const item = this.ring[this.position];
 
-      for (const subitem of asArray(item)) {
+      for (const subitem of asArray(item).reverse()) {
         subitem.restoreMemo(true);
         this.dispatchHistoryEvent({ item: subitem, isUndo: true });
       }
@@ -143,7 +143,7 @@ export class HistoryMemo {
       const newPosition = (this.position + 1) % this.size;
       const item = this.ring[newPosition];
 
-      for (const subitem of asArray(item)) {
+      for (const subitem of asArray(item).reverse()) {
         subitem.restoreMemo(false);
         this.dispatchHistoryEvent({ item: subitem, isUndo: false });
       }
--- a/packages/core/src/utilities/historyMemo/index.ts
+++ b/packages/core/src/utilities/historyMemo/index.ts
@@ -1,4 +1,5 @@
 import eventTarget from '../../eventTarget';
+import { asArray } from '../asArray';
 
 // Define Events from tools package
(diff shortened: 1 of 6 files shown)


wayfarer3130:
Is there one of the CS3D examples that can demonstrate this?  I'd prefer to test/display directly within CS3D

wayfarer3130:
@sedghi - I'm off next week, but the changes look ok, I just want to see an example in CS3D so that I can play with it/test it as a combined set.

pedrokohler:
@wayfarer3130 done. Added a new example so that you can play with it.

wayfarer3130:
@pedrokohler - I found a bug in this related to the preview mode:

- Do a segmentation annotation, accept it
- Start group recording
- Do a segmentation annotation, accept that
- Do another segmentation annotation, accept that
- End group recording
- Undo twice

The first undo marks the segmentations as in preview mode, while the second one marks the original segmentation as in preview mode additionally, but doesn't remove the segmentations after the group recording.

Compaer with/without the group recording.

packages/core/src/utilities/historyMemo/index.ts
```diff
@@ -80,25 +81,20 @@ export class HistoryMemo {
...
-            detail: {
-              isUndo: true,
-              id: item.id,
-              operationType: item.operationType || 'annotation',
-              memo: item,
-            },
-          })
-        );
+      if (Array.isArray(item)) {
+        for (const subitem of item) {
```
wayfarer3130: Can you use an asArray function that converts item to an array if it is not already an array, that way there is only one copy of the internal code - something like:
for(const subitem of asArray(item)) {
...
}
and remove the if Array.isArray?  That simplifies the total code paths and ensures they are consistent between both types.

packages/core/src/utilities/historyMemo/index.ts
```diff
@@ -113,35 +109,51 @@ export class HistoryMemo {
...
-            detail: {
-              isUndo: false,
-              id: item.id,
-              operationType: item.operationType || 'annotation',
-              memo: item,
-            },
-          })
-        );
+
+      if (Array.isArray(item)) {
```
wayfarer3130: Ditto

PR #2173, opened 2025-07-03
fix(e2e): Various improvements to the playwright tests

jbocce:
### Context
The playwright tests were lacking archived results in github.
The playwright tests were taking a bit long to run.
We wanted to be sure that the playwright tests were not giving us false negatives (i.e. a negative is a passing test).

### Changes & Results
Lower tolerances for Playwright screenshots comparisons.
Added packaging of Playwright test results to the github workflow. Removed unnecessary waits in tests to improve performance.

### Testing
Run the tests on github.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] "OS: Windows 11
- [x] "Node version: 20.9.0
- [x] "Browser: Chrome 138.0.7204.97
  


wayfarer3130:
Where do the screenshots get stored?  It seems like there are failing tests, but I'm not quite sure where to look for the results.

wayfarer3130:
Found the results - it looks like there are video files for all the tests, not just the failing ones.  Is there a way to only included failed test results, and only include all test results if a flag is set?

jbocce:
> Found the results - it looks like there are video files for all the tests, not just the failing ones. Is there a way to only included failed test results, and only include all test results if a flag is set?

@wayfarer3130 that is a great suggestion. I will try to see if there is something I can do for that.

Also, I removed a bunch of sleeps which is fine if there are existing screen shots, but when new ones are required I am afraid that we will need the sleeps again.

@wayfarer3130 will you be looking at the failed tests?

jbocce:
Ok @wayfarer3130 , I added a flag to only capture video for the first retry. This should eliminate noise and reduce the size of the packaged test results. We can add a separate config to capture more stuff in the future should we see fit. Otherwise, I think this is good for now.

PR #2180, opened 2025-07-07
feat(CrosshairsTool): add support for center gap ratio in reference lines

### LINKED ISSUE #54: [Known Issue] CrosshairsTool pointer gap
The empty space between the crosshairs lines is a fixed length currently, which makes it look huge for smaller images (canvases too I guess). Crosshairs middle gap should be configurable and be relative to canvas size. 

image

daker:
### Context

Add `referenceLinesCenterGapRatio` config option to allow the crosshairs center gap to scale proportionally with the canvas size. The gap now uses the ratio if provided, falling back to the default pixel value otherwise. Updated configuration and documentation comments accordingly.

fixes #54

### Changes & Results

| Before | After (ratio 0.05 -> 22.5px) | After(mobile/tablet ratio 0.05) |
| -------- | ------- | ------- |
| image | image | image

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] "OS: 
- [x] "Node version: 
- [x] "Browser:
  

### DIFF
--- a/packages/tools/src/tools/CrosshairsTool.ts
+++ b/packages/tools/src/tools/CrosshairsTool.ts
@@ -144,6 +144,14 @@ class CrosshairsTool extends AnnotationTool {
         // the reference lines will not be rendered. This is only used when
         // having 3 viewports in the toolGroup.
         referenceLinesCenterGapRadius: 20,
+        // The ratio is a fraction of the minimum canvas dimension (width or height).
+        // For example, if referenceLinesCenterGapRatio is set to 0.05, the gap will be 5% of the smallest side of the canvas.
+        // If set to 1, the gap will be equal to the minimum canvas dimension (which would likely hide the crosshairs).
+        // referenceLinesCenterGapRatio: null|undefined → gap is referenceLinesCenterGapRadius (default: 20 pixels)
+        // referenceLinesCenterGapRatio: 0.05 → gap is 5% of the canvas min dimension
+        // referenceLinesCenterGapRatio: 0.1 → gap is 10% of the canvas min dimension
+        // referenceLinesCenterGapRatio: 1 → gap is 100% (not recommended)
+        referenceLinesCenterGapRatio: null,
         // actorUIDs for slabThickness application, if not defined, the slab thickness
         // will be applied to all actors of the viewport
         filterActorUIDsToSetSlabThickness: [],
@@ -153,6 +161,7 @@ class CrosshairsTool extends AnnotationTool {
           enabled: false,
           opacity: 0.8,
           handleRadius: 9,
+          referenceLinesCenterGapRatio: 0.05,
         },
       },
     }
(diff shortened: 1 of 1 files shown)


packages/tools/src/tools/CrosshairsTool.ts
```diff
@@ -956,7 +965,16 @@ class CrosshairsTool extends AnnotationTool {
         canvasMinDimensionLength * 0.2
       );
       const canvasVectorFromCenterStart = vec2.create();
-      const centerGap = this.configuration.referenceLinesCenterGapRadius;
+      // Calculate center gap using ratio if provided, else fallback to pixel value
+      let centerGap = this.configuration.referenceLinesCenterGapRadius;
+      const mobileConfig = this.configuration.mobile;
+      const centerGapRatio =
```
wayfarer3130: Can you order this so that centerGap is a const?
Something like:
const { referenceLinesCenterGapRatio } = mobileConfig?.enabled ? mobileConfig : this.configuration;
const centerGrap = referenceLInesCenterGapRatio > 0 ? ....
Note you don't need to assign the ||0 as the >0 will be false for null/undefined or 0.

packages/tools/src/tools/CrosshairsTool.ts
```diff
@@ -144,6 +144,14 @@ class CrosshairsTool extends AnnotationTool {
...
         // having 3 viewports in the toolGroup.
         referenceLinesCenterGapRadius: 20,
+        // The ratio is a fraction of the minimum canvas dimension (width or height), not a 1:1 ratio.
+        // For example, if referenceLinesCenterGapRatio is set to 0.05, the gap will be 5% of the smallest side of the canvas.
+        // If set to 1, the gap will be equal to the minimum canvas dimension (which would likely hide the crosshairs).
+        // referenceLinesCenterGapRatio: 0 → gap is referenceLinesCenterGapRadius (default: 20 pixels)
+        // referenceLinesCenterGapRatio: 0.05 → gap is 5% of the canvas min dimension
+        // referenceLinesCenterGapRatio: 0.1 → gap is 10% of the canvas min dimension
+        // referenceLinesCenterGapRatio: 1 → gap is 100% (not recommended)
+        referenceLinesCenterGapRatio: 0,
```
wayfarer3130: Can you leave this undefined, so allow 0 as the gap?  I can't immediately think of why that would be useful, but it feels like something that could be used.

packages/tools/src/tools/CrosshairsTool.ts
wayfarer3130: And make it >=0 to agree with the undefined/null referenceLinesCenterGapRadius

wayfarer3130 (changes requested):
Couple of small tweaks to improve the change.

sedghi:
can we get before and after image in the PR description please

daker:
> can we get before and after image in the PR description please

@sedghi Done

sedghi:
Does this change desktop behaviour as default?

daker:
@sedghi if you don't set referenceLinesCenterGapRatio(null by default), the default 20px referenceLinesCenterGapRadius will be used.

PR #2181, opened 2025-07-08
fix(rendering): Introduce SequentialRenderingEngine to resolve canvas size limitation, performance degradation on high res monitors and enhance multi-monitor support

IbrahimCSAE:
## Overview

This PR introduces a new `SequentialRenderingEngine` to fundamentally solve the issue of viewport cropping, misalignment, and terrible performance on high-resolution displays or in layouts with many viewports.

## The Problem: Offscreen Canvas Overflow & Performance Degradation Due to Ofscreen Canvas Growth

Cornerstone3D has historically used a single, large offscreen canvas for all viewports. This canvas grows horizontally as more viewports are added. This strategy is efficient until the total width of all viewports exceeds the browser's maximum canvas size (e.g., **16,384px** in Chrome).

When this limit is breached, the offscreen canvas is silently cropped. Since the engine copies pixel data from fixed locations on the offscreen canvas to the on-screen viewports, this cropping results in severe visual artifacts, including misaligned and completely blank viewports. This issue is easily reproducible on high-DPI monitors or by manually increasing `window.devicePixelRatio`. The peformance also heavly degrades as the limit is approached.

**Visual Example of the Bug:**

CleanShot 2025-07-09 at 00 30 55@2x

**Diagram of Current RenderingEngine:**

mermaid-diagram-2025-07-09-014536

## The Journey to a Solution

Before arriving at the final solution, several alternative approaches were extensively researched and implemented. Each had significant drawbacks that made them unsuitable:

1.  **Multiple Offscreen Canvases:** This approach attempted to distribute viewports across a few offscreen canvases.
    * **Limitation:** It introduced a massive performance hit and was constrained by the requirement that all `VolumeViewports` must share the same offscreen canvas. Only `StackViewports` saw a partial benefit.
2.  **Advanced Canvas Packing:** We explored more complex packing algorithms (e.g., horizontal and vertical) to use the offscreen space more efficiently.
    * **Limitation:** This still ran into fundamental OS and browser memory limits. The theoretical maximum canvas size (e.g., 16k x 16k) is often not practically achievable, not even (8k x 8k) in some instances.
3.  **Multiple Rendering Engines:** This involved spinning up a new rendering engine for each viewport.
    * **Limitation:** This was a dead end. Browsers limit the number of active WebGL contexts to ~16 per tab, creating a hard cap of 16 viewports. It also required all volumes to share a single engine, negating the benefit for volume rendering.

## The Solution: `SequentialRenderingEngine`

This PR introduces a new, robust rendering strategy: the **`SequentialRenderingEngine`**.

Instead of a single, massive offscreen canvas that grows with each new viewport, this engine uses a fundamentally different approach. It renders each viewport **individually** to a viewport sized offscreen canvas, copies the result to the corresponding on-screen canvas, and then proceeds to the next viewport, reusing the same offscreen canvas.

This elegant solution completely bypasses the previous pain points:

-   ✅ **No More Aggregate Canvas Limits:** The browser's maximum canvas size now applies to a *single viewport*, not the combined width of all viewports.
-   ✅ **Single WebGL Context:** The engine operates with just one WebGL context, so we are no longer limited to 16 viewports.
-   ✅ **Memory:** It avoids allocating a huge, memory-intensive offscreen canvas.
-   ✅ **OS & Browser Friendly:** It steers clear of underlying OS memory and allocation limits that more aggressive strategies would hit.
-   ✅ **Enhanced Stability:** Reduces the risk of WebGL context loss associated with creating and managing huge canvas surfaces.
- ✅ Multi-Monitor Support: Running OHIF or Cornerstone across a stretched tab on multiple high-resolution monitors was practically impossible before, due to canvas size limits and severe performance degradation. With the SequentialRenderingEngine, this becomes not only feasible but smooth and performant, making the use of OHIF / Cornerstone more realistic in actual diagnostic radiology environments where 2-4 monitors are common.

**Visual Example of the Fix:**

CleanShot 2025-07-09 at 00 30 19@2x

**Diagram of Sequential RenderingEngine:**

mermaid-diagram-2025-07-09-014445

## Implementation Details

This PR includes the following key changes:

-   **`RenderingEngine` Refactor:** The original `RenderingEngine` returns an instance of either `StandardRenderingEngine` or `SequentialRenderingEngine`, based on a property that can be configured in the init function, the property is `renderingEngineMode`, which is a string that can be either `standard` or `next`
-   **`BaseRenderingEngine`:** A new abstract base class has been created to define a common rendering engine interface.
-   **`StandardRenderingEngine`:** The original `RenderingEngine` has been moved to the class `StandardRenderingEngine` which extends `BaseRenderingEngine`
-   **New `SequentialRenderingEngine`:** The new engine is implemented as a separate class, also extending `BaseRenderingEngine`.
-   **Viewport Compatibility:** The `worldToCanvas` and `canvasToWorld` methods on both `StackViewport` and `VolumeViewport` have been updated to support either rendering engine.
-   **New Examples:** Two new examples have been added to demonstrate rendering a **6x6 grid of viewports**—one using the standard engine (which will fail) and one using the new sequential engine (which will succeed). This provides a clear test of the solution.

## Timeline

- April 8th: Issue reported on some Barco monitors, eg [here](https://www.monitors.com/products/barco-coronis-mdcc-6230-k9601450?variant=32202642391129&utm_source=google&utm_medium=surfaces&utm_campaign=shopping%20feed&utm_campaign=18127936711&keyword=&gad_source=1&gclid=CjwKCAjwtdi_BhACEiwA97y8BLALY-GlPo7f6ehokIjyROJRFKHQAPATuVCm8eZUdpsefAfmnLMQEBoC304QAvD_BwE), and 4k, 6k displays, eg [here](https://www.apple.com/ca/pro-display-xdr/).
- April 9th: Try a multiple offscreen canvas approach [here](https://github.com/cornerstonejs/cornerstone3D/pull/1989)
- April 19th: Performance is horrible with multi offscreen, try a per viewport RE approach [here](https://github.com/OHIF/Viewers/pull/4976)
- May-June: Experience issues with per viewport RE, initilizing and destroying REs leads to inconsistent state and is not reliable, cornerstone3D architecture heavly relies on a single RE, many tools needed to be moved externally outside of cornerstone, such as ReferenceLines, Crosshairs, synchronizers, and core OHIF services needed to be modified heavily. Volumes need to share the same RE. Can’t have more than 16 REs, so assign 3 stacks to each RE.
- July 4th: Try canvas packing [here](https://github.com/cornerstonejs/cornerstone3D/pull/2174)
- July 7th: Not satisfied with canvas packing results, performance degradation still occurs, canvas cropping still occurs, hitting different canvas size limits on different OS/browsers
- July 9th: Implement sequential rendering engine, all issues resolved

## Usage

The new Sequential Rendering Engine is now the default in Cornerstone3D. If you wish to continue using the old rendering engine, you can override this behavior by explicitly setting the renderingEngineMode to 'standard'.

To use the old rendering engine:

```ts
import { init } from '@cornerstonejs/core';

init({
  rendering: {
    renderingEngineMode: 'standard',
  },
});

```

### OHIF

- OHIF before: https://coruscating-youtiao-10ee6d.netlify.app/viewer?StudyInstanceUIDs=1.3.12.2.1107.5.1.4.60175.30000008042114404745300000010&hangingProtocolId=@ohif/mnGrid

CleanShot 2025-07-09 at 01 00 46@2x

- Video: https://streamable.com/41h2xy, You can see me experience a context loss in the video, when I try to go back from one up to the grid, due to how huge the offscreen is.

- OHIF after: https://gorgeous-sable-7bbec5.netlify.app/viewer?StudyInstanceUIDs=1.3.12.2.1107.5.1.4.60175.30000008042114404745300000010&hangingProtocolId=@ohif/mnGrid

CleanShot 2025-07-09 at 01 03 15@2x

- Video: https://streamable.com/lyofjq
- Video with FPS while making each viewport 4120 x 1100 for fun, still very performant:  https://streamable.com/s15zyj

### Cornerstone

- CS before: https://ephemeral-raindrop-5f4044.netlify.app/live-examples/renderingengine.html

CleanShot 2025-07-09 at 00 30 55@2x

- CS After:  https://ephemeral-raindrop-5f4044.netlify.app/live-examples/renderingenginesequential.html 

CleanShot 2025-07-09 at 00 30 19@2x

### PRs

- https://github.com/cornerstonejs/cornerstone3D/pull/2159
- https://github.com/cornerstonejs/cornerstone3D/pull/1989
- https://github.com/cornerstonejs/cornerstone3D/pull/2174
- https://github.com/OHIF/Viewers/pull/4976
- https://github.com/OHIF/Viewers/pull/5180
- https://github.com/OHIF/Viewers/pull/5187

### Packing

- https://www.david-colson.com/2020/03/10/exploring-rect-packing.html
- https://github.com/mapbox/potpack
- https://mapbox.github.io/potpack/

### Canvas size limits per browser

- https://jhildenbiddle.github.io/canvas-size/#/?id=test-results

### WebGL context limits

This is different per OS and browser

- Chrome source code (may be outdated, but its still 16): https://chromium.googlesource.com/chromium/src/+/2efb8d651d7baf2313253a01df19cbfeabd690b6/third_party/WebKit/Source/modules/webgl/WebGLRenderingContextBase.cpp#106
- Discussions: https://bugzilla.mozilla.org/show_bug.cgi?id=1421481

## Thanks

Thanks to @wayfarer3130 @sedghi for their help and insights that helped me brainstorm a solution. I wouldn't have been able to resolve this without them.


IbrahimCSAE:
Deploy Preview: https://extraordinary-klepon-241bee.netlify.app/
Example Sequential: https://extraordinary-klepon-241bee.netlify.app/live-examples/renderingenginesequential.html
Example Standard: https://extraordinary-klepon-241bee.netlify.app/live-examples/renderingengine.html

packages/core/src/RenderingEngine/SequentialRenderingEngine.ts
```diff
@@ -0,0 +1,377 @@
...
+      if (this.useCPURendering) {
+        throw new Error(
+          'GPU not available, and using a viewport with no custom render pipeline.'
+        );
+      }
+
+      const { offscreenMultiRenderWindow } = this;
+      const renderWindow = offscreenMultiRenderWindow.getRenderWindow();
+
+      this._resizeOffScreenCanvasForSingleViewport(viewport.canvas);
```
JamesAPetts, line 271: Can we just keep it the same size as the largest viewport all the time?

And only resize it when any of these are resized?

IbrahimCSAE: Thats a good idea, I can also group resize operations by viewport size. should make it only two calls for a grid like tmtv or one call for a uniform grid.

There’s plenty of optimizations that can be made, I think this is a starting point and we can go from there. I’m ok with trying to optimize the resizing before merge. I will do it in the morning

It does seem pretty fast already though, would recommend giving it a test run in OHIF, the old RE is quite slow compared to it

JamesAPetts: IMO we should do it, but I agree it doesn't necessarily have to be in this PR.

IbrahimCSAE: Thanks James, Yes its def needed.

packages/adapters/src/adapters/Cornerstone3D/CircleROI.ts
```diff
@@ -183,7 +183,7 @@ class CircleROI extends BaseAdapter3D {
         points.push({ x: end[0], y: end[1], z: center[2] });
 
         const cachedStatsKeys = Object.keys(cachedStats)[0];
-        const { area, radius } = cachedStatsKeys
+        const { area = undefined, radius = 0 } = cachedStatsKeys
```
wayfarer3130, line 186: You don't need to assign undefined, it is implicit

IbrahimCSAE: It doesn't build without this for some reason, docs have been broken for months now. It asks for a default value.

https://github.com/cornerstonejs/cornerstone3D/actions/workflows/build-docs.yml

packages/adapters/src/adapters/Cornerstone3D/CobbAngle.ts
```diff
@@ -184,7 +184,9 @@ class CobbAngle extends BaseAdapter3D {
         const point4 = { x: end2[0], y: end2[1], z: end2[2] };
 
         const cachedStatsKeys = Object.keys(cachedStats)[0];
-        const { angle } = cachedStatsKeys ? cachedStats[cachedStatsKeys] : {};
+        const { angle = undefined } = cachedStatsKeys
```
wayfarer3130, line 187: If you are going to rewrite these, I'd rather see:
const angle = cachedStats[cachedStatsKeys]?.angle;
cachedStats[undefined] is just undefined and the ?.angle means it doesn't need to get an empty object.

wayfarer3130:
Can we get the adapter/tool changes reverted, unless specifically related?

packages/core/src/init.ts
```diff
@@ -15,6 +15,7 @@ const defaultConfig: Cornerstone3DConfig = {
     preferSizeOverAccuracy: false,
     strictZSpacingForVolumeViewport: true,
   },
+  renderingEngineMode: 'next',
```
sedghi: please add a comment above say what is what and what is our plan

IbrahimCSAE: comment is in the type, I can add here too if needed

sedghi: can we have a ENUM for this 
rendering engine mode or something

IbrahimCSAE: sure

sedghi (approved):
I went through all 158 examples, and they're working well. It looks like only the orientation marker is broken, which I assume you're already addressing. This is a really great pull request—one of the best we've had in recent years!

salimkanoun:
Wow that's seems really amazing, congratulations !

packages/core/examples/sequentialrenderingengine/index.ts
wayfarer3130, line 1: I made this one configurable with rows and columns, going to 12x22.  There is a significant pause about 1/3 of the way through - not that I'm concerned, just noticing it.

sedghi:
As discussed in slack, we can improve this later. I'm gonna merge

PR #2190, opened 2025-07-10
feat(rendering): Introduce WebGLContextPool for parallel rendering in ContextPoolRenderingEngine

IbrahimCSAE:
## Overview

This PR enhances the `ContextPoolRenderingEngine` (previously `SequentialRenderingEngine`) by introducing a `WebGLContextPool` class that enables parallel rendering across multiple WebGL contexts. This optimization significantly improves performance when rendering multiple viewports simultaneously while maintaining the core benefits of sequential rendering.

## WebGL Context Pooling

Following the successful resolution of canvas size limitations in [PR #2181](https://github.com/cornerstonejs/cornerstone3D/pull/2181), this PR takes the next step in performance optimization. While the sequential approach solved critical issues with viewport cropping and performance degradation on high-resolution displays, it rendered viewports one at a time. This new implementation introduces controlled parallelism to boost performance further.

### WebGLContextPool

The new `WebGLContextPool` class manages multiple WebGL contexts (default: 8) that can render viewports in parallel:

- **Configurable Context Count**: Users can specify the number of WebGL contexts via `webGlContextCount` in the init configuration
- **Smart Distribution**: Stack viewports are distributed across available contexts using round-robin assignment
- **Volume Viewport Limitation**: Volume viewports continue to use a single context (index 0) due to WebGL resource sharing constraints
- **Safe Defaults**: The default of 8 contexts provides excellent performance while staying well within browser limits (~16 contexts per tab)

## Architecture

mermaid-diagram-2025-07-10-110004

## Performance Benefits

Multiple stack viewports can now render simultaneously across different WebGL contexts, and significant performance improvement for sychronized operations, eg cine, synced scrolling.

### Configuration

Users can configure the number of WebGL contexts during initialization:

```typescript
import { init } from '@cornerstonejs/core';

init({
  rendering: {
    webGlContextCount: 8,        // Number of WebGL contexts (default: 8)
  },
});
```

### Context Assignment Strategy

- **Volume Viewports**: Always assigned to context 0 (no parallelism due to resource sharing constraints between WebGL contexts)
- **Stack Viewports**: Distributed across all available contexts using round-robin assignment

### Important Notes

1. **Context Limit**: While 8 is the safe default, users should not exceed this unless necessary. Browser limits are typically around 16 contexts per tab, but if your goal is to support as many devices as possible, it's best to stay well below this limit to avoid issues.

2. **Volume Viewport Limitation**: Volume viewports cannot currently utilize parallel rendering due to WebGL resource sharing constraints between contexts. This is a current limitation of WebGL.

### Example

https://marvelous-genie-f7fb6a.netlify.app/live-examples/webglcontextpooling

### DIFF
--- a/packages/core/src/RenderingEngine/BaseRenderingEngine.ts
+++ b/packages/core/src/RenderingEngine/BaseRenderingEngine.ts
@@ -642,6 +642,20 @@ abstract class BaseRenderingEngine {
   }
 
   /**
+   * Returns the offscreen multi-render window used for rendering.
+   */
+  public getOffscreenMultiRenderWindow(
+    viewportId?: string
+  ): VtkOffscreenMultiRenderWindow {
+    if (this.useCPURendering) {
+      throw new Error(
+        'Offscreen multi render window is not available when using CPU rendering.'
+      );
+    }
+    return this.offscreenMultiRenderWindow;
+  }
+
+  /**
    * Sets multiple viewports using custom rendering
    * pipelines to the `RenderingEngine`.
    *
--- a/packages/core/src/RenderingEngine/ContextPoolRenderingEngine.ts
+++ b/packages/core/src/RenderingEngine/ContextPoolRenderingEngine.ts
@@ -522,6 +522,23 @@ class ContextPoolRenderingEngine extends BaseRenderingEngine {
     }
     super.destroy();
   }
+
+  public getOffscreenMultiRenderWindow(
+    viewportId: string
+  ): VtkOffscreenMultiRenderWindow {
+    if (this.useCPURendering) {
+      throw new Error(
+        'Offscreen multi render window is not available when using CPU rendering.'
+      );
+    }
+
+    const contextIndex =
+      this.contextPool.getContextIndexForViewport(viewportId);
+
+    const contextData = this.contextPool.getContextByIndex(contextIndex);
+
+    return contextData.context;
+  }
 }
 
 export default ContextPoolRenderingEngine;
--- a/packages/core/src/RenderingEngine/RenderingEngine.ts
+++ b/packages/core/src/RenderingEngine/RenderingEngine.ts
@@ -106,6 +106,12 @@ class RenderingEngine {
   public destroy(): void {
     return this._implementation.destroy();
   }
+
+  public getOffscreenMultiRenderWindow(
+    viewportId?: string
+  ): VtkOffscreenMultiRenderWindow {
+    return this._implementation.getOffscreenMultiRenderWindow(viewportId);
+  }
 }
 
 export default RenderingEngine;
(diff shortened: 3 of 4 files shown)


packages/docs/docs/concepts/cornerstone-core/renderingEngine.md
```diff
@@ -30,6 +30,78 @@ canvas get updated, and at render time, we copy from offscreen to onscreen for e
...
+- Creates one massive offscreen canvas that grows horizontally as viewports are added
+- Renders all viewports to specific coordinates on this single offscreen canvas
+- Copies pixel data from the offscreen canvas to individual onscreen viewports
+
+**Limitations of StandardRenderingEngine:**
+
+- **Canvas Size Limits**: Browsers impose maximum canvas dimensions (e.g., 16,384px in Chrome). When the combined width of all viewports exceeds this limit, the offscreen canvas is silently cropped, causing severe visual artifacts, misaligned viewports, and blank viewports
+- **Performance Degradation**: As the offscreen canvas approaches size limits, performance degrades significantly, especially on high-resolution displays or layouts with many viewports
+- **Multi-Monitor Issues**: Practically impossible to use across multiple high-resolution monitors due to canvas size limitations
+- **Memory Consumption**: Allocates a huge, memory-intensive offscreen canvas regardless of actual viewport usage
```
sedghi: say benefits: it's been in production for 5 years, and for basic use cases probably is the best options, and also mobile and tablet still this is the best option imo

packages/docs/docs/concepts/cornerstone-core/renderingEngine.md
```diff
@@ -30,6 +30,78 @@ canvas get updated, and at render time, we copy from offscreen to onscreen for e
...
+
+### NextRenderingEngine (SequentialRenderingEngine)
```
sedghi: I think we just say sequentialRE, not next, since 2 years from now another thing is next

packages/core/src/RenderingEngine/BaseRenderingEngine.ts
```diff
@@ -50,7 +51,7 @@ abstract class BaseRenderingEngine {
   /**
    * @param uid - Unique identifier for RenderingEngine
    */
-  constructor(id?: string) {
+  constructor(id?: string, rendersThumbnails = false) {
```
sedghi: i don't understand why we need this rendersThumbnails, what is it different? if we really need it, it should be options: {}

IbrahimCSAE: use a single context for thumbnails, dont spin up multiple. I can make it options

IbrahimCSAE:
Deploy preview: https://remarkable-torrone-d3cd0a.netlify.app/

IbrahimCSAE:
I have tested all examples, and added multiple tests

PR #2203, opened 2025-07-15
feat(segmentation): segmentation brush interpolation

IbrahimCSAE:
### Context

This change interpolates the segmentation between the start point and end point to help with generating smooth lines instead of gaps when a quick stroke is performed.

Before: https://www.cornerstonejs.org/live-examples/labelmapsegmentationtools

After: https://lively-pie-d6c2bd.netlify.app/live-examples/labelmapsegmentationtools

### DIFF
--- a/packages/tools/src/tools/segmentation/BrushTool.ts
+++ b/packages/tools/src/tools/segmentation/BrushTool.ts
@@ -37,6 +37,7 @@ import { getStrategyData } from './strategies/utils/getStrategyData';
  */
 class BrushTool extends LabelmapBaseTool {
   static toolName;
+  private _lastDragPoint: Types.Point2 | null = null;
 
   constructor(
     toolProps: PublicToolProps = {},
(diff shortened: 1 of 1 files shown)


packages/tools/src/tools/segmentation/BrushTool.ts
```diff
@@ -371,10 +373,38 @@ class BrushTool extends LabelmapBaseTool {
...
-    this._previewData.preview = this.applyActiveStrategy(
-      enabledElement,
-      this.getOperationData(element)
-    );
+    const lastPoint = this._lastDragPoint || this._previewData.startPoint;
+    const currentPoint = currentPoints.canvas;
+
+    const distance = vec2.distance(lastPoint, currentPoint);
+
+    if (distance > 1) {
```
sedghi, line 381: what is unit of 1 ?

IbrahimCSAE: pixels

IbrahimCSAE:
todo before merge:

- preview mode is much slower now
- fix test screenshot

pedrokohler:
@IbrahimCSAE thanks a lot for the fix 😃

Issue #1526, opened 2024-10-30
RFC for 3.x


sedghi:
### What feature or change would you like to see made?

s

### Why should we prioritize this feature?

s

sedghi:
Move all viewports change set event to `VIEWPORT_NEW_IMAGE_SET`

pedrokohler:
Improve SVG visibility by slightly changing the feOffset's dx and dy:

Changing it [here](https://github.com/cornerstonejs/cornerstone3D/blob/ea3e44970a9e535db1f38c8186653ffb7376ecf9/packages/tools/src/store/addEnabledElement.ts#L94) to:

feOffset.setAttribute('dx', '1');
feOffset.setAttribute('dy', '1');

Helps improving how the SVG stands out with a slightly larger shadow when the contrast between the colors does not favor visibility.

(See: https://github.com/cornerstonejs/cornerstone3D/pull/2207/files#r2210765267 for more info)

sedghi:
we will add some codemods for this upgrade to https://github.com/OHIF/codemods

PR #2207, opened 2025-07-16
feat: improve tool styles

pedrokohler:
### Context

First step to address issues of https://github.com/OHIF/Viewers/issues/4600

1. Slightly improves SVG shadows so that they are more visible. The visual change is barely noticeable, but it really makes a difference in making the SVG stand out on the viewport.

2. Adds more configuration options to the SegmentLabelTool.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  

### DIFF
--- a/packages/tools/src/store/addEnabledElement.ts
+++ b/packages/tools/src/store/addEnabledElement.ts
@@ -91,8 +91,8 @@ function _createSvgAnnotationLayer(viewportId: string): SVGElement {
   //
   feOffset.setAttribute('result', 'offOut');
   feOffset.setAttribute('in', 'SourceGraphic');
-  feOffset.setAttribute('dx', '0.5');
-  feOffset.setAttribute('dy', '0.5');
+  feOffset.setAttribute('dx', '1');
+  feOffset.setAttribute('dy', '1');
 
   //
   feColorMatrix.setAttribute('result', 'matrixOut');
--- a/packages/tools/src/store/addEnabledElement.ts
+++ b/packages/tools/src/store/addEnabledElement.ts
@@ -91,8 +91,8 @@ function _createSvgAnnotationLayer(viewportId: string): SVGElement {
   //
   feOffset.setAttribute('result', 'offOut');
   feOffset.setAttribute('in', 'SourceGraphic');
-  feOffset.setAttribute('dx', '1');
-  feOffset.setAttribute('dy', '1');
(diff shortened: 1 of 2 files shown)


packages/tools/src/store/addEnabledElement.ts
```diff
@@ -91,8 +91,8 @@ function _createSvgAnnotationLayer(viewportId: string): SVGElement {
   //
   feOffset.setAttribute('result', 'offOut');
   feOffset.setAttribute('in', 'SourceGraphic');
-  feOffset.setAttribute('dx', '0.5');
-  feOffset.setAttribute('dy', '0.5');
+  feOffset.setAttribute('dx', '1');
```
sedghi: i think this is setting it for all of them. I don't think that is a good idea since your tool only needs it, not all other tools, fine another way

pedrokohler: It does affect all tools, but if you take a look at the impact is positive.

Sometimes the other tools also are hard to see. With this change they stand out slightly better on the viewport.

It's not a huge impact, but it definitely improves it slightly.

If you can, please check it out and let me know if it's ok.

sedghi: I'm quite sure about the positive impact, but it will break everyone's tests since it will affect pixels.
we can do it in 4.0, but usually we wait for couple of enhancements to arrive until release a breaking change. 
Put it here but revert 
https://github.com/cornerstonejs/cornerstone3D/issues/1526

pedrokohler: Great, I think it's a good compromise

@fedorov meanwhile we'll still have the possibility of customizing the tool colors and putting a box behind it to improve visibility. So hopefully that's enough for now? Let me know if you need visible examples to understand what I'm talking about.

fedorov: @pedrokohler a screenshot would be super helpful!

pedrokohler: Both default color and background box set:

image

Only background box set:

image

pedrokohler:
@sedghi I reverted the changes you asked.

fedorov:
Thank you! Please integrate any solution that @sedghi is happy with. Anything is better than what we have now for the segment label, which is nothing. Appearance can be refined later!

PR #2208, opened 2025-07-16
fix(colormap): do not return empty opacity arrays

IbrahimCSAE:
### context

prevents the returning of empty opacity arrays to not cause rendering issues

### DIFF
--- a/packages/core/src/utilities/colormap.ts
+++ b/packages/core/src/utilities/colormap.ts
@@ -100,14 +100,12 @@ function findMatchingColormap(rgbPoints, actor): ColormapPublic | null {
     }
   }
 
-  return opacity.length > 0
-    ? {
-        name: matchedColormap.Name,
-        opacity,
-      }
-    : {
-        name: matchedColormap.Name,
-      };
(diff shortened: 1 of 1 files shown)


packages/core/src/utilities/colormap.ts
```diff
@@ -100,10 +100,12 @@ function findMatchingColormap(rgbPoints, actor): ColormapPublic | null {
     }
   }
 
-  return {
+  const result = {
     name: matchedColormap.Name,
-    opacity,
+    ...(opacity.length > 0 && { opacity }),
```
sedghi: sometimes opacity is a number and not an array

IbrahimCSAE: dont think the number version is possible here, can you check the returns?

sedghi: just check the type, if array && length do it, if number do it still

IbrahimCSAE: done

sedghi (approved):

PR #2213, opened 2025-07-18
fix(viewport): viewport flicker for label map segmentation

IbrahimCSAE:
Fixes: https://github.com/OHIF/Viewers/issues/5191

and any viewport flicker related to segmentation

### DIFF
--- a/packages/tools/src/stateManagement/segmentation/SegmentationRenderingEngine.ts
+++ b/packages/tools/src/stateManagement/segmentation/SegmentationRenderingEngine.ts
@@ -50,12 +50,23 @@ class SegmentationRenderingEngine {
    * some sort of segmentation representation.
    *
    * @param viewportId - The ID of the viewport to render the segmentations on. If not provided, segmentations will be rendered on all viewports.
+   * @param immediate - If true, renders immediately without using requestAnimationFrame
    */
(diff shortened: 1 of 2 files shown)


packages/tools/src/stateManagement/segmentation/SegmentationRenderingEngine.ts
```diff
@@ -212,6 +223,28 @@ class SegmentationRenderingEngine {
       }
     );
 
+    // If immediate rendering is requested, skip the async event handling
+    if (immediate) {
```
sedghi, line 227: I believe the reason for not doing it immediately was that the segmentation actor sometimes needs the image actor. Perhaps we should hook into `pre_render` instead. Could you try that?

IbrahimCSAE: the immediate=true call happens in _imageChangeEventListener, which is listening for image rendered already, I think this means the image actor is there, or am I missing something?

Rest of the calls default to normal behavior immediate=false

sedghi: I need to test this, not yet confident

PR #2217, opened 2025-07-21
fix(security): update and pin to latest versions of eslint-plugin-prettier and eslint-config-prettier.

jbocce:


bun.lock
```diff
@@ -59,11 +59,11 @@
         "docdash": "^1.2.0",
         "dpdm": "^3.14.0",
         "eslint": "^8.39.0",
-        "eslint-config-prettier": "^8.8.0",
+        "eslint-config-prettier": "8.10.0",
```
sedghi, line 62: same comment here, fixing deps to non ^

jbocce: Same reply as before. Do we want a simple removal of "^" or something more consistent to use the versions we were using up until now?

PR #2228, opened 2025-07-24
fix(viewport): viewport should take full canvas size by default

IbrahimCSAE:
### Context

Viewports were not taking the full canvas size

Before:

CleanShot 2025-07-24 at 15 27 46@2x

After:

CleanShot 2025-07-24 at 15 28 24@2x

### DIFF
--- a/packages/core/src/RenderingEngine/Viewport.ts
+++ b/packages/core/src/RenderingEngine/Viewport.ts
@@ -98,7 +98,7 @@ class Viewport {
   /**
    * The amount by which the images are inset in a viewport by default.
    */
-  protected insetImageMultiplier = 1.1;
+  protected insetImageMultiplier = 1;
 
   protected flipHorizontal = false;
   protected flipVertical = false;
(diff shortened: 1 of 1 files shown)


wayfarer3130:
Can you update the playwright tests and unit tests?
Also, can you describe when the viewport isn't using the full canvas size by default - the conditions under which we will see this?

packages/core/src/RenderingEngine/Viewport.ts
```diff
@@ -1112,8 +1112,22 @@ class Viewport {
       imageData.indexToWorld(idx, focalPoint);
     }
 
-    const { widthWorld, heightWorld } =
-      this._getWorldDistanceViewUpAndViewRight(bounds, viewUp, viewPlaneNormal);
+    let widthWorld;
```
wayfarer3130, line 1115: My preference is to extract this into it's own function which returns a consistent result, and use const {widthWOrld, heightWorld} = this.getSizeWorld(imageData,camera)

wayfarer3130 (changes requested):

packages/core/src/RenderingEngine/Viewport.ts
IbrahimCSAE: I can do that

IbrahimCSAE:
TODO for this to be merged:

- this.getSizeWorld function to be added in viewport class, takes imageData,camera
- update screenshots for all playwright tests
- update screenshots for all karma tests

PR #2229, opened 2025-07-24
fix(widgets): disable widget renderers after render

IbrahimCSAE:
### Context

Disable widget renderers after render


sedghi (approved):
if tests passes

PR #2230, opened 2025-07-28
fix: add missing call to update spheres on on receive event from control tool

mbellehumeur:
### Context

The spheres are not moving when the control tool ( reference line) is moved because an update command was deleted by mistake.

### Changes & Results

add missing line

### Testing

Reference lines change the sphere and cropping planes,  not just the cropping planes.

### DIFF
--- a/packages/tools/src/tools/VolumeCroppingTool.ts
+++ b/packages/tools/src/tools/VolumeCroppingTool.ts
@@ -842,6 +842,7 @@ class VolumeCroppingTool extends BaseTool {
           }
         }
       }
+      this._updateCornerSpheres();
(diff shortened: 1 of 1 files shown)


wayfarer3130 (approved):

daker:
@mbellehumeur it would be good if you can add the example to list of the examples

mbellehumeur:
> @mbellehumeur it would be good if you can add the example to list of the examples

I thought I did.  Do I need to do something extra?

daker:
@mbellehumeur you are right, it's the build-and-deploy-docs step that's failling for a few months now and the website is not updated

PR #2252, opened 2025-08-05
fix floating point issues in getViewReferenceId

rodrigobasilio2022:
### Context

Previously, the getViewReferenceId method generated a unique ID for each view based on the viewPlaneNormal vector, including its full floating-point precision. This caused the cache index (targetId) for CS3D tool statistics to be overly sensitive to tiny floating-point differences in the view plane normal—even differences at the 7th or 8th decimal place. As a result, multiple redundant cached statistics could be created for what is visually the same slice, leading to unnecessary memory usage and inconsistent tool behavior. This PR fixes the issue by formatting the viewPlaneNormal values to three decimal places, ensuring that only meaningful differences in orientation result in a new cache entry, and preventing cache fragmentation due to floating-point precision errors.
#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  

### DIFF
--- a/packages/core/src/RenderingEngine/BaseVolumeViewport.ts
+++ b/packages/core/src/RenderingEngine/BaseVolumeViewport.ts
@@ -2241,9 +2241,10 @@ abstract class BaseVolumeViewport extends Viewport {
     sliceIndex ??= currentIndex;
     const { viewPlaneNormal, focalPoint } = this.getCamera();
     const querySeparator = volumeId.includes('?') ? '&' : '?';
-    return `volumeId:${volumeId}${querySeparator}sliceIndex=${sliceIndex}&viewPlaneNormal=${viewPlaneNormal.join(
-      ','
-    )}`;
+    // Format each element of viewPlaneNormal to 3 decimal places
+    // to avoid floating point precision issues
+    const formattedNormal = viewPlaneNormal.map((v) => v.toFixed(3)).join(',');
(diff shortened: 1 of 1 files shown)


packages/core/src/RenderingEngine/BaseVolumeViewport.ts
```diff
@@ -2241,9 +2241,10 @@ abstract class BaseVolumeViewport extends Viewport {
     sliceIndex ??= currentIndex;
     const { viewPlaneNormal, focalPoint } = this.getCamera();
     const querySeparator = volumeId.includes('?') ? '&' : '?';
-    return `volumeId:${volumeId}${querySeparator}sliceIndex=${sliceIndex}&viewPlaneNormal=${viewPlaneNormal.join(
-      ','
-    )}`;
+    // Format each element of viewPlaneNormal to 3 decimal places
+    // to avoid floating point precision issues
+    const formattedNormal = viewPlaneNormal.map((v) => v.toFixed(3)).join(',');
+    return `volumeId:${volumeId}${querySeparator}sliceIndex=${sliceIndex}&viewPlaneNormal=${formattedNormal}`;
```
sedghi, line 2247: i think we have this pattern elsewhere, can you please search everywhere to see if other places need to get fixed too?

rodrigobasilio2022: Searched. Didnt found any other similar pattern

PR #2258, opened 2025-08-10
getBrushToolInstances always return array

salimkanoun:
### Context

Picked up a bug where getBrushToolInstances was returning undefined when calling code was expecting array without null safe operators,
Now getBrushToolInstances always return array (empty array if no brush tool instances)

### Changes & Results
getBrushToolInstances always return array making array prototype (map, foreach) calls on return value safe


sedghi (approved):
Thanks,
I wonder if i need to have the strict flag set to true, and have an AI fix all of these?

I only see 1589 errors in packages/core :D

PR #2259, opened 2025-08-11
feat: Make jump to click can update multiple target tool groups

salimkanoun:
### Context

Until now JumpToClick was accepting only one toolgroup.
The problem is depending on stage you might have multiple viewport with multiple toolgroup non being instanciated in all stages.
This PR add the possibility to add a toolgroupId array to make JumpToClick act on multiple toolgroups

### Changes & Results

Add toolGroupIds configuration which expect array of tool group Id

### Testing

See mipJumpToClick example, add a commented toolGroupIds that does the same than viewport selection but using toolgroups

  additions or removals.

#### Tested Environment

Linux, chrome latest, node 20

### DIFF
--- a/packages/tools/src/tools/MIPJumpToClickTool.ts
+++ b/packages/tools/src/tools/MIPJumpToClickTool.ts
@@ -20,6 +20,8 @@ class MIPJumpToClickTool extends BaseTool {
       supportedInteractionTypes: ['Mouse', 'Touch'],
       configuration: {
         targetViewportIds: [],
+        toolGroupId: undefined,
+        toolGroupIds: [],
       },
     }
   ) {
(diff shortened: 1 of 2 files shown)


packages/tools/src/tools/MIPJumpToClickTool.ts
```diff
@@ -20,6 +20,8 @@ class MIPJumpToClickTool extends BaseTool {
       supportedInteractionTypes: ['Mouse', 'Touch'],
       configuration: {
         targetViewportIds: [],
+        toolGroupId: undefined,
```
sedghi, line 23: it already has targetViewportIds, and since each viewport belongs to one and only one toolGroup, this means we have targetToolGroups, why do we need to add toolGroupIds?

salimkanoun: The good things about tool groups is that you can reuse them across multiple hanging protocols

Viewport id is more likely to be unique to identify one viewport in particular and will make it harder to track across hanging protocols

PR #2260, opened 2025-08-11
[WIP] Fix invalid frame index

rodrigobasilio2022:
### Context

Fix the frame index in wadouri calls.
pixelDataFrame = frame -1

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  

### DIFF
--- a/packages/dicomImageLoader/src/imageLoader/wadouri/loadImage.ts
+++ b/packages/dicomImageLoader/src/imageLoader/wadouri/loadImage.ts
@@ -221,7 +221,7 @@ function loadImage(
   return loadImageFromPromise(
     dataSetPromise,
     imageId,
-    parsedImageId.frame,
+    parsedImageId.pixelDataFrame,
     parsedImageId.url,
     options
   );


sedghi:
how to test?

rodrigobasilio2022:
The same error is not being manifested as in the private repo. But th ideia is that when you load a multiframe image using wadouri( like /local), in here

https://github.com/cornerstonejs/cornerstone3D/blob/main/packages/dicomImageLoader/src/imageLoader/wadouri/getUncompressedImageFrame.ts#L54

frameIndex = last image index, so if the number of the images is 100, frameIndex = 100, issuing the exception

Using pixelDataFrame it works as it should

rodrigobasilio2022:
In here 

https://github.com/cornerstonejs/cornerstone3D/blob/main/packages/dicomImageLoader/src/imageLoader/wadouri/loadImage.ts#L224

it is working in OHIF because the last image is not loaded as the first time, so when an image is loaded the data is cached in dataSetCacheManager and in all the subsequent calls it uses the pixelDataFrame. But if the first frame is the last, an error occur

ReeceStevens:
For what it's worth: I want to highlight that I'm experiencing a similar issue with the frameIndex not being correct for certain datasets. I've been able to reliably reproduce this with an enhanced MR series that contains two files, each of which has 10 frames. Please see the attached zip file to this issue.

I am seeing that if I `loadAndCacheImage('dicomfile:0?frame=N')`, where N is the number of frames (e.g. I'm trying to load the last frame in the series first), I get an exception. If I explicitly pull out an earlier frame and load that first, then load the last frame, things work as expected. I think @rodrigobasilio2022 's assessment seems right, that the frame index being passed in is different (off by one) if the series is cached vs. not cached.

[series-1.zip](https://github.com/user-attachments/files/22546890/series-1.zip)

PR #2271, opened 2025-08-14
priority prefetch

m00n620:
Currently `0` is used for the image loader when prefetching for stack images. I added `priority` as a parameter to `enable` method. `0` will be used by default when omitting that parameter so it will not break others.

### DIFF
--- a/packages/tools/src/utilities/stackPrefetch/stackContextPrefetch.ts
+++ b/packages/tools/src/utilities/stackPrefetch/stackContextPrefetch.ts
@@ -11,7 +11,6 @@ import { addToolState, getToolState, type StackPrefetchData } from './state';
 import {
   getStackData,
   requestType,
-  priority,
   clearFromImageIds,
   getPromiseRemovedHandler,
 } from './stackPrefetchUtils';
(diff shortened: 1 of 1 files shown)


packages/tools/src/utilities/stackPrefetch/stackContextPrefetch.ts
```diff
@@ -62,8 +63,11 @@ const resetPrefetchDelay = 5;
  *     * Up to 100 images in the direction of travel will be prefetched
  *
  * @param element - to prefetch on
+ * @param priority - priority to be used for the request manager
  */
-const enable = (element): void => {
+const enable = (element, priority = 0): void => {
+  console.log(`CornerstoneTools.stackPrefetch - ${priority}`);
```
IbrahimCSAE: can u remove this

IbrahimCSAE:
Please add a description in the PR on what this is doing and what u are changing, I can't approve it without context.

m00n620:
> Please add a description in the PR on what this is doing and what u are changing, I can't approve it without context.

Updated the description, please check again.

PR #2275, opened 2025-08-17
Feat/3D rendering settings

mbellehumeur:
### Context

Ticket: 9756584582.

Adding sample distance multiplier to Cornerstone3D configuration.
This is used for VRT rendering.
The sample distance multiplier decreases the size of the volume when it is created. 
The other VRT  settings (opacity,shade,ect)  modify the properties of the existing volume.

### Changes & Results
Added the sampleDistanceMultiplier setting to config files.
Updated createvolumeviewport and basevolumeviewport to handle the new setting.

### Testing

Change sample distance (sampleDistanceMultiplier: 1) from 1 to 8 and notice lower GPU load and  lower resolution.

### DIFF
--- a/packages/core/src/RenderingEngine/BaseVolumeViewport.ts
+++ b/packages/core/src/RenderingEngine/BaseVolumeViewport.ts
@@ -1119,11 +1119,6 @@ abstract class BaseVolumeViewport extends Viewport {
     if (properties.preset !== undefined) {
       this.setPreset(properties.preset, volumeId, false);
     }
-
-    if (properties.preset !== undefined) {
-      this.setPreset(properties.preset, volumeId, false);
-    }
-
     this.render();
   }
 
--- a/packages/core/src/RenderingEngine/BaseVolumeViewport.ts
+++ b/packages/core/src/RenderingEngine/BaseVolumeViewport.ts
@@ -77,6 +77,8 @@ import { getCameraVectors } from './helpers/getCameraVectors';
 import { isContextPoolRenderingEngine } from './helpers/isContextPoolRenderingEngine';
 import type vtkRenderer from '@kitware/vtk.js/Rendering/Core/Renderer';
 import mprCameraValues from '../constants/mprCameraValues';
+import { setConfiguration, getConfiguration } from '@cornerstonejs/core';
+import type { Types } from '@cornerstonejs/core';
 /**
  * Abstract base class for volume viewports. VolumeViewports are used to render
  * 3D volumes from which various orientations can be viewed. Since VolumeViewports
@@ -137,6 +139,10 @@ abstract class BaseVolumeViewport extends Viewport {
     this.initializeVolumeNewImageEventDispatcher();
   }
 
+  public updateRenderingPipeline = () => {
+    this._updateRenderingPipeline();
+  };
+
   static get useCustomRenderingPipeline(): boolean {
     return false;
   }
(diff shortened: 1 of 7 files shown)


wayfarer3130:
@sedghi - for properties like the sample distance to use, should that be one of the "setProperties" on the viewport for the current value of that, while the configuration sets the current default values for a new volume.  That is, to update the value for the current viewport, one would call:

viewport.setProperties( { ... sampleDistanceMultiplier: 3 } )

where the default sampleDistanceMultiplier at the time of a viewport being created would come from the init.ts setting for the volumeRendering?  That way new viewports get the right initial settings, and the value can be set on a per-viewport basis.

I'm not quite sure how other shared settings like this are applied - I know hte gpu/cpu rendering is controlled for the renderer level, and one could argue that doing it at the renderer would be appropriate.

packages/core/src/RenderingEngine/BaseVolumeViewport.ts
```diff
@@ -1120,6 +1125,12 @@ abstract class BaseVolumeViewport extends Viewport {
     }
   }
 
+  public setSampleDistanceMultiplier(multiplier: number): void {
+    console.debug(
```
wayfarer3130: Can you just leave this as // No-op for base volume.

packages/core/src/RenderingEngine/helpers/createVolumeMapper.ts
```diff
@@ -30,7 +30,14 @@ export default function createVolumeMapper(
...
   // Set the sample distance to half the mean length of one side. This is where the divide by 6 comes from.
   // https://github.com/Kitware/VTK/blob/6b559c65bb90614fb02eb6d1b9e3f0fca3fe4b0b/Rendering/VolumeOpenGL2/vtkSmartVolumeMapper.cxx#L344
-  const sampleDistance = (spacing[0] + spacing[1] + spacing[2]) / 6;
+  let sampleDistance = (spacing[0] + spacing[1] + spacing[2]) / 6;
+
+  const volumeRenderingConfig =
+    getConfiguration().rendering?.volumeRendering || {};
+  if (volumeRenderingConfig.sampleDistanceMultiplier !== undefined) {
+    sampleDistance =
+      sampleDistance * volumeRenderingConfig.sampleDistanceMultiplier;
```
wayfarer3130: You need an || 1 on the sample distance multiplier in case it isn't defined.

packages/core/src/RenderingEngine/BaseVolumeViewport.ts
```diff
@@ -141,6 +143,10 @@ abstract class BaseVolumeViewport extends Viewport {
     return false;
   }
 
+  public updateRenderingPipeline = () => {
```
wayfarer3130: Is this method still needed to be exposed?

mbellehumeur: yes. it does not work when I remove it.

packages/core/src/RenderingEngine/helpers/createVolumeMapper.ts
mbellehumeur: ok

packages/core/src/RenderingEngine/helpers/createVolumeMapper.ts
```diff
@@ -30,7 +30,14 @@ export default function createVolumeMapper(
   const spacing = imageData.getSpacing();
   // Set the sample distance to half the mean length of one side. This is where the divide by 6 comes from.
   // https://github.com/Kitware/VTK/blob/6b559c65bb90614fb02eb6d1b9e3f0fca3fe4b0b/Rendering/VolumeOpenGL2/vtkSmartVolumeMapper.cxx#L344
-  const sampleDistance = (spacing[0] + spacing[1] + spacing[2]) / 6;
+  let sampleDistance = (spacing[0] + spacing[1] + spacing[2]) / 6;
+
+  const sampleDistanceMultiplier =
+    getConfiguration().rendering?.volumeRendering?.sampleDistanceMultiplier ||
+    1;
+  if (sampleDistanceMultiplier !== undefined) {
```
wayfarer3130: Will never be undefined as you just set it.  You can remove the if statement.

packages/core/src/RenderingEngine/BaseVolumeViewport.ts
wayfarer3130: It just tried without it - it works fine, just that the setProperties needs to get called on all the volume viewports.

wayfarer3130 (changes requested):
Lets remove the updateRenderingPipeline - it works fine without that.
Also, could you remove the unnecessary if statement?

packages/core/src/RenderingEngine/BaseVolumeViewport.ts
mbellehumeur: ok

packages/core/src/RenderingEngine/helpers/createVolumeMapper.ts
mbellehumeur: ok

wayfarer3130 (approved):

PR #2277, opened 2025-08-18
fix(viewport): adjust camera position with focal point in acquisition plane

Devu-trenser:
### Context

When the viewport launches in acquisition orientation, the focal point alignment logic updates the focal point but does not proportionally adjust the camera position.

This PR is contributed by [FlyWheel.io](https://flywheel.io/)

This results in:

- A slight camera tilt
- Z-axis variations during bounding box calculations (e.g., topLeft / bottomRight in [fillCircle.ts#L54](https://github.com/cornerstonejs/cornerstone3D/blob/v3.11.6/packages/tools/src/tools/segmentation/strategies/fillCircle.ts#L54))

Fixes issue: https://github.com/OHIF/Viewers/issues/5333

### Changes & Results
Updated focal point alignment logic to also adjust the camera position proportionally when launching in acquisition orientation.

### Testing
Verify the scenario mentioned in https://github.com/OHIF/Viewers/issues/5333

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] "OS:  Ubuntu 24
- [x] "Node version:   22
- [x] "Browser:  Chrome 134.0.0
  

### DIFF
--- a/packages/core/src/RenderingEngine/BaseVolumeViewport.ts
+++ b/packages/core/src/RenderingEngine/BaseVolumeViewport.ts
@@ -887,12 +887,16 @@ abstract class BaseVolumeViewport extends Viewport {
           [-viewPlaneNormal[0], -viewPlaneNormal[1], -viewPlaneNormal[2]],
           projectedDistance
         );
+        const focalShift = vec3.subtract(vec3.create(), newImagePositionPatient, focalPoint);
+        const newPosition = vec3.add(vec3.create(), position, focalShift);
         // this.setViewReference({
         //   ...viewRef,
         //   cameraFocalPoint: newImagePositionPatient as Point3,
         // });
         this.setCamera({
           focalPoint: newImagePositionPatient as Point3,
+          position: newPosition as Point3
+
         });
         this.render();
         return;
--- a/packages/core/src/RenderingEngine/BaseVolumeViewport.ts
+++ b/packages/core/src/RenderingEngine/BaseVolumeViewport.ts
@@ -896,7 +896,6 @@ abstract class BaseVolumeViewport extends Viewport {
         this.setCamera({
           focalPoint: newImagePositionPatient as Point3,
           position: newPosition as Point3
-
         });
         this.render();
         return;


Devu-trenser:
@wayfarer3130  Could you please take a moment to review this MR when you get a chance?

packages/core/src/RenderingEngine/BaseVolumeViewport.ts
```diff
@@ -887,12 +887,15 @@ abstract class BaseVolumeViewport extends Viewport {
...
         );
+        const focalShift = vec3.subtract(vec3.create(), newImagePositionPatient, focalPoint);
+        const newPosition = vec3.add(vec3.create(), position, focalShift);
         // this.setViewReference({
         //   ...viewRef,
         //   cameraFocalPoint: newImagePositionPatient as Point3,
         // });
         this.setCamera({
           focalPoint: newImagePositionPatient as Point3,
+          position: newPosition as Point3
```
sedghi, line 898: I wonder why we need this since this is auto recalculated at vtk internals hmmm

Devu-trenser: @sedghi 
VTK.js internally recalculates the direction of projection and distance when the focal point is updated. However, this internal recalculation the position remains same, and the direction is always computed as:
`directionOfProjection = focalPoint - position`. This can alter the current view direction.
(see [vtkCamera.setFocalPoin](https://github.com/Kitware/vtk-js/blob/bea78f5d/Sources/Rendering/Core/Camera/index.js#L78-L92)t and [vtkCamera.computeDistance](https://github.com/Kitware/vtk-js/blob/bea78f5d984ad6dad82aa8154e2010d8e4cc38db/Sources/Rendering/Core/Camera/index.js#L139-L143))


In our scenario (e.g., on acquisition-orientation launch), we want to **update the focal point without changing the current view direction**. If we only update the focal point, VTK.js will automatically adjust the direction vector based on the previous position, which alters the view direction and can lead to a tilt or Z-axis drift.

To prevent this, we update the camera position along with the focal point:

In the latest code, the issue related to [fillCircle](https://github.com/cornerstonejs/cornerstone3D/blob/v3.11.6/packages/tools/src/tools/segmentation/strategies/fillCircle.ts#L54) has been resolved, as the circle corner calculation logic was improved, resulting in a more accurate representation as see [here](https://github.com/cornerstonejs/cornerstone3D/blob/main/packages/tools/src/tools/segmentation/strategies/fillCircle.ts#L73-L75). However, a slight Z-tilt may still occur due to changes in the camera view angle, which could cause issues in the future.

wayfarer3130: I've seen this as well, and thought I'd had this change incorporated, but looks like this particular path wasn't checked fully.  The problem is that the current view **DIRECTION** does change, when the current **POSITION** is not updated.  Worse, this drifts because the value isn't calculated once and then re-used but on each update the position drifts a little further.

The right way to fix this is to store the focalShift when the viewport is reset, and to re-use the focalShift to recalculate the position internally to the setCamera call when position is not specified.

The focalShift could then be used to calculate default zoom, and would have to be updated when the zoom is updated.

I'd suggest including this PR for now, and creating a separate change request to adjust the setCamera behaviour to split out some of the behaviour to allow for independent control of relative zoom, zoom to fit/scale and different x/y scaling (stretch).

sedghi (approved):

PR #2221, opened 2025-07-22
fix(StartEndThresholdTools): Fix pointInsideVolumes calculation when multiples toolgroups

Celian-abd:
### Context
When the CircleROIStartEndThresholdTools / RectangleROIStartEndThresholdTools are added on multiples toolgroup such as [ ctToolGroup, ptToolGroup , fusionToolgroup]. The pointInsideVolume calculation was wrong and sometimes if the annotion was drawn on the ptViewport, it would take the ctViewport for the point calculation.

### Changes & Results
- Fixed the viewport selected for pointInsideVolume calculation

### Testing
Can be tested on the TMTV example

#### PR

- [] My Pull Request title is descriptive, accurate and follows the
  semantic-release format and guidelines.

#### Code

- [] My code has been well-documented (function documentation, inline comments,
  etc.)

#### Public Documentation Updates

- [] The documentation page has been updated as necessary for any public API
  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  

[blog]: https://circleci.com/blog/triggering-trusted-ci-jobs-on-untrusted-forks/
[script]: https://github.com/jklukas/git-push-fork-to-upstream-branch

PR #2280, opened 2025-08-19
add tests for circle and rectangle start stop

salimkanoun:
As discussed at office hour, I add a simple test to at least check those tools are not broken by changes,

Would need to make more extensive tests to tests manipulation of start and end slices (missing time to do it)

Thus PR https://github.com/cornerstonejs/cornerstone3D/pull/2254 and https://github.com/cornerstonejs/cornerstone3D/pull/2221 shall not break these tests


sedghi (approved):
Thanks

Issue #60, opened 2022-04-01
[Known Issue] Volume Viewport Flip


sedghi:
There is a known issue for fkipping in other acquisition axes.

Reason:
It calculates the flip based on the imageData metadata, which actually should be based on camera

sedghi:
This has been fixed

Issue #56, opened 2022-04-01
[Known Issue] Camera synchronizer not snapping to the middle of the slice


sedghi:
`CameraPositionSynchronizer` should snap the focal point to the slice's middle point

PR #2291, opened 2025-08-25
fix(dicom-loader): Correctly merge shared and per-frame functional groups for wadouri imageLoader

olhalit:
### Context

This change addresses a critical issue in the WADO-URI loader where DICOM metadata was being incorrectly merged, leading to data loss.

The root cause is the shallow merge strategy used in combineFrameInstanceDataset when combining SharedFunctionalGroupsSequence (x52009229) and PerFrameFunctionalGroupsSequence (x52009230).

When a nested sequence (e.g., PlaneOrientationSequence) exists in both the shared and per-frame groups, the per-frame version would completely overwrite the shared version.
If the per-frame sequence was missing some tags that were present in the shared one, those tags were lost.

This PR fixes the merging logic to ensure that no metadata is lost and that the frame-specific DataSet is correctly constructed.

### Changes & Results

Changes

   * Added `deepMergeElements` function: A new private function has been introduced to
     perform a recursive (deep) merge of DICOM element objects.
   * Updated Merge Logic: The core logic in combineFrameInstanceDataset was updated to use
     deepMergeElements. Instead of a shallow merge that overwrites entire sequences, the new
      logic correctly merges nested elements within sequences.
   * Preserved Attribute Priority: The implementation ensures that attributes from
     PerFrameFunctionalGroupsSequence correctly override those from
     SharedFunctionalGroupsSequence at every level of nesting, as intended by the DICOM
     standard.

  Results

   * Before: A shallow merge was causing the PerFrameFunctionalGroupsSequence to completely
     overwrite the SharedFunctionalGroupsSequence for tags that existed in both. This led to
      a loss of nested metadata attributes that were present only in the shared sequence.
   * After: The new deep merge logic correctly combines the shared and per-frame functional
     groups. Attributes from the per-frame group augment and override the shared group's
     attributes without data loss. The resulting DataSet for each frame is now complete and
     accurate.

### Testing

The changes were manually tested using a real-world DICOM file that originally exposed the
   metadata merging bug.

   1. Test Data:
       * The test was performed with a multi-frame DICOM file that contains both
         SharedFunctionalGroupsSequence (x52009229) and PerFrameFunctionalGroupsSequence
         (x52009230).
       * This file's metadata includes a nested PixelValueTransformationSequence (x00289110)
          in both the shared and per-frame functional groups.
       * The issue was discovered because the PixelSpacing tag (x00280030) was present
         within this sequence in the SharedFunctionalGroupsSequence but was absent from the
         sequence in the PerFrameFunctionalGroupsSequence, which caused it to be lost after
         the previous shallow merge logic.

   2. Test Procedure & Results:
       * The problematic DICOM file was loaded using the WADO-URI image loader.
       * The DataSet for a specific frame was then inspected after the image was fully
         loaded.
       * Verification: It was confirmed that the PixelSpacing tag is now correctly inherited
          from the shared PixelValueTransformationSequence and is present in the final
         merged metadata for the frame.

   3. Regression Testing:
       * A suite of other standard single-frame and multi-frame DICOM images were also
         loaded successfully, confirming that these changes have not introduced any
         regressions into the standard image loading process.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] "OS:  Ubuntu 24.04.2
- [x] "Node version:  20.17.0
- [x] "Browser: Chrome 139.0.7258.66
  

### DIFF
--- a/packages/dicomImageLoader/src/imageLoader/wadouri/combineFrameInstanceDataset.ts
+++ b/packages/dicomImageLoader/src/imageLoader/wadouri/combineFrameInstanceDataset.ts
@@ -81,6 +81,54 @@ function getMultiframeInformation(dataSet) {
   };
 }
 
+/**
+ * Recursively merges DICOM element objects. It gives precedence to per-frame elements,
+ * but for nested sequences (SQ), it merges their contents instead of replacing them.
+ */
+function deepMergeElements(shared, perFrame) {
+  // Start with a copy of the shared elements
+  const merged = { ...shared };
+
+  // Iterate over per-frame elements
+  for (const tag in perFrame) {
+    const sharedItem = merged[tag];
+    const perFrameItem = perFrame[tag];
+
+    // If the tag exists in both, and both are sequences (have .items),
+    // we need to merge their contents recursively.
+    if (
+      sharedItem &&
+      sharedItem.items &&
+      perFrameItem.items &&
+      sharedItem.items.length > 0 &&
+      perFrameItem.items.length > 0 &&
+      sharedItem.items[0].dataSet &&
+      perFrameItem.items[0].dataSet
+    ) {
+      // Assuming sequences in this context have only one item, which is standard.
+      // If there could be more, this logic would need to be more complex.
+      const mergedSubElements = deepMergeElements(
+        sharedItem.items[0].dataSet.elements,
+        perFrameItem.items[0].dataSet.elements
+      );
+
+      // Modify the dataSet of the sharedItem in place.
+      // This is necessary because we cannot create a new DataSet object with the correct prototype.
+      sharedItem.items[0].dataSet.elements = mergedSubElements;
+
+      // Create a new merged sequence item
+      merged[tag] = {
+        ...sharedItem,
+      };
+    } else {
+      // Otherwise, the per-frame element takes precedence (overwrite or add).
+      merged[tag] = perFrameItem;
+    }
+  }
+
+  return merged;
+}
+
 // function that retrieves specific frame metadata information from multiframe
 // metadata
 function combineFrameInstanceDataset(frameNumber, dataSet) {
@@ -102,12 +150,13 @@ function combineFrameInstanceDataset(frameNumber, dataSet) {
       frameNumber
     );
 
+    const mergedFrameElements = deepMergeElements(shared, perFrame);
+
     // creating a new copy of the dataset to remove the two multiframe dicom tags
     const newElements = {
       elements: {
         ...otherElements,
-        ...shared,
-        ...perFrame,
+        ...mergedFrameElements,
       },
     };
 


sedghi:
Do you have the dicom data to share so that we can test?

olhalit:
@sedghi Sure, I am sharing some Dicom files with you where this metadata issue is observed:
https://drive.google.com/file/d/1cNy-pLJ0crr0g4p0gctgDJdTT6jfY-Ut/view?usp=sharing

rfitzfrey:
Hi, I'm respectfully wondering if this PR is lacking something? It's affecting our application so it's something I'm tracking.

sedghi:
Sorry for the delay we will look

wayfarer3130:
The DICOM data you specified doesn't contain the PixelValueTransformationSequence (x00289110)
As well, even if it does, the transformation you suggest is illegal according to the DICOM standard:
https://dicom.nema.org/medical/dicom/current/output/html/part03.html#sect_C.7.6.16.1
specifically:
A single Functional Group shall not be included in both the Shared Functional Groups Sequence (5200,9229) and the Per-Frame Functional Groups Sequence (5200,9230).

It is certainly an error to have both, this could be dealt with a warning message, but I'm concerned about merge sub-sequences.
If I look at your provided files, the actual duplicated data is in PixelMeasuresSequence, and I would be ok with a PR that does a single level shallow merge of just the per-frame data over-writing the shared frame data, creating a warning log message.

62190: (5200,9229) SQ #64 SharedFunctionalGroupsSequence
62198: >(FFFE,E000) #56 Item #1
**62206: >(0028,9110) SQ #48 PixelMeasuresSequence**
62214: >>(FFFE,E000) #40 Item #1
62222: >>(0018,0088) DS #2 [3] SpacingBetweenSlices
**62232: >>(0028,0030) DS #22 [0.35714286\0.35714286] PixelSpacing**
62262: (5200,9230) SQ #6220 PerFrameFunctionalGroupsSequence
62270: >(FFFE,E000) #156 Item #1
62278: >(0020,9111) SQ #8 FrameContentSequence
62286: >>(FFFE,E000) #0 Item #1
62294: >(0020,9113) SQ #48 PlanePositionSequence
62302: >>(FFFE,E000) #40 Item #1
62310: >>(0020,0032) DS #32 [-124.29651\-92.541942\81.707022] ImagePositionPatient
62350: >(0020,9116) SQ #60 PlaneOrientationSequence
62358: >>(FFFE,E000) #52 Item #1
62366: >>(0020,0037) DS #44 [-0.034899498\0.99939083\0\2.220446e-016\0\-1] ImageOrientationPatient
**62418: >(0028,9110) SQ #8 PixelMeasuresSequence
62426: >>(FFFE,E000) #0 Item #1**

Note the second pixel measures sequence is empty - it would be fine just ensuring that the pixel measures sequence doesn't replace the earlier pixel measures sequence by doing a single shallow merge on those items on a per-item basis.

> @sedghi Sure, I am sharing some Dicom files with you where this metadata issue is observed: https://drive.google.com/file/d/1cNy-pLJ0crr0g4p0gctgDJdTT6jfY-Ut/view?usp=sharing

wayfarer3130:
Please also add a unit test for this.

packages/dicomImageLoader/src/imageLoader/wadouri/combineFrameInstanceDataset.ts
```diff
@@ -81,6 +81,54 @@ function getMultiframeInformation(dataSet) {
   };
 }
 
+/**
+ * Recursively merges DICOM element objects. It gives precedence to per-frame elements,
+ * but for nested sequences (SQ), it merges their contents instead of replacing them.
+ */
+function deepMergeElements(shared, perFrame) {
```
wayfarer3130, line 88: If you are going to do the merge, it has to be reasonably efficient, so you should start by using a Map containing the shared elements, and then any time a per-frame element is encountered, only perform the shallow merge on that single element.  You could even solve your current issue by doing an is-empty filter to remove any empty items in either the per-frame or shared functional groups.  Given the size of some groups, that is worthwhile - I have samples with over 300,000 per-frame entries and doing a deep merge/recreate objects on such data is extremely expensive.

wayfarer3130:
You might consider replacing the CS3D instance with the version from OHIF, as the OHIF version uses object inheritance to efficiently create the merged object in a way that would avoid needing your change at all - it does a single level replacement of just the values rather than merging the sequence objects the way the CS3D version does.  It also carefully remembers the objects at the instance level so that recreating the combined objects is efficient.

sedghi:
> You might consider replacing the CS3D instance with the version from OHIF, as the OHIF version uses object inheritance to efficiently create the merged object in a way that would avoid needing your change at all - it does a single level replacement of just the values rather than merging the sequence objects the way the CS3D version does. It also carefully remembers the objects at the instance level so that recreating the combined objects is efficient.

I like this let's do that

Here is the code https://github.com/OHIF/Viewers/blob/master/platform/core/src/utils/combineFrameInstance.ts

PR #2270, opened 2025-08-14
Range request handler did not apply `beforeSend` headers

jmhmd:
Fixes application of `beforeSend` headers for range requests, and also fixes serialization error for range requests when the retrieve config `chunkSize` parameter is a function, not a number.

#### PR

- [] My Pull Request title is descriptive, accurate and follows the
  semantic-release format and guidelines.

#### Code

- [] My code has been well-documented (function documentation, inline comments,
  etc.)

#### Public Documentation Updates

- [] The documentation page has been updated as necessary for any public API
  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  

[blog]: https://circleci.com/blog/triggering-trusted-ci-jobs-on-untrusted-forks/
[script]: https://github.com/jklukas/git-push-fork-to-upstream-branch

PR #2296, opened 2025-08-26
feat: Cornerstone3D 4.0

sedghi:
Migration Guides for 4.0

  1. Camera Field of View Changes

  - Breaking Change: Images now display edge-to-edge in viewports without the 10% padding
   that was present in 3.x
  - Migration: To restore the old behavior with padding, use:
  init({ rendering: { useLegacyCameraFOV: true } })

  2. Dynamic Volume API Changes

  - Breaking Change: Removed all deprecated timepoint-based APIs in favor of dimension
  group-based APIs
  - Removed APIs:
    - IDynamicImageVolume: Removed timePointIndex getter/setter and numTimePoints
  property
    - StreamingDynamicImageVolume: Removed timePointIndex, numTimePoints,
  getCurrentTimePointImageIds(), flatImageIdIndexToTimePointIndex(), isTimePointLoaded(),
   checkTimePointCompletion()
    - Events: Removed DYNAMIC_VOLUME_TIME_POINT_INDEX_CHANGED and
  DYNAMIC_VOLUME_TIME_POINT_LOADED
  - Migration:
    - Replace timePointIndex (0-based) with dimensionGroupNumber (1-based)
    - Replace numTimePoints with numDimensionGroups
    - Update event listeners to use DYNAMIC_VOLUME_DIMENSION_GROUP_CHANGED and
  DYNAMIC_VOLUME_DIMENSION_GROUP_LOADED
    - Note: Add 1 when converting from timePointIndex to dimensionGroupNumber

  3. Async beforeSend Callback

  - Breaking Change: The beforeSend callback in dicomImageLoader is now async and must
  return a Promise
  - Migration:
    - Wrap synchronous callbacks with Promise.resolve() or use async/await
    - Example:
    // Before (3.x)
  beforeSend: function(xhr, imageId, defaultHeaders, params) {
    return { Authorization: 'Bearer ' + token };
  }

  // After (4.x)
  beforeSend: async function(xhr, imageId, defaultHeaders, params) {
    return { Authorization: 'Bearer ' + token };
  }
  - Benefits: Enables async auth token fetching, token refresh, and integration with
  modern authentication workflows


wayfarer3130:
For the beforeSend, can you make it return either a promise, or the straight value?  A function that is awaited does not need to return a promise, only that it may return a promise, so it is fairly easy to handle either either way, and not changing it to require a promise means fewer outside changes.

packages/adapters/src/version.ts
wayfarer3130, line 1: Please do not commit adapters version.ts file as it is auto-generated.

sedghi: I think it is fine, it will auto generate them, in generate-version

const versionFilePath = path.join(srcDir, 'version.ts');
try {
  fs.writeFileSync(versionFilePath, versionContent);
  console.log(`Version exported to ${versionFilePath}`);
} catch (error) {
  console.error(`Error writing version.ts file:`, error);
  process.exit(1);
}

jmhmd:
Also regarding beforeSend, see my PR #2270 - rangeRequest.ts does not currently correctly apply `beforeSend` at all. I can update that PR to use the new async pattern, or you can just apply it here, it's just a couple lines.

wayfarer3130:
@sedghi - since the API is changing a bit here, could we fix the API for the arrow annotate label to make it consistently included inside the data object rather than having something inside the base annotation and then duplicated inside the annotation data, or worse having two different values there?

PR #2309, opened 2025-09-05
cursor pointer stroke width

m00n620:
- Introduced a new global style for cursor stroke width - `pointerStrokeWidth`
- `yarn example annotationToolModes` - you can see thick cursor pointer


IbrahimCSAE (approved):

sedghi:
As long as the default value is the current value and configurable it is acceptable, don't change default that people use though

m00n620:
I am not changing the default value here, only tested with example

IbrahimCSAE:
> I am not changing the default value here, only tested with example

thank you

PR #2315, opened 2025-09-08
feat: move cursor on resizing, showHandlesAlways config

m00n620:
- Added `showHandlesAlways` config - If true, some measurements will have handles always.
- Use `move` cursor when resizing annotations

### DIFF
--- a/packages/tools/src/eventDispatchers/mouseEventHandlers/mouseMove.ts
+++ b/packages/tools/src/eventDispatchers/mouseEventHandlers/mouseMove.ts
@@ -71,11 +71,11 @@ export default function mouseMove(evt: MouseMoveEventType) {
     }
   }
 
-  const showHandlesAlways = Boolean(
-    getStyleProperty('showHandlesAlways', {} as StyleSpecifier)
+  const showGrabCursorEnabled = Boolean(
+    getStyleProperty('showGrabCursor', {} as StyleSpecifier)
   );
 
-  if (showCrosshairsCursor && showHandlesAlways) {
+  if (showCrosshairsCursor && showGrabCursorEnabled) {
     setCursorForElement(element, 'move');
   } else {
     const toolGroup = getToolGroupForViewport(viewportId, renderingEngineId);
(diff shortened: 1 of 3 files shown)


m00n620:
Here's the demo video

https://github.com/user-attachments/assets/310cdc84-81da-4e2d-8486-6297103dd6aa

IbrahimCSAE:
merge only if CI passes

packages/tools/src/eventDispatchers/mouseEventHandlers/mouseMove.ts
```diff
@@ -48,11 +52,38 @@ export default function mouseMove(evt: MouseMoveEventType) {
...
+    setCursorForElement(element, 'move');
+  } else {
+    const toolGroup = getToolGroupForViewport(viewportId, renderingEngineId);
+    const activeTool = toolGroup?.getActivePrimaryMouseButtonTool();
+    if (activeTool) {
+      setCursorForElement(element, activeTool);
+    } else {
+      setCursorForElement(element, 'default');
     }
   }
```
sedghi: I understand the rest of the changes to the show handles always, I don't understand why we are changing this file?

m00n620: We want to show different cursor(like move, resize) when hovering the handles

tblock79:
This change breaks the "useCursors" option of the RuntimeSettings, i.e. disabling the SVG cursors is not possible anymore with:
Settings.getRuntimeSettings().set("useCursors", false);

The SVG mouse cursors can be difficult to see, so it would be important to have the ability to disable them.

Also, the code terminology is ambiguous. I understand that "showCrosshairsCursor" refers to the mouse cursor shown when hovering over a tool handle, but since there is also the Crosshairs tool (which provides a SVG cursor), it would be better to use a more clear name to avoid confusion.

Also, there could be performance implications when iterating over all annotations for every mouse move on the viewport. I would make this configurable. It would also help (a bit) to break out of the for loop once it has been identified that the move cursor should be shown.

PR #2324, opened 2025-09-12
fix(measurements): measurements were being shown in px when pixel spacing is defined

IbrahimCSAE:
## Context

using default values to decide if pixel spacing is present is wrong, cuz it gets set to true when other fields are not present that are not pixel spacing

### DIFF
--- a/packages/core/src/RenderingEngine/StackViewport.ts
+++ b/packages/core/src/RenderingEngine/StackViewport.ts
@@ -3448,10 +3448,7 @@ class StackViewport extends Viewport {
   private _getImagePlaneModule(imageId: string): ImagePlaneModule {
     const imagePlaneModule = getImagePlaneModule(imageId);
 
-    this.hasPixelSpacing =
-      !imagePlaneModule.usingDefaultValues ||
-      this.calibration?.scale > 0 ||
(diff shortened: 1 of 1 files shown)


wayfarer3130:
Can you link with OHIF and test:
* US region spacing
* ERMF spacing types
* Calibrated spacing on uncalibrated data and on re-calibrated data

wayfarer3130:
Probably adding a set of integration tests for the various cases is important.  The using default values is supposed to get set when there are no pixel spacing measures at all to use, and is supposed to not be set when there is some sort of pixel spacing measures to use.  Ideally using the enum flag for that would be better - that flag indicates the type of spacing.
It probably shouldn't also be the viewport's decision on spacing - that is really a function of the annotation position interacting with the metadata values.  The viewport only cares about spacing so that if it is displaying true size it knows the correct size to use.  Really, I'd like a change in the tools that does not rely on the viewport for that.

IbrahimCSAE:
> Can you link with OHIF and test:
> 
> * US region spacing
> * ERMF spacing types
> * Calibrated spacing on uncalibrated data and on re-calibrated data

I will do and post results, I can write a test as well for all the types

PR #2329, opened 2025-09-13
fix(coordinates): worldToCanvas and canvasToWorld should account for viewport bounds within the max viewport size canvas

IbrahimCSAE:
### Context

worldToCanvas and canvasToWorld were still assuming that a viewport consumes the full canvas, but this is not the case anymore as we default the canvas for a group of viewports to be the size of the largest one, and smaller viewports are rendered in the bottom left corner. This fixes the coordinates.


sedghi:
can you add a new test that is failing in main and passing in this branch please

sedghi (approved):

PR #2331, opened 2025-09-15
fix: NaN window level

pedrokohler:
### Context

If for some reason the voiLutModule windowWidth or windowCenter are empty arrays the fallback isn't working and it is returning NaN's.

### Changes & Results

With this change we now check if either of these values are NaN's, then we just return undefined which will allow the fallback to run and return correct window level values.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  

### DIFF
--- a/packages/core/src/RenderingEngine/helpers/setDefaultVolumeVOI.ts
+++ b/packages/core/src/RenderingEngine/helpers/setDefaultVolumeVOI.ts
@@ -91,6 +91,7 @@ function getVOIFromMetadata(imageVolume: IImageVolume): VOIRange | undefined {
         voi.voiLUTFunction = voiLutModule?.voiLUTFunction;
       }
       const { windowWidth, windowCenter } = voiLutModule;
+
       const width = Array.isArray(windowWidth) ? windowWidth[0] : windowWidth;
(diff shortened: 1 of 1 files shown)


packages/core/src/RenderingEngine/helpers/setDefaultVolumeVOI.ts
```diff
@@ -111,6 +112,11 @@ function getVOIFromMetadata(imageVolume: IImageVolume): VOIRange | undefined {
       Number(voi.windowCenter),
       voi.voiLUTFunction
     );
+
+    if (isNaN(lower) || isNaN(upper)) {
```
sedghi, line 116: How did we get to this clause? Neither the window level nor the width is zero to get here, how come the lower and upper becomes NAN?

pedrokohler: it says `voi.windowWidth ≠ 0` or `voi.windowCenter ≠ 0`

not that it must be a 0. So if it's undefined, it will pass.

Issue #2168, opened 2025-07-01
[Bug] setOptions/setDisplayArea does not fit image with respect to current transformation orientation in non-square canvas'


breaddog:
### What we are trying to do
We are trying to fit a rectangular image whose natural orientation is vertical (where h > w) in a horizontally rectangular viewport (w > h).

To achieve this we plan on performing the following steps:
1. Rotate the image to the correct orientation (vertical -> horizontal) with `setViewPresentation`
2. Fit the image into the horizontal rectangular viewport to ensue a good fit using `setOptions/setDisplayArea`

### What is the problem
Instead of the image being adjusted to fit into the viewport canvas horizontally after any transformation (i.e. rotation, flip), it has been adjusted to fit based on the original orientation of the image.

### Minimal Example Repo
We have created a minimal example repo which can be cloned that illustrates the problem this issue describes.
[Link to repo](https://github.com/breaddog/c3d-transform-aspect-ratio-html)

#### Viewports
This repo contains four viewports that helps to illustrate the problem.

![Image](https://github.com/user-attachments/assets/24581929-1491-4b18-960b-3674cee92529)

**Viewports 1-3 uses a vertical image (V), viewport 4 uses a horizontal image (H), both are the same image with the difference only being its initial orientation.**

- Viewports 1, 2 are used to describe the current behaviour based on what we are trying to do
- Viewport 3 is a control to demonstrate similar behaviour to compare with Viewports 1, 2

- Viewport 4 is a control to demonstrate the expected behaviour.

In order to avoid confusion, we will use the tags (V) and (H) to describe which image is being used.

### Replication steps
Thee base list of steps are as follows:
1. Set the canvas viewport aspect ratio to a horizontally rectangular one, we used 3:1 as an example
2. Load a vertically rectangular image (where h > w) into a Stack Viewport.
3. Rotate the image to the correct orientation using `setViewPresentation`
4. Call `setOptions` to fit the image into the display area.

### Explanation on Viewports
All viewports are loaded the same way and are based on these steps, with exception to steps 3-4 which in some cases are either swapped or sometimes omited to perform A-B and Control testing.

- Viewport 1 (V): Follows the steps above in order
- Viewport 2 (V):  1 -> 2 -> 4 -> 3
- Viewport 3 (V):  1 -> 2 -> 4
- Viewport 4 (V):  1 -> 2 -> 4

### Whats happening
[...]

PR #2334, opened 2025-09-16
fix(fov): fov was not working correctly for other planes

IbrahimCSAE:


sedghi (approved):
Looks reasonable

tblock79:
This fix seems to break the stack viewports in many cases, and I am not sure if it would work for volumes with oblique orientation. 

Using the legacy FOV calculation method for volume viewports could possibly be a quick fix:

```
if (imageData && !useLegacyMethod && (this.type !== ViewportType.ORTHOGRAPHIC)) {
    const extent = imageData.getExtent();
    widthWorld = (extent[1] - extent[0]) * spacing[0];
    heightWorld = (extent[3] - extent[2]) * spacing[1];
}
else {
    ({ widthWorld, heightWorld } = this._getWorldDistanceViewUpAndViewRight(bounds, viewUp, viewPlaneNormal));
}
```

IbrahimCSAE:
> This fix seems to break the stack viewports in many cases, and I am not sure if it would work for volumes with oblique orientation. 
> 
> 
> 
> Using the legacy FOV calculation method for volume viewports could possibly be a quick fix:
> 
> 
> 
> ```
> 
> if (imageData && !useLegacyMethod && (this.type !== ViewportType.ORTHOGRAPHIC)) {
> 
>     const extent = imageData.getExtent();
> 
>     const spacing = imageData.getSpacing();
> 
>     widthWorld = (extent[1] - extent[0]) * spacing[0];
> 
>     heightWorld = (extent[3] - extent[2]) * spacing[1];
> 
> }
> 
> else {
> 
>     ({ widthWorld, heightWorld } = this._getWorldDistanceViewUpAndViewRight(bounds, viewUp, viewPlaneNormal));
> 
> }
> 
> ```
> 
> 
> 
> 

Yeah this is broken thats why we changed the title to WIP. We could do legacy for Volume viewport but I believe it still doesn't always make the image take the full canvas. I'm leaving it up until we figure out something consistent

tblock79:
That's definitely true. Using the legacy scaling as fallback might still make sense for the interim time because the current release version is often cropping a significant number of slices in volume viewports (thus, the legacy scaling should be preferable - even if not perfect).

sedghi:
can you send us some screenshots of how much is it missing?

tblock79:
Sure, here is an example (left is the current release, right is when using the legacy method as fallback for volume viewports). As you can see, a significant number of slices is outside the FOV after a camera reset. The impact might be dataset-dependent but I noticed it quite often.
image

sedghi:
I think this is related issue https://github.com/cornerstonejs/cornerstone3D/issues/2168

wayfarer3130:
@IbrahimCSAE - can you tell me exactly how to reproduce the issue in this PR?  I've tweaked the PR to just compute the delta between your method and the current method and I'm see differences of less than 0.0001
@sedghi - unless we can get reproduction steps, I think this item can be closed.

sedghi:
@wayfarer3130 i remember the neptune study in MPR showed the issue regarding initial fov of reconstruction views

IbrahimCSAE:
> @IbrahimCSAE - can you tell me exactly how to reproduce the issue in this PR? I've tweaked the PR to just compute the delta between your method and the current method and I'm see differences of less than 0.0001 @sedghi - unless we can get reproduction steps, I think this item can be closed.

I don’t have clean repro steps beyond what Tobias originally reported, and it’s been a while since I last dug into this. I agree the delta you’re seeing is tiny.

What I’m proposing is to keep the old FOV logic for volume viewports, and keep the current/improved logic for stack viewports, so we don’t regress the stack behavior.

This is what I was proposing (and what @tblock79 also suggested):


```ts
if (imageData && viewport instanceof StackViewport) {
  const extent = imageData.getExtent();

  widthWorld = (extent[1] - extent[0]) * spacing[0];
  heightWorld = (extent[3] - extent[2]) * spacing[1];
} else {
  ({ widthWorld, heightWorld } = this._getWorldDistanceViewUpAndViewRight(
    bounds,
    viewUp,
    viewPlaneNormal
  ));
}
```

That way we don’t lose the improved FOV behavior for stack viewports, while restoring volume viewports to the old behavior.

PR #2340, opened 2025-09-18
feat(decimatedVolumeLoader):  optional downsizing and downsampling of volumes.

mbellehumeur:
### Context

ticket: 9788261802

Decimated volume loader that creates a StreamingImageVolume while delegating
 volume mutations to a plugin-style modifier chain. The loader still handles the
 responsibilities of preparing imageIds (including optional k-axis decimation),
 instantiating the streaming volume, and updating VTK/voxel-manager state, but it
 now relies on `enhancedVolumeModifiers` to encapsulate any transformations that
 are specific to projection parameters such as decimation.

 The loader currently wires a dedicated modifier set for:
 - **K-axis decimation handling**: Removes slices before metadata generation so
   that spacing reflects the decimated stack without manual scaling.
 - **In-plane decimation**: Implemented by `InPlaneDecimationModifier`, which
   updates dimensions, spacing, and DICOM metadata in a self-contained class.

 Additional modifiers can be introduced by implementing the shared
decimatedVolumeModifier` interface. For example:
 ```
 const modifiers = [
   inPlaneDecimationModifier,
   orientationModifier,         // rotates the direction matrix or redefines axis order
   partialRangeModifier,        // limits the volume to a subset of frames / slices
   windowingModifier,           // applies window/level or LUT adjustments at load time
 ];
 ```

The decimation part  of this PR  allows viewing of volumes that are too large for WebGL and to improve latency for low performance GPU hardware.  The features are:

#### Downsizing with Decimation factor [i,j,k] in enhancedVolumeLoader.

K-axis decimation: Reduces the number of slices by keeping every Nth slice
(e.g., decimation=2 keeps every other slice). This is applied before volume creation
unless already decimated at the displaySet level.

In-plane decimation: Downsamples the resolution of each slice by reducing pixel
dimensions. This is achieved by appending a decimation parameter to imageIds, which
is processed during image loading.

The BaseStreamingVolume has been adjusted to handle the desired dimensions (targetRows and targetColumns)
DecodeImageFrame is used to decode produce the PixelData. The volume spacing, dimensions, and DICOM metadata  are adjusted to reflect  the decimated resolution.

#### Downsampling with SampleDistanceMultiplier
Sets the VTK toolkit sampling distance when rendering volumes.

### Changes & Results

For downsizing:
- Added decimatedVolumeLoader which adds i,j decimation parameter to imageIDs.
- Added VolumeDecimatededLoading example.
- Modified BaseStreamingImageVolume to set the target dimensions of the image.
- Added decimation to JLS retrieve.

For downsampling:
Added sampleDistanceMultiplier to 
- ViewportProperties, 
-  Cornerstone3DConfig, 
- VolumeViewport3D,
- createVolumeMapper,
- BaseVolumeViewport,

Added sampling distance drop down to the the VolumeViewport3D example.

### Testing

Example implementation in a viewer:

https://github.com/user-attachments/assets/c576ed16-7393-47f0-978a-cbe09562f76f

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  


daker:
@mbellehumeur great feature, what's still missing to ship it ? do you need help to finish it ?

mbellehumeur:
> @mbellehumeur great feature, what's still missing to ship it ? do you need help to finish it ?

Thanks, that would be nice.  I'm wondering  if  additional smoothing can be done.
image

packages/core/src/loaders/volumeLoader.ts
```diff
@@ -28,10 +28,13 @@ import {
 import { generateVolumePropsFromImageIds } from '../utilities/generateVolumePropsFromImageIds';
 import type { StreamingDynamicImageVolume } from '../cache';
 import { cornerstoneStreamingImageVolumeLoader } from './cornerstoneStreamingImageVolumeLoader';
+import { decimateVolumeLoader } from './decimateVolumeLoader';
 
 interface VolumeLoaderOptions {
   imageIds: string[];
   progressiveRendering?: boolean;
+  kDecimation?: number; // Optional parameter for decimation factor
+  iDecimation?: number; // Optional parameter for interleave decimation
```
wayfarer3130: How about ijDecimation

wayfarer3130: Actually, using an array would be best here.

packages/core/src/utilities/decimateImagePixels.ts
```diff
@@ -0,0 +1,70 @@
+import type { IImage, PixelDataTypedArray } from '../types';
+export default function decimateImagePixels(image: IImage, factor: number) {
```
wayfarer3130: Don't forget to describe what this is doing in a comment

packages/core/src/utilities/decimateImagePixels.ts
```diff
@@ -0,0 +1,70 @@
...
+  const numComponents = image.numberOfComponents || 1;
+  const OutArrayCtor = pixelData.constructor as unknown as new (
+    length: number
+  ) => PixelDataTypedArray;
+  const out = new OutArrayCtor(newRows * newCols * numComponents);
+
+  let outIndex = 0;
+  for (let r = 0; r < newRows; r++) {
+    const inR = r * factor;
+    if (inR >= rows) break;
```
wayfarer3130: This indicates you aren't setting the final value - use Math.floor on line 10/11 above isntead and you should not need these breaks.

packages/tools/examples/volumeDecimatedLoading/index.ts
wayfarer3130, line 1: Can we call this enhancedVolumeLoading
to indicate that that it is other types of volume loading - that way it can include multiple volumes etc.

packages/core/src/loaders/decimatedVolumeLoader.ts
wayfarer3130, line 1: Can you include annotations and perhaps a segmentation, surviving the update to the volume being used - that is, show that annotations can still be successfully displayed on the 1/2 and 1/4 resolution images?

wayfarer3130:
@sedghi  - how do we want to handle segmentations and annotations in combination with the decimated volumes?  There are a few things to think about:
1. volumeId is different between volumes at different positions
2. Available world coordinates are not the same, so annotations on one volume may not correspond to image positions on a second one (eg 1/3 and 1/2 decimations only correspond every 6 frames)
3. Segmentations are also large instances, so probably need a way to create segmentations on the decimation.  Indeed having a decimated segmentation might be desirable even in the presence of full resolution data.

Over time, we will have other alternate volume representations - should there be a volume specific method to get the desired alternate volume(s), or should there be a custom actor for displaying decimated volumes, and we can add more actors in the future for other volume types?

packages/core/src/loaders/decimatedVolumeLoader.ts
mbellehumeur: Done

packages/core/src/loaders/volumeLoader.ts
mbellehumeur: array is now used.

packages/core/src/cache/classes/BaseStreamingImageVolume.ts
```diff
@@ -320,8 +320,10 @@ export class BaseStreamingImageVolume
     const { transferSyntaxUID: transferSyntaxUID } =
       metaData.get('transferSyntax', imageId) || {};
 
-    const imagePlaneModule = metaData.get('imagePlaneModule', imageId) || {};
-    const { rows, columns } = imagePlaneModule;
+    // Note: before PR 2340, rows and columns were from  imagePlaneModule = metaData.get('imagePlaneModule', imageId) .
```
wayfarer3130: Comment should be something like:
// Use the actual dimensions for this volume in order to support volumes not the same size as the raw data

packages/core/src/index.ts
```diff
@@ -64,6 +64,7 @@ import * as volumeLoader from './loaders/volumeLoader';
 import * as imageLoader from './loaders/imageLoader';
 import * as geometryLoader from './loaders/geometryLoader';
 import ProgressiveRetrieveImages from './loaders/ProgressiveRetrieveImages';
+import { enhancedVolumeLoader } from './loaders/enhancedVolumeLoader';
```
wayfarer3130: export * from './loaders/enhancedVolumeLoader'

sedghi:
> @sedghi - how do we want to handle segmentations and annotations in combination with the decimated volumes? There are a few things to think about:
> 
> 1. volumeId is different between volumes at different positions
> 2. Available world coordinates are not the same, so annotations on one volume may not correspond to image positions on a second one (eg 1/3 and 1/2 decimations only correspond every 6 frames)
> 3. Segmentations are also large instances, so probably need a way to create segmentations on the decimation.  Indeed having a decimated segmentation might be desirable even in the presence of full resolution data.
> 
> Over time, we will have other alternate volume representations - should there be a volume specific method to get the desired alternate volume(s), or should there be a custom actor for displaying decimated volumes, and we can add more actors in the future for other volume types?

This work is out of scope if it's for the "set aside" grant with FU Berlin. We only scoped 3D rendering downsampling and never proposed downsampling for regular viewports that might include segmentation and length. I initially suggested in slack we should first handle loading pixel spacings separately for regular MPR viewports and downsampled for 3D to meet the grant's aim. Then, if any funding becomes available, we can consider length and segmentation for downsampled MPR.

mbellehumeur:
> > @sedghi - how do we want to handle segmentations and annotations in combination with the decimated volumes? There are a few things to think about:
> > 
> > 1. volumeId is different between volumes at different positions
> > 2. Available world coordinates are not the same, so annotations on one volume may not correspond to image positions on a second one (eg 1/3 and 1/2 decimations only correspond every 6 frames)
> > 3. Segmentations are also large instances, so probably need a way to create segmentations on the decimation.  Indeed having a decimated segmentation might be desirable even in the presence of full resolution data.
> > 
> > Over time, we will have other alternate volume representations - should there be a volume specific method to get the desired alternate volume(s), or should there be a custom actor for displaying decimated volumes, and we can add more actors in the future for other volume types?
> 
> This work is out of scope if it's for the "set aside" grant with FU Berlin. We only scoped 3D rendering downsampling and never proposed downsampling for regular viewports that might include segmentation and length. I initially suggested in slack we should first handle loading pixel spacings separately for regular MPR viewports and downsampled for 3D to meet the grant's aim. Then, if any funding becomes available, we can consider length and segmentation for downsampled MPR.

The MPR decimation is independent of the 3D decimation when we use different volume id in the viewer.  I can demo tomorrow in office hours.

packages/core/src/loaders/enhancedVolumeLoader.ts
```diff
@@ -0,0 +1,236 @@
...
+import decimate from '../utilities/decimate';
+import VoxelManager from '../utilities/VoxelManager';
+interface IVolumeLoader {
+  promise: Promise<StreamingImageVolume>;
+  cancel: () => void;
+  decache: () => void;
+}
+
+/**
+ * Enhanced volume loader that creates a StreamingImageVolume with optional decimation
```
wayfarer3130: Can we update this comment to specify that the enhancement in this volume loader is that the volume can be loaded at different orientations, sampling or in some other way different from the standard volume loader.

Then, indicate the FIRST implementation of this is decimated volume loading, but that there are plans for allowing other orientation and partial range volume loading, as well as image access to alternate orientations?  That way this is a base class to add more implementation on top of.

packages/core/src/utilities/generateVolumePropsFromImageIds.ts
wayfarer3130, line 1: This class starts mixing the loader class information with the generate volume props.
The right way to do this is to have a class that specifies the enhanced volume information, and then knows how to generate the specific updated information here.  That way you have encapsulated the specific types of changes into the enhanced loader class, whereas here you know the internal details about the decimation from an external class and every time you add a new type of class it will bite you in that you need mroe and more complex changes.

packages/tools/src/tools/VolumeCroppingTool.ts
```diff
@@ -1086,142 +1294,139 @@ class VolumeCroppingTool extends BaseTool {
...
     const corners = [
-      [xMin, yMin, zMin],
-      [xMin, yMin, zMax],
-      [xMin, yMax, zMin],
-      [xMin, yMax, zMax],
-      [xMax, yMin, zMin],
-      [xMax, yMin, zMax],
-      [xMax, yMax, zMin],
-      [xMax, yMax, zMax],
+      indexToWorld(iMin, jMin, kMin),
```
wayfarer3130: There are existing functions in the code that generate this list.

packages/tools/src/tools/VolumeCroppingTool.ts
```diff
@@ -1086,142 +1294,139 @@ class VolumeCroppingTool extends BaseTool {
...
+    const diag0 = indexToWorld(0, 0, 0);
+    const diag1 = indexToWorld(dimensions[0], dimensions[1], dimensions[2]);
+    const diagonal = Math.sqrt(
+      Math.pow(diag1[0] - diag0[0], 2) +
+        Math.pow(diag1[1] - diag0[1], 2) +
+        Math.pow(diag1[2] - diag0[2], 2)
     );
+    const adaptiveRadius = this._calculateAdaptiveSphereRadius(diagonal);
 
+    // Add face spheres - now using 'i', 'j', 'k' instead of 'x', 'y', 'z'
```
wayfarer3130: This is a fundamental change that breaks the sagittal etc functionality.  I know it makes it a lot easier to understand/manage, but we need to understand exactly what is changing, decide whether or not that is acceptable, and if so, document it in the migrations guide, and what would be done as an alternative.
Really, the volume cropping tools should only change minimally to support these use cases.  If there are separate fixes needed for it, then lets discuss what those fixes need to be.

packages/core/src/utilities/generateVolumePropsFromImageIds.ts
wayfarer3130: One way to do this is to add various plugins that deal with different parameter types, and make the appropriate modifications to the properties.  Those would have to be applied in the correct order, but a list of modifiers could be added.  I htink that would make it a lot safer, and then you can add a unit test on those modifiers.

mbellehumeur: Removed decimation from generateVolumePropsFromImageIds and moved to ImageVolumeProps of the loader.

packages/core/src/loaders/enhancedVolumeLoader.ts
mbellehumeur: ok

packages/tools/src/tools/VolumeCroppingTool.ts
mbellehumeur: Reverted, volume cropping tool changes will be addressed in another PR

packages/tools/src/tools/VolumeCroppingTool.ts
mbellehumeur: Reverted, volume cropping tool changes will be addressed in another PR

packages/core/src/index.ts
mbellehumeur: ok

packages/core/src/loaders/enhancedVolumeLoader.ts
mbellehumeur: done

igoroctaviano:
@wayfarer3130 @sedghi could you guys review this?

sedghi:
is this separated to its own loader? the requirement is to be able to load MPR full resolution and in the same grid the 3D with down smapling, the downsampling should ONLY affect 3D

mbellehumeur:
> is this separated to its own loader? the requirement is to be able to load MPR full resolution and in the same grid the 3D with down smapling, the downsampling should ONLY affect 3D

It is up to the viewer.  I have changed the OHIF implementation so that only volume3d is affected when changing the values in the viewport control (so not like the video above).  MPR decimation only happens in the volume is too large to be rendered in the MPR viewport.

sedghi (commented):
I can't get the example to work

Image

---

sedghi (commented):
d

packages/core/examples/volumeViewport3D/index.ts
```diff
@@ -81,7 +81,8 @@ addButtonToToolbar({
     viewport.render();
   },
 });
-
+let viewport;
+let sampleDistanceMultiplier = 1;
```
sedghi: seems unnecessary please revert

packages/core/src/loaders/decimatedVolumeModifiers/applyDecimatedVolumeModifiers.ts
```diff
@@ -0,0 +1,19 @@
...
+/**
+ * Runs every modifier sequentially and returns the final props.
+ */
+export function applyEnhancedVolumeModifiers(
+  baseProps: ImageVolumeProps,
+  modifiers: EnhancedVolumeModifier[],
+  context: EnhancedVolumeModifierContext
+): ImageVolumeProps {
+  return modifiers.reduce(
+    (currentProps, modifier) => modifier.apply(currentProps, context),
```
sedghi, line 16: i don't understand all these complexity, do you have a bigger plan? since all these tiny files can be helpr functions within the same volume loader

packages/core/src/loaders/decimatedVolumeLoader.ts
```diff
@@ -0,0 +1,247 @@
...
+      decimatedResult.length &&
+      typeof decimatedResult[0] === 'number'
+        ? decimatedResult.map((idx) => options.imageIds[idx])
+        : decimatedResult;
+
+    options.imageIds = decimatedImageIds as string[];
+  }
+
+  // Apply in-plane decimation parameter to imageIds
+  if (columnDecimation > 1) {
```
sedghi, line 143: what happens to row decimation? please add a comments

packages/core/src/utilities/applyPreset.ts
```diff
@@ -8,10 +8,12 @@ import type { VolumeActor } from '../types/IActor';
  *
  * @param actor - The volume actor to apply the preset to.
  * @param preset - The preset to apply.
+ * @param volumeId - The volume ID to check for decimation.
  */
 export default function applyPreset(
   actor: VolumeActor,
-  preset: ViewportPreset
+  preset: ViewportPreset,
+  _volumeId?: string
```
sedghi: seems unnecessary change

packages/core/src/utilities/generateVolumePropsFromImageIds.ts
```diff
@@ -54,11 +54,13 @@ function generateVolumePropsFromImageIds(
...
+  const spacing = [
+    PixelSpacing[1], // column spacing (x)
+    PixelSpacing[0], // row spacing (y)
+    zSpacing,
+  ] as Point3;
 
-  // Spacing goes [1] then [0], as [1] is column spacing (x) and [0] is row spacing (y)
-  const spacing = [PixelSpacing[1], PixelSpacing[0], zSpacing] as Point3;
-  const dimensions = [Columns, Rows, numFrames].map((it) =>
+  const dimensions = [Columns, Rows, sortedImageIds.length].map((it) =>
```
sedghi: seems unnecessary

packages/core/src/index.ts
```diff
@@ -88,6 +89,8 @@ import {
   addImageSlicesToViewports,
 } from './RenderingEngine/helpers';
 
+export * from './loaders/enhancedVolumeLoader';
```
sedghi: i'm not sure if enhancedVolumeLoader is a good name, maybe decimatedVolumeLoader?

packages/tools/src/tools/displayTools/Labelmap/addVolumesAsIndependentComponents.ts
```diff
@@ -49,7 +49,7 @@ export async function addVolumesAsIndependentComponents({
   // need to remove the old actor/mapper and convert it to a new one
   // which the segmentation data is added as a second component to the volume data
   const defaultActor = viewport.getDefaultActor();
-  const { actor } = defaultActor as { actor: vtkVolume };
+  const { actor } = defaultActor as unknown as { actor: vtkVolume };
```
sedghi: if you don't need it in your build and ci passes please do not modify these

utils/ExampleRunner/example-info.json
```diff
@@ -630,6 +630,10 @@
       "volumeProgressive": {
         "name": "Volume Progressive Loading",
         "description": "Volume progressive loading inter and intra image"
+      },
+      "VolumeEnhancedLoading": {
```
sedghi: DecimatedVolumeLoading or something

packages/tools/src/tools/displayTools/Labelmap/addVolumesAsIndependentComponents.ts
mbellehumeur: removed in addOrUpdateSurfaceToElement as well.  It uses not to pass eslint but now it does.

packages/core/examples/volumeViewport3D/index.ts
mbellehumeur: ok

utils/ExampleRunner/example-info.json
mbellehumeur: it was decimate but change was requested to this name "Can we call this enhancedVolumeLoading
to indicate that that it is other types of volume loading - that way it can include multiple volumes etc."

packages/core/src/loaders/decimatedVolumeModifiers/applyDecimatedVolumeModifiers.ts
mbellehumeur: Yes, please see review comments from Bill 2 weeks ago and PR feature description : 
"Can we update this comment to specify that the enhancement in this volume loader is that the volume can be loaded at different orientations, sampling or in some other way different from the standard volume loader.

Then, indicate the FIRST implementation of this is decimated volume loading, but that there are plans for allowing other orientation and partial range volume loading, as well as image access to alternate orientations? That way this is a base class to add more implementation on top of."

packages/core/src/loaders/decimatedVolumeLoader.ts
mbellehumeur: added comment

packages/core/src/utilities/applyPreset.ts
mbellehumeur: removed

packages/core/src/utilities/generateVolumePropsFromImageIds.ts
mbellehumeur: removed

utils/ExampleRunner/example-info.json
sedghi: hmmm did i say that? feel wrong, it is doing decimation so it should be decimatedVolumeLoading or subResolutionVolumeLoading or similiar

sedghi (commented):
d

packages/core/src/loaders/decimatedVolumeModifiers/inPlaneDecimationModifier.ts
```diff
@@ -0,0 +1,51 @@
+import type Point3 from '../../types/Point3';
+import type { ImageVolumeProps } from '../../types';
+import type {
+  DecimatedVolumeModifier,
+  DecimatedVolumeModifierContext,
+} from './types';
+
+export const inPlaneDecimationModifier: DecimatedVolumeModifier = {
```
wayfarer3130, line 31: Please add comments about what this modifier is doing.  I can figure it out from the code, but the name isn't quite descriptive enough for me to understand it without reading the code.

packages/tools/examples/volumeDecimatedLoading/index.ts
```diff
@@ -0,0 +1,627 @@
+// For the low resolution image loader, what is required to do this is to create a new volume with a metadata loader that has a reduced set of images.
```
wayfarer3130: This should be a /** ... */ comment so that docs handlers can extract it as a file comment

wayfarer3130:
Please make sure the decimation=... parameter is NOT included in the URL to the back end server.

wayfarer3130:
This seems to be failing a number of the required checks - unit tests, format, docusaurus build etc.

mbellehumeur:
> Please make sure the decimation=... parameter is NOT included in the URL to the back end server.

That is the name agreed with Denny.  What is your suggestion?

sedghi (approved):

wayfarer3130:
> > Please make sure the decimation=... parameter is NOT included in the URL to the back end server.
> 
> That is the name agreed with Denny. What is your suggestion?

It isn't a standard DICOM request attribute, so you can't just add a parameter like that.  You need to change the load path, or else add a parameter to the DICOM loader to allow using custom names.  Basically you need a configuration somewhere that says whether or not to include any extra attributes when applying decimation.  If those are not provided, then the server can't respond with a decimated version, and you would have to live with the fact that the image needs to be decimated locally before displaying.

PR #2365, opened 2025-10-01
feat(Stackprefetch): add event that is triggered when the stack prefetch for a viewport has completed

tblock79:
### Context

When using the prefetch mechanism for stack viewports, it is often desirable to give the user visual feedback regarding the prefetching progress, so that the user can see when a series has been completely loaded and scrolling across the stack is possible. Currently, it is possible to monitor the load progress via events triggered by the dicom-image-loader but the images still need to be decoded afterwards, which can take a moment for large series. The IMAGE_LOAD event on the eventTarget, on the other hand, triggers for also for scroll events. Thus, hooking into this event could have performance implications. Therefore, this PR adds a new event that is triggered from the DoneCallback function of the StackPrefetch mechanism when the prefetching list has been emptied. The last-loaded imageId is provided in the event details, so that the viewport displaying the series can be easily determined. 

### Changes & Results

* Added the new event type to Events.ts
* Added the triggerEvent function to the DoneCallback function in StackPrefetch.ts

### Testing

* Add event listener for the new event to the eventTarget:
`eventTarget.addEventListener(EVENTS.STACK_PREFETCH_COMPLETE, someCallbackFunction);`
* Create a viewport and load a image series with stack prefetching enabled
* Validate that the callback function gets triggered when the prefetching is complete

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] "OS: Windows 11
- [x] "Node version: 22.17.0
- [x] "Browser: Chrome 142.0, Firefox 143.0
  

### DIFF
--- a/packages/tools/src/utilities/stackPrefetch/stackContextPrefetch.ts
+++ b/packages/tools/src/utilities/stackPrefetch/stackContextPrefetch.ts
@@ -6,6 +6,7 @@ import {
   cache,
   metaData,
   utilities,
+  triggerEvent,
 } from '@cornerstonejs/core';
 import { addToolState, getToolState, type StackPrefetchData } from './state';
 import {
@@ -14,6 +15,8 @@ import {
   clearFromImageIds,
   getPromiseRemovedHandler,
 } from './stackPrefetchUtils';
+import { Events } from '../../enums';
+import type { EventTypes } from '../../types';
 
 const { imageRetrieveMetadataProvider } = utilities;
 
@@ -221,6 +224,17 @@ function prefetch(element, priority = 0) {
           // );
         }
       }
+    }
+
+    // If all requests are complete, trigger the STACK_PREFETCH_COMPLETE event,
+    // providing the last imageId and triggering element so that the stack can
+    // be identified
+    if (stackPrefetch.indicesToRequest.length === 0) {
+      const eventDetail: EventTypes.StackPrefetchCompleteEventDetail = {
+        imageId: imageId,
+        element: element,
+      };
+      triggerEvent(eventTarget, Events.STACK_PREFETCH_COMPLETE, eventDetail);
     }
   }
 
--- a/packages/tools/src/utilities/stackPrefetch/stackContextPrefetch.ts
+++ b/packages/tools/src/utilities/stackPrefetch/stackContextPrefetch.ts
@@ -231,8 +231,8 @@ function prefetch(element, priority = 0) {
     // be identified
     if (stackPrefetch.indicesToRequest.length === 0) {
       const eventDetail: EventTypes.StackPrefetchCompleteEventDetail = {
-        imageId: imageId,
         element: element,
+        lastPrefetchedImageId: imageId,
       };
       triggerEvent(eventTarget, Events.STACK_PREFETCH_COMPLETE, eventDetail);
     }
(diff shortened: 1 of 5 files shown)


packages/tools/src/utilities/stackPrefetch/stackPrefetch.ts
```diff
@@ -127,6 +128,14 @@ function prefetch(element) {
     const imageIdIndex = stack.imageIds.indexOf(imageId);
 
     removeFromList(imageIdIndex);
+
```
sedghi, line 133: This is the old stackPrefetch which has some bugs, you should use the the one stackContextPrefetch

tblock79: Many thanks for the information! I wasn't aware that it has been replaced with another implementation.

Would it be OK if I extend the PR to trigger the event from both versions (stackPrefetch and stackContextPrefertch)? Or do you see another (inexpensive) way to get a notification on the completion of prefetching? This would greatly improve the end user experience (the scrolling can be quite bumpy if the user starts scrolling while the decoding is still ongoing).

sedghi: yes please do so
i'm not sure if that is enough even detail
you are just publishing one image Id 
you should at least publish the element

tblock79: OK, I changed the PR to add the event trigger to both the stackPrefetch and stackContextPrefetch. Moved the event definition to the Tools package (where both stackPrefetch and stackContextPrefetch reside) and added a type definition for the event details. Including the element in the event details now, as you suggested (in our case, receiving one imageId is sufficient, as we can extract the series UID from it, but including the element makes total sense).

packages/tools/src/utilities/stackPrefetch/stackContextPrefetch.ts
```diff
@@ -222,6 +225,17 @@ function prefetch(element, priority = 0) {
         }
       }
     }
+
+    // If all requests are complete, trigger the STACK_PREFETCH_COMPLETE event,
+    // providing the last imageId and triggering element so that the stack can
+    // be identified
+    if (stackPrefetch.indicesToRequest.length === 0) {
+      const eventDetail: EventTypes.StackPrefetchCompleteEventDetail = {
+        imageId: imageId,
```
sedghi: i still don't understand why we have imageId here, a stack is a set of imageIds, which image id is this?

tblock79: It is the last imageId that was loaded through the prefetch mechanism, which can be used to identify the corresponding viewport(s). Alternatively, the viewport can be identified through the element that is passed in the event details as well.

Not sure if there is value in passing the full stack data with the event (which can be a long list). In our use case, we just need a notification that prefetching has been completed. Since the same image series can be displayed in multiple viewports, it is easiest for us to get the series UID of the prefetched series, which can be extracted from any single imageId of the stack.

Shall I rename imageId to lastPrefetchedImageId to make it more clear?

tblock79: Just renamed the event-detail member, and also moved the order around (since other people might rather use the element for identifying the viewports).

sedghi (approved):

PR #2368, opened 2025-10-02
fix ellipse ROI stats

m00n620:
Fixed EllipticalROITool stats missing bug

Before
before

After
after

### DIFF
--- a/packages/tools/src/tools/annotation/EllipticalROITool.ts
+++ b/packages/tools/src/tools/annotation/EllipticalROITool.ts
@@ -5,10 +5,14 @@ import {
   VolumeViewport,
   utilities as csUtils,
   getEnabledElementByViewportId,
+  EPSILON,
 } from '@cornerstonejs/core';
 import type { Types } from '@cornerstonejs/core';
 
-import { getCalibratedLengthUnitsAndScale } from '../../utilities/getCalibratedUnits';
+import {
+  getCalibratedAspect,
+  getCalibratedLengthUnitsAndScale,
+} from '../../utilities/getCalibratedUnits';
 import throttle from '../../utilities/throttle';
 import {
   addAnnotation,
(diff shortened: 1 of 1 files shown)


IbrahimCSAE (approved):
Wait for Alireza before merging, it looks ok to me but 1 more review is safer

packages/tools/src/tools/annotation/EllipticalROITool.ts
```diff
@@ -1039,111 +1043,118 @@ class EllipticalROITool extends AnnotationTool {
...
+        const { scale, unit, areaUnit } = getCalibratedLengthUnitsAndScale(
+          image,
+          handles
+        );
+        const aspect = getCalibratedAspect(image);
+        const area = Math.abs(
+          Math.PI *
+            (worldWidth / scale / 2) *
+            (worldHeight / aspect / scale / 2)
+        );
```
sedghi, line 1107: i think this is the main change? why are we deviding by aspect? are we doing this division (by aspect) elsewhere too?

m00n620: I don't think this is the main change. I got the inspiration from CircleROITool.ts

m00n620:
```
        const ellipseObj = {
          center,
          xRadius: xRadius < EPSILON / 2 ? 0 : xRadius,
          yRadius: yRadius < EPSILON / 2 ? 0 : yRadius,
          zRadius: zRadius < EPSILON / 2 ? 0 : zRadius,
        };
```

I think this will do tricks

sedghi: Makes sense

sedghi (approved):

PR #2375, opened 2025-10-07
feat: limit pan off viewport

m00n620:
Added `limitToViewport` configuration to Pan tool so that we can limit panning off viewport.

### DIFF
--- a/packages/tools/src/tools/PanTool.ts
+++ b/packages/tools/src/tools/PanTool.ts
@@ -1,5 +1,5 @@
 import { BaseTool } from './base';
-import { getEnabledElement } from '@cornerstonejs/core';
+import { getEnabledElement, utilities as csUtils } from '@cornerstonejs/core';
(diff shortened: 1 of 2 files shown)


IbrahimCSAE (approved):

packages/tools/src/tools/PanTool.ts
```diff
@@ -39,9 +43,74 @@ class PanTool extends BaseTool {
...
       return;
     }
-    const camera = enabledElement.viewport.getCamera();
+    const viewport = enabledElement.viewport;
+    const { canvas } = viewport;
+    const camera = viewport.getCamera();
     const { focalPoint, position } = camera;
 
+    if (this.configuration.limitToViewport) {
+      const ratio = window.devicePixelRatio;
```
sedghi: please refactor to an inner method, this is noise when reading the code

IbrahimCSAE:
wait for CI then merge, thanks

PR #2378, opened 2025-10-09
fix(fov): Forcing the use of legacy FOV scaling method for orthographic viewports (temporary fix)

tblock79:
### Context

Currently, orthographic viewports do not provide correct FOV estimates on camera reset when showing reformatted orientations of the volume. As a result, the loaded image volume is cropped and a significant number of slices can be "hidden" outside the FOV (see screenshot below). 

The problem has been addressed in the in PR #2334, but the proposed solution does not work yet.

The current PR introduces a fallback to the legacy FOV scaling method for orthographic viewports. This is intended as temporary fix until another solution has been developed. The legacy method seems to work well for the orthographic viewports (I have not noticed any issues with the FOV scaling).

image

### Changes & Results

In the function resetCamera() in Viewports.ts, a condition has been added that forces the fallback to the legacy scaling method for orthographic viewports when estimating the world dimensions. 

### Testing

* Load a volumetric dataset (e.g., CT) into a volume viewport in orthographic mode with an orientation different from the acquired orientation. 
* Zoom-out to validate that there are no hidden slices outside the FOV

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] "OS: Windows 11"
- [x] "Node version: 22.17.0"
- [x] "Browser:  Chrome 142.0, Firefox 143.0"

### DIFF
--- a/packages/core/src/RenderingEngine/Viewport.ts
+++ b/packages/core/src/RenderingEngine/Viewport.ts
@@ -1134,7 +1134,13 @@ class Viewport {
     const config = getConfiguration();
     const useLegacyMethod = config.rendering?.useLegacyCameraFOV ?? false;
 
-    if (imageData && !useLegacyMethod) {
+    // Note: Force the legacy method for orthographic views as the current method
+    // does not provide correct FOV estimates for reformatted orientations.
+    if (
+      imageData &&
+      !useLegacyMethod &&
+      this.type !== ViewportType.ORTHOGRAPHIC
+    ) {
       const extent = imageData.getExtent();
       const spacing = imageData.getSpacing();
 


packages/core/src/RenderingEngine/Viewport.ts
```diff
@@ -1134,7 +1134,13 @@ class Viewport {
     const config = getConfiguration();
     const useLegacyMethod = config.rendering?.useLegacyCameraFOV ?? false;
 
-    if (imageData && !useLegacyMethod) {
+    // Note: Force the legacy method for orthographic views as the current method
+    // does not provide correct FOV estimates for reformatted orientations.
+    if (
+      imageData &&
+      !useLegacyMethod &&
+      this.type !== ViewportType.ORTHOGRAPHIC
```
sedghi, line 1142: so can't you opt out of this new FOV (until we fix it) by just setting the useLegacy to true?

tblock79: Yes, but the impact of useLegacy on the StackViewport is quite large (and StackViewports are probably used most of the time). Below is a comparison of the new and old FOV scaling for a StackViewport (the v4 scaling makes much better use of the space).

I guess it's a question of how long the final fix might take. If it will be fixed in the next few weeks, then it would be OK to leave it as is. Otherwise, might make sense to implement the fallback for now. 

I realized that I forgot to upload the updated reference screenshots for the Playwright test. I will add them just in case.

image

tblock79:
Closing this and resubmitting a new PR because somehow, it's not picking up my second commit.

PR #2379, opened 2025-10-09
fix(fov): Use of legacy FOV scaling method for orthographic viewports (temporary fix)

tblock79:
### Context

Currently, orthographic viewports do not provide correct FOV estimates on camera reset when showing reformatted orientations of the volume. As a result, the loaded image volume is cropped and a significant number of slices can be "hidden" outside the FOV (see screenshot below). 

The problem has been addressed in the in PR #2334, but the proposed solution does not work yet.

The current PR introduces a fallback to the legacy FOV scaling method for orthographic viewports. This is intended as temporary fix until another solution has been developed. The legacy method seems to work well for the orthographic viewports (I have not noticed any issues with the FOV scaling).

_Resubmitted PR because the previous PR did not pickup the commit with the added screenshots._

image

### Changes & Results

In the function resetCamera() in Viewports.ts, a condition has been added that forces the fallback to the legacy scaling method for orthographic viewports when estimating the world dimensions. 

### Testing

* Load a volumetric dataset (e.g., CT) into a volume viewport in orthographic mode with an orientation different from the acquired orientation. 
* Zoom-out to validate that there are no hidden slices outside the FOV

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] "OS: Windows 11"
- [x] "Node version: 22.17.0"
- [x] "Browser:  Chrome 142.0, Firefox 143.0"


vitblock:
Yes, the idea would be to keep the new (v4) FOV scaling for the stack viewports (since it makes much better use of the viewport space) but use the legacy scaling for the orthographic viewports because the new FOV scaling does not work correctly for anisotropic datasets (thus, slices often remain hidden outside the FOV when displaying reformats). This would be temporary until a final solution has been found.

The new PR is just because I added the updated screenshots for the Playwright tests and somehow, GitHub wasn't picking up the added commit in the initial PR (not sure why). That's why I recreated the PR (which now includes both commits).

sedghi:
please bear with us we are fixing the fov for orthogonal next week

wayfarer3130:
@tblock79  - do you have reproduction steps for this specifically?  I know you have the screen shots, but it isn't clear what study that is or what is causing the problem as the PR here is identical in results to the legacy method, update to one part in 10**12 anyways.

PR #2381, opened 2025-10-09
fix(interpolation): Fixed contour segmentation interpolation for spline contours.

jbocce:
### Context
For spline contours, only non-simplified Catmull Rom Splines were being interpolated (at all).

### Changes & Results
For simplified splines, fire an annotation modified event when the simplified spline is added so that the interpolation is triggered.
When choosing the viewport for the interpolation, choose a viewport that has the interpolated annotation's referenced image id.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] "OS: Windows 11
- [x] "Node version: 23.9.0
- [x] "Browser: Chrome 141.0.7390.66
  

### DIFF
--- a/packages/tools/src/utilities/contourSegmentation/convertContourSegmentation.ts
+++ b/packages/tools/src/utilities/contourSegmentation/convertContourSegmentation.ts
@@ -6,6 +6,7 @@ import { addAnnotation, removeAnnotation } from '../../stateManagement';
 import type { ContourSegmentationAnnotation } from '../../types';
 import { removeContourSegmentationAnnotation } from './removeContourSegmentationAnnotation';
 import { addContourSegmentationAnnotation } from './addContourSegmentationAnnotation';
+import { triggerAnnotationModified } from '../../stateManagement/annotation/helpers/state';
 
 // Default tool name used for converted contour segmentation annotations
 const DEFAULT_CONTOUR_SEG_TOOL_NAME = 'PlanarFreehandContourSegmentationTool';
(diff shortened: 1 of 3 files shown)


packages/tools/src/utilities/getViewportForAnnotation.ts
```diff
@@ -16,5 +16,15 @@ export default function getViewportForAnnotation(
 ): Types.IStackViewport | Types.IVolumeViewport | undefined {
   const viewports = getViewportsForAnnotation(annotation);
 
-  return viewports.length ? viewports[0] : undefined;
+  if (!viewports?.length) {
+    return undefined;
+  }
+
+  const viewport = viewports.find((viewport) =>
```
jbocce, line 23: Use a viewport that makes sense or else you might not see the interpolation on the viewerport you added the contours to interpolate to.

packages/tools/src/utilities/contourSegmentation/convertContourSegmentation.ts
```diff
@@ -106,6 +107,8 @@ export default function convertContourSegmentationAnnotation(
   addAnnotation(newAnnotation, annotation.metadata.FrameOfReferenceUID);
   addContourSegmentationAnnotation(newAnnotation);
 
+  triggerAnnotationModified(newAnnotation);
```
jbocce, line 110: Once simplified, trigger an event so that the interpolation kicks in for the newly added, simplified contour spline.

packages/tools/src/utilities/getViewportForAnnotation.ts
```diff
@@ -16,5 +16,15 @@ export default function getViewportForAnnotation(
...
+    viewport
+      .getImageIds()
+      .some((imageId) => imageId === annotation.metadata.referencedImageId)
```
sedghi, line 26: what happens in the mpr? i think you should also check the viewplanenormal for the camera to pick the correct viewport

jbocce: Great question. However MPR is working great probably because line 17 is calling `getViewportForAnnotation` which is alraedy filtering by `viewPlaneNormal`, `FrameOfReferenceUID` and `viewUp`. So I think we are good.

PR #2403, opened 2025-10-17
fix(dicomImageLoader): use calibrated pixel spacing (if available) when loading DICOM images via WADO-RS

tblock79:
### Context

Currently, the WADO-RS dicomImageLoader uses only the DICOM tag "PixelSpacing" (0028,0030) for setting the size of the loaded pixels. However, for projection radiography studies (DX, CR), this DICOM tag can be missing. Such DICOM files instead have the tag ImagerPixelSpacing (0018,1164) (which additionally may need to be corrected according to the EstimatedRadiographicMagnificationFactor tag). 

If the PixelSpacing tag is missing, the measurement tools currently show a value in pixels, but the shown unit is "mm". Hence, it shows incorrect values. This discrepancy arises because the annotation tools use the calibratedPixelSpacingMetadataProvider for determining which unit should be shown, while the measurement value is calculated via canvas-to-world conversions, which account only for the pixel spacing set during loading of the image.

Therefore, this PR modifies the WADO-PR loader to utilize the pixel spacing from the calibratedPixelSpacingMetadataProvider when loading the image pixel data, as the corrected pixel spacing is typically already available there (calculated with the function getPixelSpacingInformation when processing the WADO-RS metadata call). If the calibratedPixelSpacingMetadataProvider does not have an entry for the current imageId, it falls back to the PixelSpacing tag  (0028,0030), as before.

### Changes & Results

* Changed the metaDataProvider.ts of the wadors loader and changed the handler for MetadataModules.IMAGE_PLANE to read the prior-calculated pixel spacing from the calibratedPixelSpacingMetadataProvider.
* The measurement tools now show the correct metric values instead of pixel values for DX/CR studies with absent PixelSpacing DICOM tag.

### Checklist

* Load DX and/or CR studies and validate that the correct (non-pixel) values are shown if a measurement tool (e.g., the LengthTool) show the unit "mm". 

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] "OS: Windows 11"
- [x] "Node version:  22.17.0"
- [x] "Browser: Chrome 142.0, Firefox 143.0"
  

### DIFF
--- a/packages/dicomImageLoader/src/shared/ImagePlaneModule.ts
+++ b/packages/dicomImageLoader/src/shared/ImagePlaneModule.ts
@@ -11,6 +11,14 @@ const {
 
 export const PIXEL_INSTANCE = 'PixelInstanceModule';
 
+/**
+ * This is a standardized implementation of the metaData module for ImagePlaneModule.
+ * There were multiple different implementations of this previously, with updates
+ * to one not being applied to the others.  This version consistently has the same
+ * behaviour and uses an Instance object containing standard metadata required to
+ * generate the pixel spacing information, which then gets registered with the
+ * scaling module.
+ */
 export class ImagePlaneModule extends Module<Types.ImagePlaneModuleMetadata> {
   public fromNatural(instance, options?) {
     const imageId = options?.imageId;
--- a/packages/dicomImageLoader/src/shared/Module.ts
+++ b/packages/dicomImageLoader/src/shared/Module.ts
@@ -18,6 +18,22 @@ import {
   sqMetadata,
 } from './metadataHelper';
 
+/**
+ * The module class supports reading DICOM modules from various data sources
+ * in a way that defines a single requirement for the source data, and generates
+ * the same type of result data regardless of input.  This allows writing a single
+ * module provider for the metadata provider system, which helps prevent bugs
+ * when differing implementations load things in different ways.
+ *
+ * Currently there are three loaders:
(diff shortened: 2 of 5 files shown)


packages/dicomImageLoader/src/imageLoader/wadors/metaData/metaDataProvider.ts
```diff
@@ -142,73 +138,6 @@ function metaDataProvider(type, imageId) {
     };
   }
 
-  if (type === MetadataModules.IMAGE_PLANE) {
```
wayfarer3130, line 145: This code has been moved to shared code between different types of instance data.

packages/dicomImageLoader/src/shared/ImagePlaneModule.ts
```diff
@@ -0,0 +1,107 @@
...
+const {
+  toNumber,
+  getPixelSpacingInformation,
+  calibratedPixelSpacingMetadataProvider,
+} = utilities;
+
+export const PIXEL_INSTANCE = 'PixelInstanceModule';
+
+export class ImagePlaneModule extends Module<Types.ImagePlaneModuleMetadata> {
+  public fromInstance(instance) {
```
wayfarer3130: This is the standard implementation from OHIF in the fromInstance.  Making sure everyone goes through the single standard implementation allows for fixing the implementation as required for various types of new implementations.

PR #2404, opened 2025-10-17
fix: Wrong position on rehydrate annotation

wayfarer3130:
### Context

When annotations are rehydrated, they need to be able to find the metadata for the correct frame.  

### Changes & Results

Change the sop uid to image id map from straight sop uid to sop uid + frame number so that different frames retrieve the right data.

### Testing

Load the study in https://github.com/OHIF/Viewers/issues/5431 and load the SR.  Switch to MPR mode.  Click on the annotations in the measurements panel.  The three measurements should appear on different slices, the same slices as from the acquisition rendering.  Needs the equivalent fix on the OHIF side.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  

### DIFF
--- a/packages/adapters/src/adapters/Cornerstone3D/MeasurementReport.ts
+++ b/packages/adapters/src/adapters/Cornerstone3D/MeasurementReport.ts
@@ -309,11 +309,13 @@ export default class MeasurementReport {
         metadata
     }) {
         const { ReferencedSOPSequence } = SCOORDGroup.ContentSequence;
-        const { ReferencedSOPInstanceUID, ReferencedFrameNumber } =
+        const { ReferencedSOPInstanceUID, ReferencedFrameNumber = 1 } =
             ReferencedSOPSequence;
 
         const referencedImageId =
-            sopInstanceUIDToImageIdMap[ReferencedSOPInstanceUID];
+            sopInstanceUIDToImageIdMap[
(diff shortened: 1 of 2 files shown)


packages/tools/src/utilities/planar/filterAnnotationsWithinSlice.ts
wayfarer3130, line 1: This is just code cleanup to use the plane restriction correctly, rather than assuming data handles points.  It does not affect whether the code runs correctly or not.

packages/adapters/src/adapters/Cornerstone3D/MeasurementReport.ts
```diff
@@ -309,11 +309,13 @@ export default class MeasurementReport {
         metadata
     }) {
         const { ReferencedSOPSequence } = SCOORDGroup.ContentSequence;
-        const { ReferencedSOPInstanceUID, ReferencedFrameNumber } =
+        const { ReferencedSOPInstanceUID, ReferencedFrameNumber = 1 } =
             ReferencedSOPSequence;
 
         const referencedImageId =
-            sopInstanceUIDToImageIdMap[ReferencedSOPInstanceUID];
+            sopInstanceUIDToImageIdMap[
```
wayfarer3130, line 316: Just use the right key to get the correct image id.

PR #2406, opened 2025-10-20
feat(tools): Add configurable center point to CrosshairsTool

olhalit:
### Context

This feature adds a configurable center point to the CrosshairsTool, allowing users to display a dot at the intersection of the crosshair lines for improved precision. This enhancement provides better visual feedback for locating the exact center point, which is especially useful in detailed analysis. There is no associated issue number, as this is a new feature enhancement.

### Changes & Results

Changes:

   - `CrosshairsTool.ts`:
       - Added two new configuration options: centerPointColor (defaults to 'rgba(255, 255, 
         0, 0.5)') and centerPointSize (defaults to 0).
       - Added logic to the renderAnnotation method to draw a circle at the crosshair's 
         center if centerPointSize is greater than zero.
       - The size of the rendered point is capped to a maximum value to ensure it doesn't 
         become obstructive.
   - `packages/tools/examples/crossHairs/index.ts`:
       - Added a new dropdown control to the toolbar in the example.
       - This dropdown allows for dynamic adjustment of the centerPointSize, demonstrating 
         the new feature's functionality.

  Results:

   - Before: The CrosshairsTool only displayed the intersecting reference lines.
   - After: A configurable colored dot can now be displayed at the center of the crosshairs.
      Its size can be adjusted interactively in the updated example, and it is hidden by 
     default (centerPointSize: 0).

### Testing

   1. Run the project locally using `yarn start`.
   2. Navigate to the **Crosshairs** example in your browser.
   3. You should see a new dropdown in the toolbar labeled **"Center Point Size"**.
   4. Select a value greater than 0 from the dropdown (e.g., 3, 4, or 5).
   5. **Expected Result:** A yellow dot should appear at the intersection of the crosshair lines.
   6. Change the value in the dropdown.
   7. **Expected Result:** The size of the dot should update accordingly.
   8. Select 0 from the dropdown.
   9. **Expected Result:** The dot should disappear.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] "OS: Ubuntu 24.04.2"
- [x] "Node version: 20.17.0"
- [x] "Browser: Chrome 139.0.7258.66"
  

### DIFF
--- a/packages/tools/examples/crossHairs/index.ts
+++ b/packages/tools/examples/crossHairs/index.ts
@@ -255,13 +255,8 @@ addDropdownToToolbar({
       centerPoint: newCenterPointConfig,
     };
 
-    toolGroup.viewportsInfo.forEach(({ viewportId, renderingEngineId }) => {
-      const renderingEngine = getRenderingEngine(renderingEngineId);
-      const viewport = renderingEngine.getViewport(
-        viewportId
-      ) as Types.IVolumeViewport;
-      viewport.render();
-    });
+    const renderingEngine = getRenderingEngine(renderingEngineId);
+    renderingEngine.render();
   },
 });
 


packages/tools/src/tools/CrosshairsTool.ts
```diff
@@ -158,6 +158,8 @@ class CrosshairsTool extends AnnotationTool {
         filterActorUIDsToSetSlabThickness: [],
         // blend mode for slabThickness modifications
         slabThicknessBlendMode: Enums.BlendModes.MAXIMUM_INTENSITY_BLEND,
+        centerPointColor: 'rgba(255, 255, 0, 0.5)',
```
sedghi: make it one object please, centerPoint: {enabled: false, color: }

olhalit: @sedghi Updated with `centerPoint` object

packages/tools/examples/crossHairs/index.ts
```diff
@@ -230,6 +230,41 @@ addToggleButtonToToolbar({
...
+      enabled: +selectedValue > 0,
+      size: +selectedValue,
+    };
+
+    crosshairsInstance.configuration = {
+      ...oldConfiguration,
+      centerPoint: newCenterPointConfig,
+    };
+
+    toolGroup.viewportsInfo.forEach(({ viewportId, renderingEngineId }) => {
```
sedghi: instead of this you can do renderingEngine.render()

olhalit: @sedghi Updated with `renderingEngine.render()`

PR #2408, opened 2025-10-20
fix(Annotation Tools): Fix active volume to show correct cachedStats (especially on fusion viewport)

Celian-abd:
### Context
Currently, even if we set the PT volumeId into the configuration of Rectangle/Circle/EllipticalRoiTools, it will always show the cachedStats of the CT on the fusion viewport.

As we can see here : 

Capture d’écran 2025-10-20 164430

### Changes & Results
- Added a way to check if a volumeId is provided in the tool configuration
- Select the correct cachedStats to show following the volumeId

If we set PT volumeId in the tool configuration : 
Capture d’écran 2025-10-20 164845

### Testing
**Example PET-CT**
You can modify or delete the prefered  volumeId in the tool configuration : 
`fusionToolGroup.addTool(RectangleROITool.toolName, { volumeId: ptVolumeId });`

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  


packages/tools/src/tools/annotation/CircleROITool.ts
```diff
@@ -625,6 +623,22 @@ class CircleROITool extends AnnotationTool {
 
       styleSpecifier.annotationUID = annotationUID;
 
+      let targetId: string;
```
sedghi: do these in the this.getTargetId so that you don't need to repeat yourself, i think that was a parent method

Issue #2387, opened 2025-10-14
[Bug] Incorrect measurements on mammo image with ERMF


fbuonomo:
### Describe the Bug

On the attached mammographic projections, all the measurement tools give clearly incorrect values, as shown in the attached screenshot.

### Steps to Reproduce

Enable a measurement tool
Perform the measurement

### The current behavior

The values ​​of the displayed measurement differ greatly from the actual value

### The expected behavior

The measurements displayed must correspond to the actual size

### System Information

node 22.14
windows 11
chrome, firefox, edge

<img width="335" height="938" alt="Image" src="https://github.com/user-attachments/assets/22de934e-8d54-4154-8d66-2205f416d0df" />

fbuonomo:
[mammo.zip](https://github.com/user-attachments/files/22901859/mammo.zip)

fbuonomo:
In fact, in the image I attached (mammo..zip), the 00280030 tag (pixel spacing) is missing, but other viewers (like RadiAnt) display the value correctly.
If I manually add the tag in question, the measurement is displayed correctly.

fbuonomo:
According to what I read in this correction attachment, it seems to me that the measure should be treated even in the absence of the tag 00280030

[cp586_lb.pdf](https://github.com/user-attachments/files/22906255/cp586_lb.pdf)

fbuonomo:
I don't know if it's the right solution, but I solved it by modifying the management of the "imagePlane" module in metaDataProvider.ts like this:

  if (type === MetadataModules.IMAGE_PLANE) {
    //metaData = fixNMMetadata(metaData);
    let imageOrientationPatient = extractOrientationFromMetadata(metaData);
    let imagePositionPatient = extractPositionFromMetadata(metaData);
    const pixelSpacing = getNumberValues(metaData['00280030'], 2);
    const imagerPixelSpacing = getNumberValues(metaData['00181164'], 2);

    let columnPixelSpacing = null;
    let rowPixelSpacing = null;
    let rowCosines = null;
    let columnCosines = null;

    let usingDefaultValues = false;
    if (pixelSpacing) {
      rowPixelSpacing = pixelSpacing[0];
      columnPixelSpacing = pixelSpacing[1];
    } else if (imagerPixelSpacing) {
      rowPixelSpacing = imagerPixelSpacing[0];
      columnPixelSpacing = imagerPixelSpacing[1];
    } else {
      usingDefaultValues = true;
[...]

Issue #1957, opened 2025-04-02
[Bug] Wrong measurement value on Ultrasound images


idemopacs:
### Describe the Bug

We are facing issue related to wrong measurement value on some Ultrasound images. 

### Steps to Reproduce

@sedghi , I have mailed you the link of US DICOM image.
https://drive.google.com/file/d/1zupXRLYVPryIDjjC-fVcVEIE1dNp-l3T/view?usp=sharing

### The current behavior

Please see screenshot of issue.

![Image](https://github.com/user-attachments/assets/a396406b-e162-428a-a334-885b16fc029c)

### The expected behavior

Correct measurements are seen bottom right of the attached image.

### OS

Windows 10

### Node version

16.14.0

### Browser

Chrome v134.0

sedghi:
I confirm this is a bug, not sure what we are doing wrong, the file you might want to look at is packages/tools/src/utilities/getCalibratedUnits.ts inside Cornerstone3D repo

idemopacs:
Did you get a chance to look into this issue?

Thanks

karthik-durai:
Hi @idemopacs,

I’m new to this repository and interested in contributing. Could you please share the image files with me? I’d like to take a look.

idemopacs:
Yes, I can. Can you please give me your email id.

karthik-durai:
I have requested access on drive with the issue title and id.

arul-trenser:
@sedghi  Further analysis revealed that the length calculation for ultrasound (US) data exhibits inconsistent accuracy. Specifically, US data with multiple `SequenceOfUltrasoundRegions` produces a correct measurement value, while US data with a single `SequenceOfUltrasoundRegions` results in an incorrect measurement value.

Root Cause:
The core of the issue lies in the scale value, which is derived from `getCalibratedUnits` (as referenced [here](https://github.com/cornerstonejs/cornerstone3D/blob/main/packages/tools/src/utilities/getCalibratedUnits.ts#L145)). This scale value does not function as an intended unit conversion or scaling factor. Instead, for US data, it erroneously returns the `pixelSpacing` value (specifically based on `PhysicalDeltaX`) when calculating the actual physical distance.

The length is computed using the following formula:

```
const length = this._calculateLength(worldPos1, worldPos2) / scale;
```

[...]

Issue #1320, opened 2024-06-13
Calibration Issue: User-Defined Calibration Not Affecting Measurements in Cornerstone Viewport


dash-vivekpatidar:
### What feature or change would you like to see made?

Hi Support Team,

I am experiencing an issue with applying user-defined calibration in my Cornerstone viewport. I have tried using both utilities.calibratedPixelSpacingMetadataProvider.add() and utilities.calibrateImageSpacing() to set the calibration, but neither approach seems to affect the measurements as expected.

Here is the relevant part of my code:
const calibration = {
scale: 0.5,
type: cornerstone.Enums.CalibrationTypes.USER,
};

const renderingEngine = getRenderingEngine(this.renderingEngineId);
const viewport = <Types.IStackViewport>renderingEngine.getViewport(this.viewportId);

utilities.calibratedPixelSpacingMetadataProvider.add(url[0], calibration);
viewport.calibrateSpacing(url[0]);

// Also tried
utilities.calibrateImageSpacing(url[0], renderingEngine, calibration);

Expected Behavior: Measurements should be scaled according to the user-defined calibration.

Actual Behavior: Measurements remain unchanged after applying the calibration.

Below Cornerstone Version I am using
“@cornerstonejs/core”: “^1.77.0”,
“@cornerstonejs/streaming-image-volume-loader”: “^1.77.0”,
“@cornerstonejs/tools”: “^1.77.0”,

### Why should we prioritize this feature?

If the user selected the calibration tool and entered the calibration line, all annotations would be changed and work further, just like in OHIF.

sedghi:
can you reproduce the issue in our calibaration example maybe?https://www.cornerstonejs.org/live-examples/calibrationtools

sedghi:
Can you check the latest OHIF viewer, which now includes Cornerstone 3D 2.0?

Here are the migration guides:

Try OHIF locally: [https://viewer-dev.ohif.org/localbasic](https://viewer-dev.ohif.org/localbasic)<br>Try Cornerstone3D 2.0 locally: [https://www.cornerstonejs.org/live-examples/local.html](https://www.cornerstonejs.org/live-examples/local.html)

[https://docs.ohif.org/migration-guide/3p8-to-3p9/](https://docs.ohif.org/migration-guide/3p8-to-3p9/)

tokyobordel:
Have same issue on 2.12.0. My code:
```
const calibration = {
  scale: 2.0,
[...]

PR #2415, opened 2025-10-21
fix: US calibrated regions broken

wayfarer3130:
### Context

When US images have a base calibration as well as calibrated regions, then the spacing calibration for the US regions isn't being set correctly because it isn't taking into account the base calibration values.
This also exposed some issues with ERMF interactions with the calibration code, so updated the ERMF calculations as well.

Fixes:
https://github.com/OHIF/Viewers/issues/5216
https://github.com/cornerstonejs/cornerstone3D/issues/1320
https://github.com/cornerstonejs/cornerstone3D/issues/1957
https://github.com/cornerstonejs/cornerstone3D/issues/2387

### Changes & Results

* Use index coordinates for tools scaling calculations in order to perform the scaling calculations consistently according to the standard.
* Pulled out the ERMF calculation and added several variants of it
* Correctly distinguish between has pixel spacing by using both scale and the usingDefaultValues in StackViewport
* Modified US region units to detect valid length iff x and y units are identical - that is, lengths are only valid if the two measurement units for rows and columns are the same.  It is in theory possible to then detect "secs" as a length unit for this, although US probably won't be encoded that way.

### Testing

CRs in: 1.2.276.0.7230010.3.1.2.2155604110.4180.1021041295.21
Image without spacing:  Must show px, must be able to be calibrated.  When calibrated must show "User"

Image with pixel spacing, shows in "mm unknown" because it is unknown what type of calibration this is

1.3.6.1.4.1.14519.5.2.1.99.1071.55651399101931177647030363790032

Image has pixel spacing, but no imager type spacing, must show mm units Projection

1.3.46.670589.30.1.3.1.1625260923.1632320482484.1
has imager and pixel spacing, with different values, shows in ERMF

1.3.6.1.4.1.14519.5.2.1.1.84416332615988066829602832830236187384
has distance source patient/detector, shows in ERMF

If possible, find a Fiducial calibrated DX/CR to show "Calibrated" without adding a calibration.

1.3.6.1.4.1.14519.5.2.1.7085.2626.494695569589117268722281491772
Must show US Region in cm units within US region
Must show mm units outside US region
Calibration must not affect US region values, but must show mm User outside calibrated region.

1.2.840.113619.2.98.1234.1008079976.0.40
Must show US region in cm units within US region
Must show px units outside US region
Calibration must not affect US region values, but must show mm User outside calibrated region.

Added some unit tests, plan on doing one more pass of unit tests before completion to test all the changes except the px/non-px change in StackViewport which I couldn't figure out how to reasonably test)

Test with the image from https://github.com/cornerstonejs/cornerstone3D/issues/2387.
Note there is some discrepency over whether to show PROJ or ERMF - there is an extra distance tag that I'm not sure if it should be used 'Distance Source to Entrance' and then the ERMF value is 1, which is incorrect - only ERMF values larger than 1 are value for indicating ERMF has been applied.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  

### DIFF
--- a/packages/tools/examples/localAdvanced/index.ts
+++ b/packages/tools/examples/localAdvanced/index.ts
@@ -52,9 +52,10 @@ let viewport;
 const toolGroupId = 'myToolGroup';
 
 function onUpload(files) {
-  const file = files[0];
-  const imageId = dicomImageLoader.wadouri.fileManager.add(file);
-  loadAndViewImages(imageId);
+  imageIds = [...files].map((file) =>
+    dicomImageLoader.wadouri.fileManager.add(file)
+  );
+  loadAndViewImages(imageIds);
 }
 
 addUploadToToolbar({ title: 'Upload', onChange: onUpload });


packages/tools/src/utilities/getCalibratedUnits.ts
wayfarer3130, line 1: I made changes here assuming that index coordinates are being used.  That loses fractional spacing right now, and would need to be applied to the other tools, but the problem with using world coordinates is that they don't necessarily correspond to other coordinates in a computable fashion without using the full volume transform.  I don't really see a better option if we aren't going to leave bugs in this area.

wayfarer3130:
> can you please check these if they are fixed
> 
> if so please mention them in the PR description
> 
> [OHI-1962](https://linear.app/ohif/issue/OHI-1962)

This item was already fixed, is just waiting confirmation about being fixed.
Note that volume viewports do NOT apply calibration values.

> 
> [OHI-2147](https://linear.app/ohif/issue/OHI-2147)

Updated in this issue.

> 
> [OHI-2052](https://linear.app/ohif/issue/OHI-2052)
Fixed in this issue

> 
> [OHI-1989](https://linear.app/ohif/issue/OHI-1989)
Asked for more details as I'm unable to reproduce this anywhere except viewer.ohif.org

> 
> [OHI-1100](https://linear.app/ohif/issue/OHI-1100)
No idea - you asked for steps and it isn't clear whether or not it is still an issue.

> 
> [OHI-1797](https://linear.app/ohif/issue/OHI-1797)
Seems to be resolved.  Note this shows "Unknown" as the spacing type because the ERMF information isn't present - probably instead of "Unknown" we should show "Imager" to indicate it is imaging spacing.
> 
> [OHI-1778](https://linear.app/ohif/issue/OHI-1778)
Fixed in this issue

> 
> [OHI-2013](https://linear.app/ohif/issue/OHI-2013)
Updated in this issue

> 
> [OHI-2265](https://linear.app/ohif/issue/OHI-2265)


Fixed as far as I can tell

sedghi:
git pull main since ci was broken a while back

packages/tools/src/tools/annotation/LengthTool.ts
```diff
@@ -840,27 +828,30 @@ class LengthTool extends AnnotationTool {
...
-      // TODO -> Do we instead want to clip to the bounds of the volume and only include that portion?
-      // Seems like a lot of work for an unrealistic case. At the moment bail out of stat calculation if either
-      // corner is off the canvas.
+      this.isHandleOutsideImage = !LengthTool.isInsideVolume(
+        dimensions,
+        handles
+      );
 
-      // todo: add insideVolume calculation, for removing tool if outside
+      const namedLength = {
```
sedghi: where are we typing (ts) this?

packages/tools/src/tools/annotation/CircleROITool.ts
```diff
@@ -893,22 +891,58 @@ class CircleROITool extends AnnotationTool {
...
+      const { unit, areaUnit } = calibrate;
+
+      // Generates a basic cached stats.
+      const namedArea = {
+        name: 'area',
+        label: 'Area',
+        value: area,
+        unit: areaUnit,
+      };
+      const namedCircumference = {
```
sedghi: requires a enum types, that only allows certain sets

packages/tools/src/tools/base/BaseTool.ts
```diff
@@ -318,6 +318,42 @@ abstract class BaseTool {
   public static endGroupRecording() {
     DefaultHistoryMemo.endGroupRecording();
   }
+
+  /**
+   * Calculates the length between two index coordinates using the calibrate
+   * information for scaling information.
+   */
+  public static calculateLength(calibrate, indexPoints, closed = false) {
```
sedghi: i would suggest namign this calculateLengthInIndex or somethign 
also what is closed please document

wayfarer3130:
For testing this issue:
CRs in: 1.2.276.0.7230010.3.1.2.2155604110.4180.1021041295.21
one has no spacing, shows in pixels
one has pixel spacing and nothing else, shows in mm unknown

1.3.6.1.4.1.14519.5.2.1.99.1071.55651399101931177647030363790032
has pixel spacing, but no imager type spacing, shows in "Projection"

1.3.46.670589.30.1.3.1.1625260923.1632320482484.1
has imager and pixel spacing, with different values, shows in ERMF

1.3.6.1.4.1.14519.5.2.1.1.84416332615988066829602832830236187384
has distance source patient/detector, shows in ERMF

Do not have a sample showing in calibrated already or in ERMF with just pixel values

packages/tools/src/enums/MeasurementDimension.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * Identifies the dimensionality of a measurement in terms of what
+ * type of dimensions are being measured by the value.
+ */
+export enum MeasurementDimension {
```
sedghi: why not MeasurementType?

packages/tools/src/types/CalculatorTypes.ts
```diff
@@ -1,15 +1,17 @@
...
 
-type Statistics = {
+export type Statistics = {
   name: string;
   label?: string;
   value: number | number[];
   unit: null | string;
   pointIJK?: Types.Point3;
   pointLPS?: Types.Point3;
+  type?: MeasurementDimension;
```
sedghi: here we are also calling it type, so i prefer type: MeasurementType

packages/tools/src/tools/annotation/LengthTool.ts
```diff
@@ -840,27 +829,29 @@ class LengthTool extends AnnotationTool {
...
+      );
+      const calibrate = getCalibratedLengthUnitsAndScale(image, handles);
+      const { unit } = calibrate;
 
-      if (this._isInsideVolume(index1, index2, dimensions)) {
-        this.isHandleOutsideImage = false;
-      } else {
-        this.isHandleOutsideImage = true;
-      }
+      const length = LengthTool.calculateLengthInIndex(calibrate, handles);
```
sedghi, line 838: So, I thought we were only changing our calculations to use the index for calibration. Does this mean we're moving all our calculations to index space? does worldToIndex even support floating indices?

packages/tools/src/tools/base/BaseTool.ts
```diff
@@ -318,6 +318,42 @@ abstract class BaseTool {
   public static endGroupRecording() {
     DefaultHistoryMemo.endGroupRecording();
   }
+
+  /**
+   * Calculates the length between two index coordinates using the calibrate
+   * information for scaling information.
+   */
+  public static calculateLengthInIndex(calibrate, indexPoints, closed = false) {
```
sedghi, line 328: please add js doc above it

packages/tools/src/tools/annotation/LengthTool.ts
wayfarer3130: worldToIndex does support floating indices when called against the image object - the utility version rounds.
It is easier to just do all the calculations in index space as it is consistently what is calibrated for all spacing types.
The mm spacing is only calibrated for some images correctly - others use it differently.

packages/tools/src/tools/annotation/BidirectionalTool.ts
```diff
@@ -1322,41 +1314,56 @@ class BidirectionalTool extends AnnotationTool {
       }
 
       const { imageData, dimensions } = image;
-      const index1 = transformWorldToIndex(imageData, worldPos1);
-      const index2 = transformWorldToIndex(imageData, worldPos2);
-      const index3 = transformWorldToIndex(imageData, worldPos3);
-      const index4 = transformWorldToIndex(imageData, worldPos4);
+      const handles = data.handles.points.map((point) =>
```
wayfarer3130, line 1317: The worldToIndex on imageData supports floating indexes.

packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -666,8 +666,11 @@ class StackViewport extends Viewport {
   private calibrateIfNecessary(imageId, imagePlaneModule) {
     const calibration = metaData.get('calibratedPixelSpacing', imageId);
     const isUpdated = this.calibration !== calibration;
-    const { scale } = calibration || {};
-    this.hasPixelSpacing = scale > 0 || imagePlaneModule.rowPixelSpacing > 0;
+    const scale = calibration?.scale;
+    this.hasPixelSpacing =
+      scale > 0 ||
+      (!imagePlaneModule.usingDefaultValues &&
```
IbrahimCSAE, line 665: https://github.com/OHIF/Viewers/blob/afa9131da1d6fccaf5f72c66dcf3d83b2b0e5bfa/platform/core/src/classes/MetadataProvider.ts#L549

Are we still sure we want to trust `usingDefaultValues` to drive this? It seems to be set to true if any of several conditions pass here in the OHIF metadata provider.

If we will, can we make sure if that `usingDefaultValues` is `true`, then there's geniunely no pixel spacing, and it wasn't set to true due to some other missing field.

packages/core/src/utilities/getPixelSpacingInformation.ts
```diff
@@ -96,39 +136,11 @@ function calculateRadiographicPixelSpacing(instance) {
   };
 }
 
-function calculateUSPixelSpacing(instance) {
-  const { SequenceOfUltrasoundRegions } = instance;
-  const isArrayOfSequences = Array.isArray(SequenceOfUltrasoundRegions);
-
-  if (isArrayOfSequences && SequenceOfUltrasoundRegions.length > 1) {
-    console.warn(
```
IbrahimCSAE, line 111: Im assuming since you removed this its no longer an issue? if so thats amazing. (multi US region support)

wayfarer3130: Multiple regions does work correctly, but neither a single region nor multiple regions will spill over to the entire image area any longer if there isnt a base pixel spacing.
@sedghi - FYI: Previously if there was 1 region, it would act as a calibration for the entire image - EVEN if that region was in seconds or the like it would get applied as *10 mm   The regions now ONLY apply if all handles are in the region.
The PixelSpacing does now apply to the entire image as a fallback correctly, so I believe that is a better way to determine the fallback spacing.

packages/core/src/RenderingEngine/StackViewport.ts
wayfarer3130: The other changes in this PR, specifically the unit tests on getPixelSpacingInformation are fairly comprehensive, and will determine if the usingDefaultValues is set or not.  We were sometimes getting usingDefaultValues false but having a scale value and that was causing us to show in mm when in fact it was pixels.  That is a much worse case than showing in px when we could show mm.  The next time we see such a case, we can add a unit test to getPixelSpacingInformation to make sure that we actually return pixel spacing that is valid for that case.  The usingDefaultValues is intended to be the real flag that determines whether to show mm or not, excluding US and ECG.

wayfarer3130:
@sedghi - I added an advanced local test.  I haven't yet run the indivdiual tests in it to see that everything works locally properly.

PR #2418, opened 2025-10-22
fix(Annotation Tools): Fix active volume to show correct cachedStats (especially on fusion viewport)

Celian-abd:
### Context
(Remake of https://github.com/cornerstonejs/cornerstone3D/pull/2408)
Currently, even if we set the PT volumeId into the configuration of Rectangle/Circle/EllipticalRoiTools, it will always show the cachedStats of the CT on the fusion viewport.

As we can see here : 

Capture d’écran 2025-10-20 164430

### Changes & Results
- Added a way to check if a volumeId is provided in the tool configuration
- Select the correct cachedStats to show following the volumeId

If we set PT volumeId in the tool configuration : 
Capture d’écran 2025-10-20 164845

### Testing
**Example PET-CT**
You can modify or delete the prefered  volumeId in the tool configuration : 
`fusionToolGroup.addTool(RectangleROITool.toolName, { volumeId: ptVolumeId });`

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  

### DIFF
--- a/packages/tools/src/tools/base/BaseTool.ts
+++ b/packages/tools/src/tools/base/BaseTool.ts
@@ -115,12 +115,21 @@ abstract class BaseTool {
    * A function generator to test if the target id is the desired one.
    * Used for deciding which set of cached stats is appropriate to display
    * for a given viewport.
+   *
+   * This relies on the fact that the target id contains a substring which is the
+   * desired volume id when the target is a volume.
+   * It is also possible to use series query parameters such as `/series/{seriesUID}/`
+   * to generate specific series selections within a stack viewport.
    */
   public static isSpecifiedTargetId(desiredVolumeId: string) {
     // imageId including the target id is a proxy for testing if the
     // image id is a member of that volume.  This may need to be fixed in the
     // future to add more criteria.
-    return (_viewport, { imageId }) => imageId.includes(desiredVolumeId);
+    return (_viewport, { targetId }) => {
+      // target ids contain the base information for the volume, so allow specifying
+      // preference by desiredVolumeId
+      return targetId.includes(desiredVolumeId);
+    };
   }
 
   /**
@@ -292,9 +301,9 @@ abstract class BaseTool {
 
     // Check if cachedStats is available and contains the preferredVolumeId
     if (isPreferredTargetId && data?.cachedStats) {
-      for (const [imageId, cachedStat] of Object.entries(data.cachedStats)) {
-        if (isPreferredTargetId(viewport, { imageId, cachedStat })) {
-          return imageId;
+      for (const [targetId, cachedStat] of Object.entries(data.cachedStats)) {
+        if (isPreferredTargetId(viewport, { targetId, cachedStat })) {
+          return targetId;
         }
       }
     }


wayfarer3130:
I'm not at all keen on having more volume id specific tools - that means you have to know exactly what volumes to use when you create creating the tool group, and when you make any changes to the volumes available, they have to apply to the tool group.
I'd much rather see some sort of filter allowing a TYPE of volume to be applied based on some sort of criteria.  For example, in this case just having a filter of modality = PT would work.  
If it is necessary to identify the specific volumes, it should be disconnected from the tool group, and instead isolate that knowledge in a separate class.  For example, it would be possible to provide a Set class with volume ids in it, and that class could provide notifications when the volume ids in it change so that as a new volume got loaded, it could be added to the right volume set class.

sedghi:
I like the idea, but asking people to do this would be a breaking change since i think we already say use volumeId. I suggest we implement it in the next version.

salimkanoun:
Hi there I agree with both of you, the current implementation is too complex.
However using modality is probably not enough neither as I saw fused MRI on which you fuse ADC and T1/T2 series so both will be MR modality.

I would suggest to use image index in the viewport, as you can control it in the hanging protocol I think it might be possible to specify which image index is used by the tool group.

packages/tools/src/tools/base/BaseTool.ts
```diff
@@ -76,6 +90,15 @@ abstract class BaseTool {
     return utilities.deepMerge(defaultProps, additionalProps);
   }
 
+  /**
+   * A function generator to test if the target id is the desired one.
+   * Used for deciding which set of cached stats is appropriate to display
+   * for a given viewport.
+   */
+  public static isSpecifiedTargetId(desiredTargetId: string) {
```
wayfarer3130: @Celian-abd  - my preference is to provide a function that determines the preferred target id rather than hard coding the value.  That way it is possible to have more flexible determination and to work with multiple merged viewports at a single time by creating a custom function that works against both viewports.

packages/tools/src/tools/base/BaseTool.ts
```diff
@@ -76,6 +90,15 @@ abstract class BaseTool {
...
+    return (_viewport, { targetId }) => targetId.includes(desiredTargetId);
```
wayfarer3130: @sedghi - this code is wrong as the target id doesn't have to include the desired target id as a substring in general.  The right solution is to get the volume id associated with the given image id, and check for equality.  HOwever, without passing in the set of image ids associated with the volume, I don't see an easy way of doing that.  Should I just pass in a Set of the image ids that are preferred and use   imageIds.has(imageId)?

wayfarer3130:
@sedghi - I've updated the PR with changes that I think should work with the API generally.  It isn't perfect, but I think is acceptable as the API should be fairly future compatible.
@Celian-abd - can you look at the change I made and add your agreement that they resolve your issue?

Celian-abd:
@wayfarer3130 Looks good to me, thanks a lot !

sedghi:
@wayfarer3130 can you fix the volumeId and imageid thing i sent in slack and we merge this

wayfarer3130:
> @wayfarer3130 can you fix the volumeId and imageid thing i sent in slack and we merge this

@sedghi - it really wasn't clear what you wanted changed from the comment in slack

salimkanoun:
Small question does start / stop tools needing update ? I see conventional tools in this PR has a change in targetId implementation

packages/tools/src/tools/base/BaseTool.ts
```diff
@@ -76,6 +90,18 @@ abstract class BaseTool {
...
+  /**
+   * A function generator to test if the target id is the desired one.
+   * Used for deciding which set of cached stats is appropriate to display
+   * for a given viewport.
+   */
+  public static isSpecifiedTargetId(desiredVolumeId: string) {
+    // imageId including the target id is a proxy for testing if the
+    // image id is a member of that volume.  This may need to be fixed in the
+    // future to add more criteria.
+    return (_viewport, { imageId }) => imageId.includes(desiredVolumeId);
```
sedghi: @wayfarer3130 this is the place i say requires change, basically imageId includes desiredVolumeId does not makes sense.

jbocce:
I believe the following two issues might be solved by this PR...

- https://github.com/OHIF/Viewers/issues/5047
- https://github.com/OHIF/Viewers/issues/4913

@wayfarer3130  please check this. Thanks.

wayfarer3130:
@sedghi - the test in this relies on target id, not on image ids.  I've updated the PR accordingly to name things targetId.  Those will always have the volume id in them for volume targets.

packages/tools/examples/petCt/index.ts
```diff
@@ -390,7 +390,9 @@ function setUpToolGroups() {
     // Only set CT volume to MIP in the fusion viewport
     filterActorUIDsToSetSlabThickness: [ctVolumeId],
   });
-  fusionToolGroup.addTool(RectangleROITool.toolName);
+  fusionToolGroup.addTool(RectangleROITool.toolName, {
+    isPreferredTargetId: RectangleROITool.isSpecifiedTargetId(ptVolumeId),
```
sedghi, line 394: this is confusing api, what are we achieving here?

isPreferredTargetId: RectangleROITool.isSpecifiedTargetId(ptVolumeId)

wayfarer3130: It is saying that the preferred target for statistics is one belonging to the specified volume id.  If you have better naming on that, please let me know.

PR #2422, opened 2025-10-23
bump(codec): update version

IbrahimCSAE:
#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -131,7 +131,7 @@
     },
     "packages/adapters": {
       "name": "@cornerstonejs/adapters",
-      "version": "4.2.0",
+      "version": "4.6.1",
       "dependencies": {
         "@babel/runtime-corejs2": "7.26.10",
         "buffer": "6.0.3",
@@ -146,7 +146,7 @@
     },
     "packages/ai": {
       "name": "@cornerstonejs/ai",
-      "version": "4.2.0",
+      "version": "4.6.1",
       "dependencies": {
         "@babel/runtime-corejs2": "7.26.10",
         "buffer": "6.0.3",
@@ -164,7 +164,7 @@
     },
     "packages/core": {
       "name": "@cornerstonejs/core",
-      "version": "4.2.0",
+      "version": "4.6.1",
       "dependencies": {
         "@kitware/vtk.js": "32.12.1",
         "comlink": "4.4.2",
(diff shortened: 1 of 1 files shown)


sedghi (approved):
Check ci passes

bun.lock
```diff
@@ -4904,6 +4904,8 @@
 
     "detective-typescript/typescript": ["typescript@5.7.2", "", { "bin": { "tsc": "bin/tsc", "tsserver": "bin/tsserver" } }, "sha512-i5t66RHxDvVN40HfDd1PsEThGNnlMCMT3jMUuoh9/0TaqWevNontacunWyN02LA9/fIbEWlcHZcgTKb9QoaLfg=="],
 
+    "dicom-microscopy-viewer/@cornerstonejs/codec-openjpeg": ["@cornerstonejs/codec-openjpeg@1.2.4", "", {}, "sha512-UT2su6xZZnCPSuWf2ldzKa/2+guQ7BGgfBSKqxanggwJHh48gZqIAzekmsLyJHMMK5YDK+ti+fzvVJhBS3Xi/g=="],
```
jbocce, line 4907: It appears that the dicom-microscopy viewer is still using codec-openjpeg version 1.2.4. Are we ok with this? I just wanted to point this out now that I am a super lock file reading machine... 🤣Fundamentally there is no problem, but I suspect minimizing the dependencies is always a good thing.

yarn.lock
```diff
@@ -1507,7 +1507,12 @@
   resolved "https://registry.yarnpkg.com/@cornerstonejs/codec-libjpeg-turbo-8bit/-/codec-libjpeg-turbo-8bit-1.2.2.tgz#ae384b149d6655e3dd6e18b9891fab479ab5e144"
   integrity sha512-aAUMK2958YNpOb/7G6e2/aG7hExTiFTASlMt/v90XA0pRHdWiNg5ny4S5SAju0FbIw4zcMnR0qfY+yW3VG2ivg==
 
-"@cornerstonejs/codec-openjpeg@1.2.4", "@cornerstonejs/codec-openjpeg@^1.2.4":
+"@cornerstonejs/codec-openjpeg@1.3.0":
```
jbocce, line 1510: It is a bit more clear here in the yarn.lock file that we now have two dependencies on codec-open.jpeg.

bun.lock
jbocce: I suppose we could avoid this with a resolution in the package.json file. I love the word resolution because it sounds like revolution! 🤣

"you say you want a resolution, well you know, we all want to minimize dependencies..." If only John and Paul would have thought of this.

sedghi: I think we should update at the source of that lib. @igoroctaviano

PR #2426, opened 2025-10-24
fix: restrict labelmap to surface conversion

Devu-trenser:
### Context
Fixes issue: 

- https://github.com/OHIF/Viewers/issues/5427
- https://github.com/OHIF/Viewers/issues/5346

This merge request refactors the segmentation representation computation and event subscription system.

### Changes & Results
This merge request refactors the segmentation representation computation and event subscription system. The goal of this refactor is to:

- Expose event subscription and removal APIs to OHIF.
- Centralize listener management for better cleanup and memory safety.
- Introduced a dedicated segmentationEventManager module to manage segmentation listeners.
- Implemented a nested Map structure to track listeners per segmentation and representation type
- Updated computeAndAddRepresentation to use centralized event management instead of directly adding/removing listeners.
- Added utility functions to:
     - Remove a single representation listener (removeRepresentationListener).
     - Remove all listeners for a segmentation (removeAllSegmentationListeners).

### Testing

- Locally link the OHIF `Devu-trenser:fix/restrict-labelmap-to-surface-conversion` branch with the branch (Devu-trenser:fix/labelmap-to-surface-conversion)
- Launch a data in viewer and switch to a volume (MPR) viewport.
- Draw a segmentation using the brush tool.
- Switch the layout to a 3D volume viewport.
- Return to the stack/MPR viewport and continue drawing segments.
- Verify Labelmap segmentations should not be converted to Surface on each draw if there is no enabled 3d volume viewport.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  


jbocce:
I tested this PR elaborating on the instructions provided (thank you for such detail). 

In summary I locally linked the OHIF master branch with the branch (Devu-trenser:fix/labelmap-to-surface-conversion) with the fix and I found that labelmaps failed to convert to surfaces when switching from MPR to 3D four up. That is to say the third step in the instructions failed.

Here are the specific steps I was using to test this PR...
1. Checkout the branch Devu-trenser:fix/labelmap-to-surface-conversion
2. Set up my sandbox for local linking following [these instructions](https://docs.ohif.org/development/link/)
3. I ran OHIF.
4. Navigated to http://localhost:3000/segmentation?StudyInstanceUIDs=1.3.6.1.4.1.25403.345050719074.3824.20170125095438.5
5. Launched MPR with the series Body 3.0 CE
6. Add a segmentation and used the brush tool to segment the coronal view. I added just a single stroke using the brush tool
7. Launched 3D Four Up
8. Return to MPR.
9. Draw more with the brush tool and ensure each draw is NOT converted to a surface.

Unfortunately the testing failed on step 7 with the following exception(s) and failure to create a surface representation...

```
surfaceComputationStrategies.js:29 Error: Failed to convert labelmap to surface
    at computeSurfaceFromLabelmapSegmentation (surfaceComputationStrategies.js:54:1)
    at async Module.computeSurfaceData (surfaceComputationStrategies.js:22:1)
    at async computeAndAddRepresentation (computeAndAddRepresentation.js:4:1)
    at async Object.render (surfaceDisplay.js:32:1)
```

```
webWorkerManager.js:100 Error executing method 'convertLabelmapToSurface' on worker 'polySeg': TypeError: Cannot read properties of null (reading 'instance')
    at Object.convertLabelmapToSurface (vendors-node_modules_cornerstonejs_polymorphic-segmentation_dist_esm_workers_polySegConverter-544fc…:81:38)
```

Note that I tried the exact steps above using the commit prior to the one for this PR and there was no exception at step 7.

jbocce (changes requested):
Testing failed.

Devu-trenser:
@jbocce 
Thank you for the detailed testing and feedback.

I have verified the behavior again using both the latest OHIF `master` branch and the OHIF branch  `fix/restrict-labelmap-to-surface-conversion`, linked with the cs3d branch `Devu-trenser:fix/labelmap-to-surface-conversion`.

Following the exact steps mentioned in your comment, **I was unable to reproduce the issue and the labelmap to surface conversion worked as expected without any exceptions.**

A reference video demonstrating the test workflow and results has been attached for your review.

Could you please confirm if you’re still able to reproduce this issue, or if I might be missing any specific steps in your testing process?

With latest OHIF `master`:

https://github.com/user-attachments/assets/b2ce80f2-8be8-439b-ba3e-a679489a1891


With OHIF branch  `fix/restrict-labelmap-to-surface-conversion`:

https://github.com/user-attachments/assets/0f2b2a54-26e6-4c36-bb5c-326a137631fb

jbocce:
@Devu-trenser, apologies, I am not sure what is wrong with my local system so I had a colleague go through my testing steps and it passed for them. Now I will look more closely at the code.

packages/tools/src/utilities/segmentation/computeAndAddRepresentation.ts
```diff
@@ -1,30 +1,23 @@
...
+ * @param onComputationComplete - Optional callback invoked after computation completes.
  * @returns - A promise that resolves with the computed representation data.
+
  */
 async function computeAndAddRepresentation<T>(
   segmentationId: string,
   type: SegmentationRepresentations,
   computeFunction: () => Promise<T>,
-  updateFunction?: () => void,
+  updateFunction?: (segmentationId: string) => void,
```
jbocce: So here the updateFunction now takes an argument. However, I could be wrong, but most of the existing callers of `computeAndAddRepresentation` are NOT passing a function that takes this argument. Please explain.

packages/tools/src/utilities/segmentation/computeAndAddRepresentation.ts
jbocce, line 1: There are lots of changes here to better manage memory and listeners. However, I am having a very difficult time understading where precisely the fix is for avoiding the surface creation once we exit a hanging protocol/layout that shows a 3D volume. Would you be so kind as to point out the exact lines where the actual "fix" occurs? Thanks.

jbocce: Put another way, what was causing the issue before and what specific changes address it and fix it. 😊

Devu-trenser: @jbocce  We’ve slightly updated the implementation, and there’s also a dependent PR here: [OHIF/Viewers#5521](https://github.com/OHIF/Viewers/pull/5521).

**What was causing the issue before:**
Once an event listener was added for a surface, it was never removed. As a result, even after the surface representation was removed from the viewport (for example, when switching to a layout without 3D), the surface conversion continued to occur for every segmentation modification. This also caused the toaster notification to appear even when drawing in non-3D layouts.

**How this change fixes it:**
We now remove the event listener responsible for updating the surface when the corresponding surface representation is removed from the viewport(https://github.com/OHIF/Viewers/pull/5521/files#diff-ea4fdfdb42f95d2861fa7e93ad4cbe5790a936ce05feef28895fb033dcc38b1fR1967). When switching back to a layout that includes a 3D viewport, the update listener is re-added (https://github.com/OHIF/Viewers/pull/5521/files#diff-ea4fdfdb42f95d2861fa7e93ad4cbe5790a936ce05feef28895fb033dcc38b1fR334). This ensures surface updates occur only when the 3D viewport is enabled and prevents conversions or notifications in non-3d layouts.

jbocce: Thank you @Devu-trenser for the explanation. However, with the updated implementation, I would have expected to have seen listeners added and removed (here) in Cornerstone3D. I see in the accompanying https://github.com/OHIF/Viewers/pull/5521 PR that the listeners are added and removed in OHIF. However, the desired solution should be self contained to CS3D because the probelm itself is a Cornerstone issue. 

I have a minimal diff of some code that I believe would solve the issue completely in Cornerstone3D. However one thing the diff below does NOT take care of is re-adding the listener whenever a volume viewport is reintroduced. Please feel free to use the diff as part of your final solution or as a guide to your final solution.

```
diff --git a/packages/tools/src/utilities/segmentation/computeAndAddRepresentation.ts b/packages/tools/src/utilities/segmentation/computeAndAddRepresentation.ts
index 63904629d..5f511fca5 100644
--- a/packages/tools/src/utilities/segmentation/computeAndAddRepresentation.ts
+++ b/packages/tools/src/utilities/segmentation/computeAndAddRepresentation.ts
@@ -49,7 +49,7 @@ async function computeAndAddRepresentation<T>(
   }
 
   // Subscribe to any changes in the segmentation data for real-time updates
-  subscribeToSegmentationChanges(updateFunction);
+  subscribeToSegmentationChanges(segmentationId, updateFunction);
 
   // Notify other system parts that segmentation data has been modified
   triggerSegmentationModified(segmentationId);
@@ -61,7 +61,7 @@ async function computeAndAddRepresentation<T>(
  * Subscribes to segmentation changes by adding an event listener for the SEGMENTATION_DATA_MODIFIED event.
  * If there is an existing listener, it will be unsubscribed before adding the new listener.
  */
-function subscribeToSegmentationChanges(updateFunction) {
+function subscribeToSegmentationChanges(segmentationId, updateFunction) {
   const debouncedUpdateFunction = (event) => {
     _debouncedSegmentationModified(event, updateFunction);
   };
@@ -77,6 +77,17 @@ function subscribeToSegmentationChanges(updateFunction) {
     Events.SEGMENTATION_DATA_MODIFIED,
     updateFunction._debouncedUpdateFunction
   );
+
+  eventTarget.addEventListener(Events.SEGMENTATION_REPRESENTATION_REMOVED, ({ detail }) => {
+    if(detail.segmentationId === segmentationId) {
+      eventTarget.removeEventListener(
+        Events.SEGMENTATION_DATA_MODIFIED,
+        updateFunction._debouncedUpdateFunction
+      );
+
+	   computedRepresentations.delete(segmentationId);
+    }
+  });
 }
 
 const _debouncedSegmentationModified = debounce((event, updateFunction) => {
```

packages/tools/src/utilities/segmentation/computeAndAddRepresentation.ts
Devu-trenser: Updated the implementation

packages/tools/src/utilities/segmentation/computeAndAddRepresentation.ts
Devu-trenser: @jbocce  Thank you so much for reviewing and for sharing the suggestion! 
I’ve updated the implementation to add and remove the event listeners directly within Cornerstone3D. With this update, no corresponding changes are needed in the OHIF Viewer.

We’ve moved the segmentation listeners to be independent of the `computeAndAddRepresentation` method for better separation and maintainability. Previously, each representation’s render call added the event listener individually when calling [computeAndAddRepresentation](https://github.com/cornerstonejs/cornerstone3D/pull/2426/files#diff-7da41867238e893c45127c4c19f98990dc171bdb0811986cc05a8008170e29abR76). This logic has been moved in the [common render method](https://github.com/cornerstonejs/cornerstone3D/pull/2426/files#diff-cfd4acba891200903f9df866f835b35e429f778560ab64426f5c5b908ba554c6R207). Listeners are also removed when a representation is [removed](https://github.com/cornerstonejs/cornerstone3D/pull/2426/files#diff-fadb45350e9efd52d53b463e7b892294f2023396d5b4bc39fe11a4f388557379R203) and [re-added](https://github.com/cornerstonejs/cornerstone3D/pull/2426/files#diff-0153ff619168e4074ab0f409b96ef7c6cf50c3adb1f838c51d54b688c3378273R40) when it’s added back.

Could you please take a look at the latest changes when you get a chance? I’d really appreciate your feedback.

jbocce: You're welcome. However, I should give proper credit to a colleague of mine that suggested the code I gave here. Shout out to @ViniciusResende! ❤️I will now look at the code and test.

packages/tools/src/utilities/segmentation/computeAndAddRepresentation.ts
```diff
@@ -1,30 +1,20 @@
...
- * subscribes to segmentation changes, and triggers segmentation modification.
+ * Computes a segmentation representation and subscribes to future segmentation updates.
  *
  * @param segmentationId - The ID of the segmentation.
- * @param representationType - The type of the segmentation representation.
- * @param computeFunction - The function that computes the representation data.
- * @param options - Additional options for computing the representation.
+ * @param type - Representation type (e.g., LABELMAP, CONTOUR).
+ * @param computeFunction - Async function that computes representation data.
+ * @param updateFunction - Optional function to update UI/state on segmentation change.
```
jbocce: remove this param from the doc too please

packages/tools/src/tools/displayTools/Surface/surfaceDisplay.ts
```diff
@@ -145,9 +144,18 @@ async function render(
   viewport.render();
 }
 
+function getUpdateFunction(
+  viewport: Types.IVolumeViewport | Types.IStackViewport
+): (segmentationId: string) => Promise<void> {
+  const polySeg = getPolySeg();
+  return (segmentationId: string) =>
+    polySeg.updateSurfaceData(segmentationId, { viewport }).then();
```
jbocce: Do we need the `.then()` here?

packages/tools/src/stateManagement/segmentation/SegmentationRenderingEngine.ts
```diff
@@ -196,11 +196,21 @@ class SegmentationRenderingEngine {
...
+        const segmentation = getSegmentation(representation.segmentationId);
+        const existingRepresentation =
+          segmentation.representationData[representation.type] !== undefined;
 
         try {
           // @ts-ignore
-          const viewportId = display.render(viewport, representation);
-          viewportRenderList.push(viewportId);
+          display.render(viewport, representation).then(() => {
+            if (!existingRepresentation) {
```
jbocce, line 206: This is a bit curious for me. From what I can see the `display.render` also does a `addDefaultSegmentationListener` via a call to `computeAndAddRepresentation`. So why do we need this as well?

jbocce (changes requested):
This approach looks good to me. Just a few comments and questions to address. I tested the implementation and it looks great.

packages/tools/src/stateManagement/segmentation/SegmentationRenderingEngine.ts
Devu-trenser: The segmentation listener isn’t actually added via `computeAndAddRepresentation`.
Code reference: https://github.com/cornerstonejs/cornerstone3D/pull/2426/files#diff-0a9cb1d85138bdacc1c65aaab0789c7582336d654589e15c957e3f17d7e383ddL41-L56
That method only compute and add the segmentation representation, but the listener setup is handled separately.

sen-trenser:
@jbocce The comments are resolved. Could you please review and approve the changes? 
cc: @Devu-trenser

jbocce:
> @jbocce The comments are resolved. Could you please review and approve the changes? cc: @Devu-trenser

Thanks for the follow-up. We are looking forward to getting this merged soon. We are doing extra testing on our end for this PR because of how core the changes are. That said, we have tested the changes using the most recent master OHIF branch linked with the branch for this PR with the most recent changes of CS3D main merged into it. Unfortunately we found a problem. As a convenience for our own testing we have also deployed such an OHIF here: https://cs3d-pr-2426.netlify.app.

See the attached video for how the problem can be reproduced. Basically, returning from 3D four up back to MPR throws an exception. Please investigate as this problem does not appear to occur in https://viewer-dev.ohif.org/. Please feel to reach out for assistance if necessary. Be sure to use be up-to-date in both the CS3D branch and OHIF. Thank you.

FYI @Devu-trenser, @sen-trenser and @sedghi 


https://github.com/user-attachments/assets/66a743d9-5b9e-4f9d-8498-2d424a3720bf

Devu-trenser:
@jbocce Thank you for the detailed feedback and for sharing the test setup and video.

It appears that this issue is not introduced by this PR but is an existing timing-related issue.

- The problem is consistently reproducible in the latest master branch when tested locally in a freshly cloned repo (not linked with CS3D), commit: 58aa0d1eaae2e1a8103e6096864a4e5d7baf090e.
-  https://github.com/user-attachments/assets/8edebb03-a253-4e92-8088-0d2adf651692


- It occurs rarely in commit: 16233d980f17abfc30461b46c6b1888be10c7840, and can always be reproduced if the layout is switched (from 3D to 2D) before the surface conversion completes.
- https://github.com/user-attachments/assets/8decd2d9-acac-49d9-87ae-53984770eec1



Root cause:

- When computing the surface from the labelmap for a 3D viewport (an asynchronous process), if the layout changes to MPR during the computation, a subsequent async call to [createAndCacheSurfacesFromRaw](https://github.com/Devu-trenser/cornerstone3D/blob/main/packages/polymorphic-segmentation/src/Surface/createAndCacheSurfacesFromRaw.ts#L18) triggers a call to getSegmentIndexColor.

- At that point, the representation for the removed 3D viewport is no longer available due to the layout change, resulting in an empty representation array. Consequently, the segmentation index is returned as null, leading to a failure at [line 39](https://github.com/Devu-trenser/cornerstone3D/blob/main/packages/polymorphic-segmentation/src/Surface/createAndCacheSurfacesFromRaw.ts#L39).


As an initial thought, one potential quick mitigation could be to add a guard or cleanup check before invoking createAndCacheSurfacesFromRaw, ensuring that the associated viewport and segmentation representations are still valid. However, this approach has a side effect, the computed snackbar is not removed from the viewport.

I’ll continue investigating this further to identify a more robust and stable solution. Do you happen to have any suggestions or ideas that could help address this scenario more effectively?

jbocce (approved):
The changes look fine to me. Thanks for your patience with this. We did extensive testing on this. We found several issues that were NOT regression, but we will list them here for tracking purposes...

- https://github.com/OHIF/Viewers/issues/5558
- https://github.com/OHIF/Viewers/issues/5559
- https://linear.app/ohif/issue/OHI-2311

sen-trenser:
> > @jbocce The comments are resolved. Could you please review and approve the changes? cc: @Devu-trenser
> 
> Thanks for the follow-up. We are looking forward to getting this merged soon. We are doing extra testing on our end for this PR because of how core the changes are. That said, we have tested the changes using the most recent master OHIF branch linked with the branch for this PR with the most recent changes of CS3D main merged into it. Unfortunately we found a problem. As a convenience for our own testing we have also deployed such an OHIF here: https://cs3d-pr-2426.netlify.app.
> 
> See the attached video for how the problem can be reproduced. Basically, returning from 3D four up back to MPR throws an exception. Please investigate as this problem does not appear to occur in https://viewer-dev.ohif.org/. Please feel to reach out for assistance if necessary. Be sure to use be up-to-date in both the CS3D branch and OHIF. Thank you.
> 
> FYI @Devu-trenser, @sen-trenser and @sedghi
> 
>  ScreenHunter.Nov.10.14.31.mp4

@jbocce Based on our further analysis, it seems [computeAndAddRepresentation](https://github.com/Devu-trenser/cornerstone3D/blob/131f3f57ef1f6e504709acf510988587657cb25d/packages/tools/src/tools/displayTools/Surface/surfaceDisplay.ts#L76-L80) was called multiple times from render. So, it might be triggering multiple computation in background until the Surface representation data is set after first computation completes. This may cause the issue https://github.com/OHIF/Viewers/issues/5559 by other queued up conversions even if the first notification shows and prevents further notifications.

We have neither confirmed the root cause nor a fix. However, if this is causing the issue you mentioned, this may probably be fixed with a promise map in the file, and await on the same promise for each segmentation conversion request instead of triggering more computation requests.

Issue #2412, opened 2025-10-21
[Bug] Advanced Magnifying Not working


mohamedFayezPax:
### Describe the Bug

When try to open Advanced Magnifying Tool even in online examples it is not working and i got that errors

AdvancedMagnifyTool.ts:1421 Uncaught TypeError: Cannot read properties of null (reading 'viewport')
segmentIndex.ts:31 Uncaught TypeError: Cannot read properties of undefined (reading 'segments')

<img width="1900" height="285" alt="Image" src="https://github.com/user-attachments/assets/51893016-ed83-40e3-9594-b01ea2a97037" />

### Steps to Reproduce

1-open Advanced Magnifying Tool example
2- Press Ctrl + Left Click to activate the magnifying glass

### The current behavior

Nothing happen and error dsiplayed in console

### The expected behavior

Advanced Magnifying Tool to work normally

### System Information

 System:
    OS: Windows 11 10.0.22631
    CPU: (8) x64 11th Gen Intel(R) Core(TM) i5-1135G7 @ 2.40GHz
    Memory: 12.54 GB / 31.71 GB
  Binaries:
    Node: 22.16.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.22 - C:\Users\mohamed.fayez\AppData\Roaming\npm\yarn.CMD
    npm: 10.9.2 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Chrome: 141.0.7390.108
    Edge: Chromium (140.0.3485.54)
    Firefox: 143.0.4 - C:\Program Files\Mozilla Firefox\firefox.exe
    Internet Explorer: 11.0.22621.3527

sedghi:
Good catch

maxcomperatore:
anyone working on this?

jbocce:
It is assigned to me. I will be soon.

PR #2427, opened 2025-10-27
fix(AdvancedMagnifyTool): Exceptions are no longer thrown for the AdvancedMagnifyTool

jbocce:
### Context
#2412

### Changes & Results
Upon investigation, it was determined that in part changes in #2322 contributed to #2412. So we only set the active segment index if there is a segmentation representation.

Further investigation showed that the AdvancedMagnifyTool has NOT been working on volume viewports for at least since February of 2025. In general, the click to apply the AdvancedMagnifyTool on a volume viewport would immediately throw an exception. So, in AdvancedMagnifyTool, we now use better detection for segmentation actors. Now the exception is gone, the magnify glass is displayed, but behaviour is still not equivalent to that for a stack viewport. A new issue will be logged soon.

See #2428.

### Testing
See #2412 

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] "OS: Windows 11
- [x] "Node version: 23.9.0
- [x] "Browser: Chrome 141.0.7390.123
  

### DIFF
--- a/packages/tools/src/tools/AdvancedMagnifyTool.ts
+++ b/packages/tools/src/tools/AdvancedMagnifyTool.ts
@@ -68,7 +68,7 @@ const MAGNIFY_VIEWPORT_INITIAL_RADIUS = 125;
 const { Events: csEvents } = Enums;
 
 // TODO: find a better to identify segmentation actors
-const isSegmentation = (actor) => actor.uid !== actor.referencedId;
+const isSegmentation = (actor) => !!actor.representationUID;
 
 export type AutoPanCallbackData = {
   points: {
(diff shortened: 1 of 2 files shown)


packages/tools/src/tools/AdvancedMagnifyTool.ts
jbocce, line 1: The changes in this file fix an exception being thrown for volume viewports that was occurring at least since commit e1b5bc646 (Feb 28, 2025), but likely has existed longer. However even with these changes the magnify glass still does not behave like it does for stack viewports.

See #2428

maxcomperatore:
beautifully done

PR #2429, opened 2025-10-27
feat(CrosshairsTool): implement crosshairs for same-orientation viewports

nagychris:
### Context

So far, the CrosshairsTool was not available for viewports with the same orientation (e.g. three AXIAL / ACQUSITION). 

However, it might be useful to have the crosshairs also available in these cases, as we might want to display the same position on different sequences (e.g. to find a specific component of a body part that is relatively small, like lymph nodes).

### Changes & Results

For same orientation viewports, we can also render a crosshairs now. The crosshairs does not change the camera of the viewports (scroll through slices etc.), it just syncs the crosshairs between the three viewports on the current 2D (x, y) plane:

https://github.com/user-attachments/assets/fda29b9b-ded9-472c-8028-eca1d34e977a

MPR works just the same as before: 

https://github.com/user-attachments/assets/ba144a70-0e50-4eeb-a61f-987954347b29

### Testing

- checkout this PR
- change `packages/tools/examples/crossHairs/index.ts` to render two / three viewports with the same orientation (example see below)
- run the example via `yarn example crossHairs`
- check the functionality of the new crosshairs

You can use the following `viewportInputArray` (l. 306):
```ts
  const viewportInputArray = [
    {
      viewportId: viewportId1,
      type: ViewportType.ORTHOGRAPHIC,
      element: element1,
      defaultOptions: {
        orientation: Enums.OrientationAxis.AXIAL,
        background: <Types.Point3>[0, 0, 0],
      },
    },
    {
      viewportId: viewportId2,
      type: ViewportType.ORTHOGRAPHIC,
      element: element2,
      defaultOptions: {
        orientation: Enums.OrientationAxis.AXIAL, // changed to AXIAL
        background: <Types.Point3>[0, 0, 0],
      },
    },
    {
      viewportId: viewportId3,
      type: ViewportType.ORTHOGRAPHIC,
      element: element3,
      defaultOptions: {
        orientation: Enums.OrientationAxis.AXIAL, // changed to AXIAL
        background: <Types.Point3>[0, 0, 0],
      },
    },
  ];
```

#### PR

#### Code

  etc.)

  additions or removals.

-> do we need to update the docs for this? 

#### Tested Environment

- [] OS: Ubuntu 24
- [] Node version: 
- [] Browser: Chromium / Brave / Chrome

### DIFF
--- a/packages/tools/src/tools/CrosshairsTool.ts
+++ b/packages/tools/src/tools/CrosshairsTool.ts
@@ -4,7 +4,7 @@ import vtkMatrixBuilder from '@kitware/vtk.js/Common/Core/MatrixBuilder';
 
 import { AnnotationTool } from './base';
 
-import { getRenderingEngine, type Types } from '@cornerstonejs/core';
+import type { Types } from '@cornerstonejs/core';
 import {
   getEnabledElementByIds,
   getEnabledElement,
@@ -94,6 +94,11 @@ const OPERATION = {
   SLAB: 3,
 };
 
+enum CrosshairsType {
+  MPR = 1,
+  SAME_ORIENTATION = 2,
+}
+
 /**
  * CrosshairsTool is a tool that provides reference lines between different viewports
  * of a toolGroup. Using crosshairs, you can jump to a specific location in one
@@ -106,6 +111,9 @@ const OPERATION = {
 class CrosshairsTool extends AnnotationTool {
   static toolName;
 
+  isFocusedOnCanvas: boolean = false;
+  toolWorldFocus: Types.Point3;
+  viewportCrosshairsConfig: { [viewportId: string]: CrosshairsType } = {};
   toolCenter: Types.Point3 = [0, 0, 0]; // NOTE: it is assumed that all the active/linked viewports share the same crosshair center.
   // This because the rotation operation rotates also all the other active/intersecting reference lines of the same angle
   _getReferenceLineColor?: (viewportId: string) => string;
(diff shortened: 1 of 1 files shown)


packages/tools/src/tools/CrosshairsTool.ts
```diff
@@ -158,11 +166,6 @@ class CrosshairsTool extends AnnotationTool {
         filterActorUIDsToSetSlabThickness: [],
         // blend mode for slabThickness modifications
         slabThicknessBlendMode: Enums.BlendModes.MAXIMUM_INTENSITY_BLEND,
-        centerPoint: {
-          enabled: false,
-          color: 'rgba(255, 255, 0, 0.5)',
-          size: 2,
-        },
         mobile: {
           enabled: false,
```
sedghi, line 170: why is centerPoint removed

nagychris:
@sedghi I agree, this PR only adds some additional complexity to handle the specific case of 2 or 3 viewports with the same orientation. Probably the cleanest solution (long-term) would be to have a more generic CrosshairsTool that handles >3 viewports with flexible orientations (also allowing for dynamic switching).

However, I guess this would involve much more effort, and maybe even a rewrite of the current tool. As I am highly  interested in an improved Crosshairs with same-orientation support, are there any plans of implementing this in the near future? Otherwise, are there any downsides of integrating this into the library until the refactoring & extension takes place? Or can we maybe modify this PR in a way that suits your plan in a better way?

packages/tools/src/tools/CrosshairsTool.ts
nagychris: This was an error during merging, I could re-add it again if we decide to move on with the PR.

PR #2431, opened 2025-10-29
feat(decoders): allow preloadinng

IbrahimCSAE:
### Context

Allows preloading decoders, default -> no network request for all. configured to preload -> network request for all, configured with array of names of decoders -> only network requests for those.


packages/dicomImageLoader/src/shared/decoders/preloadDecoders.ts
```diff
@@ -0,0 +1,102 @@
+import type { LoaderDecodeOptions } from '../../types';
+
+/**
+ * Available decoder names for preloading
+ */
+export type DecoderName =
```
sedghi: can you make this enum and take it outside?

packages/dicomImageLoader/src/shared/decoders/preloadDecoders.ts
wayfarer3130, line 1: Can you add a general "Decoders" class, and have it own a Map of decoder name to decoder configuration.
Add a "registerDecoder" to add a new one
Add configurations for all 6 of these here
Add an "initialize" method for creating a new instance of the decoder, which will get stored in the Map itself.  That instance will have a function that does that decode on a standard set of attributes.
The standard/default configuration should include the WASM path to load from (at least as an over-ride).
It should also include a flag as to whether to pre-load that item

When calling preload, ONLY the items flagged for preloading will get preloaded.
Calling load on an item will return the previously created async wait function, OR will create one as required.  That is, it has to be thread safe for preloading starting multiple times and not load the WASM multiple times.

As part of the configuration for each one, have the WASM path and the

packages/dicomImageLoader/src/types/LoaderOptions.ts
```diff
@@ -31,4 +31,11 @@ export interface LoaderOptions {
   errorInterceptor?: (error: LoaderXhrRequestError) => void;
   strict?: boolean;
   decodeConfig?: LoaderDecodeOptions;
+  /**
+   * Preload image decoders during initialization.
+   * - true: preload all available decoders
+   * - string[]: array of decoder names to preload (e.g., ['htj2k', 'jpeg2000', 'jpegls'])
+   * - false/undefined: no preloading (default)
+   */
+  preloadDecoders?: boolean | string[];
```
wayfarer3130, line 40: Suggest making this a configuration object for the loaders that UPDATES the existing loader configuration, on a per item basis, maybe with an overall flag to preload everything.  That way a completely new decoder can be added just by configuring it here.  Also allows specfying the wasm path and other required custom attributes depending on configuration.

Issue #2287, opened 2025-08-22
[Feature Request] Support for Axis-Based Image Stretching


Devu-trenser:
This feature request is on behalf of [FlyWheel.io](https://flywheel.io/)

### What feature or change would you like to see made?

- Add support for a stretch option that allows images to be stretched along specific axes.

### Why should we prioritize this feature?

- Provide a more consistent viewing experience across different imaging modalities.

cc: @kgoettler

sedghi:
we discussed this in office hours i guess?

wayfarer3130:
I created a separate PR that is related to this one, but not the same.  However, both should be done as a single change:
[#2350] describes a bug in the camera positions drifting.
The fix is to have original camera information as one component, and then compute fixed delta in a consistent fashion, so that every update uses the new information as an absolute update rather than being as a specific "set it to this value" which can cause drifts.
When this is done, adding a non uniform anisotropic scaling could be included as additional properties.
Without doing it that way, the existing calculations of focal point/position become problematic as they directly interact with the non uniform scaling.

That is, the scaling should be something like:

scalingDirection = [1,0]  (for example to scale x non-uniformly, and leave y identical)
aspectScale = 3

This would give a 3:1 scaling on the x:y direction.  The x direction would change scale values separately from the y, ONLY for canvas to index/world coordinates, without any changes in index to world coordinates.

Other linear and non-linear transforms are possible, but it would be very helpful to have an explanation of how they get used as skew isn't a common operation when measurements are included.  Not saying it is impossible to use for registration, just tricky.

Devu-trenser:
@wayfarer3130 I haven’t seen that PR yet. Could you please share the link so I can take a look?

wayfarer3130:
> [@wayfarer3130](https://github.com/wayfarer3130) I haven’t seen that PR yet. Could you please share the link so I can take a look?

Sorry, should have said, proposed change rather than a PR.  The idea is to do some of that type of scaling capability, to fix a number of outstanding bugs.

nithin-trenser:
@wayfarer3130 I’m currently analyzing the feasibility of implementing this feature and will share updates once I have more details.

nithin-trenser:
@wayfarer3130 

[...]

PR #2443, opened 2025-11-04
Feat: Axis-based Image Stretching

arul-trenser:
### Context

- PR for the Feature Request: **Axis-based Image Stretching** #2287 
- This PR has been incorporated by [FlyWheel.io](https://flywheel.io/).

### Changes & Results
This PR introduces **axis-based image stretching** using camera projection matrix manipulation.

#### Usage Example
**Stretch along the X-axis:**
```js
const { aspectRatio }  = viewport.getCamera()
viewport.setCamera({
    acpectRatio: [2, aspectRatio[1]]
})
```

**Stretch along the X-axis:**
```js
const { aspectRatio }  = viewport.getCamera()
viewport.setCamera({
    acpectRatio: [aspectRatio[0], 2]
})
```

### Testing

Basic Stretching along X and Y axis:

https://github.com/user-attachments/assets/20b3d859-fd03-4a7e-ac22-c08bd811723f

Segmentation Brush Tool Behavior in Stretched Viewport:

https://github.com/user-attachments/assets/9a5162e7-5ebe-496d-8f74-bbdd80ece85a

Circle ROI Tool Behavior in Stretched Viewport

https://github.com/user-attachments/assets/17ced2a7-09a9-47f3-879b-f3c80621c331

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] "OS:  Windows 11
- [x] "Node version:  22.19.0
- [x] "Browser: Chrome  141.0.7390.123
  


arul-trenser:
@sedghi when you have a moment, could you kindly review this PR at your earliest convenience?
Thanks!

sedghi:
how we can test it?

packages/core/src/types/CPUFallbackViewport.ts
```diff
@@ -26,6 +27,7 @@ interface CPUFallbackViewport {
   colormap?: CPUFallbackColormap;
   displayedArea?: CPUFallbackViewportDisplayedArea;
   modality?: string;
+  aspectRatio?: Point2;
```
sedghi, line 30: are you adding aspect ratio for both cpu and gpu pipelines?

arul-trenser:
> how we can test it?

@sedghi  Added an example demonstrating the stretch behavior in `axialBasedImageStretching`
Also attached videos in the PR description

sedghi:
Are you aware of other viewers how they do this feature? i'm trying to evaluate the result at least to some other library or product that has this feature

jmannau:
Hi,
We work with systems that have different vertical/horizontal scales.  For example, with Optical Coherence Topography (https://eyewiki.org/Optical_Coherence_Tomography) scans, it is common to display the image at 2:1 (vertical/horizontal) aspect ration.  The scans often have 2x->4x vertical resolution compared to the horizontal scale. This allows better visualisation of pathology in the vertical/depth axis. We also allow users to change the aspect ratio between 1:1 and 4:1 to account for different vendor devices and scan types. 
Below are 2 images that demonstrate this.

**1:1 Aspect Ratio**
Screenshot 2025-11-12 at 11 23 17 am

**the same scan at 2:1 Aspect Ratio**
Screenshot 2025-11-12 at 11 23 30 am

sedghi:
@jmannau lovely, would you be able to test this PR against the viewer you have?

jmannau:
Happy to. But we’re very short on time until the new year. I’ll see what we can do

sen-trenser:
@jmannau We are aiming to include this change along with the CS3D dependent version of the upcoming OHIF 3.12 release, which is expected to be released later this year.
When you have the opportunity, could you please prioritize reviewing this MR and share any feedback? Your input would be greatly appreciated and help us to timely include it in the next release.

Thank you!

packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -2683,7 +2691,7 @@ class StackViewport extends Viewport {
     });
   }
 
-  private resetCameraGPU({ resetPan, resetZoom }): boolean {
+  private resetCameraGPU({ resetPan, resetZoom, resetAspectRatio }): boolean {
```
wayfarer3130: Can you just pass the reset options to the parent rather than needing to specifically list them?  It means that all new reset options will just flow through unless specifically editted.
eg:
resetCameraGPU(resetOptions?: ResetOptions): boolean

...
return super.resetCamera({
   resetToCenter,
   ...resetOptions
})

packages/core/src/RenderingEngine/Viewport.ts
```diff
@@ -1391,6 +1396,36 @@ class Viewport {
...
   }
 
+  /**
+   * Returns the current aspect ratio of the viewport as a 2D point `[widthRatio, heightRatio]`.
+   *
+   * @returns The current aspect ratio `[widthRatio, heightRatio]`
+   *          based on the active camera settings.
+   */
+  public getAspectRatio(): Point2 {
+    const { aspectRatio } = this.getCamera();
```
wayfarer3130, line 1412: The initial aspect ratio probably needs to come from the viewport properties so that a viewport can be laid out with initial aspect ratio on first display.  Otherwise you need to think about what happens when the aspect ratio changes and you want to fit the image to the area.

wayfarer3130: @sedghi - what are your thoughts on the initial aspect ratio being a property of the viewport, and then the current aspect ratio gets applied later?  This interacts with some of our thoughts on how to manage the viewport state in OHIF.

packages/core/src/RenderingEngine/Viewport.ts
```diff
@@ -1391,6 +1396,36 @@ class Viewport {
...
+    return aspectRatio;
+  }
+
+  /**
+   * Sets the aspect ratio of the viewport using the provided 2D point `[widthRatio, heightRatio]`.
```
wayfarer3130: This changes the transform for world to/from canvas mapping - I assume that is handled by the vtk transform?

wayfarer3130: Could you document how it affects that?
Also, you need to document that it does NOT affect camera renders.

packages/tools/src/utilities/getCenterAndRadiusInCanvas.ts
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * Calculates the center point and radius in canvas coordinate of a circle
+ * from a set of world coordinates within the given viewport.
+ *
+ * The function projects the provided world coordinates into the
+ * viewport's canvas coordinates and returns both the
+ * calculated center and radius in canvas coordinates.
+ *
+ * @param {Types.Point3[]} points - The list of 3D points defining the circle - center and point on circle.
```
wayfarer3130: Typescript convention is to list just the name, so
@param points - The list of ...
as the type is implicit in the function declaration.

packages/tools/src/tools/annotation/CircleROITool.ts
```diff
@@ -238,14 +243,84 @@ class CircleROITool extends AnnotationTool {
     const { viewport } = enabledElement;
 
     const { points } = annotation.data.handles;
-    const canvasHandles = points.map((p) => viewport.worldToCanvas(p));
-    const canvasCenter = canvasHandles[0];
-    const radius = getCanvasCircleRadius([canvasCenter, canvasHandles[1]]);
-    const radiusPoint = getCanvasCircleRadius([canvasCenter, canvasCoords]);
+    // Get the radius in world units from the drag distance
+    const ellipseWorldCoordinates = getEllipseWorldCoordinates(
+      [points[0], points[1]],
```
wayfarer3130: Just pass in points or points.slice(0,1), but I believe you will only have two points.

packages/core/src/RenderingEngine/Viewport.ts
```diff
@@ -2002,9 +2044,17 @@ class Viewport {
...
     }
+    const currentAspectRatio = this.getAspectRatio();
+    target.aspectRatio = currentAspectRatio;
     if (pan) {
-      target.pan = this.getPan();
-      vec2.scale(target.pan, target.pan, 1 / initZoom);
+      const currentPan = this.getPan();
+      const [aspectX, aspectY] = currentAspectRatio;
+
+      // Normalize pan to remove effect of zoom and stretch
```
wayfarer3130: Need to document this in the return if you are doing it.

packages/tools/src/utilities/getCenterAndRadiusInCanvas.ts
wayfarer3130, line 1: This function shouldn't use the aspect ratio, and there are three variants of it depending on what you are trying to do:
- circle in world coordinates
- circle in canvas coordinates
- circle in image coordinates

You HAVE to know which one you are doing in order to figure out what this function returns and also how to draw the appropriate object.  Each one has it's use, but they are quite different in implementation, as well as in serialization and deserialization.

What the adapters uses for image co-planar annotations is circle in image coordinates
For MPR annotations, it uses circle in world coordinates
The radius reported is always in calibrated coordinates, which is a specified transform of image coordinates.

wayfarer3130:
We need to define some requirements as part of this change for what it is that the circle, sphere and elliptical tools are doing.
The sphere I believe should always be a sphere in world coordinates, as otherwise it does weird things to the shape.  That means it should show as an ellipse if displaying world coordinates does not result in an aspect ratio of 1:1 for the rows/columns.

Circle and circle brushes are harder because of the differences in how things get saved.  The circle brush is never saved in image coordinates, so I think we could fairly easily display a circle in the current transformed coordinates.
Circle annotations are saved as both MPR in world coordinates and as index coordinates.
This needs some documentation in a requirements section for exactly what it is that is meant by "sphere" "circle brush" and "circle tool" under MPR/co-planar modes with index, canvas and world coordinates at different aspect ratios.

wayfarer3130:
For the volume (mpr and 3d) use cases, we should define what the aspect ratio means and how it applies in terms of the transform.  In general, one can have 3 vectors that the transform is applied along, and those 3 are orthonormal, with a scaling factor defining them.  I think this is something we need to figure out how to incorporate into the transform directly rather than trying to figure out which axial/sagittal/coronal it is closest to.  That may actually be easier if we used a generic canvas stretch direction vector and orthogonal scale vector rather than x,y scaling.  That way the transform for rotation simply rotates the stretch vector.

packages/core/src/RenderingEngine/StackViewport.ts
arul-trenser: Updated as suggested.

packages/core/src/RenderingEngine/Viewport.ts
arul-trenser: Updated it.

packages/tools/src/tools/annotation/CircleROITool.ts
arul-trenser: Updated.

packages/core/src/RenderingEngine/Viewport.ts
arul-trenser: Added.

sen-trenser:
> For the volume (mpr and 3d) use cases, we should define what the aspect ratio means and how it applies in terms of the transform. In general, one can have 3 vectors that the transform is applied along, and those 3 are orthonormal, with a scaling factor defining them. I think this is something we need to figure out how to incorporate into the transform directly rather than trying to figure out which axial/sagittal/coronal it is closest to. That may actually be easier if we used a generic canvas stretch direction vector and orthogonal scale vector rather than x,y scaling. That way the transform for rotation simply rotates the stretch vector.

@wayfarer3130 Thank you for the detailed review and for outlining a more generalised approach to handling aspect ratio and scaling for volume, MPR, and 3D use cases. We appreciate the architectural insight and the time spent analysing this area.

We would like to clarify that this pull request is intentionally scoped to enabling stretch behavior for the acquisition orientation only. The suggestion to introduce a generic stretch direction vector, and rotate it based on orthogonal scaling vectors across MPR viewports represents a broader design change that goes beyond the original intent of this contribution.

Incorporating this approach would require a rework of the existing transform model and interaction logic, effectively constituting a separate feature rather than an incremental improvement to the current implementation. It may also refined for behaviors related to other tools in MPR views. Addressing these changes extend beyond the intended scope of this contribution and cannot be incorporated within this pull request without significantly expanding its complexity and impact. 

Given this, we would like to ask whether it would be acceptable to proceed with this PR as scoped, without incorporating the suggested generalised stretch behavior.

Thank you again for the review and for maintaining the project.

packages/tools/src/utilities/getCenterAndRadiusInCanvas.ts
arul-trenser: Updated as suggested

packages/core/src/RenderingEngine/Viewport.ts
arul-trenser: Added aspectRatio property for viewport.

packages/tools/src/utilities/getCenterAndRadiusInCanvas.ts
arul-trenser: Updated the function without using aspect ratio.

arul-trenser:
> We need to define some requirements as part of this change for what it is that the circle, sphere and elliptical tools are doing. The sphere I believe should always be a sphere in world coordinates, as otherwise it does weird things to the shape. That means it should show as an ellipse if displaying world coordinates does not result in an aspect ratio of 1:1 for the rows/columns.
> 
> Circle and circle brushes are harder because of the differences in how things get saved. The circle brush is never saved in image coordinates, so I think we could fairly easily display a circle in the current transformed coordinates. Circle annotations are saved as both MPR in world coordinates and as index coordinates. This needs some documentation in a requirements section for exactly what it is that is meant by "sphere" "circle brush" and "circle tool" under MPR/co-planar modes with index, canvas and world coordinates at different aspect ratios.

@wayfarer3130  In stretched viewports, annotation tools operate in world coordinates, so a Circle ROI defined by a world-space center and radius may render as an ellipse, preserving true physical geometry. The segmentation brush cursor is computed in canvas coordinates and always appears as a perfect circle regardless of stretching. The brush should draw perfect circles, not ellipses, regardless of image stretching. When the image is stretched or shrunk, the drawn segments should stretch or shrink proportionally with the image.

For clarity, the attached video shows the current behavior of the Circle tool, Circle brush, and Sphere brush in a stretched viewport, as implemented in this PR.


https://github.com/user-attachments/assets/1f4ed234-074a-482d-9b05-c16273f9f2f6

arul-trenser:
@wayfarer3130 Thanks for the feedback! I’ve applied the suggested changes and included the documentation. Looking forward to your thoughts; please review and approve if everything looks good.

packages/core/src/RenderingEngine/Viewport.ts
```diff
@@ -1391,6 +1398,58 @@ class Viewport {
...
+    if (this.options?.aspectRatio) {
+      return this.options.aspectRatio;
+    }
+
+    const { aspectRatio } = this.getCamera();
+    return aspectRatio || [1, 1];
+  }
+
+  /**
+   * Sets the aspect ratio of the viewport using the provided 2D point
```
wayfarer3130: Sets the aspect ratio of the viewport (canvas) using the provided 2D point

packages/core/src/RenderingEngine/Viewport.ts
```diff
@@ -1391,6 +1398,58 @@ class Viewport {
...
   }
 
+  /**
+   * Returns the current aspect ratio of the viewport as a 2D point `[widthRatio, heightRatio]`.
+   *
+   * @returns The current aspect ratio `[widthRatio, heightRatio]`
+   *          based on the active camera settings.
+   */
+  public getAspectRatio(): Point2 {
+    if (this.options?.aspectRatio) {
```
wayfarer3130: This will return the wrong value if the options had an aspect ratio initially, but setAspectRatio has been called subsequently to update the value.

packages/core/src/RenderingEngine/Viewport.ts
```diff
@@ -1454,7 +1513,7 @@ class Viewport {
   }
 
   protected getCameraNoRotation(): ICamera {
-    const vtkCamera = this.getVtkActiveCamera();
+    const vtkCamera = this.getVtkActiveCamera() as extendedVtkCamera;
```
wayfarer3130, line 1521: @sedghi - how can the default return type of getVtkActiveCamera be setup so that it has the extended methods for aspect ratio?

packages/core/src/RenderingEngine/vtkClasses/extendedVtkCamera.ts
```diff
@@ -0,0 +1,115 @@
...
+   */
+  publicAPI.getProjectionMatrix = (aspect, nearZ, farZ) => {
+    const matrix = superGetProjectionMatrix(aspect, nearZ, farZ);
+
+    const [sx, sy] = model.aspectRatio;
+
+    if (sx !== 1.0 || sy !== 1.0) {
+      const viewUp = publicAPI.getViewUp();
+      const viewPlaneNormal = publicAPI.getViewPlaneNormal();
+      const { idxX, idxY } = getProjectionScaleIndices(viewUp, viewPlaneNormal);
```
wayfarer3130: Doesn't this already handle the aspect ratio as the "aspect" parameter?  Otherwise you should be able to create the generic multiplier by getting `viewUp*sy` and `viewRight*sx` and using them as the matrix multiplier - that is:
```
[ux,uy,uz] = viewUp * sy
viewRight = vec3.cross(viewUp,viewNormal)
[rx,ry,rz] = viewRight * sz

Then your matrix is something like:
[
  rx,ry,rz,
  ux,uy,uz,
  0,0,1
]

```

and you matrix multiply the previous matrix by the normalization matrix.

You have to get the ordering/setup right, but given that you have it working for 3 existing orientations, it shouldn't be too hard to figure out where the values all go.

packages/core/src/RenderingEngine/helpers/getProjectionScaleIndices.ts
wayfarer3130, line 1: Modify this as requested to return the scaling matrix so that it works generically for any orientation, not just axial/sagittal/coronal.

packages/core/src/RenderingEngine/vtkClasses/extendedVtkCamera.ts
wayfarer3130, line 1: Hoping this can move into the base definitions so you don't need to case the camera types.

packages/tools/examples/calibrationTools/index.ts
wayfarer3130, line 1: These changes allow testing with aspect ratio applied to the canvas AND the aspect ratio applied to the underlying size.  This all seems to work as far as I can tell.

packages/tools/examples/resize/index.ts
wayfarer3130, line 1: This allows both the viewport size and the canvas aspect ratio to be varied independently and then resizing applied.  It demonstrates that the size changes aren't working correctly any longer with resize.
This can be seen by switching between the 0.5:1 and 1:2 ratios - both of these should appear exactly identical as they only affect aspect ratio, not zoom, but right now they end up modifying the zoom.

wayfarer3130:
@sedghi - I think now that it is clarified that the aspect ratio applies only to the display and not to the coordinate/world space, I think this becomes a lot like rotation or flip.  There are still some bugs to fix in terms of the resize operation relative to the aspect ratio, where it no long applies the zoom to fit or scale size.

packages/core/src/RenderingEngine/Viewport.ts
arul-trenser: Updated the comment as suggested

packages/core/src/RenderingEngine/Viewport.ts
arul-trenser: Updated to return the correct aspectRatio

packages/tools/examples/resize/index.ts
arul-trenser: Fixed the issue.

https://github.com/user-attachments/assets/3a81ff51-1e07-4836-866a-4d464a2283ed

arul-trenser:
@wayfarer3130  Updated the PR based on the review feedback and added additional unit tests to validate the projection matrix across all orientations and rotated viewports. Requesting review of the latest changes.

wayfarer3130:
This is too much of a change for the OHIF 3.12 release.
I will do a final review once 3.12 is released and then get it merged at that time.

sen-trenser:
@wayfarer3130 Could you please review and share your feedback on this changes?

arul-trenser:
@wayfarer3130 Could you please review and share your feedback for this PR?

packages/core/src/RenderingEngine/helpers/getProjectionScaleMatrix.ts
wayfarer3130, line 1: The calculated transform doesn't work properly when a rotation is applied to the viewport.

packages/core/src/RenderingEngine/vtkClasses/extendedVtkCamera.ts
```diff
@@ -0,0 +1,122 @@
...
+     * Get the aspectRatio of the viewport
+     *  @defaultValue [1, 1]
+     */
+    getAspectRatio(): [x: number, y: number];
+
+    /**
+     * Set the aspectRatio of the viewport
+     * @param aspectRatio - aspectRatio of the viewport in x and y axis
+     */
+    setAspectRatio(aspectRatio: [x: number, y: number]): boolean;
```
wayfarer3130, line 46: The aspect ratio should apply to the canvas space, regardless of orientation of the image.  You probably need to document that.  You MIGHT choose to instead apply a single aspect stretch value, and a vector in canvas space along which to apply it.  That would then allow for arbitrary orientation stretches, and it becomes possible to choose to stretch the image rather than the viewport by applying rotation/flip to the orientation of the stretch.

packages/core/src/RenderingEngine/Viewport.ts
wayfarer3130, line 1: None of the aspect ratio stuff applies to the CPU based rendering - there isn't any reason it can't, and it is important to allow screen captures/display with such information on it.

packages/core/src/utilities/getNormalizedAspectRatio.ts
```diff
@@ -0,0 +1,23 @@
...
+const EPSILON = 1e10;
+
+/**
+ * Normalizes a pair of dimensions into a standardized aspect ratio array.
+ * @param aspectRatio - An array containing [width, height].
+ * @returns A normalized array [w, h] where at least one value is 1.
+ */
+export const getNormalizedAspectRatio = (aspectRatio: Point2): Point2 => {
+  const [width, height] = aspectRatio;
+  if (width === height) {
```
wayfarer3130: Use isEqual here please.

wayfarer3130 (changes requested):
The rotation of images is such a fundamental part of MPR viewing that I think it needs to be fixed.
Also, the scale of the image area should be updated so that scale to fit applies the specified transform on the image aspect so as to scale to fit after the transform.  If you can figure out scale to fit fixes for rotated images at the same time, I wouldn't mind having that included as well, but it isn't directly part of this PR so it doesn't need to be included.

packages/core/src/utilities/getNormalizedAspectRatio.ts
arul-trenser: Modified as requested in https://github.com/cornerstonejs/cornerstone3D/pull/2443/commits/a30d0e331f7a5d94792fa946f1128a80bd19a476

packages/core/src/RenderingEngine/helpers/getProjectionScaleMatrix.ts
arul-trenser: Fixed rotation issue in https://github.com/cornerstonejs/cornerstone3D/pull/2443/commits/a30d0e331f7a5d94792fa946f1128a80bd19a476

packages/core/src/RenderingEngine/Viewport.ts
arul-trenser: Added stretch for CPU based rendering in https://github.com/cornerstonejs/cornerstone3D/pull/2443/commits/a30d0e331f7a5d94792fa946f1128a80bd19a476

arul-trenser:
@wayfarer3130
I have addressed all the review comments.

The implementation has been updated to use canvas-space stretching, ensuring that the stretch is applied relative to the canvas axes regardless of the MPR view rotation. This maintains consistent stretching behavior even when the image is rotated.

Could you please re-review the PR and share your feedback?

sen-trenser:
@wayfarer3130 Could you please take a look at this PR and provide your feedbacks?

wayfarer3130 (changes requested):
It is looking close - I added the ability to test with CPU rendering, and to have different viewport types, with the crosshairs bound to ctrl+click to allow orientation changes.

I am still noticing the changing the aspect ratio does not fit to viewport.  I suspect you need a reset camera operation that has a don't change aspect ratio and that it should get called when you change the aspect ratio.  Then, the fit to viewport needs to take into account the updated aspect ratio size.

sen-trenser:
@wayfarer3130  
> I am still noticing the changing the aspect ratio does not fit to viewport.

> Then, the fit to viewport needs to take into account the updated aspect ratio size.

Keeping the image fitted to the viewport after applying aspect ratio changes would cause the image to shrink rather than stretch. The intended behavior of the stretch feature is to keep the image centered and expand it outward along the selected axis while preserving the original size of the non-stretched axis. This allows users to view and annotate details more accurately in parallel to the stretched direction.

Could you share the use case for refitting the image to the viewport after stretching? Understanding this use case will help us assess whether supporting this behavior as the default would be beneficial to end users.

Thanks!

cc: @kgoettler

arul-trenser:
> changing the aspect ratio does not fit to viewport.

Fixed the above scenario  by adding a configuration option to control whether the viewport should fit after applying the stretch. The current default behavior is to fit the viewport after stretch.

@wayfarer3130 Could you please take a look at this PR and provide your feedbacks?

wayfarer3130:
@arul-trenser - could you fix the OHIF downstream tests?  This should be passing for those.

wayfarer3130:
> @wayfarer3130
> 
> > I am still noticing the changing the aspect ratio does not fit to viewport.
> 
> > Then, the fit to viewport needs to take into account the updated aspect ratio size.
> 
> Keeping the image fitted to the viewport after applying aspect ratio changes would cause the image to shrink rather than stretch. The intended behavior of the stretch feature is to keep the image centered and expand it outward along the selected axis while preserving the original size of the non-stretched axis. This allows users to view and annotate details more accurately in parallel to the stretched direction.
> 
> Could you share the use case for refitting the image to the viewport after stretching? Understanding this use case will help us assess whether supporting this behavior as the default would be beneficial to end users.
> 
> Thanks!
> 
> cc: @kgoettler

The point is to be able to visually see fractures or other features that are narrow in the direction of stretch and wide orthogonal  For example, rib fractures.   If the zoom factor is reset so that the image stays in view, then the user can then use the zoom and pan tools to go through the entire image to look for such features.    If they have adjusted it and don't initially see the full image, it is possible that a small portion of the image will not be visible AND will not be noticed to be offscreen.  So, in terms of safety I would prefer to start with displaying the full image after changing the aspect ratio.
If one really wants a single-click tool, then probably the best option would be a way to draw a line that automatically orients itself up-down on the screen or left/right depending on the original orientation, AND stretches the line to 2:1 or something like that.  That way the clinician would know exactly what they are seeing, and if they wanted a full-image view, then the line would start out full-range.  For now, I'd like the scale to fit to work to prevent issues.

wayfarer3130 (approved):
Thank you for sticking with this - it has taken a while to get this through. 
Will just waiting for all the tests to complete - I updated to origin/main

arul-trenser:
@wayfarer3130  Hi, I faced an issue where the scroll wasn't working properly with the 'stretch + fit to viewport' setting. I have fixed the scrolling issue in commit [7ad379c](https://github.com/cornerstonejs/cornerstone3D/pull/2443/commits/7ad379c518286be0270091ac543224fa355c8e7b). Could you please review the changes? Apologies for missing this earlier. 

Thanks in advance!

Issue #2225, opened 2025-07-23
[Bug] Unable to save Segmentation in Multi-frame data


Devu-trenser:
### Describe the Bug

**Issue 1: Overlapping segment issues:**

- The saved SEG file displays a warning:  “Unsupported overlapping segments detected, segmentation rendering results may be incorrect.”
- Segmentations are not rendered correctly in the viewport.
- For example, segments drawn across different slices appear on the same slice.
- Clicking on an ROI causes the segment in the viewport to disappear.

**Issue 2: Not Handling 8-bit Data for Multiframe Support**

The segmentation pipeline is currently designed for 16-bit. Not supporting for 8 bit multi frame data

### Steps to Reproduce

- Load a multi-frame data  in viewer https://viewer-dev.ohif.org/segmentation?StudyInstanceUIDs=1.2.276.0.7230010.3.1.2.447481088.1.1669202398.851612
- Draw segmentation
- Export the segmentation using viewer. (**Note:** after temporarily fixing ImageOrientationPatient undefined issue )
- Observe that all segments appear on a single slice, causing overlap.

### The current behavior

Unable to correctly export segmentation in the multi frame data from viewer

### The expected behavior

 Segmentation should be correctly exported for multi-frame data.

### OS

Ubuntu 24.04

### Node version

22.16.0

### Browser

Chrome 137.0.0

Devu-trenser:
**Root cause:**  
The issue 1 occurs during offset calculation using the following code: byteOffset = sliceLength * imageIdIndex * TypedArrayConstructor.BYTES_PER_ELEMENT, ([code](https://github.com/cornerstonejs/cornerstone3D/blob/main/packages/adapters/src/adapters/Cornerstone/Segmentation_4X.js#L1214))
In this case, the same offset is being computed for all segments drawn on different slices. The root cause is that imageIdIndex is derived using imageId ([code link](https://github.com/cornerstonejs/cornerstone3D/blob/main/packages/adapters/src/adapters/Cornerstone/Segmentation_4X.js#L1182)), which is incorrectly determined via sopUIDImageIdIndexMap [(reference code)](https://github.com/cornerstonejs/cornerstone3D/blob/main/packages/adapters/src/adapters/Cornerstone/Segmentation_4X.js#L348). Since the SOP Instance UID is the same for all image frames, the logic retrieves only the last frame's imageId. As a result, all segments are written to the same location in the labelmap (the last frame), leading to overlapping segment data.

[...]

PR #2445, opened 2025-11-04
fix: Save RTSS from contour segmentation and SEG multiframe/singleframe volume save and RTSS save

wayfarer3130:
### Context

Allow saving contour segmentations to RTSS
Allows saving segmentations on non-volumetric data as well as on multi-frame data correctly.
Avoids verify errors caused by using incorrect header data for study/series references.

Also fixes:
[#2225]

### Changes & Results

- Add Annotations based RTSS to Normalized DICOM converter
  - Passes DCIODVFY tests except for ReferencedSeriesSequence, which I believe DCIODVFY is wrong about.
  - Creates correct single frame, multiframe and volume based RTSS objects
  - Works against CS3D metadata providers as well as OHIF metadata provider
- Add segmentation creation on non-volumetric data
  -  Uses the new metadata provider framework so this works now against CS3D and OHIF providers
  - Correct use of referenced frame to get frame specific references
  - Works around a number of bugs in dcmjs (TODO - fix the dcmjs bugs too)
  - Fixes segmentation generation on 8 bit or 1 bit data

### Known Bugs
- Does not fix any issues with contour/segmentations on time based data, nor does it allow saving of any time specific annotations
- Reloading a contour segmentation will not show the segmentation filled

### Testing

Find some studies that are:
1. Volume based 
2. Volumetric type modality, but single frame  (eg Juno single frame series)
3. Non-volumetric but stack based (eg CR/DX type series)
4. Multi-frame volumetric studies

Work against a back end permitting saving (eg Orthanc or Static DICOMweb local)

Create a contour annotation on each of 1-4, saving each one
Rename the segmentation name and segment labels
Change the colour of the segment

Ensure that the contour annotation creates a new RTStruct for each series saved, and that this can be reloaded 

Create a labelmap segmentation on each of 1-4, saving each one
Rename/colour as per contour segmentations

Ensure that the SEG data can be reloaded and is correclty labelled/coloured AND that the segmentations save apply to the correct image.

See known bugs above

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  


package.json
```diff
@@ -40,6 +40,7 @@
     "docs:watch": "npx lerna run docs:watch",
     "install:update-lockfile": "yarn install && bun install --config=./bunfig.update-lockfile.toml",
     "install:yarn:frozen-lockfile": "yarn install --frozen-lockfile",
+    "install:frozen": "yarn install --frozen-lockfile",
```
wayfarer3130, line 43: We actually don't want to embed the fact htat it is yarn inside our build script, so adding a install:frozen to remove that exposure of what we are doing internally.

packages/adapters/.prettierrc
wayfarer3130, line 1: Switching to the standard prettier code format for adapters.

packages/adapters/src/adapters/Cornerstone3D/RTStruct/utilities/getReferencedSeriesSequence.ts
wayfarer3130, line 1: This will generate multi-series/multi-instance references for exactly the set of frames which are referenced.

packages/adapters/src/adapters/Cornerstone3D/Segmentation/generateSegmentation.ts
wayfarer3130, line 1: Use the new metadata modules which correctly reference frames in multiframes, and has a bunch of work arounds for bugs in dcmjs.
Intent is to ALSO resolve the dcmjs bugs to avoid needing these work arounds.

packages/adapters/src/adapters/Cornerstone3D/Segmentation/generateSegmentation.ts
```diff
@@ -1,22 +1,46 @@
...
+  metadata,
+  options: Options = {}
+) {
+  const segmentation = _createMultiframeSegmentationFromReferencedImages(
+    images,
+    metadata,
+    options
+  );
+  const segmentationResult = fillSegmentation(segmentation, labelmaps, options);
+  const predecessorImageId = options?.predecessorImageId;
```
wayfarer3130, line 35: This provides a correct replacement reference when one segmentation replaces another.

packages/adapters/src/adapters/Cornerstone3D/Segmentation/generateSegmentation.ts
```diff
@@ -30,36 +54,66 @@ function generateSegmentation(images, labelmaps, metadata, options = {}) {
...
-            SOPClassUID: instance.SopClassUID || instance.SOPClassUID,
-            SOPInstanceUID: instance.SopInstanceUID || instance.SOPInstanceUID,
-            PixelData: image.voxelManager.getScalarData(),
-            _vrMap: {
-                PixelData: "OW"
-            },
-            _meta: {}
-        };
-    });
+  const studyData = metadata.get(MetadataModules.STUDY_DATA, images[0].imageId);
```
wayfarer3130: Getting the study data this way prevents leaking of invalid tag values from the referenced images into the segmentation object and resolves a number of issues overall around validity of the new data.

packages/adapters/src/adapters/Cornerstone3D/Segmentation/generateSegmentation.ts
```diff
@@ -30,36 +54,66 @@ function generateSegmentation(images, labelmaps, metadata, options = {}) {
...
+  const datasets = images.map((image) => {
+    const { imageId } = image;
+    const seriesData = metadata.get(MetadataModules.SERIES_DATA, imageId);
+    const imageData = metadata.get(MetadataModules.IMAGE_DATA, imageId);
+    return {
```
wayfarer3130, line 67: This type of image reference can be created against any of the metadata providers, allowing creation of segmentation against the wado uri and dicom part 10 providers as well.

packages/adapters/src/adapters/Cornerstone3D/Segmentation/generateSegmentation.ts
```diff
@@ -30,36 +54,66 @@ function generateSegmentation(images, labelmaps, metadata, options = {}) {
...
+      ...studyData,
+      ...seriesData,
+      ...imageData,
+      PixelData: image.voxelManager.getScalarData(),
+      // Declaring it as 16 bits allows the normalizer to work on 8 bit data
+      BitsAllocated: 16,
```
wayfarer3130, line 73: Fixes the 8 bit data reference issue.

packages/adapters/src/adapters/Cornerstone3D/Segmentation/generateSegmentation.ts
```diff
@@ -30,36 +54,66 @@ function generateSegmentation(images, labelmaps, metadata, options = {}) {
...
-        );
-    }
+  if (!multiframe) {
+    throw new Error(
+      'Failed to normalize the multiframe dataset, the data is not multi-frame.'
+    );
+  }
 
-    return new SegmentationDerivation([multiframe], options);
+  multiframe.SharedFunctionalGroupsSequence ||= {};
```
wayfarer3130, line 102: Bunch of bugs caused by single frame data not being normalizeable by dcmjs, so these items fix the response.

packages/adapters/src/adapters/Cornerstone3D/Segmentation/generateSegmentation.ts
```diff
@@ -30,36 +54,66 @@ function generateSegmentation(images, labelmaps, metadata, options = {}) {
...
+      PlaneOrientationSequence: {},
+    };
+  }
+  if (isSingleNonMultiFrame) {
+    multiframe.PerFrameFunctionalGroupsSequence = [
+      multiframe.PerFrameFunctionalGroupsSequence[0],
+    ];
+    multiframe.NumberOfFrames = 1;
+  }
+  return new SegmentationDerivation([multiframe], options);
```
wayfarer3130, line 117: Using segmentation derivation prevents errors when segmentations exist on different SOP classes.

packages/adapters/src/utilities/referencedMetadataProvider.ts
wayfarer3130, line 1: A good bit of the change to make both Segmentation and RTSS work reliably on both single and multi-frame data is included in this file, as it allows consistently getting different types of references/frame data for different parts of the destination object.

packages/core/src/metaData.ts
wayfarer3130, line 1: Just moving shared utilities here.

packages/tools/src/tools/base/ContourBaseTool.ts
```diff
@@ -30,6 +29,30 @@ abstract class ContourBaseTool extends AnnotationTool {
...
+
+    const ContourData = [];
+    for (const point of data.contour.polyline) {
+      for (const v of point) {
+        ContourData.push(v.toFixed(2));
+      }
+    }
+
+    const { referencedImageId } = toolData.metadata;
+    const ContourImageSequence = metadataProvider.get(
```
wayfarer3130, line 43: Here is an example usage of the referenced metadata provider - doing it this way makes it consistent across multiple types while still being customizeable.

package.json
sedghi: we already have a commad above this line doing the same thing, maybe rename that then?

packages/core/src/enums/MetadataModules.ts
```diff
@@ -36,6 +42,42 @@ enum MetadataModules {
...
+   */
+
+  /** References a given frame*/
+  IMAGE_SOP_INSTANCE_REFERENCE = 'ImageSopInstanceReference',
+  REFERENCED_SERIES_REFERENCE = 'ReferencedSeriesReference',
+  PREDECESSOR_SEQUENCE = 'PredecessorSequence',
+
+  STUDY_DATA = 'StudyData',
+  SERIES_DATA = 'SeriesData',
+  IMAGE_DATA = 'ImageData',
```
sedghi, line 109: please add documentation on each of them what they are actually

packages/core/src/metaData.ts
```diff
@@ -84,4 +84,64 @@ function getMetaData(type: string, ...queries): any {
...
+ */
+export const lowerTag = (tag: string) => {
+  if (tag.startsWith('SOP')) {
+    return `sop${tag.substring(3)}`;
+  }
+  if (tag.endsWith('ID')) {
+    tag = `${tag.substring(0, tag.length - 2)}Id`;
+  }
+  return tag.charAt(0).toLowerCase() + tag.slice(1);
+};
```
sedghi, line 145: maybe move to a utility, not this file

package.json
wayfarer3130: The issue with that one is it is used elsewhere right now as part of hte build/test, and it contains too much detail about how it is implemented in that it is yarn specific.
I wanted one that would replace the one above it, and the one above would eventually go away in a few months.

packages/core/src/metaData.ts
```diff
@@ -84,4 +84,64 @@ function getMetaData(type: string, ...queries): any {
...
+        }
+        Object.assign(result, capitalizedData);
+      }
+    } catch (error) {
+      console.error(`Error retrieving ${t} data:`, error);
+    }
+  }
+
+  return result;
+}
```
sedghi, line 119: can we just use the same as above with query of 'instance' ? so we use the same `metadata.get(imageId, 'instance')` , imagine in the function above getMeatdata (which we already have), we just add a check that is if (query==='instance') , then we don't change our public/one way of getting metadata, and _getInstanceModule remains as internal util

packages/tools/src/tools/base/ContourBaseTool.ts
```diff
@@ -30,6 +29,30 @@ abstract class ContourBaseTool extends AnnotationTool {
     super(toolProps, defaultToolProps);
   }
 
+  static getContourSequence(toolData, metadataProvider) {
+    const { data } = toolData;
+
+    const ContourData = [];
+    for (const point of data.contour.polyline) {
```
sedghi, line 36: what if there is no contour.polyline on data, are you sure all of our contors have polyline? don't think so, can you check spline and livewire?

packages/tools/src/tools/base/ContourBaseTool.ts
```diff
@@ -30,6 +29,30 @@ abstract class ContourBaseTool extends AnnotationTool {
...
+    const { referencedImageId } = toolData.metadata;
+    const ContourImageSequence = metadataProvider.get(
+      'ImageSopInstanceReference',
+      referencedImageId
+    );
+
+    return {
+      NumberOfContourPoints: ContourData.length / 3,
+      ContourImageSequence,
+      ContourGeometricType: 'CLOSED_PLANAR',
```
sedghi, line 51: Why are we forcing it to be closed planar when we know a contour in `ContourBaseTool` can be open? Maybe we should rename the function to indicate it's specifically for `closed_planar` contours.

packages/adapters/.prettierrc
sedghi: Thanks

packages/core/src/metaData.ts
wayfarer3130: Can I rename it to something like getNormalGroup
The point is that it allows getting different sets of normalized data consistently without having to have getInstance, which doesn't really work very reliably.

packages/core/src/metaData.ts
wayfarer3130: It is a metaData utility, so it belongs on the metaData instance.  I specifically want it shared as part of the normal metaData handling so that we don't have multiple versions of this code.

packages/adapters/src/adapters/helpers/copySeriesTags.ts
```diff
@@ -1,28 +1,27 @@
...
-    };
-    for (const tagKey of seriesTags) {
-        const value = src[tagKey];
-        if (value === undefined) {
-            continue;
-        }
-        study[tagKey] = value;
+export function copySeriesTags(src): Record<string, unknown> {
+  const result = {};
+  for (const tagKey of seriesTags) {
```
sedghi, line 21: kind of agree with the copilot comment

packages/adapters/src/utilities/InstanceTypes.ts
```diff
@@ -0,0 +1,42 @@
...
+  ROIContourSequence: NormalModule[];
+  RTROIObservationsSequence: NormalModule[];
+  ReferencedFrameOfReferenceSequence: NormalModule[];
+  Modality: 'RTSTRUCT';
+  PositionReferenceIndicator: string;
+  StructureSetLabel: string;
+  StructureSetName: string;
+  StructureSetDate: string;
+  StructureSetTime: string;
+};
```
sedghi: The file inconsistently uses `interface` and `type`. Could you please choose one to use throughout?

packages/adapters/src/utilities/referencedMetadataProvider.ts
```diff
@@ -0,0 +1,170 @@
...
+      StructureSetDate: DicomMetaDictionary.date(),
+      StructureSetTime: DicomMetaDictionary.time(),
+    };
+  },
+
+  [MetadataModules.NEW_INSTANCE_DATA]: (imageId) => {
+    const studyData = metaData.get(MetadataModules.STUDY_DATA, imageId);
+    return {
+      ...studyData,
+      SeriesNumber: '50000',
```
sedghi, line 166: seems arbitrary number, no? can you document this on top of this file as CONST

packages/adapters/src/utilities/referencedMetadataProvider.ts
```diff
@@ -0,0 +1,170 @@
...
+    const frameModule = metaData.get(MetadataModules.FRAME_MODULE, imageId);
+    const { sopClassUID, sopInstanceUID, frameNumber, numberOfFrames } =
+      frameModule;
+    if (numberOfFrames > 1) {
+      return {
+        ReferencedSOPClassUID: sopClassUID,
+        ReferencedSOPInstanceUID: sopInstanceUID,
+        ReferencedFrameNumber: frameNumber,
+      };
+    }
```
sedghi, line 63: Why can't we add these metadata module providers to wadors and wadouri? Having them inside feels a bit odd.

packages/core/src/metaData.ts
wayfarer3130: Renamed to combineNormalizeModules so it describes what it is actually doing.

packages/tools/src/tools/base/ContourBaseTool.ts
wayfarer3130: Spline and livewire both have the polyline, just that it gets regenerated.  However, I haven't registered either spline or livewire, nor do I handle holes yet.  Don't want to go too far in the first pass.

packages/tools/src/tools/base/ContourBaseTool.ts
wayfarer3130: This is only called for segmentations, which are always closed.  There is a ton more work to add open contours to RTSS and other types, and I'm not even sure we should be saving them that way.  The CLOSED_PLANAR will get replaced with something else once we add hole saving too.

packages/adapters/src/adapters/helpers/copySeriesTags.ts
wayfarer3130: Not worth doing - the copySeriesTags is going away entirely.

packages/adapters/src/utilities/InstanceTypes.ts
wayfarer3130: No, it specifically uses interface and type distinctly to get a desired behaviour.
interface can be extended externally, whereas type can be union/intersection.  Unfortunately, they didn't allow both in a single declaration so it ends up using both items.

packages/adapters/src/utilities/referencedMetadataProvider.ts
wayfarer3130: It is arbitrary, added a comment indicating it is for replacement.

packages/adapters/src/utilities/referencedMetadataProvider.ts
wayfarer3130: The point is that they depend on OTHER metadata modules, so they work the same way against:
OHIF, Wado URI, WADO RS, JSON metadata.
They absolutely can't go in the wadors and wadouri modules since they are intended to add onto the existing behaviour.  That fundamentally fixes bugs in creating new instances because it means we don't have to create new modules and then forget about doing it in one of the places.
It SHOULD work against fakeMetadataProvider as well, allowing us to actually test some of this against the fake metadata instances.

packages/adapters/src/utilities/InstanceTypes.ts
wayfarer3130: I differentiate the instance/single data types with the I prefix specifically so that external parties can extend the interface externally.

packages/adapters/src/utilities/referencedMetadataProvider.ts
wayfarer3130: This is the plan for how to fix the per-frame metadata modules and the calibration/spacing information as well - I will add another metadata module provider that knows how to deal with those types of operations as a combined provider.  Then, for instance normalized data it will just work without having to have the combine instance handling that breaks so many things in CS3D.

packages/adapters/src/utilities/InstanceTypes.ts
wayfarer3130: Hmm, I'm just going to make everything optional in the NormalModule, but declared to the right type, and get rid of all the itnerfaces.

packages/core/src/metaData.ts
```diff
@@ -84,4 +84,64 @@ function getMetaData(type: string, ...queries): any {
...
+  if (tag.endsWith('Id')) {
+    tag = `${tag.substring(0, tag.length - 2)}ID`;
+  }
+  return tag.charAt(0).toUpperCase() + tag.slice(1);
+};
+
+/**
+ * Converts a tag name to lowerCamelCase
+ */
+export const lowerTag = (tag: string) => {
```
sedghi: this is not lower really, it is rather uncapitalize tag, since it does not lower all chars.

How about we call them, toUpperCamelTag, toLowerCamelTag

packages/tools/src/tools/displayTools/Contour/contourDisplay.ts
```diff
@@ -213,20 +213,20 @@ function _checkContourGeometryMatchViewport(
...
-  // get the normal from two points
-  let normal = vec3.cross(
-    vec3.create(),
-    vec3.sub(vec3.create(), point2, point1),
-    vec3.sub(vec3.create(), point3, point1)
-  );
-  normal = vec3.normalize(vec3.create(), normal);
-  const dotProduct = vec3.dot(normal, viewportNormal);
-  return Math.abs(dotProduct) > 0.9;
+  const { points } = contours[0];
```
wayfarer3130, line 216: The old code was invalid if the first three points happened to be colinear.  The new version is much closer to correct is it would require 25 colinear points (or all the points for less than 25 points) to fail to be orthogonal with respect to point 0.
It also works for 0,1 and 2 point contours.

packages/core/src/utilities/getPixelSpacingInformation.ts
```diff
@@ -41,9 +43,14 @@ function calculateRadiographicPixelSpacing(instance) {
 
   if (!PixelSpacing) {
     if (!EstimatedRadiographicMagnificationFactor) {
-      console.warn(
-        'EstimatedRadiographicMagnificationFactor was not present. Unable to correct ImagerPixelSpacing.'
-      );
+      if (!alreadySeenWarn.has(instance.SeriesInstanceUID)) {
+        console.warn(
```
wayfarer3130, line 47: Not part of this change really, just that I wasn't able to debug one particular study issue because of a wall of messages.

sedghi (approved):

PR #2446, opened 2025-11-05
fix(segmentation): Re-adding a segmentation representation with contour annotations was not rendering the contour annotations; OHI-2287 and OHI-2288

jbocce:
### Context
If a segmentation representation with contour annotations was removed from a viewport (but was NOT deleted) and then later re-added to a viewport the contour annotations were NOT rendered.

### Changes & Results
Do not remove contour annotations from the annotation state when a segmentation representation is removed.
Only remove the contour annotations from the annotation state when the segmentation itself is removed. When rendering contour annotations, filter out the annotations that are not part of any segmentation representation for the viewport being rendered.

### Testing
Remove segmentation representations containing contour annotations from a viewport. Then re-add the same segmentation representation. The contour annotations should render.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] "OS: Windows 11
- [x] "Node version: 23.9.0
- [x] "Browser: Chrome 142.0.7444.60
  

### DIFF
--- a/packages/tools/src/tools/base/ContourSegmentationBaseTool.ts
+++ b/packages/tools/src/tools/base/ContourSegmentationBaseTool.ts
@@ -228,8 +228,10 @@ abstract class ContourSegmentationBaseTool extends ContourBaseTool {
       const segmentationId = (annotation as ContourSegmentationAnnotation)?.data
         ?.segmentation?.segmentationId;
 
+      // If there is no segmentationId, then we don't now how to filter it
+      // here so we return true to keep it.
       if (!segmentationId) {
-        return false;
+        return true;
       }
 
       return !!defaultSegmentationStateManager.getSegmentationRepresentation(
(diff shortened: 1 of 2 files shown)


packages/tools/src/init.ts
```diff
@@ -183,6 +189,11 @@ function _removeCornerstoneToolsEventListeners() {
     annotationSelectionListener
   );
 
+  eventTarget.removeEventListener(
```
jbocce, line 192: This event was previously added but not removed.

packages/tools/src/tools/displayTools/Contour/contourDisplay.ts
```diff
@@ -40,8 +40,6 @@ function removeRepresentation(
 
   const { viewport } = enabledElement;
 
-  removeContourFromElement(viewportId, segmentationId);
```
jbocce, line 43: No longer do we removed the contour (annotations) from the viewport element when the representation is removed. They are only removed when the segmentation itself is removed.

packages/tools/src/tools/annotation/PlanarFreehandROITool.ts
```diff
@@ -441,6 +441,12 @@ class PlanarFreehandROITool extends ContourSegmentationBaseTool {
       return;
     }
 
+    const baseFilteredAnnotations =
```
jbocce, line 444: Invoke the base filter first to remove any contour annotations that do NOT apply to this viewport.

PR #2450, opened 2025-11-05
Fix Segment Opacity Update When Adjusting Fill in Segmentation Styles

### LINKED ISSUE #4977: [Bug] Segmentation Fill is not working after loading seg file and clicking on the segments - OSS 3.10
### Describe the Bug

When changing the segmentation fill after clicking on segments, the opacity of the clicked segments does not change

https://github.com/user-attachments/assets/e1f78d45-2bb2-415e-bd48-46ada088a7b2

### Steps to Reproduce

1. Launch the study that contains segmentation.
2. Drag and drop the segmentation.
3. Click on some segments.
4. Adjust the Fill in the segmentation settings.
5. Observe that the opacity of the clicked segment does not change.

### The current behavior

When changing the segmentation fill after clicking on segments, the opacity of the clicked segments does not change

### The expected behavior

The opacity of all segments in the active segmentation must change when the opacity is adjusted.

### OS

Windows 10

### Node version

20.9.0

### Browser

Chrome 135.0.0

Devu-trenser:
### Context
This PR addresses the issue where, after clicking on segments and adjusting the segmentation's fill, the opacity of the clicked segments doesn't update as expected. The root cause was that the fillAlpha value was being applied at the per-segment level during highlighting, but the corresponding per-segment entry in the config was not being cleared afterward. As a result, the global-level opacity (fillAlpha) setting was not taking effect correctly.
 
Fixes: https://github.com/OHIF/Viewers/issues/4977

Dependent PR : https://github.com/OHIF/Viewers/pull/4978

### Changes & Results

- Added a new style restoration mechanism for segmentation configurations.
- Implemented createRestoreFunction to capture and restore style values for specific keys.
- Added internal helpers:
      - `_removeStyleFromHigherLevels` — Removes higher-level overrides before restoration.
      - `_applyStyleValueAtLevel` — Reapplies captured style values at their original configuration level.
      - `_getCurrentValueForRestore` — Determines current style value and its hierarchy level.
      - `_getViewportLevelValue, _getSegmentationLevelValue, _getGlobalLevelValue` — Level-specific lookup helpers.

Before

- Clicking on a segment applied highlight at the per-segment level, but opacity was applied at the global level.
- This caused mismatches where opacity updates were not reflected correctly after interaction.

After

- Highlight and opacity changes now behave consistently across interactions.
- Temporary style overrides (e.g., during highlight) are cleaned up when restored.

### Testing

1. Locally link the OHIF `abhijith-trenser:fix/segmentation-fill-not-working-after-loading-seg-and-click-segments` branch with the cornerstone3D branch `Devu-trenser:fix/seg-fill-not-working-after-segment-click`.
2. Launch the study that contains segmentation and load it or draw new segmentation.
3. Click on some segments.
4. Adjust the Fill in the segmentation settings.
5. Verify that the opacity of the clicked segment is correctly changing.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] "OS: Linux 6.11 Ubuntu 24.10 24.10 (Oracular Oriole)
- [x] "Node version: 22.16.0
- [x] "Browser: Chrome: 139.0.7258.154

### DIFF
--- a/packages/tools/src/stateManagement/segmentation/SegmentationStyle.ts
+++ b/packages/tools/src/stateManagement/segmentation/SegmentationStyle.ts
@@ -12,7 +12,30 @@ import { utilities } from '@cornerstonejs/core';
 
 export type RepresentationStyle = LabelmapStyle | ContourStyle | SurfaceStyle;
 
+export type RepresentationStyleKeys =
+  | keyof LabelmapStyle
+  | keyof ContourStyle
+  | keyof SurfaceStyle;
+
 export type BaseRepresentationStyle = BaseLabelmapStyle | BaseContourStyle;
+
+type StyleSpecifier = {
+  type: SegmentationRepresentations;
+  segmentIndex?: number;
+  segmentationId?: string;
+  viewportId?: string;
+};
+
+enum ConfigLevel {
+  Default = 0,
+  Global = 1,
+  Segmentation = 2,
+  Viewport = 4,
+}
+interface RestoreEntry {
+  value: unknown;
+  level: ConfigLevel;
+}
 
 interface SegmentationStyleConfig {
   global: {
@@ -44,6 +67,8 @@ interface SegmentationStyleConfig {
     };
   };
 }
+
+const ALL_SEGMENTATIONS_KEY = '__allSegmentations__';
 
 /**
  * This class handles the configuration of segmentation styles. It supports
@@ -84,15 +109,7 @@ class SegmentationStyle {
    * @param specifier.segmentIndex - Optional. The index of the specific segment to style.
    * @param styles - The styles to set.
    */
-  setStyle(
-    specifier: {
-      type: SegmentationRepresentations;
-      viewportId?: string;
-      segmentationId?: string;
-      segmentIndex?: number;
-    },
-    styles: RepresentationStyle
-  ): void {
+  setStyle(specifier: StyleSpecifier, styles: RepresentationStyle): void {
     const { viewportId, segmentationId, type, segmentIndex } = specifier;
 
     const currentStyles = this.getStyle(specifier);
@@ -155,7 +172,6 @@ class SegmentationStyle {
         }
       } else {
         // Viewport-specific style for all segmentations of a type
-        const ALL_SEGMENTATIONS_KEY = '__allSegmentations__';
         if (!representations[ALL_SEGMENTATIONS_KEY]) {
           representations[ALL_SEGMENTATIONS_KEY] = {};
         }
@@ -236,12 +252,7 @@ class SegmentationStyle {
    * @param specifier.segmentIndex - Optional. The index of the specific segment.
    * @returns An object containing the combined style and renderInactiveSegmentations flag for the viewport.
    */
-  getStyle(specifier: {
-    viewportId?: string;
-    segmentationId?: string;
-    type?: SegmentationRepresentations;
-    segmentIndex?: number;
-  }): RepresentationStyle {
+  getStyle(specifier: StyleSpecifier): RepresentationStyle {
     const { viewportId, segmentationId, type, segmentIndex } = specifier;
     let combinedStyle = this.getDefaultStyle(type);
     let renderInactiveSegmentations = false;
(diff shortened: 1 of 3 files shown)


Devu-trenser:
@jbocce Could you please review this when you have a moment?

jbocce (changes requested):
Thank you for this PR. To be honest, this is a bit much for something that can likely be solved  by introducing a flag to the `SegmentationStyle.setStyle` method that specifies if the style being set should do a hard replace or a merge with the existing style. The current `SegmentationStyle.setStyle` merges the style to set with what is existing, so the replace would be the new functionality. Then the change in OHIF would simply have to add the `replace` flag on [this line](https://github.com/OHIF/Viewers/blob/5c44fe06dc7a9c12769841edc762cba537022b89/extensions/cornerstone/src/services/SegmentationService/SegmentationService.ts#L1734).

@sedghi what are your thoughts?

Take as an example the following diff...

```
@@ -90,28 +90,31 @@ class SegmentationStyle {
       viewportId?: string;
       segmentationId?: string;
       segmentIndex?: number;
+      replace?: boolean;
     },
     styles: RepresentationStyle
   ): void {
-    const { viewportId, segmentationId, type, segmentIndex } = specifier;
+    const {
+      viewportId,
+      segmentationId,
+      type,
+      segmentIndex,
+      replace = false,
+    } = specifier;

     const currentStyles = this.getStyle(specifier);

+    const mergedStyles = replace ? styles : { ...currentStyles, ...styles };
+
     let updatedStyles: RepresentationStyle;

     if (!viewportId && !segmentationId) {
       // Global style setting
-      updatedStyles = {
-        ...currentStyles,
-        ...styles,
-      };
+      updatedStyles = mergedStyles;
     } else {
       // Per segmentation or per viewport style setting
       updatedStyles = this.copyActiveToInactiveIfNotProvided(
-        {
-          ...currentStyles,
-          ...styles,
-        },
+        mergedStyles,
         type
       );
     }
```

jbocce:
Regarding https://github.com/cornerstonejs/cornerstone3D/pull/2450#pullrequestreview-3449089911, perhaps a better name for `replace` is `mergeAndSet` with a default value of `true` for the current functionality and `false` when someone does NOT want to merge from the hierarchy but simply wants to set the set style with the given specifier.  Later when someone uses `getStyle`, it does the merging of the hierarchy. Actually this is quite ideal in my opinion.

So for the case you are trying to solve, OHIF would pass a `mergeAndSet` value of `false`.

Devu-trenser:
@jbocce Thank you for the suggestion,  adding a replace flag is a great idea and would certainly work in the current scenario.
 However, there are a few cases where it might not fully work, which is why we introduced the restore function. Below listed the cons of this approach:

- Applying a hard replace could cause issues if per-segment style configurations are introduced in the future (I’m not sure if per-segment styles are currently used anywhere other than in the highlight animation), as it would clear those settings once the highlight operation completes.


- Capturing and reapplying styles before and after the highlight isn’t currently feasible from OHIF, since the getStyle() method returns a merged view of all configuration levels (viewport, segmentation, and global). This makes it difficult to identify which level each value originally came from.


The restore function captures both the effective value and its corresponding configuration level, allowing us to safely revert temporary style changes without affecting other styles or configuration hierarchies.


I’d love to hear your thoughts on whether it might make sense to restore the original styles and configuration levels after the highlight animation completes.

jbocce:
@Devu-trenser , thank you for the feedback. We discussed this PR internally and we think that the best approach is to add a `removeStyle` method to `SegmentationStyle`. `removeStyle` would NEVER merge, but would simply remove the style entry at the specified level. This we feel best aligns with the principles of Cornerstone3D where we want to focus on general solutions and not necessarily ones that are particular to OHIF. We appreciate the solution proposed by this PR, but it is too specific to segment highlighting in OHIF and goes beyond what is needed in CS3D.

FYI @sedghi

jbocce:
@Devu-trenser , thank you for the feedback. We discussed this PR internally and we think that the best approach is to add a boolean parameter to SegmentationStyle.setStyle that determines whether to merge the styles before setting. This we feel best aligns with the principles of Cornerstone3D where we want to focus on general solutions and not necessarily ones that are particular to OHIF. We appreciate the solution proposed by this PR, but it is too specific to segment highlighting in OHIF and goes beyond what is currently needed in CS3D.

Consider simply naming the parameter `merge`. I checked with AI and they thought it was the best name for the parameter. By default for backwards compatibility, the default value for `merge` should be `true`.

FYI @sedghi

Devu-trenser:
@jbocce  Thanks for the feedback! I’ve updated the code to add the merge parameter to SegmentationStyle.setStyle, with a default value of true for backward compatibility.

packages/tools/src/stateManagement/segmentation/SegmentationStyle.ts
```diff
@@ -90,28 +91,30 @@ class SegmentationStyle {
       viewportId?: string;
       segmentationId?: string;
       segmentIndex?: number;
+      merge?: boolean;
```
jbocce: Instead of part of the specifier, merge should be another, separate parameter after the styles parameter.

jbocce (changes requested):
One small change.

jbocce (approved):

packages/tools/src/stateManagement/segmentation/SegmentationStyle.ts
Devu-trenser: Thank you, I have updated it.

PR #2453, opened 2025-11-07
fix(tools): Prevent infinite update loop in SplineROITool

olhalit:
### Context

When adding or editing a contour segmentation using any of the SplineROI based tools (e.g., CatmullRomSplineROI, LinearSplineROI, BSplineROI), an infinite loop of SEGMENTATION_DATA_MODIFIED events was triggered. 

This issue was caused within the renderAnnotationInstance method of the SplineROITool. During every render pass, the tool would recalculate the spline's polyline and call updateContourPolyline. This process unconditionally marked the annotation as invalidated, which in turn triggered a new SEGMENTATION_DATA_MODIFIED event and a subsequent re-render, leading to a feedback loop.
### Changes & Results

Modified `SplineROITool.ts`:
* A new private method, _arePolylinesEqual, has been added. This method performs a comparison of two polyline point arrays in canvas space to check if they are geometrically identical within a small tolerance.
* The logic in renderAnnotationInstance has been updated. It now uses _arePolylinesEqual to compare the newly generated spline polyline with the one currently stored in the annotation's data.
* The call to updateContourPolyline is now conditional and only executes if the polyline's geometry has actually changed.

Effect of the change:
* Before: Any render pass involving a SplineROI annotation would unconditionally trigger a data modification event (SEGMENTATION_DATA_MODIFIED), leading to an infinite re-render loop.
* After: The modification event is now correctly triggered only when the spline's control points are moved or its geometry is altered. The infinite loop is resolved, and the application remains responsive during spline manipulation.
### Testing

This fix was tested locally by monitoring the SEGMENTATION_DATA_MODIFIED event to ensure it no longer fires in a loop.
#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] "OS: Ubuntu 24.04.2
- [x] "Node version: v20.19.5
- [x] "Browser: Chrome 139.0.7258.66
  

### DIFF
--- a/packages/tools/src/tools/annotation/SplineROITool.ts
+++ b/packages/tools/src/tools/annotation/SplineROITool.ts
@@ -758,6 +758,25 @@ class SplineROITool extends ContourSegmentationBaseTool {
     element.removeEventListener(Events.TOUCH_TAP, this._mouseDownCallback);
   };
 
+  // Compare polylines in canvas space with a small tolerance
+  private _arePolylinesEqual = (
+    a: Types.Point2[] | undefined,
+    b: Types.Point2[] | undefined,
+    eps = 0.001
+  ) => {
+    if (!a || !b) return false;
+    if (a.length !== b.length) return false;
+    for (let i = 0; i < a.length; i++) {
+      if (
+        Math.abs(a[i][0] - b[i][0]) > eps ||
+        Math.abs(a[i][1] - b[i][1]) > eps
+      ) {
+        return false;
+      }
+    }
+    return true;
+  };
+
   protected isContourSegmentationTool(): boolean {
     // Disable contour segmentation behavior because it shall be activated only
     // for SplineContourSegmentationTool
@@ -816,16 +835,26 @@ class SplineROITool extends ContourSegmentationBaseTool {
       const spline = this._updateSplineInstance(element, annotation);
       const splinePolylineCanvas = spline.getPolylinePoints();
 
-      this.updateContourPolyline(
-        annotation,
-        {
-          points: splinePolylineCanvas,
-          closed: data.contour.closed,
-          targetWindingDirection: ContourWindingDirection.Clockwise,
-        },
-        viewport,
-        { updateWindingDirection: data.contour.closed }
-      );
+      // Compare with existing contour polyline (converted to canvas)
+      const existingPolylineCanvas =
+        annotation.data.contour.polyline?.map(worldToCanvas) || [];
+
+      // Only update/invalidate if geometry actually changed
+      if (
+        !this._arePolylinesEqual(existingPolylineCanvas, splinePolylineCanvas)
+      ) {
+        const isClosed = data.contour.closed;
+        this.updateContourPolyline(
+          annotation,
+          {
+            points: splinePolylineCanvas,
+            closed: isClosed,
+            targetWindingDirection: ContourWindingDirection.Clockwise,
+          },
+          viewport,
+          { updateWindingDirection: isClosed }
+        );
+      }
     });
 
     // Let the base class render the contour


olhalit:
Hi team! 👋
Just wanted to kindly follow up on this PR.

The fix resolves an infinite SEGMENTATION_DATA_MODIFIED loop in SplineROI-based tools and significantly improves app stability during contour editing.

Please let me know if there is anything I can improve or adjust to help move the review forward.

Thanks a lot for your time and for maintaining this great project!

sedghi:
Thanks for the PR, we will look soon

packages/tools/src/tools/annotation/SplineROITool.ts
```diff
@@ -758,6 +758,25 @@ class SplineROITool extends ContourSegmentationBaseTool {
     element.removeEventListener(Events.TOUCH_TAP, this._mouseDownCallback);
   };
 
+  // Compare polylines in canvas space with a small tolerance
+  private _arePolylinesEqual = (
```
wayfarer3130, line 762: This should be a utility in the tools module

packages/tools/src/tools/annotation/SplineROITool.ts
```diff
@@ -758,6 +758,25 @@ class SplineROITool extends ContourSegmentationBaseTool {
...
+    a: Types.Point2[] | undefined,
+    b: Types.Point2[] | undefined,
+    eps = 0.001
+  ) => {
+    if (!a || !b) return false;
+    if (a.length !== b.length) return false;
+    for (let i = 0; i < a.length; i++) {
+      if (
+        Math.abs(a[i][0] - b[i][0]) > eps ||
```
wayfarer3130, line 771: Why not use isEqual(a[i],b[i])?

packages/tools/src/tools/annotation/SplineROITool.ts
```diff
@@ -816,16 +835,26 @@ class SplineROITool extends ContourSegmentationBaseTool {
...
-          points: splinePolylineCanvas,
-          closed: data.contour.closed,
-          targetWindingDirection: ContourWindingDirection.Clockwise,
-        },
-        viewport,
-        { updateWindingDirection: data.contour.closed }
-      );
+      // Compare with existing contour polyline (converted to canvas)
+      const existingPolylineCanvas =
+        annotation.data.contour.polyline?.map(worldToCanvas) || [];
```
wayfarer3130, line 840: You don't need the || [] at the end  - the compare function works without it.

packages/tools/src/tools/annotation/SplineROITool.ts
```diff
@@ -816,16 +835,26 @@ class SplineROITool extends ContourSegmentationBaseTool {
...
+      if (
+        !this._arePolylinesEqual(existingPolylineCanvas, splinePolylineCanvas)
+      ) {
+        const isClosed = data.contour.closed;
+        this.updateContourPolyline(
+          annotation,
+          {
+            points: splinePolylineCanvas,
+            closed: isClosed,
+            targetWindingDirection: ContourWindingDirection.Clockwise,
```
wayfarer3130, line 852: Is it possible that the target winding direction will update/smooth the contour, resulting in infinite loop anyways?

packages/tools/src/tools/annotation/SplineROITool.ts
```diff
@@ -816,16 +835,26 @@ class SplineROITool extends ContourSegmentationBaseTool {
...
+      // Compare with existing contour polyline (converted to canvas)
+      const existingPolylineCanvas =
+        annotation.data.contour.polyline?.map(worldToCanvas) || [];
+
+      // Only update/invalidate if geometry actually changed
+      if (
+        !this._arePolylinesEqual(existingPolylineCanvas, splinePolylineCanvas)
+      ) {
+        const isClosed = data.contour.closed;
+        this.updateContourPolyline(
```
wayfarer3130, line 847: Consider moving the check for is same into the updateContourPolyline itself, perhaps under a flag.
Also, the isSamePolyline with the changes I suggested works fine for either 2d or 3d point data, so you might not need to convert the old one to canvas points if the new one is being converted anyways, and might have new changes inside updateContourPolyline.

PR #2455, opened 2025-11-11
feat(AnnotationTool) Add margin, border radius, linkLineColor to AnnotationTool styles

steph603:
### Context

New styles have been added to the AnnotationTool text box, as well as have extended the existing styling to allow separate coloring for the `textBoxLinkLine`

- Added optional `textBoxBorderRadius`, defaults to `0` to fallback to current behaviour
- Added optional `textBoxBorderMargin`, defaults to `0` to fallback to current behaviour
- Added optional `textBoxLinkLineColor`, if not provided, uses `textBoxColor` to fallback to current behaviour 
- `_drawTextBackground` in `drawTextBox.ts` now uses the height of the text elements to calculate width and height rather than it's parent
- `<tspan>` elements in the textBackground are centred within the box using the margin as an offset

### Changes & Results
Screenshot 2025-11-11 at 4 13 51 pm
Screenshot 2025-11-11 at 4 14 45 pm

### Testing

Tested using `volumeAnnotationTools` and `stackAnnotationTools` examples

- Ensure all annotation tools and labels are working as previously and are backwards compatible
- In `ToolStyle.ts`, add the below styles to the `defaultConfig` 
```
 // Turn background on to show radius/margin              
 textBoxBackground: 'rgb(255, 255, 255)',                 
 // New properties:                                       
 textBoxLinkLineColor: 'rgb(255, 0, 255)',                
 textBoxLinkLineColorSelected: 'rgb(0, 255, 255)',        
 textBoxBorderRadius: '10',                                
 textBoxMargin: 10,                                       
```
- Ensure all annotation tools display the new styles

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] "OS:  macOS Tahoe 26.1, 
- [x] "Node version: 22.15
- [x] "Browser: Chrome 142.0.7444.135, Firefox 143.0.1, Safari 26.1 
  

### DIFF
--- a/packages/tools/src/drawingSvg/drawTextBox.ts
+++ b/packages/tools/src/drawingSvg/drawTextBox.ts
@@ -61,16 +61,16 @@ function _drawTextGroup(
     fontFamily,
     fontSize,
     background,
-    borderRadius,
-    margin,
+    textBoxBorderRadius,
+    textBoxMargin,
   } = options;
 
   let textGroupBoundingBox;
   const [x, y] = [position[0] + padding, position[1] + padding];
   const backgroundStyles = {
     color: background,
-    borderRadius,
-    margin,
+    textBoxBorderRadius,
+    textBoxMargin,
   };
   const svgns = 'http://www.w3.org/2000/svg';
   const svgNodeHash = _getHash(annotationUID, 'text', textUID);
@@ -193,7 +193,11 @@ function _createTextSpan(text): SVGElement {
 }
 
 function _drawTextBackground(group: SVGGElement, backgroundStyles) {
-  const { color, borderRadius = '0', margin = '0' } = backgroundStyles;
+  const {
+    color,
+    textBoxBorderRadius = 0,
+    textBoxMargin = 0,
+  } = backgroundStyles;
   let element = group.querySelector('rect.background');
   const textElement = group.querySelector('text').getBBox();
 
@@ -215,27 +219,29 @@ function _drawTextBackground(group: SVGGElement, backgroundStyles) {
   }
 
   // Get the text groups's bounding box and use it to draw the background rectangle
-  // use the text box dimensions to apply the margin
+  // use the text box dimensions to apply the textBoxMargin
   const bBox = group.getBBox();
 
   const attributes = {
     x: `${bBox.x}`,
     y: `${bBox.y}`,
-    width: `${textElement.width + Number(margin) * 2}`,
-    height: `${textElement.height + Number(margin) * 2}`,
+    width: `${textElement.width + Number(textBoxMargin) * 2}`,
+    height: `${textElement.height + Number(textBoxMargin) * 2}`,
     fill: color,
-    rx: borderRadius,
-    ry: borderRadius,
+    rx: textBoxBorderRadius,
+    ry: textBoxBorderRadius,
   };
 
-  // Add offset to the text spans to centre them within the margin
-  const tSpans = Array.from(
-    group.querySelector('text').querySelectorAll('tspan')
-  );
-  tSpans.forEach((tspan, i) => {
-    i === 0 && tspan.setAttribute('y', margin);
-    tspan.setAttribute('x', margin);
-  });
+  if (textBoxMargin) {
+    // Add offset to the text spans to centre them within the textBoxMargin
+    const tSpans = Array.from(
+      group.querySelector('text').querySelectorAll('tspan')
+    );
+    tSpans.forEach((tspan, i) => {
+      i === 0 && tspan.setAttribute('y', textBoxMargin);
+      tspan.setAttribute('x', textBoxMargin);
+    });
+  }
 
   setAttributesIfNecessary(attributes, element);
 
--- a/packages/tools/src/tools/base/AnnotationTool.ts
+++ b/packages/tools/src/tools/base/AnnotationTool.ts
@@ -357,13 +357,13 @@ abstract class AnnotationTool extends AnnotationDisplayTool {
         specifications,
         annotation
       ),
-      borderRadius: this.getStyle(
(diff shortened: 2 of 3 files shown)


sen-trenser:
It would be great if we have `textBoxBackgroundOpacity` style control as well

wayfarer3130:
> It would be great if we have `textBoxBackgroundOpacity` style control as well

Opacity should be able to be done with rgba, eg:
textBoxBackground: '#ffffff80'
or
textBoxBackground: 'rgba(255,255,255,0.5)'

steph603:
Yes, opacity is pre-existing 😄 
`textBoxBackground: 'rgba(0,0,0,0.5)'`
Screenshot 2025-11-11 at 5 16 38 pm

packages/tools/src/tools/base/AnnotationTool.ts
```diff
@@ -357,6 +357,17 @@ abstract class AnnotationTool extends AnnotationDisplayTool {
         specifications,
         annotation
       ),
+      borderRadius: this.getStyle(
```
sedghi: i suggest calling it textboxBorderRadius

packages/tools/src/tools/base/AnnotationTool.ts
```diff
@@ -357,6 +357,17 @@ abstract class AnnotationTool extends AnnotationDisplayTool {
...
+        'textBoxBorderRadius',
+        specifications,
+        annotation
+      ),
+      margin: this.getStyle('textBoxMargin', specifications, annotation),
```
sedghi: same

packages/tools/src/tools/base/AnnotationTool.ts
```diff
@@ -357,6 +357,17 @@ abstract class AnnotationTool extends AnnotationDisplayTool {
...
+      linkLineColor: this.getStyle(
```
sedghi: same

packages/docs/docs/concepts/cornerstone-tools/annotation/config.md
```diff
@@ -38,22 +38,26 @@ level to check.
...
-  textBoxColorLocked: 'rgb(255, 255, 0)',
-  textBoxBackground: '',
-  textBoxLinkLineWidth: '1',
-  textBoxLinkLineDash: '2,3',
+    colorHighlighted: 'rgb(0, 255, 0)',
+    colorSelected: 'rgb(0, 220, 0)',
+    colorLocked: 'rgb(209, 193, 90)',
+    lineWidth: '1',
+    lineDash: '',
+    shadow: true,
```
sedghi, line 46: which shadow is this? if textbox shadow, let's call it textbox shadow

packages/tools/src/drawingSvg/drawTextBox.ts
```diff
@@ -202,16 +215,28 @@ function _drawTextBackground(group: SVGGElement, color: string) {
...
-    height: `${bBox.height}`,
+    width: `${textElement.width + Number(margin) * 2}`,
+    height: `${textElement.height + Number(margin) * 2}`,
     fill: color,
+    rx: borderRadius,
+    ry: borderRadius,
   };
 
+  // Add offset to the text spans to centre them within the margin
+  const tSpans = Array.from(
```
sedghi: shouldn't this only happen when there is a margin?

packages/docs/docs/concepts/cornerstone-tools/annotation/config.md
steph603: I think it is - but I have not added or modified this code, this is just an update to the docs as they are out of date with the default props referred to in `ToolStyle.ts`
I'm happy to update this and all usages to `textBoxShadow` if you think it should be all aligned, but it will be a breaking change for others already using this.

packages/tools/src/drawingSvg/drawTextBox.ts
steph603: Probably, yes, but margin and borderRadius both default to `'0'`, so they will never be falsy (and need to have a default or the other styles break). 

Confirmed multi-line behaviour is unaffected when no margin is provided

Image

I could instead default to `0` and wrap this block in a condition if you think its worthwhile?

packages/tools/src/tools/base/AnnotationTool.ts
steph603: I'm happy to do this, my thoughts process here was following the pattern on the other props.
eg. `'textBoxLinkLineWidth'` is provided to the config, and this function turns it into `lineWidth` (same with `color, fontSize, fontFamily`, etc. 
If I change this, especially for `textBoxLinkLineColor`, it will break the pattern for a few props and maybe be more confusing. 
Or I could change all prop names to match?

What do you think?

sedghi: yeah probably lineWidth is too generic too, i rather be specific

packages/tools/src/drawingSvg/drawTextBox.ts
sedghi: Thanks

steph603: Done! 👍🏻

packages/tools/src/tools/base/AnnotationTool.ts
steph603: Yep - I agree with you on being specific in naming. 
I've updated the new props to be specific with the `textBox` prefix etc as you suggested. 

I had a look at the other generic names too - eg. `lineWidth` - this particular prop is referenced 190 times, and I know sometimes it refers to the textBoxLine, and sometimes the Line of the annotation/measurement itself. I'm nervous to go making so many changes in this PR so have left existing names as is. 🙏🏻

PR #2474, opened 2025-11-19
feat(decoders): decode color baseline JPEGs in worker

IbrahimCSAE:
## Summary
- Add an OffscreenCanvas-based JPEG Baseline color decoder to the worker decoders.
- Route 8-bit color JPEGs through it so the full decode stays in the worker instead of bouncing back to the main thread
- Simplify the browser entry point since it no longer needs to pass a canvas down to the worker

That keeps the fast path but avoids running the browser decoder on the main thread for color JPEGs.


IbrahimCSAE:
Might need to make this optional due to offscreen canvas support in browsers, will get to it.

sedghi:
yes please make it optional default false

PR #2479, opened 2025-11-20
fix(SegmentationState): only activate first segment if it exists

nagychris:
### Context

This change was already there in a previous version: https://github.com/cornerstonejs/cornerstone3D/blob/b1979f6092bd59358ba840a4ba4356ef915b8073/packages/tools/src/stateManagement/segmentation/internalAddSegmentationRepresentation.ts#L37-L44 

But was removed again in the current `main` branch (via https://github.com/cornerstonejs/cornerstone3D/commit/a1ed21861779b52257557ab5c3928c492fd238a8): 
https://github.com/cornerstonejs/cornerstone3D/blob/e7b0cea10f4ceb71710d3f01eaead75152180b7a/packages/tools/src/stateManagement/segmentation/internalAddSegmentationRepresentation.ts#L53-L60

-> this breaks the expected behavior of segmentations, as they now always load with an empty segment with index 1, because `activateSegmentIndex` is called and this creates the non-existing segment with an empty label `''`. We should not do this, and rather expect the application to handle the case of an initially empty segmentation. 

If we add an empty segment automatically, it is harder for the application to handle the initial state for empty segmentations -> they might need to always remove this segment if it exists, e.g., if they want to check if no segment exists yet. Otherwise the user might start drawing with a non-existent segment index.

### Changes & Results

Before: empty segmentations are loaded with an empty segment with index 1.
After: empty segmentations are empty after loading (contain no segments).

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] "OS:  ubuntu 24
- [x] "Node version:  v22.7.0
- [x] "Browser: Brave / Chromium (latest)
  

### DIFF
--- a/packages/tools/src/stateManagement/segmentation/internalAddSegmentationRepresentation.ts
+++ b/packages/tools/src/stateManagement/segmentation/internalAddSegmentationRepresentation.ts
@@ -54,8 +54,8 @@ function internalAddSegmentationRepresentation(
       const segmentKeys = Object.keys(segmentation.segments);
       if (segmentKeys.length > 0) {
         firstSegmentIndex = segmentKeys.map((k) => Number(k)).sort()[0];
+        setActiveSegmentIndex(segmentationId, firstSegmentIndex);
       }
-      setActiveSegmentIndex(segmentationId, firstSegmentIndex);
     }
   }
 


packages/tools/src/stateManagement/segmentation/internalAddSegmentationRepresentation.ts
```diff
@@ -54,8 +54,8 @@ function internalAddSegmentationRepresentation(
       const segmentKeys = Object.keys(segmentation.segments);
       if (segmentKeys.length > 0) {
         firstSegmentIndex = segmentKeys.map((k) => Number(k)).sort()[0];
+        setActiveSegmentIndex(segmentationId, firstSegmentIndex);
```
jbocce, line 57: Unfortunately this change will break this [example](https://www.cornerstonejs.org/live-examples/interpolationcontoursegmentation) and likely others. I used this PR branch, ran the example and it throws an exception when clicking to add a contour because no segment is added. I think the existing behaviour is desired and relied upon in many scenarios. @sedghi thoughts?

sedghi: Hey @nagychris, can you tell me more about the problem you're running into, without getting too deep into the code details? For example, are you loading a labelmap, and then seeing something unexpected happen?

nagychris: Hi @sedghi, the issue is that my application expects the initial state of a segmentation to have no segments, because the user should first select which type of segment they want to create before they can draw. So the issue is related to how the workflow works in my application. 

The current workaround I see is to remove this empty segment as soon as the segmentation is loaded, but it's quite hacky, so I wondered if we can remove this again. 

But if you say that this empty / default segment is required for other cornerstone3D features, then we can also keep it. Just wanted to raise this topic, as it might also be an issue for other applications using cornerstone3D.

sedghi: So the issue with your currrent proposal is that sometimes (specifically in RTSS) we skip segments, so the first one becomes segment 2 for instance

nagychris: @sedghi  but that sounds like a more general bug to me, that should not be related to whether you add a default segment here?

PR #2484, opened 2025-11-21
fix: sculpt tool whorl effects

wayfarer3130:
#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  

### DIFF
--- a/packages/tools/examples/SculptorTool/index.ts
+++ b/packages/tools/examples/SculptorTool/index.ts
@@ -75,6 +75,8 @@ content.appendChild(viewportGrid);
 
 const instructions = document.createElement('p');
 instructions.innerText = `
+Left mouse button is initially bound to planar freehand and right button to contour segmentation tool
+
 Drawing:
 
 - Select PlanarFreehandROI from dropdown button to draw freehandROIs.
@@ -83,7 +85,8 @@ Drawing:
 
 Editing:
 
-- Select SculptorTool from dropdown ,then adjustable cursor will appear.
+- Sculptor tool is initially configured as the right drag tool, so you can just use it immediately
+- Select SculptorTool from dropdown
 - Nearest freehand ROI/Freehand Contour Segmentation will be selected while clicking, and toolsize can be adjusted by moving cursor near to selected annotation.
 - Moving the cursor closer to the active ROI reduces the tool diameter.
 - Moving the cursor away from the active increases the tool diameter.
@@ -99,6 +102,8 @@ const toolsNames = [
   SculptorTool.toolName,
 ];
 let selectedToolName = toolsNames[0];
+
+let activeSegmentIndex = 0;
 
 function updateActiveSegmentIndex(segmentIndex: number): void {
   activeSegmentIndex = segmentIndex;


wayfarer3130:
@sedgi 
The intersections look like below.  The idea is to iterate through all of them, and basically split the
entry point .... cursor edge data ... nearest angular exit point

This will produce one set per entry.
Then, this is combined with the nearest angular exit point data, finding the next entry point which is in the loop after the nearest exit point.

This can produce one or many intersections
These need to each be checked for degenerate cases

intersections= [
  {
    "annotationUID": "306a7c51-b043-4b53-a977-54c3af6a9f28",
    "isEnter": true,
    "index": 7,
    "point": {
      "0": -96.79963684082031,
      "1": -53.24888610839844,
      "2": 78
    },
    "angle": -0.15526077561331691
  },
  {
    "annotationUID": "306a7c51-b043-4b53-a977-54c3af6a9f28",
    "isEnter": false,
    "index": 310,
    "point": {
      "0": -21.13036346435547,
      "1": -52.24995040893555,
      "2": 78
    },
    "angle": 0.04187563893500747
  },
  {
    "annotationUID": "306a7c51-b043-4b53-a977-54c3af6a9f28",
    "isEnter": true,
    "index": 722,
    "point": {
      "0": -18.770809173583984,
      "1": -59.526981353759766,
      "2": 78
    },
    "angle": 0.06546338935345554
  },
  {
    "annotationUID": "306a7c51-b043-4b53-a977-54c3af6a9f28",
    "isEnter": false,
    "index": 736,
    "point": {
      "0": -100.28927612304688,
      "1": -75.1943588256836,
      "2": 78
    },
    "angle": -0.10851506169480139
  }
]

package.json
```diff
@@ -196,6 +196,7 @@
   "resolutions": {
     "axios": "1.12.0",
     "buffer": "6.0.3",
+    "node-forge": "1.3.2",
```
wayfarer3130, line 199: @jbocce - this fixes the audit message to allow the build to proceed.

playwright.config.ts
```diff
@@ -77,7 +77,7 @@ export default defineConfig({
     },
   ],
   webServer: {
-    command: 'cross-env COVERAGE=true nyc yarn build-and-serve-static-examples',
+    command: 'cross-env COVERAGE=true nyc bun build-and-serve-static-examples',
```
wayfarer3130, line 80: @jbocce - this fixes the build in the current versions - something changed in the deployment such that yarn isn't available.

jbocce: That's weird, but ok. I wonder if this will be need in OHIF too?

package.json
jbocce: Could you please make a similar change in OHIF?

wayfarer3130: Yes, will do as part of the commit to include this change.

wayfarer3130:
> * It looks like the sculptor tool activates after a mouse-up event in the demo for freehand drawing, which isn't working right. I guess the demo shouldn't handle that kind of transition.

I added a right-click active freehand, will add notes to that effect

> * The text box shifts when you're using the sculpture tool. It's not really an issue right now, but let's create a ticket to address it later. The reason it's okay for now is that we don't actually use the text box during segmentation.

Created https://github.com/cornerstonejs/cornerstone3D/issues/2491

> * I'm really enjoying how this is handled and interacted with now. Just be sure the undo function works for every action. The demo doesn't have undo, so I can't test it out.

Adding undo/redo buttons to the sculptor tool example to test this.

PR #2485, opened 2025-11-24
fix(Volumes): Allow loading 1-slice volumes

paulglx:
### Context

Currently, Cornerstone isn't able to load single-slice volume in a volume viewport.

However, the stack viewport in vtk is more or less a single slice vtk volume viewport, so it is be possible to enable single slice volumes in the volume viewport. (thanks @wayfarer3130)

There is logic in place that assumes multiple slices. This PR fixes it, allowing single slice volumes in CS3D.

### Changes & Results

- Handle 1-slice case in util files for loading volumes
	- Use pixel spacing in columns/rows to make cubic voxels
- Prevent scrolling if 1 slice only
- Small additional style changes

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  

### DIFF
--- a/packages/core/src/utilities/calculateSpacingBetweenImageIds.ts
+++ b/packages/core/src/utilities/calculateSpacingBetweenImageIds.ts
@@ -6,7 +6,33 @@ import { getConfiguration } from '../init';
  * Default spacing value (in mm) used as fallback when spacing cannot be calculated
  * or retrieved from DICOM metadata
  */
-const DEFAULT_SPACING = 1;
+const DEFAULT_THICKNESS_SINGLE_SLICE = 1;
+
+/**
+ * Gets pixel spacing value for creating cubic voxels with equal side lengths.
+ *
+ * @param metadata - Image plane module metadata
+ * @returns The best available pixel spacing value, or undefined if none available
+ */
+function getPixelSpacingForCubicVoxel(metadata: {
+  columnPixelSpacing?: number;
+  rowPixelSpacing?: number;
+  pixelSpacing?: number[];
+}): number | undefined {
+  if (metadata.columnPixelSpacing !== undefined) {
+    return metadata.columnPixelSpacing;
+  }
+  if (metadata.rowPixelSpacing !== undefined) {
+    return metadata.rowPixelSpacing;
+  }
+  if (metadata.pixelSpacing?.[1] !== undefined) {
+    return metadata.pixelSpacing[1];
+  }
+  if (metadata.pixelSpacing?.[0] !== undefined) {
+    return metadata.pixelSpacing[0];
+  }
+  return undefined;
+}
 
 /**
  * Calculates the spacing between images in a series based on their positions
@@ -23,14 +49,27 @@ export default function calculateSpacingBetweenImageIds(
   } = metaData.get('imagePlaneModule', imageIds[0]);
 
   if (imageIds.length === 1) {
-    const { sliceThickness, spacingBetweenSlices } = metaData.get(
-      'imagePlaneModule',
-      imageIds[0]
-    );
+    const {
+      sliceThickness,
+      spacingBetweenSlices,
+      columnPixelSpacing,
+      rowPixelSpacing,
+      pixelSpacing,
+    } = metaData.get('imagePlaneModule', imageIds[0]);
 
     if (sliceThickness) return sliceThickness;
     if (spacingBetweenSlices) return spacingBetweenSlices;
-    return DEFAULT_SPACING;
+
+    const pixelSpacingValue = getPixelSpacingForCubicVoxel({
+      columnPixelSpacing,
+      rowPixelSpacing,
+      pixelSpacing,
+    });
+    if (pixelSpacingValue !== undefined) {
+      return pixelSpacingValue;
+    }
+
+    return DEFAULT_THICKNESS_SINGLE_SLICE;
   }
 
   // Calculate scan axis normal from image orientation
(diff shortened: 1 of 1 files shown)


packages/core/src/utilities/calculateSpacingBetweenImageIds.ts
```diff
@@ -2,6 +2,12 @@ import { vec3 } from 'gl-matrix';
 import * as metaData from '../metaData';
 import { getConfiguration } from '../init';
 
+/**
+ * Default spacing value (in mm) used as fallback when spacing cannot be calculated
+ * or retrieved from DICOM metadata
+ */
+const DEFAULT_SPACING = 1;
```
wayfarer3130: You need to be more specific about what spacing this is - how about
DEFAULT_THICKNESS_SINGLE_SLICE

packages/core/src/utilities/calculateSpacingBetweenImageIds.ts
```diff
@@ -16,6 +22,17 @@ export default function calculateSpacingBetweenImageIds(
...
 
+  if (imageIds.length === 1) {
+    const { sliceThickness, spacingBetweenSlices } = metaData.get(
+      'imagePlaneModule',
+      imageIds[0]
+    );
+
+    if (sliceThickness) return sliceThickness;
+    if (spacingBetweenSlices) return spacingBetweenSlices;
+    return DEFAULT_SPACING;
```
wayfarer3130: I would use the pixel spacing in the column direction, or the two rather than having a default - the problem is the default could be way to big or way to small, whereas creating a cubic voxel with equal side length is a better idea.

packages/core/src/utilities/calculateSpacingBetweenImageIds.ts
```diff
@@ -142,7 +156,7 @@ export default function calculateSpacingBetweenImageIds(
       console.debug(
         'Could not calculate spacing. The VolumeViewport visualization is compromised. Setting spacing to 1 to render'
       );
-      spacing = 1;
+      spacing = DEFAULT_SPACING;
     }
```
wayfarer3130: I would use the same logic here for spacing using the spacing between columns, and drop an error message.  This is actually quite a serious issue in that it can be non obvious what is going on.    If you don't want to touch this one right now, then just leave it spacing = 1;

packages/core/src/utilities/getClosestImageId.ts
```diff
@@ -35,7 +35,7 @@ export default function getClosestImageId(
   const { direction, spacing, imageIds } = imageVolume;
   const { ignoreSpacing = false } = options || {};
 
-  if (!imageIds?.length) {
+  if (imageIds?.length <= 1) {
```
wayfarer3130: if (!imageIds?.length)
is the canonical check for defined

packages/core/src/utilities/isValidVolume.ts
```diff
@@ -10,7 +10,7 @@ import isEqual from './isEqual';
  * @returns true if the imageIds form a valid volume, false otherwise.
  */
 function isValidVolume(imageIds: string[]): boolean {
-  if (imageIds.length <= 1) {
+  if (imageIds.length === 0) {
```
wayfarer3130: Same here
imageIds.length

packages/tools/src/utilities/segmentation/getOrCreateImageVolume.ts
```diff
@@ -13,7 +13,7 @@ import {
 function getOrCreateImageVolume(
   referencedImageIds: string[]
 ): Types.IImageVolume | undefined {
-  if (!referencedImageIds || referencedImageIds.length <= 1) {
+  if (!referencedImageIds || referencedImageIds.length === 0) {
```
wayfarer3130: Same

wayfarer3130 (changes requested):
Couple of tweaks here

packages/core/src/utilities/getClosestImageId.ts
paulglx: Thanks for the feedback, this is fixed at [8686596](https://github.com/cornerstonejs/cornerstone3D/pull/2485/commits/8686596dafde21dad939d14c9d6649e291ddaa42)

packages/core/src/utilities/isValidVolume.ts
paulglx: Same as above ([8686596](https://github.com/cornerstonejs/cornerstone3D/pull/2485/commits/8686596dafde21dad939d14c9d6649e291ddaa42))

packages/tools/src/utilities/segmentation/getOrCreateImageVolume.ts
paulglx: Same as above ([8686596](https://github.com/cornerstonejs/cornerstone3D/pull/2485/commits/8686596dafde21dad939d14c9d6649e291ddaa42))

packages/core/src/utilities/calculateSpacingBetweenImageIds.ts
paulglx: Makes sense, this is renamed in [2075620](https://github.com/cornerstonejs/cornerstone3D/pull/2485/commits/2075620c85ef96fe88c401db5a42407fabd9647d) and used as last resort if all methods fail

packages/core/src/utilities/calculateSpacingBetweenImageIds.ts
paulglx: Makes sense! This is implemented at [2075620](https://github.com/cornerstonejs/cornerstone3D/pull/2485/commits/2075620c85ef96fe88c401db5a42407fabd9647d). It tries using `columnPixelSpacing`, `rowPixelSpacing`, or data from the `pixelSpacing` array if the former aren't defined

packages/core/src/utilities/calculateSpacingBetweenImageIds.ts
paulglx: Same as above, using the same method ([2075620](https://github.com/cornerstonejs/cornerstone3D/pull/2485/commits/2075620c85ef96fe88c401db5a42407fabd9647d))

paulglx:
> Couple of tweaks here

Comments addressed, any feedback is welcome! @wayfarer3130

wayfarer3130 (approved):

PR #2486, opened 2025-11-25
fix: Export seg

wayfarer3130:
### Context

There was a missing global Buffer that is now being hit since we were calling a different save method that pre-bundles the result using Buffer.  Also updated the example to prevent double loading of the segmentation which fixes the first error reported (which previously existed).

### Changes & Results

Provide a global Buffer in examples
Prevent double click on segmentation load

### Testing

Run segmentationStack example, load image, load segmentation, save seg

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  

### DIFF
--- a/packages/dicomImageLoader/src/imageLoader/wadors/metaData/metaDataProvider.ts
+++ b/packages/dicomImageLoader/src/imageLoader/wadors/metaData/metaDataProvider.ts
@@ -87,9 +87,9 @@ function metaDataProvider(type, imageId) {
       seriesNumber: getNumberValue(metaData['00200011']),
       studyInstanceUID: getValue<string>(metaData['0020000D']),
       seriesDate: getValue<string>(metaData['00080021']),
-      seriesTime: getValue<string>(metaData['00080031'], 0, ''),
+      seriesTime: getValue<string>(metaData['00080031']),
       acquisitionDate: getValue<string>(metaData['00080022']),
-      acquisitionTime: getValue<string>(metaData['00080032'], 0, ''),
+      acquisitionTime: getValue<string>(metaData['00080032']),
     };
   }
 


packages/dicomImageLoader/src/imageLoader/wadors/metaData/metaDataProvider.ts
```diff
@@ -88,16 +86,10 @@ function metaDataProvider(type, imageId) {
...
-      seriesTime: dicomParser.parseTM(
-        getValue<string>(metaData['00080031'], 0, '')
-      ),
-      acquisitionDate: dicomParser.parseDA(
-        getValue<string>(metaData['00080022'])
-      ),
-      acquisitionTime: dicomParser.parseTM(
-        getValue<string>(metaData['00080032'], 0, '')
-      ),
+      seriesDate: getValue<string>(metaData['00080021']),
```
wayfarer3130, line 89: Gets the raw value to prevent throwing an exception on creating the instance object - otherwise the created object is invalid.

utils/ExampleRunner/template-config.js
```diff
@@ -50,6 +50,9 @@ module.exports = {
         },
       ],
     }),
+    new rspack.ProvidePlugin({
```
wayfarer3130, line 53: This change fixes the actual save.

jbocce:
@wayfarer3130  Is this something that our testing team should look at too?

yarn.lock
```diff
@@ -4215,9 +4222,9 @@ add-stream@^1.0.0:
   integrity sha512-qQLMr+8o0WC4FZGQTcJiKBVC59JylcPSrTtk6usvmIDFUOCKegapy1VHQwRbFMOFyb/inzUVqHs+eMYKDM1YeQ==
 
 adm-zip@^0.5.10:
```
jbocce, line 4224: Just pointing this out as an FYI... looks like yarn decided to start using adm-zip version 0.5.16 while bun has been using it for a while.

packages/adapters/examples/segmentationStack/index.ts
jbocce, line 1: Is this file just format changes?

packages/adapters/examples/segmentationVolume/utils.ts
jbocce, line 1: Formatting only?

wayfarer3130:
> @wayfarer3130 Is this something that our testing team should look at too?

It was something they found, but I think just test that saving doesn't throw an exception locally, then we will merge and retest in the released environment as it needs to be tested there regardless.

packages/dicomImageLoader/src/imageLoader/wadors/metaData/metaDataProvider.ts
jbocce: Are there places where `seriesDate` was NOT expected to be the raw value and now need to change?

wayfarer3130:
> Just some comments to answer/address. I ran the example and could download the SEG without error. However I only tried: loading the DICOM, loading the SEG, drawing a SEG and then exporting. Let me know if there is more you would like tested.

That is really all I was looking to have checked as the issue was really a packaging issue caused by a change in how the download was packaged.

packages/dicomImageLoader/src/imageLoader/wadors/metaData/metaDataProvider.ts
wayfarer3130: Every place I reviewed, and I think I reviewed every one except examples, the expected values was the raw value and not the formatted value.  There are a lot of places needing the raw values.

packages/adapters/examples/segmentationStack/index.ts
wayfarer3130: No, there is a duplicate load prevention below

packages/adapters/examples/segmentationStack/index.ts
```diff
@@ -1,226 +1,233 @@
...
+    state.referenceImageIds = await createImageIdsAndCacheMetaData(
+      state.devConfig.fetchDicom
+    );
+
+    loadDicom();
+  },
+  container: group1,
 });
 
+let segmentationLoaded = false;
```
wayfarer3130, line 152: This remembers whether the segmentation has been loaded, and prevents the stack exception reported in the issue - this was a pre-existing issue that occurs on a double click on that button.  No other changes.

packages/adapters/examples/segmentationVolume/utils.ts
```diff
@@ -11,218 +11,213 @@ const { downloadDICOMData } = cornerstoneAdapters.helpers;
...
+      segmentationId,
+      segmentIndex
     );
-
-    const labelmap3D = {
-        segmentsOnLabelmap: Array.from(new Set(allSegmentsOnLabelmap.flat())),
-        metadata: [],
-        labelmaps2D
+    const RecommendedDisplayCIELabValue = dcmjs.data.Colors.rgb2DICOMLAB(
+      color.slice(0, 3).map((value) => value / 255)
```
wayfarer3130, line 155: This prevents a dciodvfy error caused by the alpha channel being included.  This is the only change here.

packages/adapters/examples/segmentationVolume/utils.ts
wayfarer3130: There is a warning fix below causing invalid DICOM to be generated (although interpretable generally).

PR #2489, opened 2025-11-26
fix(preview): do not iterate through volume if no previewSegmentIndex is set

nagychris:
… is set

### Context

I am trying to develop a connected-component labeling algorithm that needs to modify the voxel data / scalar data of the segmentation (via voxelManager.setCompleteScalarDataArray).
However, I noticed after doing this for the first time, interestingly the activation of tools are delayed by some seconds. For example, switching between different variants of  BrushTool  (CircularBrush and CircularEraser) freezes the UI for a few seconds before it actually emits the TOOL_MODE_CHANGED event and updates the selection in the UI.
I was also able to reproduce this in the labelmapSegmentationTools example with the following code after the run(); call in line 446:
```
let timeout = null;
eventTarget.addEventListener(csToolsEnums.Events.SEGMENTATION_DATA_MODIFIED, (evt) => {
  if (timeout) {
    clearTimeout(timeout);
  }
  // Debounce updating the voxel data to avoid doing it too frequently
  timeout = setTimeout(() => {
    timeout = null;
    const { segmentationId } = evt.detail;
    const segmentationVolume = cache.getVolume(segmentationId);
    const {voxelManager} = segmentationVolume;

    const scalarDataArray = voxelManager.getCompleteScalarDataArray();
    // Imagine we are doing some modifications to the segmentation data here...
    // Update the voxel data afterwards (e.g. because we split/merged segments)
    voxelManager.setCompleteScalarDataArray(scalarDataArray);
  }, 200);
});

eventTarget.addEventListener(csToolsEnums.Events.TOOL_MODE_CHANGED, (evt) => {
  console.log('Tool mode changed: ', evt.detail);
});
```

Edit: after debugging for a while, I noticed the delay is coming from the `rejectPreview` inside `BrushTool.disableCursor`, more specifically this line where we iterate over the whole volume: [https://github.com/cornerstonejs/cornerstone3D/blob/c19287ba13ddf06c16f2852153b636[…]tools/src/tools/segmentation/strategies/compositions/preview.ts](https://github.com/cornerstonejs/cornerstone3D/blob/c19287ba13ddf06c16f2852153b6362df0095643/packages/tools/src/tools/segmentation/strategies/compositions/preview.ts#L134)

Observation: We dont need to do this in the case where preview is disabled and `previewSegmentIndex` is `null`.
This fixes the above issue with the delay (at least if preview is disabled).

### Changes & Results

- if no preview is enabled, we dont iterate through the whole volume anymore when disabling / switching between brush tool instances
- preview works same as before (if enabled, `previewSegmentIndex` will be set)

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS:  Ubuntu 24
- [] "Node version:  v22.7.0
- [] "Browser:
   Brave/Chrome


sedghi:
can you rebase so that tests run again

PR #2493, opened 2025-11-27
fix: Cursor disappearing when multiple tools active

wayfarer3130:
### Context

The cursor for one of the tools would disappear when there were multiple tools active.  This was caused by duplicate rendering of the tool cursor, and depending on which render happened last the cursor would disappear or fail to move.

### Changes & Results

Added a `primary` value on a tool to know whether it is currently marked as having the primary button active.

Added a static `activeCursorTool` on a tool to allow figuring out which tool should render the cursor.

Added shift + left mouse button to erase sections in labelmapSegmentationTools example

Only update the cursor in the active cursor tool for the brush tool.

### Testing

Run the labelmapSegmentationTools
Select either the sphere or circular brush tools
Draw with left button
Erase with shift+left button

Cursor should be visible throughout.

Select circular erase
Move on window, cursor should be visible.

Select threshold circle, check it doesn't throw immediately.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  

### DIFF
--- a/packages/tools/src/tools/base/BaseTool.ts
+++ b/packages/tools/src/tools/base/BaseTool.ts
@@ -14,7 +14,21 @@ const { DefaultHistoryMemo } = csUtils.HistoryMemo;
 abstract class BaseTool {
   static toolName;
 
-  /** Set to the tool that is currently drawing the active cursor */
+  /**
+   * Set to the tool that is currently drawing the active cursor.  This
+   * will be either primary mouse button tool if no tool is currently
+   * being directly interacted with, OR the tool that is directly interacted
+   * with.  This logic ensures that there is only a single tool at a time
+   * drawing, which prevents tools not getting mouse updates from over-writing
+   * the cursor.
+   *
+   * - If the tool bound to the primary button is a cursor drawing tool,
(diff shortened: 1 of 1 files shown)


packages/tools/src/tools/base/BaseTool.ts
```diff
@@ -21,6 +25,9 @@ abstract class BaseTool {
   public toolGroupId: string;
   /** Tool Mode - Active/Passive/Enabled/Disabled/ */
   public mode: ToolModes;
+  /** Primary tool - this is set to true when this tool is primary */
+  public primary = false;
```
sedghi: isPrimary makes it more clear

wayfarer3130:
> can we not hold these two states in the tool instance and ask the toolGroup about who is active? and who has the activeCursor? Each tool instance belongs to one toolGroup at a time, and therefore you can do this.toolGroup.getWhatever

It is held in the tool instance, but it is set from the tool group because it is the tool group that knows about when something is primary or not.

PR #2494, opened 2025-11-28
fix: calibration fails with local upload

wayfarer3130:
### Context

Calibration fails when using local upload

### Changes & Results

Replaced the metadata providers for wadors and wadouri with shared instances that use a common listener pattern.

The plan is to split out the metadata handling into a separate cs3d module, but this is planned as a secondary step shortly after merging this PR.

### Testing

Run calibrationTools
Upload a file which has no pixel spacing (shows in px by default)
Select a calibration from the pull down
Expect to see measurements in 'mm'

Run advancedLocal
Test with a variety of the locally uploaded/wadors instances
They should be consistent with the expected measurement units for that type, in both the local upload and the wadors version.

Note that most of this is tested with existing and new unit/karma tests

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  


packages/core/src/utilities/Tags.ts
wayfarer3130, line 1: @sedghi - I have most of the common tags we are using available now in a way that works across different source data types.  I'm seeing quite a few differences in handling between different systems - can we talk about this PR next week?  It is still trying to snowball more than I like, but it is sort of getting there.

PR #2513, opened 2025-12-11
fix: Fix potential divide by 0 in Viewport.resetCamera

abustany:
getExtent returns array indices, not dimensions. This fixes a division by zero when displaying 1x1 pixel images.

### Context

Displaying 1x1 pixel images fails

### Changes & Results

`Viewport.resetCamera` computes widths/heights using VTK's `getExtent` function, but that function return indices, not dimensions. This PR fixes the computation to correct the off-by-one error.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] OS: macOS 
- [x] Node version: 22 
- [x] Browser: Firefox 146
  

### DIFF
--- a/packages/core/src/RenderingEngine/Viewport.ts
+++ b/packages/core/src/RenderingEngine/Viewport.ts
@@ -1138,8 +1138,8 @@ class Viewport {
       const extent = imageData.getExtent();
       const spacing = imageData.getSpacing();
 
-      widthWorld = (extent[1] - extent[0]) * spacing[0];
-      heightWorld = (extent[3] - extent[2]) * spacing[1];
+      widthWorld = (1 + extent[1] - extent[0]) * spacing[0];
+      heightWorld = (1 + extent[3] - extent[2]) * spacing[1];
     } else {
       ({ widthWorld, heightWorld } = this._getWorldDistanceViewUpAndViewRight(
         bounds,


sedghi:
Do you have a 1x1 pixel image that we can test this with?

packages/core/src/RenderingEngine/Viewport.ts
```diff
@@ -1138,8 +1138,8 @@ class Viewport {
       const extent = imageData.getExtent();
       const spacing = imageData.getSpacing();
 
-      widthWorld = (extent[1] - extent[0]) * spacing[0];
-      heightWorld = (extent[3] - extent[2]) * spacing[1];
+      widthWorld = (1 + extent[1] - extent[0]) * spacing[0];
```
sedghi, line 1141: I checked and yeah seems like this was wrong, the extent is index so start at 0
All tests need to be updated as a result i guess (which we can do)
Double checking with @wayfarer3130

abustany:
> Do you have a 1x1 pixel image that we can test this with?

In our case it's a placeholder image that is generated (basically we build the pixel data by hand and have a custom loader). I did generate one using The Gimp though, attaching it here.

[1px.dcm.zip](https://github.com/user-attachments/files/24105844/1px.dcm.zip)

PR #2523, opened 2025-12-17
fix: 4d for timeslot vector

sedghi:
This pull request adds support for splitting multiframe 4D images (such as NM Multi-frame SPECT/PET) by leveraging DICOM's `TimeSlotVector` and `SliceVector` tags. The main focus is on enabling correct grouping of image frames by time slots and slices for advanced 4D image handling. Additionally, it updates the `ImageVolume` class to improve its handling of time points and dynamic volume detection.

**Multiframe 4D support and image grouping:**

* Added logic in `splitImageIdsBy4DTags.ts` to detect and handle multiframe 4D images using the DICOM `TimeSlotVector` (0054,0070) and `SliceVector` (0054,0080) tags, generating frame-specific `imageIds` and grouping them by time slot. This enables correct splitting of NM multi-frame images for 4D visualization and analysis. [[1]](diffhunk://#diff-aa38dd841463a4614454bcaabd2259bc9b7b624b53c5262745ac0d294014fbc0R14-R130) [[2]](diffhunk://#diff-aa38dd841463a4614454bcaabd2259bc9b7b624b53c5262745ac0d294014fbc0R287-L185)
* Introduced the `handleMultiframe4D` function and supporting types to encapsulate the splitting logic and result structure.
* Updated the main `splitImageIdsBy4DTags` function to use the new multiframe splitting logic before falling back to previous grouping methods.

**ImageVolume class improvements:**

* Refactored the deprecated `numTimePoints` property in `ImageVolume` to a getter that dynamically returns the number of dimension groups (time points), defaulting to 1 if not available.
* Improved the `isDynamicVolume` method in `ImageVolume` to use the new `numTimePoints` getter, providing more robust detection of 4D (dynamic) volumes.

### DIFF
--- a/packages/core/src/utilities/splitImageIdsBy4DTags.ts
+++ b/packages/core/src/utilities/splitImageIdsBy4DTags.ts
@@ -32,7 +32,7 @@ interface MultiframeSplitResult {
  *
  * @param baseImageId - The base imageId that must contain a "/frames/" pattern followed by a digit.
  *   Expected format: e.g., "wadouri:http://example.com/image/frames/1" or "wadors:/path/to/image.dcm/frames/1".
- *   The pattern "/frames/\d+" will be replaced with "/frames/{frameNumber}".
+ *   The pattern "/frames/\d+" will be replaced with "/frames/" + frameNumber.
  * @param frameNumber - The frame number to use (1-based)
  * @returns The imageId with the frame number replaced
  * @throws {Error} If baseImageId does not contain the required "/frames/" pattern, throws an error


wayfarer3130 (commented):
Can you add testing notes as to which study to try loading to test this?  Even if it has to link with OHIF to do it, I'd like to be able to run this.  I can upload a new dataset if required.
Can you address the copilot issues - for the frame numbers, I would suggest adding an endpoint to the metadata which returns all imageIds for the given sop class, and then just use the index.  Otherwise you could pass the imageIds list into this function instead of trying to parse/generate image ids.

sedghi:
> Can you add testing notes as to which study to try loading to test this? Even if it has to link with OHIF to do it, I'd like to be able to run this. I can upload a new dataset if required. Can you address the copilot issues - for the frame numbers, I would suggest adding an endpoint to the metadata which returns all imageIds for the given sop class, and then just use the index. Otherwise you could pass the imageIds list into this function instead of trying to parse/generate image ids.

This issue has the data https://github.com/OHIF/Viewers/issues/5554
It requires another PR in OHIF, to also handle the multiframe NM 4D. I suggest we merge this after review, and then in my PR to OHIF you test

Update: here is the OHIF PR https://github.com/OHIF/Viewers/pull/5666

wayfarer3130 (approved):

wayfarer3130:
@sedghi  - I'm not sure what is going on with the docusaurus build

PR #2532, opened 2026-01-01
add patient sex O in allowed values

salimkanoun:
### Context

PatientSex attribute was missing 'O' value for other, we have error thrown in CS while it is properly handled in calculateSUV
https://dicom.nema.org/dicom/2013/output/chtml/part03/sect_C.2.html#table_C.2-3

### Changes & Results
PatientSex O should not throw error as defined in DICOM standard


sedghi:
@jbocce can you please check ci

jbocce:
> @jbocce can you please check ci

@salimkanoun please update your branch with latest commits and I believe the ci problems will be solved. Thanks so much.

FYI @sedghi

salimkanoun:
Yes sounds better ^

Issue #2317, opened 2025-09-09
[Bug] How to adapt to container size to guarantee no spillage


yuxi-ovo:
### Describe the Bug

I rendered MPR images in the project, and some images will be rendered beyond the container size, and the width must be manually adjusted to about 150px before the image size will gradually return to the middle of the container

<img width="1137" height="464" alt="Image" src="https://github.com/user-attachments/assets/2a0b5566-c46f-418e-bacf-5d6920a402a3" />

### Steps to Reproduce

1

### The current behavior

1

### The expected behavior

1

### System Information

1

sedghi:
Can you check you are using latest version? since recently we fixed this [https://www.cornerstonejs.org/docs/migration-guides/4x/camera-fov](https://www.cornerstonejs.org/docs/migration-guides/4x/camera-fov)

yuxi-ovo:
My version is
<img width="400" height="104" alt="Image" src="https://github.com/user-attachments/assets/837168e5-802f-4c8c-a134-0db9f990396c" />

yuxi-ovo:
Now there are two solutions, one is to set the zoom, and the other is to adjust the width of the container, but these two methods will be very different on different devices, and they can't adapt well to the container size
<img width="1555" height="899" alt="Image" src="https://github.com/user-attachments/assets/717fafd4-3a8e-49a6-b70b-40ae4e5231ca" />

yuxi-ovo:
@sedghi  Can you help me take a look? I'm in a hurry

PR #2537, opened 2026-01-06
fix: One size calculation to address various FOV issues.

wayfarer3130:
### Context

In CS3D 4.0, there was a change to update the size of the image area to use just the width/height.  That contained a few bugs causing the sizing of some images to fail to be correct, but there are a lot of tests looking at the exact size.
This PR is compatible with that change for all stack views, but uses the correct sizing/location for MPR views.

### Changes & Results

Distort the correct positioning just a tiny bit to agree with the new full-view display, but otherwise just use a single view calculation for displaying the image data.

### Testing

Run 
bun example resize

Turn off the "store Display Area" (it causes some problems when switching stored display areas - it can otherwise be used)

Use the various display areas to show the image in various locations/sizes - by default, the image should just touch the edges - this is incorrect by up to 2 pixels in a few cases.

Addresses: [#2317]

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  

### DIFF
--- a/packages/tools/examples/mprReformat/index.ts
+++ b/packages/tools/examples/mprReformat/index.ts
@@ -117,8 +117,6 @@ const viewportColors = {
   [viewportId3]: 'rgb(0, 200, 0)',
 };
 
-let synchronizer;
-
 const viewportReferenceLineControllable = [
   viewportId1,
   viewportId2,
--- a/tests/MPRReformat.spec.ts
+++ b/tests/MPRReformat.spec.ts
@@ -10,7 +10,7 @@ test.beforeEach(async ({ page }) => {
   await visitExample(page, 'mprReformat', 2000);
 });
 
-test.describe('MPR Reformat Visual Tests shouldUpdateThis', () => {
+test.describe('MPR Reformat Visual Tests', () => {
   test('should match screenshot before interaction', async ({ page }) => {
     const screenshotLocator = page.locator('#content > div');
 


packages/core/src/RenderingEngine/Viewport.ts
```diff
@@ -1129,23 +1114,34 @@ class Viewport {
...
 
-      widthWorld = (extent[1] - extent[0]) * spacing[0];
-      heightWorld = (extent[3] - extent[2]) * spacing[1];
-    } else {
-      ({ widthWorld, heightWorld } = this._getWorldDistanceViewUpAndViewRight(
-        bounds,
-        viewUp,
-        viewPlaneNormal
-      ));
+      const extent = imageData.getExtent();
```
wayfarer3130, line 1130: The rest of this code is just temporary to show that the width world/height world are nearly exactly the same as the preivous calculation.

packages/tools/examples/resize/index.ts
```diff
@@ -412,10 +412,8 @@ async function run() {
 
   // Get Cornerstone imageIds for the source data and fetch metadata into RAM
   const imageIds = await createImageIdsAndCacheMetaData({
-    StudyInstanceUID:
-      '1.3.6.1.4.1.14519.5.2.1.7009.2403.334240657131972136850343327463',
-    SeriesInstanceUID:
-      '1.3.6.1.4.1.14519.5.2.1.7009.2403.226151125820845824875394858561',
+    StudyInstanceUID: '1.3.6.1.4.1.25403.345050719074.3824.20170125095438.5',
```
wayfarer3130, line 415: Use a better example of a non-cube volume to show sizing of volumes.

wayfarer3130:
@tblock79 and @IbrahimCSAE  - does this PR address your concerns?  I know it doesn't yet pass quite all the tests, still need to update a few, but it is mostly compatible as well as working correctly for multiple view orientations.

PR #2539, opened 2026-01-07
fix(Volume viewport): Image shift in mpr when measurement is loaded.

nithin-trenser:
### Context
This PR resolves image shifting in MPR viewports when loading or selecting measurements. This issue was observed with multiframe datasets.

Fixes:  https://github.com/OHIF/Viewers/issues/5684

This PR has been incorporated by [FlyWheel.io](https://flywheel.io/)

### Changes & Results
 The setViewPlane() method was updated to compute a view plane that is the correct distance but centered on the same spot as it was previously.  This does not resolve the underlying issue, but is a correct fix for newly loaded SRs that can change the focal position relative to the image.
 
 Before:
 

https://github.com/user-attachments/assets/3de6706d-4d6c-43a4-8731-5e5b77d6ba91

After:

https://github.com/user-attachments/assets/398a8e68-a272-4a5b-a7bf-a04d7f67e376

### Testing

- Launch a data : https://viewer-dev.ohif.org/segmentation?StudyInstanceUIDs=1.2.826.0.1.3680043.2.1125.1.11608962641993666019702920539307840
- Switch to Advanced Layout (multi-viewport).
- Load SR to the viewport.
- Confirm that the image loads as expected.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] OS: Windows 11 10.0.22631
- [x] Node version: 20.17.0
- [x] Browser: Chrome: 143.0.7499.170
  

### DIFF
--- a/packages/core/src/RenderingEngine/BaseVolumeViewport.ts
+++ b/packages/core/src/RenderingEngine/BaseVolumeViewport.ts
@@ -804,19 +804,14 @@ abstract class BaseVolumeViewport extends Viewport {
     this.setBestOrentation(inPlaneVector1, inPlaneVector2);
 
     const { focalPoint, viewPlaneNormal } = this.getCamera();
-    const normalizedViewPlaneNormal = vec3.normalize(
-      vec3.create(),
-      viewPlaneNormal
-    );
     const deltaFocal = vec3.subtract(vec3.create(), point, focalPoint);
-    const alongNormal = vec3.dot(deltaFocal, normalizedViewPlaneNormal);
+    const alongNormal = vec3.dot(deltaFocal, viewPlaneNormal);
     const deltaNormal = vec3.scaleAndAdd(
       vec3.create(),
       focalPoint,
-      normalizedViewPlaneNormal,
+      viewPlaneNormal,
       alongNormal
     ) as Point3;
-
     this.setViewReference({
       FrameOfReferenceUID,
       cameraFocalPoint: deltaNormal,


nithin-trenser:
### Fix Summary

The issue occurred because `setViewReference()` was called without `viewPlaneNormal`, which triggered fallback logic that invokes `setViewPlane()`.

`setViewPlane()` then recursively calls `setViewReference()` while passing `viewPlaneNormal`, but omits `referenceImageId`. As a result, subsequent logic in `setViewReference()` that depends on `referenceImageId` fails due to the value being undefined.

The fix ensures `referenceImageId` is passed to `setViewPlane()`, so it is correctly forwarded during the recursive `setViewReference()` call.

@sedghi could you please review this PR ?

packages/core/src/RenderingEngine/BaseVolumeViewport.ts
```diff
@@ -807,6 +807,7 @@ abstract class BaseVolumeViewport extends Viewport {
       FrameOfReferenceUID,
       cameraFocalPoint: point,
```
wayfarer3130: The proposed solution only works if one has a referenced image id.  I believe the correct behaviour is to use the existing focal point from the camera and modify it with the difference between the current view plane normal and the specified value.
That is:
get the view plane normal/focal point from the camera right now
const deltaFocal = vec3.subtract(vec3.create(),point, focalPoint);
const alongNormal = vec3.dot(deltaFocal,viewPlaneNormal)
const deltaNormal = vec3.scaleAndAdd(vec3.create(), focalPoint, viewPlaneNormal, alongNormal)
then use deltaNormal as the new point.

wayfarer3130: You can test the difference between solutions by saving annotations in a non-acquisition orientation and then seeing the jumping for those ones.

nithin-trenser:
@wayfarer3130  The code has been refactored as per your instructions. Could you please verify the changes?

wayfarer3130 (approved):

PR #2541, opened 2026-01-07
fix: Livewire interpolation is weird

wayfarer3130:
### Context

Fixes [#5641]

Livewire interpolation currently interpolates the contour closure, which is a straight line originally.  That causes display issues.

### Changes & Results

Don't interpolate closure

### Testing

Open segmentation mode on a volume
Active livewire with interpolation on (see 5641)
Draw contours around the outside of a body or other hard defined object separated by several slices
View the intermediate slices - the contour closure should be a straight line if the original contour was closed by a straight line and livewire if it was clsoed by clicking on the starting point.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  

### DIFF
--- a/packages/tools/src/tools/annotation/LivewireContourTool.ts
+++ b/packages/tools/src/tools/annotation/LivewireContourTool.ts
@@ -384,9 +384,6 @@ class LivewireContourTool extends ContourSegmentationBaseTool {
       movingTextBox: false,
     };
 
-    const enabledElement = getEnabledElement(element);
-    const { renderingEngine } = enabledElement;
-
     this._activateModify(element);
     triggerAnnotationRenderForViewportIds(viewportIdsToRender);
     evt.preventDefault();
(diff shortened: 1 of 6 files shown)


wayfarer3130:
@jbocce - my test cases were a little too consistent - when I vary the number of control points more it seems that it doesn't always interpolate between the control points correctly to lave a straight line across the end.  The current change improves things, but I will need to think if there is a solution generally or not.

wayfarer3130:
@aimee-ferreira - I just merged this PR into CS3D.  The test class to use is  interpolationcontoursegmentation.
You need to use the livewire interpolation tool.  If you close by double click, with the tool showing a straight line (isUShapedOpenContour 'lineSegment'), then you need to match that with another contour several images away ALSO with a straight line segment.  
If you close by clicking on the first point, then you need to match with another contour also clicking on the same point.

PR #2544, opened 2026-01-08
fix: add pixel spacing extraction fallback to Imager Pixel Spacing tag

julianlinaress:
### Context

In some MG datasets, `PixelSpacing (0028,0030)` is missing but
`ImagerPixelSpacing (0018,1164)` is present (e.g., 0.1, 0.1). The loader
currently only checks `0028,0030`, so `imagePlaneModule.usingDefaultValues`
becomes `true`, `hasPixelSpacing` is `false`, and measurement tools fall back
to `px`. This change adds a fallback to read `0018,1164` so pixel spacing is
recognized and measurements use real-world units.

No issue was filed; this was identified while debugging `LengthTool` units
showing `px` despite valid spacing in the DICOM header.

### Changes & Results

- Add a fallback to read `ImagerPixelSpacing (0018,1164)` when
  `PixelSpacing (0028,0030)` is missing in WADO-URI metadata extraction.
- Before: `hasPixelSpacing = false`, measurements display in `px`.
- After: `hasPixelSpacing = true`, measurements use mm based on 0.1 spacing.
- No UI or public API changes.

### Testing

Manual:
1. Load an MG study that lacks `0028,0030` but has `0018,1164 = [0.1, 0.1]`.
2. Inspect `imagePlaneModule.pixelSpacing` and verify it is `[0.1, 0.1]`.
3. Use `LengthTool` and confirm the unit is `mm` (not `px`).
4. Load a study that already provides `0028,0030` and confirm behavior is unchanged.

Automated tests not added (metadata extraction path only).

#### PR

#### Code

  etc.).

  additions or removals. (N/A, no public API changes)

#### Tested Environment

- [x] "OS: Linux 6.6.87.2-microsoft-standard-WSL2"
- [x] "Node version: 24.5.0"
- [x] "Browser: N/A (not tested)"

### DIFF
--- a/packages/dicomImageLoader/src/imageLoader/wadouri/metaData/extractPositioningFromDataset.ts
+++ b/packages/dicomImageLoader/src/imageLoader/wadouri/metaData/extractPositioningFromDataset.ts
@@ -158,6 +158,12 @@ function extractSpacingFromDataset(dataSet: DataSet) {
     );
   }
 
+  // If pixelSpacing not valid to this point, trying to get the spacing
+  // from the Imager Pixel Spacing tag
+  if (!pixelSpacing && dataSet.elements.x00181164) {
+    pixelSpacing = getNumberValues(dataSet, 'x00181164', 2);
+  }
+
   return pixelSpacing;
 }
 


julianlinaress:
Oh, ok. I did not see any warning in the viewer I used for comparison. Since there’s no existing pattern, how would you want this represented in Cornerstone’s data model? Should we add a calibration marker/flag for Imager Pixel Spacing usage, or should we avoid using that tag entirely?

wayfarer3130:
See: https://viewer-dev.ohif.org/viewer?StudyInstanceUIDs=1.3.6.1.4.1.14519.5.2.1.4792.2001.105216574054253895819671475627
It shows as "Proj", meaning projected.  There is a full PR that fixes this, but it isn't quite ready for merge yet:
https://github.com/cornerstonejs/cornerstone3D/pull/2494
You could do just the minimal required changes, but I still think it will take a while.

What I'm hoping to do is split the changes out into `@cornerstone/metadata` and share them with the various versions so it works consistently for all data sources, and that we can release a version but not use it int he OHIF 3.12 release until we have time to fully test it.

packages/dicomImageLoader/src/imageLoader/wadouri/metaData/extractPositioningFromDataset.ts
wayfarer3130, line 1: This is invalid behaviour and is quite dangerous.
The problem is that there is no warning that this measurement isn't a body measurement at all, but is strictly larger than body measurement.   I'm aware of at least 1 case of operating on the wrong spot on the body because of this type of issue.  Look at ERMF settings/values.
There will be a display of this in CS3D 5.0, but it will include the appropriate warnings to indicate that this is imager measurement and not body measurement.

PR #2557, opened 2026-01-13
feat: add results for different viewports on the fusion view for rectangle/circle ROI

wayfarer3130:
### Context

This PR adds display of measurements from multiple volumes onto a fused display.

Note this currently only works for acquisition orientation on volume viewports because the restrictions for display of the annotations are such that they don't work on the secondary viewports for non-acquisition.  That is being fixed as a separate PR/issue.

### Changes & Results

Add a generic method for get text lines
Add support for displaying stats from different volumes or multiple volumes
Add a getMeasurementTargets to allow getting a list of volumes to use for this.

### Testing

Run tmtv mode before/after the change
Create a rectangle roi on each of the three orientations.
In the old version, only one measurement from both volumes would appear, whereas now both of them appear.

Note: Area measurement is unchanged - it is incorrect on both old/new.

Try again with circular measurement.

For annotations drawn on the center row, you may need to navigate the top row some as the distance between annotations and thus the distance that a single annotation might be shown changes depending on the differing slice thickness.

Annotations must have been displayed in both top and second row at some point in order to allow displaying measurements in the bottom row.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  

## Summary by CodeRabbit

* **New Features**
  * Added dynamic TMTV layouts, including fusion and oblique arrangements, with layout-aware synchronization.
  * Added modality- and volume-based measurement targeting for multi-volume ROI statistics.
  * ROI measurements now display statistics across multiple fused volumes.
  * Added Generic Dynamic Volume and Generic Multi-Volume Measurements examples.
* **Bug Fixes**
  * Improved volume-specific image data resolution and annotation reference handling.
  * Improved ROI statistics caching and boundary detection.
* **Documentation**
  * Added guidance for configuring measurement targets and multi-volume statistics.


packages/core/src/RenderingEngine/BaseVolumeViewport.ts
```diff
@@ -639,7 +639,7 @@ abstract class BaseVolumeViewport extends Viewport {
   ): ViewReference {
     const target = super.getViewReference(viewRefSpecifier);
     const volumeId = this.getVolumeId(viewRefSpecifier);
-    if (viewRefSpecifier?.forFrameOfReference !== false) {
+    if (viewRefSpecifier?.forFrameOfReference !== true) {
```
wayfarer3130, line 682: This was a dumb reversal of the attribute check that we never tested or used previously, but now that we do it needed to be fixed to get the right value back.

packages/core/src/RenderingEngine/Viewport.ts
```diff
@@ -1126,22 +1126,6 @@ class Viewport {
       // stack viewports, but is technically incorrect.
       widthWorld = Math.max(spacing[0], widthWorld - spacing[0]);
       heightWorld = Math.max(spacing[1], heightWorld - spacing[1]);
-
```
wayfarer3130: This was unused code for demonstrating that the new/old methods of FOR were the same, just cleaning it up.

packages/tools/src/tools/annotation/BidirectionalTool.ts
wayfarer3130, line 1: There are a number of corrections around typing needed to allow getting hte right types on attributes.

packages/tools/src/utilities/planar/filterAnnotationsWithinSlice.ts
```diff
@@ -113,7 +113,7 @@ export default function filterAnnotationsWithinSlice(
   // Annotation should be within the slice, which means that it should be between
   // camera's focalPoint +/- spacingInNormalDirection.
 
-  const halfSpacingInNormalDirection = spacingInNormalDirection / 2;
+  const halfSpacingInNormalDirection = spacingInNormalDirection / 2 + EPSILON;
```
wayfarer3130, line 119: If the spacing is exactly half for the CT and PT, then the annotations don't appear in the center row - this fixes that issue.  Note you get annotations from all three nearby CT's on the PT image.

utils/ExampleRunner/example-info.json
```diff
@@ -535,6 +535,10 @@
         "name": "PET-CT Fusion + MIPLayout",
         "description": "PT-CT fusion layout with Crosshairs, and synchronized cameras, CT W/L and PET threshold"
       },
+      "tmtv": {
```
sedghi: don't we already have two ptct one? maybe do it in those examples? and we don't add one here?

packages/core/src/RenderingEngine/BaseVolumeViewport.ts
```diff
@@ -639,7 +639,7 @@ abstract class BaseVolumeViewport extends Viewport {
...
       target.volumeId = volumeId;
     }
     if (typeof viewRefSpecifier?.sliceIndex !== 'number') {
```
sedghi, line 685: please add a comment above this line explaining why it should be true in the code

packages/core/src/RenderingEngine/VolumeViewport.ts
```diff
@@ -757,7 +757,7 @@ class VolumeViewport extends BaseVolumeViewport {
   ): ViewReference {
     const viewRef = super.getViewReference(viewRefSpecifier);
     if (!viewRef?.volumeId) {
-      return;
+      return viewRef;
```
sedghi, line 794: please add a comment why (in code)

packages/tools/src/utilities/planar/filterAnnotationsWithinSlice.ts
sedghi: please add this exact comment in the code

packages/tools/src/tools/base/BaseTool.ts
```diff
@@ -319,6 +320,50 @@ abstract class BaseTool {
     );
   }
 
+  /**
```
sedghi, line 394: here getTargetId should be converted to getTargetIds, or make the return string[] and then render loop can loop over targetIds in their calculateCahcedStats

packages/tools/src/tools/base/BaseTool.ts
```diff
@@ -290,39 +427,310 @@ abstract class BaseTool {
...
+      candidates.push({
+        targetId,
+        referencedId,
+        displaySet: displaySetInfo?.displaySet,
+        displaySetUID: displaySetInfo?.displaySetUID,
+        instance,
+        modality: (instance?.Modality as string) ?? volume.metadata?.Modality,
+        imageIds,
+        index,
+      });
```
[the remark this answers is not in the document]
wayfarer3130: It IS the actor index that is the display set position.

sedghi (commented):
d

sedghi:
The target-selection API has been reworked per the review feedback - see #2802 for the alternative implementation (targets this branch, so its diff is just the architecture change). `targetsFilter` is now a `{ key, options }` spec resolved through the `annotationTargetFilter` metadata providers, and the static `targetFilters` helpers are removed from the tool classes.

packages/docs/docs/concepts/cornerstone-tools/annotation/measurementTargets.md
```diff
@@ -0,0 +1,255 @@
...
+display sets of the viewport and can therefore also cause statistics to be
+computed for targets no viewport has computed yet. For backward
+compatibility, a configured `isPreferredTargetId` is honoured **before** the
+filter (including the ROI tools' default filter), so configurations that
+predate `targetsFilter` keep their behaviour.
+
+:::note
+Multi-target selection currently only works for volumes displayed on screen.
+Stack-based fusion and targets not currently displayed are not yet supported.
+:::
```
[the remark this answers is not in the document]
wayfarer3130: The comment by coderabbitai here is wrong.

packages/docs/docs/concepts/cornerstone-tools/annotation/measurementTargets.md
```diff
@@ -135,15 +150,27 @@ those two modalities:
...
+  targetPredicate: measurementTargetFilters.forModality('CT', 'PT'),
+});
+```
+
+Just the first pixel-data target (the pre-5.x single-target behaviour),
+referencing the chooser by name:
+
+```ts
+toolGroup.addTool(CircleROITool.toolName, {
+  targetsFilter: 'firstPixelData',
```
sedghi: these should all be typed and not strings

wayfarer3130: It can be a named string or a function - let me update the PR to use the function version, and add a comment to indicate it can be the string name of it in order to allow declarative applications.

wayfarer3130: Removed the string setting capability - will do that in OHIF

sedghi (approved):
i think docs should be updated though

PR #2560, opened 2026-01-15
Feat: Allow the ROI StartEndThreshold tools to choose their measurement target

theoc0702:
### Context

The `RectangleROIStartEndThresholdTool` and the `CircleROIStartEndThresholdTool` could not
select a specific volume in a fusion viewport, for example a PET/CT viewport. The two tools
used the default target of the viewport. Therefore the user could not run a threshold
segmentation on the PET volume without manual work.

This pull request keeps that intent. The branch is now rebased on `main`, and the two tools
use the measurement target selection that #2557 added. Each tool no longer has its own
`getTargetId`. Both tools inherit `BaseTool.getMeasurementTargets`, which builds the
candidate targets from the actors and the display sets of the viewport. The
`targetsFilter` chooser and the `targetPredicate` then select among the candidate targets.
See `measurementTargetFilters`. This mechanism replaces the deprecated
`isPreferredTargetId` configuration.

This mechanism also removes a second, parallel selection mechanism, and it makes the
feature work in a case that the first implementation could not. The deprecated
`isPreferredTargetId` reads only the statistics that the tool already computed
(`data.cachedStats`). Therefore `isPreferredTargetId` could never apply to an annotation
that the user just drew. The candidates behind `targetsFilter` come from the viewport.
Therefore the tool selects the PT volume at the first draw.

### How the two selectors get defined, and what the two selectors do

`packages/tools/src/tools/base/measurementTargetFilters.ts` defines the ready made
selectors. `packages/tools/src/types/IBaseTool.ts` defines the types. The package exports
the selectors as `measurementTargetFilters`.

The selection has two halves, and each half is a plain function. You set each half on the
tool configuration:

```ts
import { measurementTargetFilters } from '@cornerstonejs/tools';

toolGroup.addTool(RectangleROIStartEndThresholdTool.toolName, {
  targetsFilter: measurementTargetFilters.firstPixelData,
  targetPredicate: measurementTargetFilters.forModality('PT'),
});
```

- `targetPredicate` is a `MeasurementTargetPredicate`. The function reads one candidate, and
  the function returns `true` when the tool can measure that candidate. The function decides
  the eligibility of one candidate only.
- `targetsFilter` is a `MeasurementTargetsFilter`. The function reads the full list of
  candidates and the options, and the function returns the subset that the tool measures.
  The function decides the number of the targets.

**The candidates.** `BaseTool.getMeasurementTargetCandidates` builds the list. The method
reads the actors of the viewport in viewport order, and the method makes one candidate for
each actor that has a `referencedId` and a cached volume. A segmentation representation
stays in the list, and that candidate holds a `representationUID`. If the viewport has no
such actor, the method makes one candidate from the default view reference of the viewport.
Each candidate holds the `targetId`, the `referencedId`, the `index`, and the fields of the
display set where the metadata gives them: `displaySet`, `displaySetUID`, `instance`,
`modality` and `imageIds`. A candidate that comes from an unknown display set holds no
`modality`.

**The order of the steps.** `BaseTool.getMeasurementTargets` does this:

1. The deprecated `isPreferredTargetId` runs first, but only when the configuration holds
   `isPreferredTargetId` and the annotation holds `cachedStats`. The method returns the
   first targetId that `isPreferredTargetId` accepts.
2. If the configuration holds a `targetsFilter`, the method builds the candidates, and the
   method calls `targetsFilter(candidates, options)`. The `options` hold the viewport, the
   tool configuration and the annotation data. The method returns the `targetId` of each
   candidate that the chooser gives back.
3. If the configuration holds no `targetsFilter`, the method returns the default view
   reference of the viewport, which is the first volume actor.

`getTargetId` returns element 0 of that list. The two tools of this pull request measure one
target, so the two tools use `getTargetId`. An empty list gives no target, and then the tool
computes no statistics.

**The ready made choosers** (`targetsFilter`):

| Chooser | Result | Does the chooser use `targetPredicate`? |
| --- | --- | --- |
| `firstPixelData` | The first eligible candidate | Yes |
| `allPixelData` | Every eligible candidate | Yes |
| `first` | The first candidate | No |
| `all` | Every candidate | No |

**The ready made predicates** (`targetPredicate`):

| Predicate | The predicate keeps a candidate when |
| --- | --- |
| `isPixelData` | The candidate holds no `representationUID`, and the `modality` of the candidate is not one of `SEG`, `RTSTRUCT`, `RTPLAN`, `SR`, `PR`, `KO`. A candidate with no `modality` stays. |
| `forModality('PT')` | The `modality` of the candidate is one of the given modalities. A candidate with no `modality` goes out. |
| `forId(id)` | The `displaySetUID`, the `referencedId` or the `targetId` of the candidate holds the given text. The test is a test of a substring. |

`firstPixelData` and `allPixelData` call `isEligible`. `isEligible` applies `isPixelData`
first, and then `isEligible` applies the configured `targetPredicate`. Therefore these two
choosers never measure a segmentation, whatever result the `targetPredicate` gives. `first`
and `all` do not call `isEligible`, so `first` and `all` ignore the `targetPredicate`.

**The configuration of this pull request.** The example configures `firstPixelData` with
`forModality('PT')`. On the `CT + PT fusion` view the candidates are the CT volume and then
the PT volume. `forModality('PT')` refuses the CT candidate, `firstPixelData` gives back the
PT candidate, and the tool measures the PT volume. On the `PT only` view the PT volume is
the one candidate, and the same configuration gives the same volume that the default
selection gives.

### Changes & Results

**`AnnotationTool.ts`**

- The `getTargetVolume` helper is here. `RectangleROITool` and `CircleROITool` both extend
  `AnnotationTool`, so `AnnotationTool` is the first shared parent of the two
  StartEndThreshold tools. The helper resolves the targetId and the volume of the target
  together, and the helper returns `null` when the viewport holds no target for the tool.
- The abstract `addNewAnnotation` returns `Annotation | null`. `null` means that the tool
  created no annotation. A tool must not return `undefined`, so the caller can tell a
  deliberate empty result from a missing return value.

**`RectangleROIStartEndThresholdTool.ts` and `CircleROIStartEndThresholdTool.ts`**

- The two tools drop the tool specific `getTargetId` override. The inherited
  implementation selects the target, and the implementation obeys the
  `targetsFilter` and the `targetPredicate` configuration. With no configured filter the
  behaviour does not change: the tool uses the default view reference of the viewport.
- `addNewAnnotation`, `_endCallback` and `_calculateCachedStatsTool` call the inherited
  `getTargetVolume`. The helper takes the viewport alone. The `cachedStats` of these two
  tools is a flat `VolumeStats` (`pointsInVolume`, `statistics`), and not a map with a
  targetId as the key. Therefore the annotation holds no `volumeId:` key that the target
  selection can reuse, and the viewport alone decides the target.
- `addNewAnnotation` resolves the target before the tool sets any drawing state, and
  `addNewAnnotation` returns `null` when a configured filter selects no target on the
  viewport. An example is a PT only filter on a CT viewport of the same tool group. The
  tool computes no statistics, the tool does not throw an error, and the tool is not left
  in a draw state that `cancel()` cannot unwind.
- `renderAnnotation` asks the viewport with `hasVolumeId` instead of a loop over
  `getAllVolumeIds()`.
- `_calculateCachedStatsTool` clears `annotation.invalidated` when the method computes no
  statistics. The render callback calls the method for every invalidated annotation, so a
  flag that stays set repeats the work on every frame.
- The statistics read the `Modality` from the volume that the tool measures. The annotation
  metadata holds no `Modality`, and the PT statistics were unitless instead of SUV.
- `getCalibratedLengthUnitsAndScale` gets the index coordinates of the handles in both
  tools. The function compares each handle against the pixel bounds of the ultrasound
  regions of the image, and the other ROI tools give the function index coordinates. The
  Rectangle tool gave the name `data.habdles`, which is a spelling mistake and gave the
  function `undefined`. The Circle tool gave the world points. A volume of the modality CT
  or PT holds no ultrasound region, so the calibrated `area` and `areaUnit` of these
  examples do not change.
- `getVolumeId` replaces the `targetId.split(/volumeId:|\?/)[1]` parsing. The public
  `getAllVolumeIds()` and `hasVolumeId()` replace the private `viewport.volumeIds` access
  that needed a `@ts-ignore`.

**`mouseDownActivate.ts` and `touchStartActivate.ts`**

The two dispatchers accept a tool that returns `null` from `addNewAnnotation`. A tool
returns `null` when the tool has no target on the viewport, which is not an error. The two
dispatchers logged `Error adding new annotation, viewport not ready` for that no-op.

**`rectangleROIStartEndThresholdWithSegmentation` example**

- The example has a `View` dropdown. The default view shows the PT volume alone, and the
  `CT + PT fusion` view adds the CT of the same study.
- The fusion view cannot be the default view. The CT covers more of the patient than the PT,
  so a viewport that holds both volumes resets its camera to the combined bounds, and a
  given slice index shows a different PT slice.
  `tests/rectangleROIThresholdStatisticsMIM.spec.ts` compares the statistics of this example
  against MIM reference values at fixed slice indices, and those values apply to the PT
  volume alone. The example keeps the water phantom study, which holds the PT series that
  MIM measured.
- The example configures `measurementTargetFilters.firstPixelData` with
  `forModality('PT')`. The fusion view adds the CT volume before the PT volume, so the CT
  volume is the default measurement target of the viewport, and this configuration selects
  the PT volume instead. The configuration matches how the `petCt` example configures the
  same selection.
- The `Run Segmentation` button stops when the user drew no annotation, and when the
  annotation covers no voxel of the PT volume. A relative threshold has no maximum in the
  second case, and the relative value is not an absolute threshold.

**`measurementTargets.jest.js`**

The regression tests cover both tools:

- the default target when the configuration holds no filter,
- the selection of the PT volume on a fusion viewport by modality,
- no target when the filter matches nothing.

**`rectangleROIStartEndThresholdFusionTarget.spec.ts`**

This new test selects the `CT + PT fusion` view, and the test confirms two facts. The CT
volume is the first actor of the viewport, and therefore the default measurement target.
The statistics of an annotation that the test draws hold `Modality: PT`. The two mobile
projects do not run this test, as the two mobile projects already do not run the MIM test.

**`PlanarLegacyCompatibilityController.ts` (a fix of the core package)**

The `Playwright (compatibility)` check showed a fault of the core package. The compatibility
mode opens each example with `?type=next`, which puts `PlanarViewportLegacyAdapter` behind
the legacy API. In that mode `setVolumesForViewports` threw when the call mounted a volume
that the viewport already held. The viewport then kept no actor, and the page showed this
error:

```
[PlanarViewport] No registered planar dataset metadata for __planar_v2__:PT_AXIAL:volume:cornerstoneStreamingImageVolume:PT_VOLUME_ID
```

The smallest case uses no fusion view, and the smallest case uses neither of these two
tools. A second call of `setVolumesForViewports`, with the volume that the viewport already
holds, gave these results before the fix:

| Mode | Result of the call | The actors of the viewport after the call |
| --- | --- | --- |
| legacy | the call succeeds | the PT volume |
| compatibility | the call throws the error above | none |

`mountVolumes` registers the dataset of each volume, and then `mountVolumes` calls
`setDisplaySets`. `setDisplaySets` removes each binding of the viewport before the viewport
mounts the new bindings, and `PlanarViewportLegacyAdapter.removeData` removes the
registration of the dataset. Therefore the registration of a volume that the caller mounts
again was gone, and `loadPlanarData` threw. `setStack` does not have this fault, because
`setStack` removes the bindings before `setStack` registers the dataset.

`mountVolumes` now marks each data id of the mount, and `removeData` keeps the registration
of a marked data id. A finally block clears each mark, because the catch of `mountVolumes`
removes the data of a mount that fails.

The effect on this pull request: the fusion view held the CT volume alone, the candidates
held no PT volume, `forModality('PT')` selected no target, and the tool created no
annotation. The fusion view now holds the CT volume, the PT volume and the labelmap, and the
test passes in the compatibility mode.

**`PlanarLegacyCompatibilityController.jest.js`**

Two new tests cover the fix of the core package:

- the registration of a volume survives a mount of that volume that repeats,
- a mount that fails removes the data of the volume.

**Effect**

The user can now configure these two tools to select the correct volume, for example the PT
series, for the statistics and for the threshold operations in a fusion viewport.

*Note: the two tools still measure one target only. `CircleROITool` and `RectangleROITool`
show the statistics of the CT volume and of the PT volume together, but these two tools do
not, because these two tools override `renderAnnotation` and `getTextLines` completely. A
change to that behaviour is a larger follow up.*

### Testing

1. Run the example `rectangleROIStartEndThresholdWithSegmentation`.
2. Select `CT + PT fusion` in the `View` dropdown.
3. Draw an ROI on the fused image.
4. Look at the console and at the UI. The tool calculates the statistics and does the
   segmentation on the **PT volume**, as `index.ts` configures, and not on the CT volume.
5. `npx jest packages/tools/test/measurementTargets.jest.js` covers the target selection of
   both tools.
6. `npx playwright test rectangleROIThresholdStatisticsMIM rectangleROIStartEndThresholdFusionTarget`
   covers the MIM reference values of the default view, and the PT target of the fusion
   view. The 5 tests pass on the `chromium` project. The `webkit` project is not tested
   here, because this machine holds no WebKit browser of Playwright.
7. `PLAYWRIGHT_FORCE_COMPAT=true npx playwright test rectangleROIStartEndThresholdFusionTarget`
   covers the compatibility mode. This test failed before the fix of the core package.
8. `npx jest packages/core/test/PlanarLegacyCompatibilityController.jest.js` covers the fix
   of the core package.

*Note: this pull request changes the Circle tool for consistency of the code. The
verification used the Rectangle tool in the example.*

#### PR

#### Code

#### Tested Environment

- [x] OS: Linux Ubuntu 24.04.3 LTS
- [x] Node version: v24.11.0
- [x] Browser: Chrome v143.0.7499.169

## Summary by CodeRabbit

* **New Features**
  * Added CT + PT fusion viewing to the rectangle ROI threshold example, with PT-only mode selected by default.
  * Added a perfusion color display preset for fusion visualization.
  * ROI measurements in fusion mode now target the PT volume for threshold statistics.

* **Bug Fixes**
  * Prevented annotation and measurement errors when no suitable volume is available.
  * Corrected calibration and modality handling so measurements use appropriate units.
  * Added safeguards for empty annotations and invalid measurement results.
  * Improved volume remounting reliability and cleanup after failed mounts.


salimkanoun:
hmmm we will need help, really don't understand why tests are failing this PR shouldn't change behavior if you don't use isPreferredTargetId

packages/tools/src/tools/segmentation/CircleROIStartEndThresholdTool.ts
```diff
@@ -113,6 +113,30 @@ class CircleROIStartEndThresholdTool extends CircleROITool {
     }
   }
 
+  protected getTargetId(
```
wayfarer3130: This should generally be in the parent class, not be specific to the tool as it is a general behaviour.

packages/tools/src/tools/segmentation/RectangleROIStartEndThresholdTool.ts
```diff
@@ -111,6 +111,30 @@ class RectangleROIStartEndThresholdTool extends RectangleROITool {
     }
   }
 
+  protected getTargetId(
```
wayfarer3130: If it were in the base class, it would just get used by both tools and you wouldn't be duplicating code.

salimkanoun: I'm not sure it is feasible, the RectangleTool uses getTargetId for baseTools which is based on imageId.
Here as we are in 3D we work on a volume, there is no imageId for each slice (can be reconstructed axis)

I think we are missing a layer of abstraction, some tools that are expected to work on image each one with an imageId should have a different base class than 3D tools that will rely on volume definition.

Unfortunately it needs to rework all tools codes to make correct inheritance ..

wayfarer3130 (approved):

wayfarer3130 (approved):

PR #2561, opened 2026-01-15
fix(drawLine): SVG line disappeared on Firefox

Marsevil:
### Context

**On Firefox only**

When updating the image index shown in a `StackViewport` with a reference line (not set as current viewport), the reference line disappeared in Firefox.

> Can't reproduce in a minimal example. Seems to depend on the redrawing order.

### Changes & Results

By setting the filter through the `filter` attribute instead of the `style` attribute, everything is displayed correctly.

I would like to know if there was a reason the `filter` was defined through the `style` attribute. Could there be a bug somewhere else ?

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- OS: Windows 11
- Node version: 24.12
- Browser: Firefox 
  


sedghi (approved):
Makes sense

PR #2567, opened 2026-01-19
fix(security): Fix CVE-2024-28863 and CVE-2026-23745. Also updated lerna to fix circleci npm publish error.

jbocce:
### Context
Title says most of it.

See https://github.com/advisories/GHSA-f5x3-32g6-xq36 and https://github.com/advisories/GHSA-8qq5-rm4j-mr97

Also see https://app.circleci.com/pipelines/github/cornerstonejs/cornerstone3D/9419/workflows/fb1f24ed-a231-465e-9f4f-3849bfc5cb74/jobs/24959
 

### Changes & Results
Updated lerna to 9.0.3

I did an analysis with AI to ensure that the only dependencies introduced relate to the change in lerna

[bun-lerna-dependency-report.md](https://github.com/user-attachments/files/24719349/bun-lerna-dependency-report.md)
[yarn-lerna-dependency-report.md](https://github.com/user-attachments/files/24719351/yarn-lerna-dependency-report.md)

### Testing
It appears that at runtime, tar is used via @itk-wasm/dam to do segment interpolation. So we should ensure the following still behave as expected. Note that the deploy links below were created based on the patched code.

https://cs3ddocs.netlify.app/live-examples/labelmapinterpolation
https://cs3ddocs.netlify.app/live-examples/interpolationcontoursegmentation

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

  System:
    OS: Windows 11 10.0.26200
    CPU: (20) x64 12th Gen Intel(R) Core(TM) i7-12700H
    Memory: 4.16 GB / 31.68 GB
  Binaries:
    Node: 22.17.0 - C:\Users\joebo\AppData\Local\fnm_multishells\37464_1768400684303\node.EXE
    Yarn: 1.22.22 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
    npm: 11.6.2 - C:\Users\joebo\AppData\Local\fnm_multishells\37464_1768400684303\npm.CMD
    bun: 1.2.23 - C:\Users\joebo\.bun\bin\bun.EXE
  Browsers:
    Chrome: 143.0.7499.193
    


wayfarer3130:
The tar usage of @itk-wasm/dam occurs during the transfer of data between webworker/main system.  I believe that is NOT how we are doing the transfer itself, which means we could potentially just exclude @itk-wasm/dam  - @sedghi  - would that be a better option, and does it work?  As well, since it is in-memory and not to file, neither vulnerability applies.

jbocce:
> The tar usage of @itk-wasm/dam occurs during the transfer of data between webworker/main system. I believe that is NOT how we are doing the transfer itself, which means we could potentially just exclude @itk-wasm/dam - @sedghi - would that be a better option, and does it work? As well, since it is in-memory and not to file, neither vulnerability applies.

Thanks @wayfarer3130. Excluding a transitive dependency sounds like overkill and might not work. I would say the alternative is to dismiss/ignore the vulnerability as we already do for another vulnerability.

Issue #2458, opened 2025-11-12
[Bug] Clicking VOI buttons for the volumevoisigmoid example do nothing


jbocce:
### Describe the Bug

For the Sigmoid VOI Volume example, clicking either the Set Linear VOI or the Set Sigmoid VOI buttons does NOTHING.

### Steps to Reproduce

1. Visit [https://www.cornerstonejs.org/live-examples/volumevoisigmoid](<https://www.cornerstonejs.org/live-examples/volumevoisigmoid>)
2. Click the Set Linear VOI button.
3. Click the Set Sigmoid VOI button.

### The current behavior

Clicking does nothing for either button. No exceptions in the console.

### The expected behavior

The VOI should change.

### System Information

System:

OS: Windows 11 10.0.26200<br>CPU: (20) x64 12th Gen Intel(R) Core(TM) i7-12700H<br>Memory: 9.55 GB / 31.68 GB

Binaries:

Node: 23.9.0 - \~\\AppData\\Local\\fnm_multishells\\14296_1762952391211\\node.EXE<br>Yarn: 1.22.22 - C:\\Program Files (x86)\\Yarn\\bin\\yarn.CMD<br>npm: 10.9.2 - \~\\AppData\\Local\\fnm_multishells\\14296_1762952391211\\npm.CMD<br>bun: 1.2.23 - \~.bun\\bin\\bun.EXE

Browsers:

Chrome: 142.0.7444.163 (Official Build) (64-bit)

justinterveystech:
I went through the code and saw that

      1.Neither button changes window width/center
      2.Neither button resets or constrains the VOI range
      3.The WindowLevelTool continues to dominate visual behaviour ( Window/Level interaction is enabled by default via WindowLevelTool)

So visually:

Buttons don't change anything upon clicked
Dragging causes contrast change in all cases
The difference between linear and sigmoid LUTs is subtle
Users reasonably conclude “nothing changed”

Why does no button appear “active”?

Button state is not tracked , No UI feedback exists , No overlay or label indicates the active VOI mode
So even though internal state does change, the UI gives zero confirmation.


=======SUGGESTION=======

The suggestion was to apply a deliberately constrained VOI range (That suits the functions) when switching modes:

viewport.setProperties({
  VOILUTFunction: Enums.VOILUTFunctionType.SAMPLED_SIGMOID,
  voiRange: { lower: 300, upper: 1200 }
});

and

[...]

PR #2579, opened 2026-01-27
 Applied a deliberately constrained VOI range

justinterveystech:
Context

bugfix for https://github.com/cornerstonejs/cornerstone3D/issues/2458 
I went through the code and saw that

      1.Neither button changes window width/center
      2.Neither button resets or constrains the VOI range
      3.The WindowLevelTool continues to dominate visual behaviour ( Window/Level interaction is enabled by default via WindowLevelTool)

So visually:

Buttons don't change anything upon clicked
Dragging causes contrast change in all cases
The difference between linear and sigmoid LUTs is subtle
Users reasonably conclude “nothing changed”

Why does no button appear “active”?

Button state is not tracked , No UI feedback exists , No overlay or label indicates the active VOI mode
So even though internal state does change, the UI gives zero confirmation.

Changes & Results

User get the direct feed back of change when explicitly setting voiRange: Input range becomes fixed. Only the mapping function changes. The effect of the LUT becomes observable
It makes Sigmoid VOI perceptually distinguishable

By constraining voiRange:

Linear VOI → harsh clipping + uniform contrast
Sigmoid VOI → smooth transitions + mid-range emphasis
It creates an immediate, visible response to button clicks

Testing
User can go to the same volumesigmoidexample and click either buttons .Change is directly visible and they can still do drag operations

Checklist

Code

  etc.)

Tested Environment
OS:Winodws
Browser :Chrome and edge

### DIFF
--- a/packages/core/examples/volumeVoiSigmoid/index.ts
+++ b/packages/core/examples/volumeVoiSigmoid/index.ts
@@ -50,8 +50,10 @@ addButtonToToolbar({
     ) as Types.IVolumeViewport;
 
     // Set a range to highlight bones
-    viewport.setProperties({ VOILUTFunction: Enums.VOILUTFunctionType.LINEAR });
-
+    viewport.setProperties({
+      VOILUTFunction: Enums.VOILUTFunctionType.LINEAR,
+      voiRange: { lower: -1000, upper: 3000 },
+    });
     viewport.render();
   },
 });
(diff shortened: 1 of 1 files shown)


sedghi:
The LINEAR button sets voiRange {lower:-1000, upper:3000} while the SIGMOID button sets {lower:300, upper:1200}. Since the PR's stated purpose is to make the linear-vs-sigmoid LUT difference observable, using two different ranges confounds the comparison: the visual difference is now driven mostly by the differing range, not by the LUT function. Using the same range for both would isolate the LUT effect the example is meant to demonstrate.

Am i missing something?

PR #2586, opened 2026-01-29
Feat/orientation control tool

mbellehumeur:
### Context

ticket:  18376011052

'Orientation controller' for volume 3D viewports.

Note:  The example for this tool is in the volume cropping tool example:
bun run example volumeCroppingTool

### Changes & Results

Attempted to follow  VTK.js-based widget architecture to facilitate porting  to VTK.js:.

1. **`vtkOrientationControllerWidget`** (`packages/tools/src/utilities/vtkjs/OrientationControllerWidget/`)
   - Core widget managing VTK actors and interaction
   - Creates and manages `vtkActor` instances for the polyhedron
   - Handles mouse picking via `vtkCellPicker`
   - Manages actor lifecycle (add/remove from viewports)
   - Handles positioning and sizing relative to viewport corners

2. **`AnnotatedRhombicuboctahedronActor`** (`packages/tools/src/utilities/vtkjs/AnnotatedRhombicuboctahedronActor/`)
   - Generates the 26-faced polyhedron geometry
   - Creates VTK actors with textured faces and labels
   - Supports main faces (6), edge faces (12), and corner faces (8)
   - Applies anatomical labels (L/R, A/P, S/I) using LPS convention

3. **`OrientationController` Tool** (`packages/tools/src/tools/OrientationController.ts`)
   - Cornerstone3D tool extending `BaseTool`
   - Wraps the VTK widget for integration
   - Manages tool lifecycle (enable/disable)
   - Handles configuration updates and viewport synchronization

### Testing

The example was added to the cornerstone volume cropping tool example since it applies only to VRT viewport.

Test in OHIF here:
https://youtu.be/OV3VP9rfzz4

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  

### DIFF
--- a/packages/tools/examples/volumeCroppingTool/index.ts
+++ b/packages/tools/examples/volumeCroppingTool/index.ts
@@ -173,7 +173,7 @@ instructions.innerText = `
   - Use the scroll wheel to scroll through the slices in the orthographic viewports.
   - Toggle the clipping planes, handles, and rotate clipping planes on drag.
   - Click on the faces/edges/corners of the beveled cube orientation widget to change the orientation.
-  URL params: numViewports=1|2|3, colorScheme=gray|rgy|marker, letterColorScheme=mixed|white|black, keepOrientationUp=true|false
+  URL params: numViewports=1|2|3
   `;
 
 content.append(instructions);
@@ -263,50 +263,6 @@ function getNumViewportsFromUrl() {
     return num;
   }
   return 3; // default
-}
-
-/**
- * Get the color scheme from the URL (?colorScheme=gray|rgy|marker)
- */
-function getColorSchemeFromUrl(): 'gray' | 'rgy' | 'marker' {
-  const params = new URLSearchParams(window.location.search);
-  const value = params.get('colorScheme');
-  if (value === 'gray' || value === 'rgy' || value === 'marker') {
-    return value;
-  }
-  // Check for legacy grayColors parameter for backward compatibility
-  const grayColors = params.get('grayColors');
-  if (grayColors === 'true') {
-    return 'gray';
-  }
-  return 'rgy'; // default
-}
-
-/**
- * Get the keepOrientationUp value from the URL (?keepOrientationUp=true|false)
- */
-function getKeepOrientationUpFromUrl(): boolean {
-  const params = new URLSearchParams(window.location.search);
-  const value = params.get('keepOrientationUp');
-  if (value === 'true') {
-    return true;
-  }
-  if (value === 'false') {
-    return false;
-  }
-  return true; // default
-}
-
-/**
- * Get the letter color scheme from the URL (?letterColorScheme=mixed|white|black)
- */
-function getLetterColorSchemeFromUrl(): 'mixed' | 'awhite' | 'black' {
-  const params = new URLSearchParams(window.location.search);
-  const value = params.get('letterColorScheme');
-  if (value === 'mixed' || value === 'white' || value === 'black') {
-    return value;
-  }
-  return 'mixed'; // default
 }
 
 /**
@@ -446,10 +402,9 @@ async function run(numViewports = getNumViewportsFromUrl()) {
     ],
   });
 
-  // Get color scheme from URL (needed for both orthographic and 3D viewports)
-  const colorScheme = getColorSchemeFromUrl();
-  const keepOrientationUp = getKeepOrientationUpFromUrl();
-  const letterColorScheme = getLetterColorSchemeFromUrl();
+  const colorScheme: 'rgy' | 'gray' | 'marker' = 'rgy';
+  const keepOrientationUp = true;
+  const letterColorScheme: 'mixed' | 'white' | 'black' = 'mixed';
 
   toolGroup.addTool(OrientationControllerTool.toolName, {
     colorScheme,
@@ -492,29 +447,19 @@ async function run(numViewports = getNumViewportsFromUrl()) {
   // Enable OrientationControllerTool after viewport is added and volume is loaded
   toolGroupVRT.setToolEnabled(OrientationControllerTool.toolName);
 
-  // Add dropdown for orientation control colors (reloads page with URL param)
+  // Add dropdown for orientation control colors
   const colorSchemeValues: string[] = ['rgy', 'gray', 'marker'];
   const colorSchemeLabels = ['RGY', 'Gray', 'Marker'];
-  // Ensure colorScheme is valid, default to 'rgy' if not
-  const validColorScheme = colorSchemeValues.includes(colorScheme)
-    ? colorScheme
-    : 'rgy';
 
   addDropdownToToolbar({
     labelText: 'Orientation Control: Colors',
     options: {
       values: colorSchemeValues,
-      defaultValue: validColorScheme,
+      defaultValue: colorScheme,
       labels: colorSchemeLabels,
     },
     container: controlRow,
     onSelectedValueChange: (selectedValue) => {
-      const url = new URL(window.location.href);
-      url.searchParams.set('colorScheme', String(selectedValue));
-      // Remove legacy parameter if present
-      url.searchParams.delete('grayColors');
-      window.history.replaceState({}, '', url);
-
       const toolGroupVRT =
         cornerstoneTools.ToolGroupManager.getToolGroup(toolGroupIdVRT);
       const orientationControllerTool = toolGroupVRT.getToolInstance(
@@ -529,28 +474,19 @@ async function run(numViewports = getNumViewportsFromUrl()) {
     },
   });
 
-  // Add dropdown for letter color scheme (reloads page with URL param)
+  // Add dropdown for letter color scheme
   const letterColorSchemeValues: string[] = ['mixed', 'white', 'black'];
   const letterColorSchemeLabels = ['Mixed', 'White', 'Black'];
-  const validLetterColorScheme = letterColorSchemeValues.includes(
-    letterColorScheme
-  )
-    ? letterColorScheme
-    : 'mixed';
 
   addDropdownToToolbar({
     labelText: 'Letter Colors',
     options: {
       values: letterColorSchemeValues,
-      defaultValue: validLetterColorScheme,
+      defaultValue: letterColorScheme,
       labels: letterColorSchemeLabels,
     },
     container: controlRow,
     onSelectedValueChange: (selectedValue) => {
-      const url = new URL(window.location.href);
-      url.searchParams.set('letterColorScheme', String(selectedValue));
-      window.history.replaceState({}, '', url);
-
       const toolGroupVRT =
         cornerstoneTools.ToolGroupManager.getToolGroup(toolGroupIdVRT);
       const orientationControllerTool = toolGroupVRT.getToolInstance(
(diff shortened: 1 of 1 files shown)


daker:
@mbellehumeur Hi, not sure it can help but i have already ported vtkPlatonicSolidSource to vtk.js https://github.com/Kitware/vtk-js/pull/3308

mbellehumeur:
> @mbellehumeur Hi, not sure it can help but i have already ported vtkPlatonicSolidSource to vtk.js [Kitware/vtk-js#3308](https://github.com/Kitware/vtk-js/pull/3308)

You are welcome to add the OrientationController widget to VTK.js as I have not contributed to that repo yet.

packages/tools/examples/volumeCroppingTool/index.ts
```diff
@@ -363,45 +436,248 @@ async function run(numViewports = getNumViewportsFromUrl()) {
...
+
+  // Get keepOrientationUp from URL
+  const keepOrientationUp = getKeepOrientationUpFromUrl();
+
+  // Get letter color scheme from URL
+  const letterColorScheme = getLetterColorSchemeFromUrl();
+
+  // Configure faceColors based on URL parameter
+  let faceColors;
+  if (colorScheme === 'gray') {
```
wayfarer3130: Can you create a map of color scheme to face colours in the tools package - probably as a static variable on the tool itself, and return the face colors as a function from there?  This seems like something that should be shared/available as a setting.

packages/tools/examples/volumeCroppingTool/index.ts
```diff
@@ -363,45 +436,248 @@ async function run(numViewports = getNumViewportsFromUrl()) {
...
+    colorScheme,
+    faceColors,
+    keepOrientationUp,
+    letterColorScheme,
+  });
+  // Enable OrientationControllerTool after viewport is added and volume is loaded
+  toolGroupVRT.setToolEnabled(OrientationControllerTool.toolName);
+
+  // Add dropdown for orientation control colors (reloads page with URL param)
+  const colorSchemeValues: string[] = ['rgy', 'gray', 'marker'];
```
wayfarer3130, line 451: If you create a map with the color schemes, you can use const colorSchemeValues = [...colorSchemes.keys()]

packages/tools/src/tools/OrientationControllerTool.ts
```diff
@@ -0,0 +1,724 @@
...
+          topBottom: provided.topBottom,
+          frontBack: provided.frontBack,
+          leftRight: provided.leftRight,
+        };
+      }
+    }
+
+    // Otherwise, use colorScheme to determine colors
+    switch (colorScheme) {
+      case 'marker':
```
wayfarer3130: These are also defined in the example - please create a static map instance with the values to use rather than implementing a poor man's map using switch.

packages/tools/src/tools/OrientationControllerTool.ts
```diff
@@ -0,0 +1,724 @@
...
+      case 'all-white': // Backward compatibility
+        return {
+          zMinus: [255, 255, 255],
+          zPlus: [255, 255, 255],
+          yMinus: [255, 255, 255],
+          yPlus: [255, 255, 255],
+          xMinus: [255, 255, 255],
+          xPlus: [255, 255, 255],
+        };
+      case 'black':
```
wayfarer3130: Ditto

packages/tools/src/tools/OrientationControllerTool.ts
```diff
@@ -0,0 +1,724 @@
...
+    }
+
+    const actors = this.widget.getActors(viewportId);
+
+    if (!actors) {
+      return;
+    }
+
+    const { viewport } = enabledElement;
+    if (viewport.type !== Enums.ViewportType.VOLUME_3D) {
```
wayfarer3130: Why do you disallow other viewport types?  There isn't any real reason you couldn't allow orthographic viewports, and even a stack viewport could be set to read only to display the orientation.  
The plan for stack/orthographic viewports is to combine them and let the viewport itself indicate if it can change orientations.  You could add a base method isOrientationChangeable defaulting to return false, and for ortho/3d viewports return true.  That is far better than guessing the volume type.

packages/tools/src/tools/OrientationControllerTool.ts
```diff
@@ -0,0 +1,724 @@
...
+    const startUp = vec3.fromValues(
+      startViewUpArray[0],
+      startViewUpArray[1],
+      startViewUpArray[2]
+    );
+    const startRight = vec3.create();
+    vec3.cross(startRight, startUp, startForward);
+    vec3.normalize(startRight, startRight);
+
+    const startMatrix = mat4.create();
```
wayfarer3130: I believe mat4 has a create from values that you can just put these in the right order.  Not that it really matters, but it does work.

packages/tools/src/tools/OrientationControllerTool.ts
```diff
@@ -0,0 +1,724 @@
...
+    startMatrix[2] = startRight[2];
+    startMatrix[4] = startUp[0];
+    startMatrix[5] = startUp[1];
+    startMatrix[6] = startUp[2];
+    startMatrix[8] = startForward[0];
+    startMatrix[9] = startForward[1];
+    startMatrix[10] = startForward[2];
+    startMatrix[15] = 1;
+
+    const targetForward = vec3.fromValues(
```
wayfarer3130: If you are not updating targetForward, then just use targetViewPlaneNOrmal

wayfarer3130: That is, a Point3 is a valid argument for a vec3 operation.

packages/tools/src/tools/OrientationControllerTool.ts
```diff
@@ -0,0 +1,724 @@
...
+      targetUp = vec3.fromValues(
+        targetViewUp[0],
+        targetViewUp[1],
+        targetViewUp[2]
+      );
+    } else {
+      // Keep current viewUp, but project it onto the plane perpendicular to targetForward
+      // to ensure orthogonality
+      const currentUp = vec3.fromValues(startUp[0], startUp[1], startUp[2]);
+      vec3.normalize(currentUp, currentUp);
```
wayfarer3130: const currentUp = vec3.normalize(vec3.create(), startUp);

packages/tools/src/tools/OrientationControllerTool.ts
```diff
@@ -0,0 +1,724 @@
...
+    targetMatrix[5] = targetUp[1];
+    targetMatrix[6] = targetUp[2];
+    targetMatrix[8] = targetForward[0];
+    targetMatrix[9] = targetForward[1];
+    targetMatrix[10] = targetForward[2];
+    targetMatrix[15] = 1;
+
+    const startQuat = quat.create();
+    const targetQuat = quat.create();
+    mat4.getRotation(startQuat, startMatrix);
```
wayfarer3130: I generally prefer:
const startQuat = mat4.getRotation(quat.create(), startmatrix);
The other gl-matrix operaitons all work like that, but I haven't quite worked with mat4 enough to know if it does too.

packages/tools/src/tools/OrientationControllerTool.ts
```diff
@@ -0,0 +1,724 @@
...
+    mat4.getRotation(targetQuat, targetMatrix);
+
+    let dotProduct = quat.dot(startQuat, targetQuat);
+    if (dotProduct < 0) {
+      targetQuat[0] = -targetQuat[0];
```
wayfarer3130: quat.scale(targetQuat,targetQuat,-1)

packages/tools/src/tools/OrientationControllerTool.ts
```diff
@@ -0,0 +1,724 @@
...
+      const t = currentStep / steps;
+      const easedT = t < 0.5 ? 2 * t * t : 1 - Math.pow(-2 * t + 2, 2) / 2;
+
+      const interpolatedQuat = quat.create();
+      quat.slerp(interpolatedQuat, startQuat, targetQuat, easedT);
+
+      const interpolatedMatrix = mat4.create();
+      mat4.fromQuat(interpolatedMatrix, interpolatedQuat);
+
+      const interpolatedForward = vec3.fromValues(
```
wayfarer3130: You can use slice to get interpolatedForward.

packages/tools/src/tools/OrientationControllerTool.ts
```diff
@@ -0,0 +1,724 @@
...
+        interpolatedMatrix[10]
+      );
+      const interpolatedUp = vec3.fromValues(
+        interpolatedMatrix[4],
+        interpolatedMatrix[5],
+        interpolatedMatrix[6]
+      );
+
+      viewport.setCamera({
+        viewPlaneNormal: Array.from(interpolatedForward) as [
```
wayfarer3130: interpolatedForward as Point3 should be sufficient.

packages/tools/src/tools/OrientationControllerTool.ts
```diff
@@ -0,0 +1,724 @@
...
+          number,
+          number,
+          number,
+        ],
+        viewUp: Array.from(interpolatedUp) as [number, number, number],
+      });
+      viewport.resetCamera({
```
wayfarer3130: Extract a constant for the parameters, outside the function.

packages/tools/src/utilities/vtkjs/AnnotatedRhombicuboctahedronActor/index.js
```diff
@@ -0,0 +1,482 @@
...
+      }
+    }
+
+    const actors = [];
+
+    const hexToRgb = (hex) => {
+      const result = /^#?([a-f\d]{2})([a-f\d]{2})([a-f\d]{2})$/i.exec(hex);
+      return result
+        ? [
+            parseInt(result[1], 16),
```
wayfarer3130: return toNumber(result) || [255,255,255]

wayfarer3130: Hmm, you are extracting attributes, so
toNumber(result.slict(1,4)) || [255,255,255]

wayfarer3130 (changes requested):
A number of simplifications, but generally looks good - once you update those I will approve and merge.

wayfarer3130:
@mbellehumeur - can you add notes on the URL parameters and how to see the example in cornerstone?  You should be able to enable the new tool in the volume cropping example.
Ideally I'd like to see it enabled for all 4 viewports, even if we don't plan on using it that way right now.

packages/tools/examples/volumeCroppingTool/index.ts
```diff
@@ -359,6 +446,19 @@ async function run(numViewports = getNumViewportsFromUrl()) {
     ],
   });
 
+  // Get color scheme from URL (needed for both orthographic and 3D viewports)
+  const colorScheme = getColorSchemeFromUrl();
+  const keepOrientationUp = getKeepOrientationUpFromUrl();
+  const letterColorScheme = getLetterColorSchemeFromUrl();
+
+  toolGroup.addTool(OrientationControllerTool.toolName, {
```
wayfarer3130, line 409: I never see this actually rendering - I don't quite know what is required to get it working.
It doesn't seem like you enabled the maintainer commits, so even if I fix it locally, I still can't push it to your branch.

wayfarer3130:
Whether the marker is visible or not is quite width dependent.  Wider views of the browser fail to display it.
image

I've never seen it on the main 3d view, and it doesn't work on any of the other views.

Also, it shouldn't load/reload based on the URL parameters - just set the parameters inline on the response rather than redirecting/reloading the view.

PR #2588, opened 2026-01-29
fix: redo not restoring cachedStats

wayfarer3130:
### Context

The cachedStats sometimes contains some large point arrays, which aren't needed for restoring cached stats but which had prevented the cloning of cachedStats.  This causes OHIF to not display the restored stats.

### Changes & Results

Add a safeStructuredClone that clones objects recursively, ignoring certain fields and setting objects to null when they don't clone successfully.  This allows an easier clone in the history state that includes the cached stats but omits the large arrays of float data.

### Testing

Link to OHIF
Create one of every annotation type
Look at the screen capture for what is shown for each annotation
Undo until all are gone
Redo until all are restored

Restored values should list the same data on the right hand panel as before undo.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  

### DIFF
--- a/packages/tools/src/utilities/safeStructuredClone.ts
+++ b/packages/tools/src/utilities/safeStructuredClone.ts
@@ -55,9 +55,8 @@ function omitUncloneableKeys(
     if (value === null || value === undefined || typeof value !== 'object') {
       result[key] = value;
     } else if (Array.isArray(value)) {
-      // The omit keys is not inside objects in arrays right now, so just copy the array as is
       result[key] = value.map((value) =>
-        omitUncloneableKeys(value as Record<string, unknown>)
+        safeStructuredClone(value as Record<string, unknown>)
       );
     } else {
       result[key] = omitUncloneableKeys(value as Record<string, unknown>);
@@ -82,6 +81,9 @@ export function safeStructuredClone<T>(value: T): T {
   }
   if (typeof value !== 'object') {
     return value;
+  }
+  if (Array.isArray(value)) {
+    return value.map((item) => safeStructuredClone(item)) as unknown as T;
   }
   return omitUncloneableKeys(value as Record<string, unknown>) as T;
 }
--- a/packages/tools/test/utilities/safeStructuredClone.jest.js
+++ b/packages/tools/test/utilities/safeStructuredClone.jest.js
@@ -1,0 +1,101 @@
+import { safeStructuredClone } from '../../src/utilities/safeStructuredClone';
+import { describe, it, expect } from '@jest/globals';
+
+describe('safeStructuredClone', () => {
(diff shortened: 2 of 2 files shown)


jbocce:
Verification passed as per the testing notes and as per the OHIF issue https://github.com/OHIF/Viewers/issues/5749.

High level scenarios tested:
- added one of every measurement type (tracked) on a single image; undid and redid every single one; used screen shot to compare before and after redo
- added one of every measurement type (untracked) on various images of various display sets; undid and redid every single one
- added one of every measurement type (tracked) on various images of various display sets; undid and redid every single one

One issue discovered and logged: [[Bug] Images shift when navigating measurements](https://github.com/OHIF/Viewers/issues/5761)

packages/tools/src/utilities/safeStructuredClone.ts
```diff
@@ -0,0 +1,113 @@
...
+    return value;
+  }
+  if (typeof value !== 'object') {
+    return value;
+  }
+  if (Array.isArray(value)) {
+    try {
+      return structuredClone(value) as T;
+    } catch {
+      return [] as T;
```
jbocce: Should this be null instead of []? I think this is the only one that goes against the "replaces any value that cannot be cloned with null" statement.

packages/tools/src/utilities/safeStructuredClone.ts
```diff
@@ -0,0 +1,113 @@
...
+): Record<string, unknown> {
+  const result: Record<string, unknown> = {};
+  for (const [key, value] of Object.entries(obj)) {
+    if (OMIT_KEYS.has(key)) {
+      continue;
+    }
+    if (value === null || value === undefined || typeof value !== 'object') {
+      result[key] = value;
+    } else if (Array.isArray(value)) {
+      result[key] = tryCloneValue(value);
```
jbocce: Would it ever make sense to do a `cloneRecursiveWithFallback` on each element of the array should `null` be returned here?

jbocce (approved):
A couple of minor comments. Consider adding @sedghi as a reviewer.

packages/tools/src/utilities/safeStructuredClone.ts
```diff
@@ -0,0 +1,113 @@
...
+  'pointsInVolume',
+  'projectionPoints',
+  'contour',
+  'spline',
+]);
+
+/**
+ * Recursively copies an object, omitting known large/non-cloneable keys.
+ */
+function omitUncloneableKeys(
```
sedghi, line 43: can you ask ai to add unit tests for this file

PR #2591, opened 2026-01-30
fix: Wrong calculation of rectangle size using indexes

wayfarer3130:
### Context

Rectangles were incorrectly using both scale and world units to calculation area.  Switched to use index coordinates.

### Testing

Run tmtv example in cs3d, check that circle and rectangle have roughly the same value on all three rows.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  

### DIFF
--- a/packages/core/src/RenderingEngine/BaseVolumeViewport.ts
+++ b/packages/core/src/RenderingEngine/BaseVolumeViewport.ts
@@ -2373,6 +2373,9 @@ abstract class BaseVolumeViewport extends Viewport {
       volumeId = actorEntries.find(
         (actorEntry) => actorEntry.actor.getClassName() === 'vtkVolume'
       )?.referencedId;
+      if (!volumeId) {
+        return;
+      }
     }
 
     const currentIndex = this.getSliceIndex();
--- a/packages/tools/src/eventDispatchers/mouseEventHandlers/mouseDownActivate.ts
+++ b/packages/tools/src/eventDispatchers/mouseEventHandlers/mouseDownActivate.ts
@@ -29,7 +29,11 @@ export default function mouseDownActivate(
   }
 
   if (activeTool.addNewAnnotation) {
-    const annotation = activeTool.addNewAnnotation(evt, 'mouse');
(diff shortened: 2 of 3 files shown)


jbocce:
Unit tests are failing. Please have a look.

packages/core/examples/tmtv/index.ts
```diff
@@ -45,20 +46,16 @@ const { createCameraPositionSynchronizer, createVOISynchronizer } =
 
 // Study IDs
 const FirstStudyID = `1.3.6.1.4.1.14519.5.2.1.7009.2403.871108593056125491804754960339`;
-const SecondStudyID =
-  '1.3.6.1.4.1.14519.5.2.1.7009.2403.334240657131972136850343327463';
-const ThirdStudyID = `1.3.6.1.4.1.9328.50.17.15423521354819720574322014551955370036`;
+// const SecondStudyID =
```
jbocce, line 51: If not needed then please remove.

jbocce (changes requested):
The code looks fine - just one super minor comment. However I'm not sure about the testing results. Look at the circle results versus the rectangle.

Here is the circle...

Image

And here is the rectangle...

Image

Notice the large discrepancy in area.

wayfarer3130:
> The code looks fine - just one super minor comment. However I'm not sure about the testing results. Look at the circle results versus the rectangle.
> 
> Here is the circle...
> 
> Image
> And here is the rectangle...
> 
> Image
> Notice the large discrepancy in area.

The circle value is quite wrong, but I wasn't able to reproduce that.  It did indicate there was a bug in the rectangle code, which I have fixed.  The values when I test them are now very similar, and I added the bidirectional so I could get length measurements.

wayfarer3130:
The radius/area are consistent with each other, but the radius is complete garbage - that would be a head size of about 2/3 of a meter, which is way too big.  That looks to me like it is using the index pixels from the PT, but everything else from the CT.

wayfarer3130:
OK, I can reproduce this now reliably, looking into it a bit more.

PR #2595, opened 2026-02-04
fix: Newer dcmjs version for seg save

wayfarer3130:
### Context

Update dcmjs version to fix a seg save issue

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  

### DIFF
--- a/packages/docs/bun.lock
+++ b/packages/docs/bun.lock
@@ -20,7 +20,6 @@
         "@svgr/webpack": "8.1.0",
         "clsx": "1.1.1",
         "cross-env": "7.0.3",
-        "dcmjs": "0.49.4",
         "dicom-parser": "1.8.21",
         "dicomweb-client": "0.10.4",
         "docusaurus-plugin-copy": "0.1.1",
--- a/packages/docs/package.json
+++ b/packages/docs/package.json
@@ -54,7 +54,6 @@
     "@svgr/webpack": "8.1.0",
     "clsx": "1.1.1",
     "cross-env": "7.0.3",
-    "dcmjs": "0.49.4",
     "dicom-parser": "1.8.21",
     "dicomweb-client": "0.10.4",
     "docusaurus-plugin-copy": "0.1.1",


package.json
```diff
@@ -66,6 +66,8 @@
     "webpack:watch": "npx lerna run webpack:watch"
   },
   "devDependencies": {
+    "baseline-browser-mapping": "^2.9.19",
```
wayfarer3130: Often on install:update-lockfile it would get a mixed version of this, so updating to a consistent version as a resolution.

package.json
```diff
@@ -66,6 +66,8 @@
...
+    "caniuse-lite": "^1.0.30001767",
```
wayfarer3130: caniuse-lite had several internal versions depending on components used, so making this consistent.

package.json
```diff
@@ -190,15 +192,14 @@
     "not op_mini all"
   ],
   "packageManager": "yarn@1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e",
-  "dependencies": {
-    "caniuse-lite": "1.0.30001734"
-  },
   "resolutions": {
+    "@babel/helper-create-class-features-plugin": "7.28.3",
```
wayfarer3130, line 194: Need to force these two for consistency

package.json
```diff
@@ -190,15 +192,14 @@
...
+    "caniuse-lite": "^1.0.30001767",
     "axios": "1.12.0",
     "buffer": "6.0.3",
     "node-forge": "1.3.2",
     "canvas": "3.2.0",
-    "dcmjs": "0.48.0",
+    "dcmjs": "0.49.4",
```
wayfarer3130, line 200: Need this version of dcmjs to fix segmentation save.

packages/docs/package.json
```diff
@@ -53,7 +53,7 @@
     "@svgr/webpack": "8.1.0",
     "clsx": "1.1.1",
     "cross-env": "7.0.3",
-    "dcmjs": "0.48.0",
+    "dcmjs": "0.49.4",
```
jbocce: Please update the bun.lock and yarn.lock files for the docs package as well.

package.json
```diff
@@ -190,15 +192,14 @@
     "not op_mini all"
   ],
   "packageManager": "yarn@1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e",
-  "dependencies": {
-    "caniuse-lite": "1.0.30001734"
-  },
   "resolutions": {
+    "@babel/helper-create-class-features-plugin": "7.28.3",
+    "caniuse-lite": "^1.0.30001767",
```
jbocce: Please use exact versioning.

package.json
jbocce: No need for this to be here since it is in the resolutions. Please remove.

package.json
jbocce: I am confused to see this package at all listed in the package.json file. It does NOT appear in either of the previous bun.lock or yarn.lock files. So I don't understand why you would ever see a mixed version of this. In fact, I tried running `npm ls baseline-browser-mapping` in a clean CS3D branch and the dependency tree was EMPTY. I also tried running it on this branch and there were no transitive packages relying on it. Considering you made no coding changes in this PR, it is a bit surprising to me why it is here. I would recommend removing this from here entirely, but I suspect you might have a reason why it is here. At the very least, exact versioning should be used, but I am hoping we can just remove it.

package.json
jbocce: Thank you for using exact versioning.

packages/docs/yarn.lock
jbocce, line 1: This looks like way too many changes in this file. I think we should stick to using `yarn install` to update this!

packages/docs/package.json
```diff
@@ -21,6 +21,7 @@
     "docs:ci": "cross-env NODE_OPTIONS=--max_old_space_size=16384 && yarn run build:docs",
     "docs:ci-no-tests": "cross-env NODE_OPTIONS=--max_old_space_size=16384 && yarn run prep-no-tests && yarn run build:docs",
     "dev:watch": "cross-env NODE_OPTIONS=--max_old_space_size=16384 TYPEDOC_WATCH=true docusaurus start",
+    "install:update-lockfile": "bun install --yarn",
```
jbocce: Can we please stick to what we do for the main CS3D project? Thanks.

packages/docs/yarn.lock
wayfarer3130: Still working on the PR - will let you know when it is ready for review.

packages/docs/package.json
wayfarer3130: Switched this to a full yarn install && bun install pair.

packages/docs/bun.lock
```diff
@@ -2802,6 +2803,8 @@
 
     "@babel/traverse/debug": ["debug@4.4.0", "", { "dependencies": { "ms": "^2.1.3" } }, "sha512-6WTZ/IxCY/T6BALoZHaE4ctp9xm+Z5kY/pzYaCHRFeyVhojxlrm+46y68HA6hr0TcwEssoxNiDEUJQjfPZ/RYA=="],
 
+    "@cornerstonejs/adapters/dcmjs": ["dcmjs@0.48.0", "", { "dependencies": { "@babel/runtime-corejs3": "^7.22.5", "adm-zip": "^0.5.10", "gl-matrix": "^3.1.0", "lodash.clonedeep": "^4.5.0", "loglevel": "^1.8.1", "ndarray": "^1.0.19", "pako": "^2.0.4" } }, "sha512-hxDxLkfs3roAx3KaDue4aexeFUXCT073wuepvQ7FdS6gZs+WfYYSv/IDW/ejMfVAazs8lLgRfit2reEbKmQYsQ=="],
```
jbocce: It looks like adapters is using the old version of dcmjs. Is this intentional?

packages/docs/yarn.lock
```diff
@@ -1185,14 +1185,14 @@
...
-  integrity sha512-5ZSB/2cXaewn+8pFLZS/o/znkYOdTtgDJ9k5XjOny3gzarTszcaVr7aTiTayN8TkTyra5TgAC4jc6XOkQV+YeQ==
+"@cornerstonejs/adapters@4.15.29":
+  version "4.15.29"
+  resolved "https://registry.yarnpkg.com/@cornerstonejs/adapters/-/adapters-4.15.29.tgz#cd8a60c72db55a76f8e0489ad825bf0f28b0f625"
+  integrity sha512-OcG3YtCQ+C7X/f7buK3dsMOc8NyqsgH/LZRT/JbvppUnqhhfyKIpBTtJG+2+roMyLsi3QrsCeI9Eu6KaQgFFvw==
   dependencies:
     "@babel/runtime-corejs2" "7.26.10"
     buffer "6.0.3"
-    dcmjs "0.45.0"
+    dcmjs "0.48.0"
```
jbocce, line 1195: Here it is again.

jbocce: That is, see my comment in bun.lock about this.

package.json
```diff
@@ -39,6 +39,7 @@
     "docs": "npx lerna run docs",
     "docs:watch": "npx lerna run docs:watch",
     "install:update-lockfile": "yarn install && bun install --config=./bunfig.update-lockfile.toml",
+    "install:update": "echo Use for faster dev deploy, do not commit yarn lock file && bun install --config=./bunfig.update-lockfile.toml --yarn",
```
jbocce: I didn't add this originally because I think Alireza didn't want it because yarn is still the official tool for dev building.

packages/docs/yarn.lock
wayfarer3130, line 1195: This one doesn't want to update, even with a resolution

wayfarer3130: That is the declared dependency in that version of @cornerstonejs adapters - it is the resolution down at the bottom that maps 0.48.0 to the newer version of dcmjs.

package.json
wayfarer3130: Removed.

packages/docs/yarn.lock
```diff
@@ -4641,10 +4641,10 @@ data-uri-to-buffer@^6.0.2:
   resolved "https://registry.yarnpkg.com/data-uri-to-buffer/-/data-uri-to-buffer-6.0.2.tgz#8a58bb67384b261a38ef18bea1810cb01badd28b"
   integrity sha512-7hvf7/GW8e86rW0ptuwS3OcBGDjIi6SZva7hCyWC0yYry2cOPmLIjXAUHI6DK2HsnwJd9ifmt57i8eV2n4YNpw==
 
-dcmjs@0.45.0:
-  version "0.45.0"
-  resolved "https://registry.yarnpkg.com/dcmjs/-/dcmjs-0.45.0.tgz#ce53dd06fc345e8231e28c340a36a8094a62422a"
-  integrity sha512-MFtACBLQB6vj/3lzHEf6iGBLFC8f/QysoDO+DVUJeLrKYQi4mXYzzBbHmZCpMXqIpj8qXXLvnP/u2ugd9DVu7Q==
+dcmjs@0.48.0, dcmjs@0.49.4:
```
wayfarer3130, line 4644: This is the resolution that maps the dcmjs version for the docs - the earlier 0.48.0 includes don't change, because they are what the released package contains.

packages/docs/package.json
```diff
@@ -95,6 +96,7 @@
   "resolutions": {
     "node-forge": "1.3.2",
     "qs": "6.14.1",
+    "dcmjs": "0.49.4",
```
jbocce, line 98: Do we really want to list this twice - here and in the dependencies?

PR #2607, opened 2026-02-09
perf(rendering): add VolumeSliceViewport using ImageResliceMapper 

sedghi:
There are work need to be done,
major thing is labelmap outline is not yet in vtk js for volume reslice mapper but some work here https://discourse.vtk.org/t/label-outline-rendering-in-vtk-jss-imagereslicemapper/16224/2

---

This PR introduces a new `VolumeSliceViewport` that renders volumetric data as 2D slices using `vtkImageResliceMapper` instead of the traditional `vtkVolumeMapper`. The key difference is the rendering technique: the existing `VolumeViewport` uses **ray tracing** (ray casting through a 3D texture with clipping planes to produce a thin slab), while the new `VolumeSliceViewport` uses **rasterization** via `vtkImageResliceMapper`. Both approaches sample from a 3D texture, but rasterization is a simpler algorithm that produces equivalent results for 2D slice viewing and is more broadly compatible across GPU hardware and shader implementations.

### Motivation

At the time the original volume viewport was implemented, vtk.js did not have a mature `vtkImageResliceMapper` IIRC. The workaround was to use the full volume rendering pipeline (`vtkVolumeMapper` + `vtkVolume` + `vtkSlabCamera`) with clipping planes to extract thin slabs — effectively ray tracing through the volume to produce a 2D slice. While this works, ray tracing shaders are more complex and can be less compatible across devices.

The `vtkImageResliceMapper` offers a rasterization-based alternative that:
- Produces the same visual output for standard 2D slice viewing
- Uses simpler, more portable GPU shaders with broader device compatibility
- Still operates on the same underlying 3D texture
- Natively supports slab thickness and blend modes (MIP, MinIP, Average) through its own slab type system

### New Viewport: `VolumeSliceViewport`
- New viewport class (`ViewportType.VOLUME_SLICE`) extending `VolumeViewport` that uses `vtkImageResliceMapper` for rasterization-based 2D slice rendering
- Uses standard `vtkCamera` instead of `vtkSlabCamera` (no slab projection needed)
- Custom `canvasToWorld` / `worldToCanvas` coordinate transformations adapted for the reslice mapper pipeline
- Camera `setCamera` override that projects the focal point onto the slice plane for accurate slice positioning
- Slice plane synchronization: camera movements automatically update the reslice plane origin/normal on all actors
- Blend mode support via `vtkImageResliceMapper` slab types (MIP, MinIP, Average)
- Slab thickness support through the reslice mapper

### New VTK Classes
- **`vtkSharedImageResliceMapper`**: Extends `vtkImageResliceMapper` to carry a shared scalar texture reference, avoiding memory duplication across multiple reslice mappers
- **`vtkStreamingOpenGLImageResliceMapper`**: Extends `vtkOpenGLImageResliceMapper` to support progressive/streaming texture updates (matching the streaming texture pattern used by the existing volume mapper)
- Updated `vtkStreamingOpenGLViewNodeFactory` to register the new OpenGL mapper

### New Helper: `createVolumeSliceActor`
- Creates `vtkImageSlice` actors with `vtkSharedImageResliceMapper` for a given volume
- Handles VOI initialization, color transfer functions, and multi-component volumes
- Fires `VOI_MODIFIED` events on actor creation

### Rendering Engine Updates
- **`ContextPoolRenderingEngine`** and **`TiledRenderingEngine`**: Added `VOLUME_SLICE` viewport type routing to `VolumeSliceViewport`
- **`RenderingEngine`**: Simplified mode selection — `ContextPool` is now the default (removed explicit `ContextPool` case and warning for unknown modes)
- **`BaseRenderingEngine`**: `VOLUME_SLICE` viewports get default axial orientation; null-safe `canvas.getContext('2d')` call

### BaseVolumeViewport Refactoring
- Changed several `private` members to `protected` to allow `VolumeSliceViewport` to override behavior: `_FrameOfReferenceUID`, `_setVolumeActors`, `_isValidVolumeInputArray`, `_addVolumeId`, `initializeColorTransferFunction`
- Replaced `actorIsA(entry, 'vtkVolume')` with `isImageActor(entry)` throughout — this correctly matches both `vtkVolume` and `vtkImageSlice` actors
- Uses standard `vtkCamera` for `VOLUME_SLICE` type, `vtkSlabCamera` for others
- Added preset guard: `applyPreset` is skipped for non-`vtkVolume` actors with a warning
- Updated type annotations from `vtkVolume` to `ImageActor | VolumeActor` where applicable

### VOI & Labelmap Support
- **`setDefaultVolumeVOI`**: Extended to handle `vtkImageSlice` actors — sets `colorWindow`/`colorLevel` and initializes RGB transfer function points when empty
- **`labelmapDisplay`**: Added `vtkImageSlice`-specific labelmap rendering setup (lookup table scalar range, color window/level)
- **`performVolumeLabelmapUpdate`**: Reset updated frames array before marking individual slices for streaming texture updates

### Examples
- **`volumeResliceViewport`**: New example demonstrating the `VOLUME_SLICE` viewport with annotation tools and a built-in rendering benchmark
- **`volumeViewportLegacy`**: Companion example using the traditional `ORTHOGRAPHIC` viewport for side-by-side comparison
- Updated **`crossHairs`**, **`petCt`**, and **`labelmapSegmentationTools`** examples to use `VOLUME_SLICE` viewport type
- Added PET opacity slider to the PET/CT example

### Other
- `Viewport.getRenderer()` simplified to call `renderingEngine.getRenderer()` directly instead of going through `offscreenMultiRenderWindow`
- `ViewportType` enum: added `VOLUME_SLICE = 'volumeSlice'`
- `IVolumeInput` callback type updated to accept `ImageActor`
- `getCameraVectors` helper updated to accept `VOLUME_SLICE` viewports

## Test Plan

- [ ] Open the **Volume Reslice Viewport** example — verify axial/sagittal/coronal slices render correctly with annotation tools (Length, Probe, ROI, etc.)
- [ ] Open the **Legacy Volume Viewport** example — verify the traditional orthographic viewport still works identically
- [ ] Run the built-in benchmark on both examples and compare rendering performance
- [ ] Open the **PET/CT** example — verify volume fusion renders correctly with the new viewport type, test the PET opacity slider
- [ ] Open the **Crosshairs** example — verify crosshair tool works with the new viewport type
- [ ] Open the **Labelmap Segmentation Tools** example — verify segmentation overlays render correctly on `VOLUME_SLICE` viewports
- [ ] Verify `canvasToWorld` / `worldToCanvas` coordinate accuracy by placing annotations and confirming they appear at the correct world positions
- [ ] Verify scrolling through slices (mouse wheel) works correctly on all orientations
- [ ] Test window/level adjustments on the new viewport type


IbrahimCSAE:
Nice

.husky/pre-commit
```diff
@@ -1 +1,7 @@
+#!/usr/bin/env sh
```
wayfarer3130, line 1: This won't work in general, requires sh to be able to run.

packages/core/examples/volumeResliceViewport/index.ts
```diff
@@ -0,0 +1,463 @@
...
+async function run() {
+  await initDemo({
+    core: {
+      rendering: {
+        renderingEngineMode: Enums.RenderingEngineModeEnum.ContextPool,
+      },
+    },
+  });
+
+  cornerstoneTools.addTool(LengthTool);
```
wayfarer3130, line 235: Can you add the tools using the helper methods as having all of this repeated distracts from the core goal of this example, which is to show the reslicing setup/behaviour.

packages/core/examples/volumeViewportLegacy/index.ts
wayfarer3130, line 1: Ditto on this file.

wayfarer3130:
@sedghi - I like the changes generally, but we are starting to see a proliferation of viewport types and that makes testing as well as viewing somewhat problematic.  I'd really rather have 1 viewport type for "planar" images, whether stack or volume based and then a 3d viewport, the video viewport etc, but not having different viewport types.  That would make the fixes much more general.  Then, if there was some sort of "actor" implementation and perhaps information on whether the stack of images is:  
1. Non-volumetric=> use same actors, but always single slice and throw on trying to use non-acquisition orientations.
2. Volumetric => allow non-acquisition/thickness options
3. CPU rendered => provide a CPU fallback
4. Preferences/settings to determine how to manage non-acquisition/thickness orientations:
    - multiple volumes, only load required one(s)
    - orthogonal volumes, only load closest - resample to orthogonal
    - recreated fractional volume => Allow slice thickness up to the specified number of slices, and resample to given orientation/number of frames.

Note how this actually allows a lot more flexibility in rendering, since it works with the CPU rendering almost immediately, and on low end can simple create resample volumes for display.  Navigation within the resampled data is relatively easy as the memory is simply copied (which is fast) or re-indexed (faster).

It is quite east to create a fractional/mpr volume using nearest neighbor to start with.
It is trivially easy to create that for a partial volume in acquisition orientation, so that can be the base behaviour and then it just upgrades to various volume types as required/configured.  Minimal case is to CPU render the resampled data or to vtk render the resampled data.

What we would really like to have is the move to a single viewport type for all MPR/stack views  and deprecate the old distinction between stack and mpr viewports.  

There should be NO difference in the isReferenceViewable, getViewReference etc, and users should be able to cleanly switch to MPR views as needed.

sedghi:
closing in favor of https://github.com/cornerstonejs/cornerstone3D/pull/2666

PR #2610, opened 2026-02-10
fix palette color image loading

rodrigobasilio2022:
### Context

The convertPALETTECOLOR function was using Promise.all().then() without returning or awaiting the promise. This caused the function to complete immediately before the palette color data was fully fetched and the pixel data was properly converted. As a result, images with PALETTE COLOR photometric interpretation would sometimes render as black images because the pixel data was being used before it was ready.

Solution

Convert the color space conversion pipeline to use async/await pattern to ensure proper synchronization:

convertPALETTECOLOR - Changed from synchronous to async function that awaits palette data fetching and conversion
convertColorSpace - Converted to async function and awaits the palette color conversion when needed
createImage - Updated the decode promise handler to be async and await color space conversion

Changes

Modified convertPALETTECOLOR.ts: Changed function signature to async and replaced .then() with await
Modified convertColorSpace.ts: Made function async and added await for PALETTE COLOR conversion
Modified createImage.ts: Made the decode callback async to properly await color conversion

Impact

This ensures that palette color images are fully processed before rendering, eliminating the race condition that caused black images to appear intermittently.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  

### DIFF
--- a/packages/dicomImageLoader/src/imageLoader/colorSpaceConverters/convertPALETTECOLOR.ts
+++ b/packages/dicomImageLoader/src/imageLoader/colorSpaceConverters/convertPALETTECOLOR.ts
@@ -46,7 +46,7 @@ export default function convertPaletteColor(
 
   const start = imageFrame.redPaletteColorLookupTableDescriptor[1];
   const bitsStored = imageFrame.redPaletteColorLookupTableDescriptor[2];
-  const shift = bitsStored > 8 || rData.every((num) => num < 256) ? 0 : 8;
+  const shift = bitsStored > 8 || rData.some((num) => num > 255) ? 8 : 0;
 
   const rDataCleaned = convertLUTto8Bit(rData, shift);
   const gDataCleaned = convertLUTto8Bit(gData, shift);


packages/dicomImageLoader/src/imageLoader/createImage.ts
```diff
@@ -91,7 +91,7 @@ function createImage(
   return new Promise<DICOMLoaderIImage | Types.IImageFrame>(
     (resolve, reject) => {
       // eslint-disable-next-line complexity
-      decodePromise.then(function (imageFrame: Types.IImageFrame) {
+      decodePromise.then(async function (imageFrame: Types.IImageFrame) {
         // If we have a target buffer that was written to in the
```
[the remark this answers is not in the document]
wayfarer3130: This comment needs to be addressed.

packages/dicomImageLoader/src/imageLoader/convertColorSpace.ts
```diff
@@ -23,7 +23,11 @@ function convertYBRFull(imageFrame, colorBuffer, useRGBA) {
   }
 }
 
-export default function convertColorSpace(imageFrame, colorBuffer, useRGBA) {
```
wayfarer3130, line 26: I think I'd prefer to leave convertColorSpace as non-async, and instead add a function that ensures that required bulkdata for applying luts and so on has been loaded successfully.  That function should be called before the convert palette color stuff gets called, ideally as soon as images start loading.

packages/dicomImageLoader/src/imageLoader/createImage.ts
```diff
@@ -91,7 +91,7 @@ function createImage(
   return new Promise<DICOMLoaderIImage | Types.IImageFrame>(
```
wayfarer3130, line 102: You should return the result of an async function here, that is called with the decodePromise as input.  That preserves the error handling and then you can call the palette colour handling correctly.

packages/dicomImageLoader/src/imageLoader/colorSpaceConverters/convertPALETTECOLOR.ts
```diff
@@ -104,6 +67,57 @@ export default function (
...
+ * with automatic fetching of palette color lookup tables if not already loaded.
+ * This is useful for users who want to convert palette color frames without
+ * manually fetching the palette data first.
+ *
+ * @param imageFrame - The ImageFrame to convert
+ * @param colorBuffer - The buffer to write the converted pixel data to
+ * @param useRGBA - Whether to output RGBA (true) or RGB (false)
+ * @returns Promise that resolves when conversion is complete
+ */
+export async function convertPALETTECOLORWithFetch(
```
wayfarer3130: Can you externalize this function to "fetchLUTForInstance"
It should return a promise only if there is LUT data to fetch, and should also check the other two externalizable LUTS (modality and voi).
Then that one gets awaited before doing the regular convertPaletteColor.

packages/dicomImageLoader/src/imageLoader/colorSpaceConverters/convertPALETTECOLOR.ts
```diff
@@ -12,84 +13,46 @@ function convertLUTto8Bit(lut: number[], shift: number) {
...
+ * Note: Palette bulkdata must be loaded on the imageFrame before calling this function
  *
- * @param imageFrame - The ImageFrame to convert
+ * @param imageFrame - The ImageFrame to convert (must have palette data loaded)
  * @param colorBuffer - The buffer to write the converted pixel data to
+ * @param useRGBA - Whether to output RGBA (true) or RGB (false)
  * @returns
  */
-export default function (
+export default function convertPALETTECOLOR(
```
wayfarer3130: Name this convertPaletteColor for consistency.

wayfarer3130 (approved):

sedghi (changes requested):
it is not a request change but just to grab your attention

- can you go over this example, and compare your local version of rendering to the deploy and make sure we don't have regression (i'm not sure if we have test for these)
https://www.cornerstonejs.org/live-examples/dicomimageloaderwadouri

- also can you connect it to OHIF and make sure US studies work fine, actually run the OHIF tests with your local version of cornerstone

rodrigobasilio2022:
Look in to the example: one difference. In deflate option it gives an error and OSS silent give the same error, so i think the PR is better in not hiding problems. OHIF appears to look ok (didn't open all US examples)

wayfarer3130 (changes requested):
When I link to OHIF and run, I get an exception that hte palette colour hasn't been loaded yet.  I suspect the load palette data is a little too late.

packages/dicomImageLoader/src/imageLoader/colorSpaceConverters/convertPALETTECOLOR.ts
```diff
@@ -104,6 +67,48 @@ export default function (
...
+ * with automatic fetching of palette color lookup tables if not already loaded.
+ * This is useful for users who want to convert palette color frames without
+ * manually fetching the palette data first.
+ *
+ * @param imageFrame - The ImageFrame to convert
+ * @param colorBuffer - The buffer to write the converted pixel data to
+ * @param useRGBA - Whether to output RGBA (true) or RGB (false)
+ * @returns Promise that resolves when conversion is complete
+ */
+export async function convertPaletteColorWithFetch(
```
wayfarer3130, line 104: I think you should assume this is too late in the process and that the fetchLUT should have already been called.  I would suggest deleting this method and ensuring the data is fetched early enough and added to the actual instance.

packages/dicomImageLoader/src/imageLoader/colorSpaceConverters/fetchLUTForInstance.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  if (needsPaletteLUT) {
+    // Fetch palette data if not already loaded
+    const palettePromises = Promise.all([
+      fetchPaletteData(imageFrame, 'red', null),
+      fetchPaletteData(imageFrame, 'green', null),
+      fetchPaletteData(imageFrame, 'blue', null),
+    ]).then(([redData, greenData, blueData]) => {
+      // Attach palette data to imageFrame
+      imageFrame.redPaletteColorLookupTableData = redData;
```
wayfarer3130, line 32: It probably needs to be fetched in the instance metadata instead so it can be stored there.

packages/dicomImageLoader/src/imageLoader/createImage.ts
```diff
@@ -18,10 +18,11 @@ import isColorImageFn from '../shared/isColorImage';
 import removeAFromRGBA from './removeAFromRGBA';
 import isModalityLUTForDisplay from './isModalityLutForDisplay';
 import setPixelDataType from './setPixelDataType';
+import { fetchPaletteData } from './colorSpaceConverters/fetchPaletteData';
 
 let lastImageIdDrawn = '';
 
-function createImage(
+async function createImage(
```
wayfarer3130, line 25: You can't make the overall function async.  The fetch should really happen at the time the instance is about to be added to the metadata tree, as you have access to the full data there an can convert inline binary as well - and inline binary will be a common format for that data.

PR #2615, opened 2026-02-13
feat(tools): implement orthogonalT and lineSegment modes for PlanarFreehandROI open U-shape contours

TFRadicalImaging:
### Context

The PlanarFreehandROITool supports open U-shaped contours (e.g. LA horseshoe shapes) with a "T"
overlay: a dashed chord connecting the two endpoints and a dashed line from the chord midpoint to a
"peak" point. Currently, the peak is always the **farthest point** on the contour from the midpoint
(in any direction).

For LA cardiac workflows, the T line should instead be **perpendicular to the chord**, ending where
that perpendicular intersects the contour. The type `'orthogonalT'` already existed in the type
system but had no implementation.

Additionally, a `'lineSegment'` mode was needed to draw only the closing chord without any T-line,
and the enclosed region (contour + chord) should optionally be filled/shaded.

### Changes & Results

- **`orthogonalT` mode**: T-line endpoint is computed via ray-contour intersection perpendicular to
  the chord midpoint, instead of the farthest point. Falls back to farthest-point if no intersection
  is found.
- **`lineSegment` mode**: Draws only the dashed chord connecting the two open endpoints — no T-line
  or peak calculation.
- **Shaded fill**: The enclosed region (contour + chord) is now filled when `fillOpacity > 0`,
  reusing the existing `drawPathSvg` fill mechanism from closed contours.
- **Refactor**: Centralized variant dispatch into `resolveVectorToPeak` /
  `resolveVectorToPeakOnRender` in `findOpenUShapedContourVectorToPeak.ts`, eliminating duplicated
  if/else branching across `drawLoop.ts`, `openContourEditLoop.ts`, and `renderMethods.ts`. Adding
  future variants now requires editing only one function.
- **Example updated**: `planarFreehandROITool` example now has a dropdown to select between
  `farthestT`, `orthogonalT`, and `lineSegment` modes.

**Files changed:**

- `packages/tools/src/tools/annotation/planarFreehandROITool/findOpenUShapedContourVectorToPeak.ts`
- `packages/tools/src/tools/annotation/planarFreehandROITool/renderMethods.ts`
- `packages/tools/src/tools/annotation/planarFreehandROITool/drawLoop.ts`
- `packages/tools/src/tools/annotation/planarFreehandROITool/openContourEditLoop.ts`
- `packages/tools/examples/planarFreehandROITool/index.ts`

### Testing

1. Run `bun example planarFreehandROITool`
2. Draw an open U-shaped (horseshoe) contour
3. Select the contour
4. Pick a mode from the **U-Shape Mode** dropdown:
   - `farthestT` → T-line goes to the farthest point from chord midpoint (unchanged behavior)
   - `orthogonalT` → T-line is perpendicular to the chord
   - `lineSegment` → only the dashed chord, no T-line
5. Click **"Apply U-Shape to selected open contour"**
6. Verify the T overlay renders correctly for each mode
7. Edit the contour and verify the peak recalculates correctly
8. To test fill shading, set `fillOpacity > 0` in annotation style

#### PR

#### Code

      etc.)

      additions or removals.

#### Tested Environment

- [x] OS: macOS 25.1.0
- [x] Node version: >=20
- [x] Browser: Versão 144.0.7559.134 (Oficial Version) (arm64)

### DIFF
--- a/packages/tools/examples/planarFreehandROITool/index.ts
+++ b/packages/tools/examples/planarFreehandROITool/index.ts
@@ -4,14 +4,14 @@ import {
   Enums,
   volumeLoader,
   getRenderingEngine,
-  getEnabledElement,
 } from '@cornerstonejs/core';
 import {
   initDemo,
   createImageIdsAndCacheMetaData,
   setTitleAndDescription,
   addButtonToToolbar,
-  addDropdownToToolbar,
+  addFillOpacityDropdownToToolbar,
+  addUShapeModeDropdownToToolbar,
   createInfoSection,
 } from '../../../../utils/demo/helpers';
 import * as cornerstoneTools from '@cornerstonejs/tools';
@@ -28,20 +28,17 @@ const {
   ZoomTool,
   ToolGroupManager,
   Enums: csToolsEnums,
-  annotation,
 } = cornerstoneTools;
 
 const { ViewportType } = Enums;
 const { MouseBindings } = csToolsEnums;
-const { selection } = annotation;
-const defaultFrameOfReferenceSpecificAnnotationManager =
-  annotation.state.getAnnotationManager();
 
 // Define a unique id for the volume
 const volumeName = 'CT_VOLUME_ID'; // Id of the volume less loader prefix
 const volumeLoaderScheme = 'cornerstoneStreamingImageVolume'; // Loader id which defines which volume loader to use
 const volumeId = `${volumeLoaderScheme}:${volumeName}`; // VolumeId with loader id + volume id
 const renderingEngineId = 'myRenderingEngine';
+const toolGroupId = 'STACK_TOOL_GROUP_ID';
 
 const viewportIds = ['CT_STACK', 'CT_VOLUME_SAGITTAL'];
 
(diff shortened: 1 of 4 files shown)


packages/tools/src/tools/annotation/planarFreehandROITool/findOpenUShapedContourVectorToPeak.ts
```diff
@@ -63,14 +63,138 @@ export default function findOpenUShapedContourVectorToPeak(
...
+          bestIntersection = [
+            centerOfFirstToLast[0] + t * dir[0],
+            centerOfFirstToLast[1] + t * dir[1],
+          ];
+        }
+      }
+    }
+  }
+
+  // If no intersection found, fall back to the existing farthest-point logic.
```
wayfarer3130: Fallback to no result and add a warn message.

packages/tools/src/tools/annotation/planarFreehandROITool/findOpenUShapedContourVectorToPeak.ts
```diff
@@ -63,14 +63,138 @@ export default function findOpenUShapedContourVectorToPeak(
...
+    -firstToLastUnitVector[0],
+  ];
+
+  // Find the center of the chord.
+  const centerOfFirstToLast: Types.Point2 = [
+    (first[0] + last[0]) / 2,
+    (first[1] + last[1]) / 2,
+  ];
+
+  // Cast rays in both normal directions and find the farthest intersection
```
wayfarer3130: First intersection is acceptable actually - where first is defined as a point on the curve that is orthogonal to the closing line within isEqual limits

wayfarer3130: Probably leave this code as is, just thinking that first item found would make it fast to return and simplify the code.

packages/tools/examples/planarFreehandROITool/index.ts
wayfarer3130, line 1: Can you make this apply automatically on changing the pulldown?  I'd also like a false/empty on the open to remove it.
Then, please add an opacity setting pulldown so that you can test your other change.

packages/tools/src/tools/annotation/planarFreehandROITool/findOpenUShapedContourVectorToPeak.ts
```diff
@@ -63,14 +63,138 @@ export default function findOpenUShapedContourVectorToPeak(
...
+  canvasPoints: Types.Point2[],
+  viewport: Types.IStackViewport | Types.IVolumeViewport
+): Types.Point3[] {
+  // Find chord from first to last point.
+  const first = canvasPoints[0];
+  const last = canvasPoints[canvasPoints.length - 1];
+
+  const firstToLastUnitVector = vec2.create();
+
+  vec2.set(firstToLastUnitVector, last[0] - first[0], last[1] - first[1]);
```
wayfarer3130: The preferred form is:
const firstToLastUnitVector = vec2.sub(vec2.create(), last, first);

packages/tools/src/tools/annotation/planarFreehandROITool/findOpenUShapedContourVectorToPeak.ts
```diff
@@ -63,14 +63,138 @@ export default function findOpenUShapedContourVectorToPeak(
...
+    (first[0] + last[0]) / 2,
+    (first[1] + last[1]) / 2,
+  ];
+
+  // Cast rays in both normal directions and find the farthest intersection
+  // with the contour segments.
+  let bestT = -1;
+  let bestIntersection: Types.Point2 | null = null;
+
+  for (const dir of [normalVector1, normalVector2]) {
```
wayfarer3130: I think this is worth fixing as you end up doing a double loop and the loops can get quite big so avoiding a full double loop is worthwhile.
```
const delta = vec2.create();
for( const p of canvasPoints) {
   vec2.sub(delta,p,centerOfFirstToLast);
   vec2.normalize(delta,delta)
   const dp = vec2.dot(delta,normalVector)
   if( isEqual(dp,0) ) { return p }
}
```

wayfarer3130: You would need to return the right type value, but I think that loop is a lot shorter/faster, and since this can get updated interactively it is worth optimizing it some.

packages/tools/src/tools/annotation/planarFreehandROITool/findOpenUShapedContourVectorToPeak.ts
```diff
@@ -63,14 +63,125 @@ export default function findOpenUShapedContourVectorToPeak(
...
+  // the perpendicular from the midpoint crosses the contour. Interpolate
+  // between the two bracketing points for a precise intersection.
+  const delta = vec2.create();
+  let prevDp: number | null = null;
+  let prevPoint: Types.Point2 | null = null;
+  let orthogonalPoint: Types.Point2 | null = null;
+
+  for (const p of canvasPoints) {
+    vec2.sub(delta, p, centerOfFirstToLast);
+    const dp =
```
wayfarer3130: This is vec2.dot(chordDir,delta)

Issue #2623, opened 2026-02-19
[Bug] "WindowLevelTool._getMultiplierFromDynamicRange" missing BitsStored capping for stack viewports


zainmo11:
### Describe the Bug

The `WindowLevelTool._getMultiplierFromDynamicRange` method applies `BitsStored` metadata capping to limit the dynamic range multiplier for **volume viewports**, but does **not** apply the same capping for **stack viewports**. This causes the WW/WL values to explode to extremely large numbers (e.g., `WW: 48,786,120`) when interacting with high-dynamic-range images (CR/DX X-rays) in stack viewports, making the tool unusable.

<img width="1459" height="804" alt="Image" src="https://github.com/user-attachments/assets/6a34b473-e276-4526-87b5-cd4c2280b11d" />

<img width="1449" height="849" alt="Image" src="https://github.com/user-attachments/assets/7c39ca19-fc96-4509-b5bb-4deca32dc1c2" />

### Steps to Reproduce

1. Load a CR or DX X-ray image using a **stack viewport** (`viewport.setStack(imageIds)`)
2. Activate the `WindowLevelTool`
3. Drag the mouse to adjust WW/WL
4. Observe that WW/WL values jump to extremely large numbers (millions+), and the image becomes all white

### The current behavior

In [`WindowLevelTool.ts` — `_getMultiplierFromDynamicRange`](https://github.com/cornerstonejs/cornerstone3D/blob/main/packages/tools/src/tools/WindowLevelTool.ts):

```js
_getMultiplierFromDynamicRange(viewport, volumeId) {
    let imageDynamicRange;
    if (volumeId) {
        //  Volume viewports: BitsStored capping is applied
        const BitsStored = imageVolume?.metadata?.BitsStored;
        const metadataDynamicRange = BitsStored ? 2 ** BitsStored : Infinity;
        imageDynamicRange = Math.min(calculatedRange, metadataDynamicRange);
    }
    else {
        //  Stack viewports: NO BitsStored capping — raw pixel range used directly
        imageDynamicRange = this._getImageDynamicRangeFromViewport(viewport);
    }
    // ...
}
```

[...]

PR #2624, opened 2026-02-19
fix(WindowLevelTool): cap image dynamic range using BitsStored for stack viewports

zainmo11:
fixes https://github.com/cornerstonejs/cornerstone3D/issues/2623

This PR aligns the dynamic range calculation for stack viewports with volume viewports by implementing BitsStored capping. This prevents window/level values from becoming excessively large when calculating the multiplier for images with wide dynamic ranges (e.g. 16-bit images with burned-in pixels), ensuring consistent behavior across viewport types.

### DIFF
--- a/packages/tools/src/tools/WindowLevelTool.ts
+++ b/packages/tools/src/tools/WindowLevelTool.ts
@@ -174,7 +174,6 @@ class WindowLevelTool extends BaseTool {
 
   _getMultiplierFromDynamicRange(viewport, volumeId) {
     let imageDynamicRange;
-    let BitsStored;
 
     if (volumeId) {
       const imageVolume = cache.getVolume(volumeId);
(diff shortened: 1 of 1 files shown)


sedghi:
Can you add what is behaviour before and after, maybe share a screenshot on how it looks?

zainmo11:
this is the case 
Screenshot From 2026-02-19 15-59-26(1)
when i change WL
Screenshot From 2026-02-19 15-59-34(1)


after edit it works well 

Screenshot From 2026-02-20 00-07-44


this is issue too for more details 
https://github.com/cornerstonejs/cornerstone3D/issues/2623

packages/tools/src/tools/WindowLevelTool.ts
```diff
@@ -201,6 +202,16 @@ class WindowLevelTool extends BaseTool {
         : Math.min(calculatedRange, metadataDynamicRange);
     } else {
       imageDynamicRange = this._getImageDynamicRangeFromViewport(viewport);
+      // Cap using BitsStored metadata (same logic as the volume viewport path above)
+      const imageId = viewport.getCurrentImageId?.();
```
sedghi: Could we refactor that out? It seems like we should apply it to both the volume and stack viewports, so there's no need to repeat it; just take it out of the if/else block.

zainmo11: okay it is done now 
u can review it again

zainmo11:
If you’d like, I can share the case

packages/tools/src/tools/WindowLevelTool.ts
```diff
@@ -188,28 +190,35 @@ class WindowLevelTool extends BaseTool {
...
+        BitsStored = imagePixelModule?.bitsStored;
+      }
+    }
+
+    // Burned in Pixels often use pixel values above the BitsStored.
+    // This results in a multiplier which is way higher than what you would
+    // want in practice. Thus we take the min between the metadata dynamic
+    // range upper value and actual middle slice dynamic range.
+    if (BitsStored) {
+      const metadataDynamicRange = 2 ** BitsStored;
```
wayfarer3130: This is the wrong value for signed data - it will be 2 ** (BitsStored-1) for signed data.  ALSO, you should clip the lower range to the minimum value of  -(2 ** (BitsStored-1))  for signed data

packages/tools/src/tools/WindowLevelTool.ts
```diff
@@ -188,28 +190,35 @@ class WindowLevelTool extends BaseTool {
...
-      // This results in a multiplier which is way higher than what you would
-      // want in practice. Thus we take the min between the metadata dynamic
-      // range upper value and actual middle slice dynamic range.
-      imageDynamicRange = !Number.isFinite(calculatedRange)
-        ? metadataDynamicRange
-        : Math.min(calculatedRange, metadataDynamicRange);
+      imageDynamicRange = calculatedDynamicRange[1] - calculatedDynamicRange[0];
+      BitsStored = imageVolume?.metadata?.BitsStored;
     } else {
       imageDynamicRange = this._getImageDynamicRangeFromViewport(viewport);
```
wayfarer3130, line 213: Can you add the clipping into getImageDYnamicRangeFromViewport directly rather than having logic int he window level tool - that avoids spreading the knowledge around.

wayfarer3130: It looks like it is getting applied twice, so ignore this comment.

packages/tools/src/tools/WindowLevelTool.ts
zainmo11: okay, done

wayfarer3130:
> If you’d like, I can share the case

If you have it anonymized, and are able to share, that would be appreciated.

zainmo11:
https://drive.google.com/file/d/1L4uOTkaTb8yge_g_ewSp6Sog_6DYGTh6/view?usp=sharing
@wayfarer3130

PR #2625, opened 2026-02-20
feat: Add metadata module to single-source metadata

wayfarer3130:
### Context

Create a metadata module to have a single spot for parsing/handling of different types of DICOM data.

ohif_ref: feat/use-beta-5.0-cs3d

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  


packages/dicomImageLoader/src/imageLoader/wadors/register.ts
```diff
@@ -2,8 +2,21 @@ import { metaData, registerImageLoader, type Types } from '@cornerstonejs/core';
...
   registerImageLoader('wadors', loadImage as unknown as Types.ImageLoaderFn);
+
+  if (options?.useMetadataProvider) {
+    return;
+  }
+
+  /**
+   * @deprecated The wadors metadata provider is deprecated.
+   * Use addDicomwebInstance from @cornerstonejs/metadata to register
+   * DICOMweb metadata directly into the INSTANCE_ORIG cache instead.
```
sedghi: we should include a link later to say see this link for more information or guide or something

packages/dicomImageLoader/src/types/LoaderOptions.ts
```diff
@@ -31,4 +31,10 @@ export interface LoaderOptions {
   errorInterceptor?: (error: LoaderXhrRequestError) => void;
   strict?: boolean;
   decodeConfig?: LoaderDecodeOptions;
+  /**
+   * When true, skips registering the legacy wadouri/wadors metadata providers.
+   * Use addBinaryDicomInstance and addDicomwebInstance from
+   * @cornerstonejs/metadata to populate the INSTANCE_ORIG cache instead.
+   */
+  useMetadataProvider?: boolean;
```
sedghi: i prefer the other way around, if we are thinking of deprecating that in future version -> useLegacyMetadataProvider

utils/demo/helpers/createImageIdsAndCacheMetaData.js
```diff
@@ -70,34 +80,58 @@ export default async function createImageIdsAndCacheMetaData({
...
-  imageIds.forEach((imageId) => {
-    let instanceMetaData =
-      cornerstoneDICOMImageLoader.wadors.metaDataManager.get(imageId);
-
-    if (!instanceMetaData) {
-      return;
+    // For multiframe expansions, store INSTANCE_ORIG for each frame imageId
+    if (useMetadata && imageIds.length !== originalImageIds.length) {
+      for (const imageId of imageIds) {
+        if (metaData.get('instanceOrig', imageId)) {
```
sedghi: i can't fine instanceOrig anywhere in the code, what is it? naming seems strange

packages/metadata/src/utilities/metadataProvider/dicomSplit.ts
```diff
@@ -0,0 +1,7 @@
+export function dicomSplit(value) {
```
sedghi, line 15: docs pleaes

packages/metadata/src/utilities/metadataProvider/makeArrayLike.ts
```diff
@@ -0,0 +1,28 @@
+export function makeArrayLike(obj) {
```
sedghi, line 15: docs please

utils/demo/helpers/createImageIdsAndCacheMetaData.js
sedghi: Pick from any of these pairs

- sourceInstance + instance
- rawInstance + instance
- baseInstance + resolvedInstance (or instance)

packages/metadata/src/utilities/Tags.ts
```diff
@@ -0,0 +1,182 @@
...
+  vr: string;
+  primaryGroup: string;
+  groups: string[];
+}
+
+/**
+ * Creates a tag entry defining module membership only.
+ * VR and VM are resolved from the dcmjs dictionary in addTag().
+ */
+export function tag(hexTag: string, ...groups: string[]): TagEntry {
```
sedghi: can we rename to indicate it creates actually? something like `createTagEntry`

packages/metadata/src/utilities/Tags.ts
```diff
@@ -0,0 +1,182 @@
...
+  hexTag: string
+): { name: string; vr: string; vm: string } | undefined {
+  return dicomDictionary[hexTag.toUpperCase()];
+}
+
+/**
+ * Looks up the hex tag code for a natural tag name (e.g. "SOPInstanceUID" → "00080018").
+ * Uses the mapTagInfo registry which is populated from the Tags definitions.
+ */
+export function lookupTagHex(name: string): string | undefined {
```
sedghi: maybe `getTagHexByName` or ``getTagHexByKeyword`?

packages/metadata/src/utilities/metadataProvider/addInstance.ts
```diff
@@ -0,0 +1,58 @@
...
+ *
+ * Parses the ArrayBuffer using dcmjs AsyncDicomReader with
+ * NaturalTagListener to produce a naturalized instance, then stores it
+ * in the INSTANCE_ORIG cache.
+ *
+ * @param imageId - The imageId to associate with this instance
+ * @param arrayBuffer - The DICOM Part 10 binary data
+ * @returns A promise that resolves to the naturalized instance object
+ */
+export async function addBinaryDicomInstance(
```
sedghi: what does binary here mean?

packages/metadata/src/utilities/metadataProvider/cacheData.ts
```diff
@@ -0,0 +1,77 @@
...
+
+const mapCacheData = new Map<string, Map<string, unknown>>();
+
+/**
+ * Creates a typed provider that caches results for the given type.
+ *
+ * Options can include: noCache to not cache this value or use the cached value,
+ * and reCache to get a new value and add it to the cache.
+ */
+export function cacheDataForType(type: string) {
```
sedghi: createTypeCacheProvider?

sedghi (commented):
The draft looks good and I understand most of it, but it would be helpful to add more to the documentation regarding how `addDicomwebInstance`, `addBinaryDicomInstance`, and `setCacheData` work together. I noticed `setCacheData` is used within `addDicomwebInstance`, but since `createImageIdsAndCacheMetadata` uses both, I'm a bit confused on how they interact.

utils/demo/helpers/createImageIdsAndCacheMetaData.js
wayfarer3130: I renamed to natural using a constant shared between OHIF and CS3D (since the OHIF platform/core now depends on the metadata module it makes sense to share those from metadata).

.agents/skills/metadata-package.md
```diff
@@ -0,0 +1,116 @@
+# @cornerstonejs/metadata Package
```
sedghi: This reads more like a architecture than an agent skill. 
Can you use the skill creator skill to create this skill https://github.com/anthropics/skills/blob/main/skills/skill-creator/SKILL.md 

It should have YAML frontmatter. 
Skill should be about what to do when editing/reviewing this area of code not explaining the architecture. 

Something like 
- When you work on metadata check these things
- do this don't do that 
- add tests everytime you do something here 

I suggest we move architecture to docs.

.circleci/config.yml
```diff
@@ -159,23 +160,19 @@ workflows:
...
           requires:
             - CHECKOUT
-      - HOLD_FOR_APPROVAL:
-          type: approval
-          requires:
-            - BUILD
       - NPM_PUBLISH:
           requires:
-            - HOLD_FOR_APPROVAL
+            - BUILD
```
sedghi, line 177: This is intentional as i like to double check when doing a breaking change

packages/adapters/src/adapters/Cornerstone3D/LabelData.ts
```diff
@@ -17,155 +17,125 @@ const {
...
+    /**
+     * This adds the additional information from the annotation data for
+     * the label data.
+     */
+    public contentItem() {
+        const contentEntries = this.tid300Item.contentItem();
+        const { label, handles } = this.annotation.data;
+
+        if (label) {
+            contentEntries.push(this.createQualitativeLabel(label));
```
sedghi: seems like we don't fix the entry anymore? please check if intentinoal

packages/adapters/src/adapters/Cornerstone3D/PlanarFreehandROI.ts
```diff
@@ -2,38 +2,31 @@ import MeasurementReport from './MeasurementReport';
...
 
 /** Contour/polyline SR logic is shared by LivewireContour, registered as a subtype. */
 class PlanarFreehandROI extends BaseAdapter3D {
   public static closedContourThreshold = 1e-5;
 
   static {
-    this.init('PlanarFreehandROI', ControlPointPolyline);
-    PlanarFreehandROI.registerSubType(PlanarFreehandROI, 'LivewireContour');
-    PlanarFreehandROI.registerSubType(PlanarFreehandROI, 'SplineROI');
+    this.init('PlanarFreehandROI', TID300Polyline);
```
sedghi: previously it registered planar, and live wire and spline, now it is only planar, please check if intentional

packages/adapters/src/adapters/Cornerstone3D/PlanarFreehandROI.ts
```diff
@@ -141,7 +108,6 @@ class PlanarFreehandROI extends BaseAdapter3D {
     return {
       /** From cachedStats */
       points,
-      controlPoints,
       area,
```
sedghi, line 145: please check if intentional removal, and whether requires migration guides

packages/adapters/package.json
```diff
@@ -84,7 +85,7 @@
   "dependencies": {
     "@babel/runtime-corejs2": "7.26.10",
     "buffer": "6.0.3",
-    "dcmjs": "0.49.4",
+    "dcmjs": "0.50.1",
```
sedghi, line 87: please add this to migration guides FYI, we updated dcmjs etc etc

packages/ai/package.json
```diff
@@ -26,6 +26,7 @@
   },
   "scripts": {
     "prebuild": "node ../../scripts/generate-version.js ./",
+    "prebuild:esm": "node ../../scripts/generate-version.js ./",
```
sedghi: looks the same as prebuild line above

packages/core/examples/ecg/index.ts
```diff
@@ -1,6 +1,7 @@
 import type { Types } from '@cornerstonejs/core';
 import { RenderingEngine, Enums } from '@cornerstonejs/core';
 import * as cornerstoneTools from '@cornerstonejs/tools';
+import { addPart10Instance } from '@cornerstonejs/metadata/utilities/metadataProvider';
```
sedghi: I recommend `addDicomPart10Instance`

packages/core/examples/ecg/index.ts
```diff
@@ -185,6 +189,37 @@ async function run() {
     },
   });
 
+  // Upload ECG button — load a DICOM waveform file and display it
+  addButtonToToolbar({
```
sedghi, line 193: upload sounds like we upload to our server and people might not like it, maybe use local ecg or something

packages/core/src/RenderingEngine/ContextPoolRenderingEngine.ts
```diff
@@ -194,11 +194,6 @@ class ContextPoolRenderingEngine extends BaseRenderingEngine {
 
   /**
    * Resizes viewports that use VTK.js for rendering.
-   * Only updates the VTK offscreen size when the displayed size (canvas client size in device pixels)
-   * differs from the current rendered size (canvas.width/height). The on-screen canvas dimensions
-   * are updated only when the VTK render result is copied in _copyToOnscreenCanvas, avoiding
-   * flicker during resize. If a render is already scheduled, resize is deferred until the next
-   * resize() call.
    */
   protected _resizeVTKViewports(
```
sedghi, line 203: please check if intentionally removed function comment from before

packages/core/src/RenderingEngine/ContextPoolRenderingEngine.ts
```diff
@@ -420,8 +412,8 @@ class ContextPoolRenderingEngine extends BaseRenderingEngine {
     );
     const maxSize = this.contextPool.getMaxSizeForContext(contextIndex);
 
-    const viewportWidth = viewport.sWidth;
-    const viewportHeight = viewport.sHeight;
+    const viewportWidth = viewport.canvas.width;
```
sedghi: not sure why metadata PR is touching the viewport canvas, and moving from sWidth to width. Is this intentional? document migration if needed? and add comment on top of this line why sWidth is wrong if so

packages/core/src/utilities/logger.ts
```diff
@@ -1,96 +1,20 @@
...
-
-if (typeof window !== 'undefined') {
-  (window as unknown as WindowLog).log = loglevel;
-}
-
-export type Logger = LogLevelLogger & {
-  getLogger: (...categories: string[]) => Logger;
-};
-
+import { logging } from '@cornerstonejs/metadata';
```
sedghi: strange our logging is now from metadata? instead of core?

packages/core/package.json
```diff
@@ -69,6 +69,7 @@
   },
   "scripts": {
     "prebuild": "node ../../scripts/generate-version.js ./",
+    "prebuild:esm": "node ../../scripts/generate-version.js ./",
```
sedghi: looks the same as prebuild

packages/core/package.json
```diff
@@ -86,6 +87,12 @@
     "gl-matrix": "3.4.3",
     "loglevel": "1.9.2"
   },
+  "devDependencies": {
+    "@cornerstonejs/metadata": "4.18.4"
+  },
+  "peerDependencies": {
+    "@cornerstonejs/metadata": ">=4.18.0 <6.0.0-0"
+  },
```
sedghi, line 96: why dev and peer is not the same version

packages/dicomImageLoader/src/imageLoader/wadouri/loadImage.ts
```diff
@@ -169,62 +171,179 @@ function getLoaderForScheme(scheme: string): LoadRequestFunction {
...
+  }
+  return result;
+}
+
+/**
+ * Loads an image from the NATURAL path: ensures NATURAL is populated (fetch +
+ * addPart10Instance when needed), gets frame pixel data via getMetaData(MetadataModules.COMPRESSED_FRAME_DATA, imageId, `{ frameIndex }`),
+ * then creates IImage. Does not use dataSetCacheManager.
+ */
+function loadImageFromNatural(
```
sedghi: I recommend `loadImageFronNaturalizedMetadata` to fit better with dcmjs naturalize

packages/dicomImageLoader/src/imageLoader/wadouri/loadImage.ts
```diff
@@ -169,62 +171,179 @@ function getLoaderForScheme(scheme: string): LoadRequestFunction {
...
+  const promise = (async (): Promise<DICOMLoaderIImage> => {
+    const start = Date.now();
+    console.log('[dicomImageLoader/wadouri] loadImageFromNatural: start', {
+      imageId,
+      scheme: parsedImageId.scheme,
+      url: parsedImageId.url,
+      frameIndex,
+    });
+
+    const NATURAL = MetadataEnums.MetadataModules.NATURAL;
```
sedghi: maybe `MetadataModules.NATURALIZED` not sure on this one

packages/dicomImageLoader/src/decodeImageFrameWorker.js
```diff
@@ -151,11 +151,7 @@ export function postProcessDecodedPixels(
       maxAfterScale = scaledValues.max;
     }
   } else if (disableScale) {
-    imageFrame.preScale = {
-      enabled: true,
-      scaled: false,
-    };
-
+    // Do not set imageFrame.preScale when scaling is disabled (e.g. identity slope/intercept).
```
sedghi, line 154: I think requires migration as a FYI

packages/dicomImageLoader/package.json
```diff
@@ -111,12 +112,14 @@
...
   },
-  "peerDependencies": {
+  "devDependencies": {
     "@cornerstonejs/core": "5.0.0-beta.1",
-    "dicom-parser": "1.8.21"
+    "@cornerstonejs/metadata": "4.18.4"
   },
-  "devDependencies": {
-    "@cornerstonejs/core": "5.0.0-beta.1"
+  "peerDependencies": {
```
sedghi, line 119: not sure about this range stuff, why can't we just add the version since all of these will go as 5.0

packages/labelmap-interpolation/package.json
```diff
@@ -26,6 +26,7 @@
   },
   "scripts": {
     "prebuild": "node ../../scripts/generate-version.js ./",
+    "prebuild:esm": "node ../../scripts/generate-version.js ./",
```
sedghi: ditto

packages/labelmap-interpolation/package.json
```diff
@@ -54,13 +55,13 @@
...
-  "peerDependencies": {
-    "@cornerstonejs/core": "5.0.0-beta.1",
-    "@cornerstonejs/tools": "5.0.0-beta.1",
-    "@kitware/vtk.js": "34.15.1"
-  },
   "devDependencies": {
     "@cornerstonejs/core": "5.0.0-beta.1",
     "@cornerstonejs/tools": "5.0.0-beta.1"
+  },
+  "peerDependencies": {
```
sedghi, line 61: ditto

packages/metadata/src/enums/MetadataModules.ts
```diff
@@ -0,0 +1,211 @@
...
+   *
+   * In the legacy metadata modules, this could be a call time generated object
+   * where it combines all the individual modules back into an instance.  However,
+   * in the newer metadata module, this is a base object used to create other modules
+   * assuming this item has already been stored.
+   *
+   * Typically this object will be created from the 'natural' object using the
+   * combineFramesInstance utility function.
+   */
+  INSTANCE = 'instance',
```
sedghi, line 100: sounds like a migration guide FYI

packages/metadata/src/utilities/metadataProvider/cacheData.ts
```diff
@@ -0,0 +1,88 @@
...
+export function addCacheForType(type: string, options?) {
+  addTypedProvider(type, createTypeCacheProvider(type), {
+    priority: 50_000,
+    clear: clearTypedCacheData.bind(null, type) as () => void,
+    clearQuery: clearTypedCacheData.bind(null, type),
+    ...options,
+  });
+}
+
+export function registerCacheProviders() {
```
sedghi, line 329: how about the general module? I can see that we have a constant call on mouse move to get FOR which is crazy, maybe we should add it by default for general too?

packages/dicomImageLoader/src/imageLoader/createImage.ts
```diff
@@ -283,9 +405,11 @@ async function createImage(
           windowWidth: voiLutModule.windowWidth
             ? voiLutModule.windowWidth[0]
             : undefined,
-          voiLUTFunction: voiLutModule.voiLUTFunction
-            ? voiLutModule.voiLUTFunction
-            : undefined,
+          voiLUTFunction:
+            (voiLutModule.voiLUTFunction?.length &&
```
sedghi, line 409: if the voiLUTFunction is a string like "SIGMOID" this will return S

packages/metadata/src/utilities/metadataProvider/combineFrameInstance.ts
```diff
@@ -0,0 +1,245 @@
...
+      value: frameNumber,
+      writable: true,
+      enumerable: true,
+      configurable: true,
+    });
+    return newInstance;
+  }
+
+  // For RTDOSE datasets
+  if (instance.GridFrameOffsetVector) {
```
sedghi, line 138: AI review on this file gave this please check it out



> DICOM RTDOSE GridFrameOffsetVector has two valid meanings: relative offsets when the first value is 0, or absolute patient z positions when the first value equals ImagePositionPatient[2] for axial grids. This code always treats every value as a relative offset from ImagePositionPatient, so absolute-offset RTDOSE objects get shifted by the origin again; frame 1 at z=100 with offsets [100, 102, ...] becomes z=200. Detect the absolute form before scaleAndAdd, or only apply the current relative calculation when the first offset is zero.

packages/metadata/src/utilities/metadataProvider/combineFrameInstance.ts
```diff
@@ -0,0 +1,245 @@
...
+ * Combine the Per instance frame data, the shared frame data
+ * and the root data objects.
+ * The data is combined by taking nested sequence objects within
+ * the functional group sequences.  Data that is directly contained
+ * within the functional group sequences, such as private creators
+ * will be ignored.
+ * This can be safely called with an undefined frame in order to handle
+ * single frame data. (eg frame is undefined is the same as frame===1).
+ */
+export const combineFrameInstance = (frame, instance) => {
```
sedghi, line 17: AI review of this file gave this 

> ImagePositionPatientToUse is initialized from the root instance before the per-frame functional group is merged, then assigned back after merging. If a multiframe object has a root ImagePositionPatient plus per-frame PlanePositionSequence values, this overwrites every frame's merged per-frame position with the root position. The fallback should only be applied when the merged frame still has no ImagePositionPatient, or when the detector-derived NM fallback was actually computed.

packages/metadata/src/utilities/metadataProvider/uriModule.ts
```diff
@@ -0,0 +1,43 @@
+import { MetadataModules } from '../../enums';
+import { addTypedProvider } from '../../metaData';
+
+export interface UriModule {
```
sedghi, line 4: maybe we include frame of reference UID?

packages/metadata/src/utilities/modules/clinicalTrial.ts
```diff
@@ -0,0 +1,8 @@
+import type { ModuleTagEntry } from './index';
+
+/** Clinical Trial module tags. */
+export const tags: ModuleTagEntry[] = [
+  'ClinicalTrialSponsorName',
+  'ClinicalTrialSiteID',
+  'ClinicalTrialSiteName',
```
sedghi, line 7: ClinicalTrialSubjectID is used quite often too

packages/metadata/src/utilities/modules/ptImage.ts
```diff
@@ -0,0 +1,7 @@
+import type { ModuleTagEntry } from './index';
+
+/** PET Image module tags. */
+export const tags: ModuleTagEntry[] = [
```
sedghi, line 4: the modality name is PT, maybe ptImage?

packages/metadata/src/utilities/modules/ptIsotope.ts
```diff
@@ -0,0 +1,4 @@
+import type { ModuleTagEntry } from './index';
+
+/** PET Isotope module tags. */
+export const tags: ModuleTagEntry[] = [['RadiopharmaceuticalInfo', '00540016']];
```
sedghi, line 4: if you decide to rename pet this file too

packages/metadata/src/utilities/modules/ptSeries.ts
```diff
@@ -0,0 +1,8 @@
+import type { ModuleTagEntry } from './index';
+
+/** PET Series module tags. */
+export const tags: ModuleTagEntry[] = [
```
sedghi, line 4: if you decide to rename pet this file too

packages/metadata/src/utilities/isEqual.ts
```diff
@@ -0,0 +1,110 @@
+function areNumbersEqualWithTolerance(
+  num1: number,
+  num2: number,
+  tolerance: number
+): boolean {
+  return Math.abs(num1 - num2) <= tolerance;
+}
+
+function areArraysEqual(
```
sedghi: it is very strange that these are getting moved to metadata and core is also importing from metadata

packages/metadata/src/utilities/logging.ts
```diff
@@ -0,0 +1,57 @@
+/**
+ * Logging utilities. Uses the log exported by dcmjs so that metadata and
+ * other packages share the same log hierarchy.
+ */
+import { log } from 'dcmjs';
+
+export type Logger = {
```
sedghi: same as this, i am fine if you want to create a @cornerstonejs/utils new package that everyone import from there honestly than logging is imported from metadata in core

packages/metadata/src/utilities/toNumber.ts
```diff
@@ -0,0 +1,24 @@
+/**
+ * Converts a value to a finite number, returning undefined if the value is not finite.
+ *
+ * @param value - The value to convert to a finite number
+ * @returns The finite number value, or undefined if the value is not finite
+ */
+export function toFiniteNumber(value: number | undefined): number | undefined {
```
sedghi: same

packages/metadata/src/registerDefaultProvider.ts
```diff
@@ -0,0 +1,94 @@
...
+ * - Combine frame provider (handles multiframe instances)
+ * - Tag modules (converts instance data to module-specific results)
+ * - Image plane calibrated provider
+ * - Calibration module provider
+ * - Pixel data update provider (palette color handling)
+ * - Transfer syntax provider
+ *
+ * Call at start or after removeAllProviders() to re-establish typed providers; duplicates are skipped.
+ */
+export function registerDefaultProvider() {
```
sedghi: registerDefaultProviders i guess since it is doing bunch of them

utils/demo/helpers/createImageIdsAndCacheMetaData.js
```diff
@@ -31,11 +42,9 @@ export default async function createImageIdsAndCacheMetaData({
   wadoRsRoot,
   client = null,
   convertMultiframe = true,
+  useMetadata = true,
```
sedghi: what does useMetadata mean? can we not use metadata?

sedghi: i think we can remove this new flag, and derive it from if uselegacy is on or off

packages/metadata/src/metaData.ts
```diff
@@ -0,0 +1,336 @@
...
+  if (!newProvider) {
+    throw new Error(`newProvider is empty for ${type}`);
+  }
+  typedProviderMap.set(type, newProvider);
+}
+
+/**
+ * A provider for the general typed providers.
+ */
+export function typedProviderProvider(type: string, query: string, options) {
```
sedghi: metadataModuleProvider?

utils/demo/helpers/createImageIdsAndCacheMetaData.js
```diff
@@ -70,34 +79,54 @@ export default async function createImageIdsAndCacheMetaData({
...
+        if (frameIdx >= 0) {
+          const baseImageId = imageId.substring(0, frameIdx) + '/frames/1';
+          const baseInstance = metaData.get('instanceOrig', baseImageId);
+          if (baseInstance) {
+            setCacheData(MetadataModules.NATURAL, imageId, baseInstance);
+          }
+        }
+      }
     }
+  }
```
sedghi: I am not a fan of this api, can we do something simpler for people to not know about the narual, setCachedData etc etc

I can see addDicomwebInstance already does all of this 

```
export function addDicomwebInstance(
  imageId: string,
  metadata: Record<string, unknown>
) {
  const iterator = new MetaDataIterator(metadata);
  const listener = NaturalTagListener.createMetadataListener();
  listener.startObject();
  iterator.syncIterator(listener);
  const instance = listener.pop();
  setCacheData(MetadataModules.NATURAL, imageId, instance);
  return instance;
}
```

can't we just handle multiframe there with the new uri module?

```
const instanceImageId = getUriModule(imageId)?.baseImageId ?? imageId;
setCacheData(MetadataModules.NATURAL, instanceImageId, natural);
```

sedghi (changes requested):
I think I understand the intent of the cache layer now. My concern is that the current API exposes that cache too directly. Callers can write `setCacheData` and read `getCacheData`, but some cached values are derived from other values. If `NATURAL` changes, the derived `INSTANCE`, `IMAGE_PLANE`, `FRAME_MODULE`, etc. entries are not automatically invalidated, so the cache can silently become stale or inconsistent.

I think metadata should follow the same model we already use for image and volume loading. For images, callers use `loadImage` / `loadAndCacheImage`. The cache is still there, but it is owned by the loader: it checks for existing load objects, deduplicate works, writes to the cache, and handles decache/purge. Users do not manually populate image or volume cache entries to make loading work, i mean we don't have addImageToCache for example. 

I think metadata should have the same shape. The public API should be semantic ingestion/removal APIs, not cache mutation APIs. For example:

- `addDicomWebMetadata(imageId, metadata)`
- `addDicomPart10ArrayBuffer(imageId, arrayBuffer)`
- `addNaturaliuzedMetadata(imageId, naturalInstance)` as an advanced/internal-style maybe?
- `removeMetadata(imageId)`
- `clearMetadata()`

Internally, we can still have a source metadata store and a derived metadata cache, but the package should own invalidation. So:

1. Store source metadata once per instance imageId.
2. Resolve requested imageIds to `{ requestedImageId or originalImageid or just imageId, instanceImageId, frameNumber }`.
3. Let providers derive modules from source metadata plus frame context.
4. Cache derived provider results internally.
5. When source metadata is added/replaced/removed, clear the derived entries for that instance/frame set.

The frame resolution should happen inside the metadata layer, the same way image/volume cache behavior happens inside the loader layer.

.agents/skills/metadata-package.md
sedghi: test1

packages/core/src/RenderingEngine/ContextPoolRenderingEngine.ts
```diff
@@ -207,18 +202,16 @@ class ContextPoolRenderingEngine extends BaseRenderingEngine {
   ) {
     const devicePixelRatio = window.devicePixelRatio || 1;
 
+    const canvasesDrivenByVtkJs = vtkDrivenViewports.map(
```
sedghi: testd

.circleci/config.yml
```diff
@@ -159,23 +160,19 @@ workflows:
           requires:
             - CHECKOUT
 
-  # MERGE TO MAIN/BETA
+  # MERGE TO MAIN
```
sedghi, line 163: test1234

packages/core/src/utilities/logger.ts
wayfarer3130: Metadata needs logging, and is included into core, so to make it available to metadata, it needs to be there or in something included from metadata.

packages/core/package.json
wayfarer3130: devDependency controls build order and thus needs exact same version to force build order

peerDependency controls linking, so if you want to allow bun link in a different version than original, you need to have the range.

This change fixes bugs in using debug bun linked versions as well as in ohif-downstream checks.

packages/dicomImageLoader/package.json
wayfarer3130: That breaks bun link when the version included is not the version linked.

packages/metadata/src/utilities/isEqual.ts
wayfarer3130: I'm trying hard to NOT change the API for core, so that it can be deprecated in core for a version or two before it gets removed from core.

sedghi (commented):
Thanks for the update, it is much better, i have the following commetns 

1. get() is doing ingestion for metadata which is quite not expected, i prefer having separate .add and .get methods. also can we rename to be more explicit naming
Also we don't need to separate the sync from async. 

  metaData.add(NATURALIZED, imageId, { dicomwebJson })   // sync source path
  metaData.add(NATURALIZED, imageId, { part10Buffer })   // async source path

This way we don't need another `asyncNaturalized` module and consumer doesn't need to remember to use which

---

2. instead of subclassing the cache, can we do factory instead 

// before
  class CalibrationCache extends CacheData {
    static setCalibration(imageId: string, value: CalibrationValue) {
      this.setCacheDataInternal('calibration', imageId, value);
    }
  }
  export const calibration = CalibrationCache;

  // after
  export const calibration = createWritableCache<CalibrationValue>({
    type: MetadataModules.CALIBRATION,
    validate: (v) => /* optional shape check */,
  });

  // usage
  calibration.write(imageId, value);
  calibration.read(imageId);

packages/metadata/src/utilities/metadataProvider/uriModule.ts
wayfarer3130: It is primarily used by the combine frame handling to decide on which frame to combine -mostly as a way to get the frame number/base uri without having to have parsing code.  Probably I should shatter it down into just the frame number, and leave the existing base image id handling for the image id part of it.    That way it becomes easy to ask for just "give me the frame number for this imageId"

packages/metadata/src/utilities/metadataProvider/combineFrameInstance.ts
wayfarer3130: Resolved, but leaving the comment for you to look at.

packages/adapters/src/adapters/Cornerstone3D/LabelData.ts
wayfarer3130: Yeah, I had a merge go wrong and didn't realize it in time.  Reverting to origin/main for all of adapters except the build bits.

.agents/skills/metadata-migration.md
```diff
@@ -0,0 +1,44 @@
+---
+name: metadata-migration
+description: Migrate codebases from legacy CS3D metadata flows to the metadata branch model with required init/provider updates and recommended provider/cache/imageId changes.
+---
+
+# Metadata Migration Skill (`origin/main` -> `metadata`)
+
+Use this skill to migrate codebases from the legacy CS3D metadata flow in
```
sedghi, line 8: I think this should say which style, to which style. Like code bases (old style) to new style. The agent doesn't need to know about branches

.agents/skills/metadata-migration.md
```diff
@@ -0,0 +1,44 @@
...
+name: metadata-migration
+description: Migrate codebases from legacy CS3D metadata flows to the metadata branch model with required init/provider updates and recommended provider/cache/imageId changes.
+---
+
+# Metadata Migration Skill (`origin/main` -> `metadata`)
+
+Use this skill to migrate codebases from the legacy CS3D metadata flow in
+`origin/main` to the `metadata` branch model.
+
+## Mandated changes
```
sedghi, line 11: These are all very high level and not informative for a working agent. Can you ask claude to read the diff between the this branch and beta (or main) and create a skill using skill-creator skill for migration

packages/core/src/utilities/isEqual.ts
```diff
@@ -1,110 +1,11 @@
...
- */
-export function isNumber(n: number[] | number): boolean {
-  if (Array.isArray(n)) {
-    return isNumber(n[0]);
-  }
-  return isFinite(n) && !isNaN(n);
-}
-
-export default isEqual;
+/** @deprecated Import from `@cornerstonejs/utils` instead. */
```
sedghi, line 1: Thanks

packages/core/src/utilities/isValidVolume.ts
```diff
@@ -48,12 +53,14 @@ function isValidVolume(imageIds: string[]): boolean {
 
   for (let i = 0; i < imageIds.length; i++) {
     const imageId = imageIds[i];
-    const { modality, seriesInstanceUID } = metaData.get(
-      'generalSeriesModule',
-      imageId
-    );
-    const { imageOrientationPatient, pixelSpacing, columns, rows } =
-      metaData.get('imagePlaneModule', imageId);
+    const general = metaData.get('generalSeriesModule', imageId);
```
sedghi: shouldn't these be using the ENUM modules

utils/demo/helpers/createImageIdsAndCacheMetaData.js
```diff
@@ -70,34 +77,45 @@ export default async function createImageIdsAndCacheMetaData({
...
-      instanceMetaData
-    );
+    if (useLegacyWadoRs) {
+      cornerstoneDICOMImageLoader.wadors.metaDataManager.add(
+        imageId,
+        instanceMetaData
+      );
+    }
+
+    if (useTypedMetadata) {
```
sedghi: talked over zoom

utils/demo/helpers/createImageIdsAndCacheMetaData.js
```diff
@@ -70,34 +77,45 @@ export default async function createImageIdsAndCacheMetaData({
...
+    }
+
     return imageId;
   });
 
   // if the image ids represent multiframe information, creates a new list with one image id per frame
   // if not multiframe data available, just returns the same list given
   if (convertMultiframe) {
+    const originalImageIds = [...imageIds];
     imageIds = convertMultiframeImageIds(imageIds);
```
sedghi: talked about doing something this in another PR 

imageIds = addDicomWebInstance(imageId, instance)

to handle multiframe more natively

sedghi (approved):
Let's fix those other things on another PR as discussed

PR #2627, opened 2026-02-21
fix: move crosshairs implementation to absolute camera instead of delta

sedghi:
There's a persistent bug in OHIF where sudden viewport changes, like an HP stage change, can cause the crosshairs to go out of sync. The real issue is crosshair camera stuff were using delta instead of absolute computation from camera

# Before

https://github.com/user-attachments/assets/0970f786-a427-46df-93ad-6226f3103689

# After 

https://github.com/user-attachments/assets/dc568966-4dbd-4702-93b8-d5da06e6c46b


packages/tools/src/tools/CrosshairsTool.ts
```diff
@@ -2858,6 +2851,278 @@ class CrosshairsTool extends AnnotationTool {
...
+  };
+
+  _calculateToolCenterFromAbsoluteCameras = (): Types.Point3 | null => {
+    const viewportsInfo = this._getViewportsInfo();
+    const uniquePlanes: Array<{
+      normal: Types.Point3;
+      point: Types.Point3;
+    }> = [];
+
+    for (let i = 0; i < viewportsInfo.length; i++) {
```
jbocce: If this can be changed to a forEach, consider doing that.

packages/tools/src/tools/CrosshairsTool.ts
```diff
@@ -2858,6 +2851,278 @@ class CrosshairsTool extends AnnotationTool {
...
+      return null;
+    }
+
+    const hasChanged = !csUtils.isEqual(this.toolCenter, toolCenter, 1e-3);
+    if (!hasChanged) {
+      return toolCenter;
+    }
+
+    if (updateViewportCameras) {
+      this.setToolCenter(toolCenter, !emitEvent);
```
jbocce, line 3111: Do we not want to update `this.toolCenter` for this branch of the if?

jbocce:
Looks like you need to update your branch to get the security patch so CI can run.

packages/tools/src/tools/CrosshairsTool.ts
sedghi: It does inside the this.setToolCenter no?

sedghi:
@jbocce can you check again

PR #2641, opened 2026-03-03
WIP cornerstone3d viewport planar

sedghi:
Not yet ready

**Plan:**
* Replace the stacked CPU/GPU volume viewports with a single planar viewport that can switch between them.
* Add a CPU-based planar rendering mode (2D only).

### DIFF
--- a/packages/core/src/RenderingEngine/PlanarActorMapper/StackCPUActorMapper.ts
+++ b/packages/core/src/RenderingEngine/PlanarActorMapper/StackCPUActorMapper.ts
@@ -12,13 +12,13 @@ import type {
   IImage,
   ImageActor,
   Point3,
+  IStackActorMapper,
+  IStackCPUActorMapperContext,
   IStackInput,
 } from '../../types';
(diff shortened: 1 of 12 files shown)


wayfarer3130:
I definitely think we should switch to a single stack/volume viewport as there are too many bugs introduced by having them different.  Using metadata registered display sets from @cornerstonejs/metadata would make this a lot easier as it could then be read directly to determine if off-acquisition orientations were possible and/or MIP views.
Then, using a voxel manager which can map to interpolated off-acquisition orientations allows the viewport to show those without loading the full volume dataset into the GPU memory for display, as well as directly allowing display of that data - that single implementation in the voxel manager side would help enormously to enable different types of views optimized for CPU/GPU and different memory sizes.

sedghi:
Yeah i wait for your metadata stuff to arrives, then i would do that, but viewports still require work regardless.

sedghi:
close in favor of https://github.com/cornerstonejs/cornerstone3D/pull/2666

PR #2662, opened 2026-03-16
feat(crosshair): toggle slab handles while keeping slab lines

daker:
### Context

I want to manage slab thickness programmatically via a select list, so while we still want the dashed slab lines to show the current thickness, we need the handles to stay hidden when the UI makes the change for us. The existing getReferenceLineSlabThicknessControlsOn callback kept the whole thickness interaction tied together, so handles would disappear only when we disabled the entire thickness controls.

### Changes & Results

Added _slabThicknessHandlesEnabled and a slabThicknessHandles configuration flag so the dashed slab lines keep rendering while the handle rendering/hit testing is gated independently. The SLAB logic still consults getReferenceLineSlabThicknessControlsOn, but we now only block handles when the new flag is false while leaving the lines visible for reference.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  

### DIFF
--- a/packages/tools/src/tools/CrosshairsTool.ts
+++ b/packages/tools/src/tools/CrosshairsTool.ts
@@ -112,6 +112,7 @@ class CrosshairsTool extends AnnotationTool {
   _getReferenceLineControllable?: (viewportId: string) => boolean;
   _getReferenceLineDraggableRotatable?: (viewportId: string) => boolean;
   _getReferenceLineSlabThicknessControlsOn?: (viewportId: string) => boolean;
+  _slabThicknessHandlesEnabled = true;
   _volumeViewportNewVolumeListeners = new Map<
     string,
     {
@@ -151,6 +152,8 @@ class CrosshairsTool extends AnnotationTool {
         handleRadius: 3,
         // Enable HDPI rendering for handles using devicePixelRatio
         enableHDPIHandles: false,
+        // Toggle for slab thickness handles
+        slabThicknessHandles: true,
         // radius of the area around the intersection of the planes, in which
         // the reference lines will not be rendered. This is only used when
         // having 3 viewports in the toolGroup.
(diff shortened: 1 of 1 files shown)


daker:
@sedghi or @wayfarer3130 Could you review the PR when you get a chance?

packages/tools/src/tools/CrosshairsTool.ts
wayfarer3130, line 1: Can you add a control for this on one of the crosshairs tool examples?

packages/tools/src/tools/CrosshairsTool.ts
```diff
@@ -193,6 +196,9 @@ class CrosshairsTool extends AnnotationTool {
     this._getReferenceLineDraggableRotatable =
       toolProps.configuration?.getReferenceLineDraggableRotatable ||
       defaultReferenceLineDraggableRotatable;
+    this._slabThicknessHandlesEnabled =
```
wayfarer3130, line 199: I think i would rather see an empty slab thickness handles array than having another value, and specifically clear the handles list whenever they would otherwise get updated.  That ensures people don't try to render/display any of that, and the click handling doesn't try to activate the slab thickness.  
It also means one can change the configuration dynamically, which can be useful.

wayfarer3130:
I don't see a good way to test this - there should be at least one example using a pulldown list for slab thickness and a button to turn the configuration on/off.

packages/tools/src/tools/CrosshairsTool.ts
daker: @wayfarer3130 Thanks for the review! Could you clarify what content should be included in the handles list? handles coordinates are calculated inside the tool

daker: @wayfarer3130 sorry for the noise but could you clarify what content should be included in the handles list? handles coordinates are calculated inside the tool

sedghi (commented):
isn't 

this._getReferenceLineSlabThicknessControlsOn =
      toolProps.configuration?.getReferenceLineSlabThicknessControlsOn ||

doing exactly what you want? it shows it on and off no?

daker:
@sedghi no i want to keep the tickness dashed lines visible and hide the handles, the thinckness is controlled via a predefined list not via handles, using _getReferenceLineSlabThicknessControlsOn will hide the lines and handles

sedghi (approved):

daker:
@sedghi can this be merged if it's ok?

PR #2666, opened 2026-03-17
cornerstone3d redo viewports

sedghi:
ohif_ref: ohifohifupdateplaywightimgaes

## Why This Is Long Overdue

Cornerstone's viewport layer has been carrying too many responsibilities for too long. The existing `StackViewport`, `VolumeViewport`, `VideoViewport`, WSI, ECG, CPU, VTK, and segmentation paths all grew around real product needs, but the ownership boundaries became blurry: viewport classes load data, choose render implementations, preserve camera state, manage actors, project overlays, dispatch image events, and special-case segmentation behavior.

That made common workflows harder than they should be. A stack image and a volume slice can represent the same plane but still travel through different APIs. Fusion overlays and labelmaps need to know whether the base viewport is stack-like, volume-like, CPU-backed, or VTK-backed. Camera fields are used both as user-facing navigation state and renderer commands. Adding a new rendering path requires touching viewport behavior that should not care how the pixels are drawn.

This PR introduces the **Generic Viewport** architecture to cleanly separate those concerns. The new model keeps the existing rendering power, but moves it behind explicit boundaries: logical data ids, data providers, render paths, viewport data bindings, semantic view state, and per-data presentation. It is also designed for incremental adoption through compatibility adapters instead of forcing every existing example and tool to migrate at once.

## How To Try It

Existing examples can be routed through the Next viewport adapters with:

- `?type=next` to enable `rendering.useViewportNext`
- `?type=next&cpu=true` to force CPU planar rendering paths where supported

There are also dedicated Next examples for stack, volume slice, scale/position, video, ECG, WSI, multi-volume, annotation tools, and labelmap workflows.

## Viewport

This adds the new viewport family and compatibility path:

- `ViewportType.PLANAR_NEXT` for 2D stack-like and volume-slice workflows
- `ViewportType.VIDEO_NEXT`, `ECG_NEXT`, `WHOLE_SLIDE_NEXT`, and `VOLUME_3D_NEXT`
- compatibility adapters so legacy `STACK`, `ORTHOGRAPHIC`, `VIDEO`, `ECG`, `WHOLE_SLIDE`, and `VOLUME_3D` inputs can be backed by Next implementations when `useViewportNext` is enabled

The main change is `PlanarViewport`: one viewport model for stack images, volume slices, CPU image rendering, CPU volume sampling, VTK image rendering, and VTK volume-slice rendering. Applications bind logical data; the viewport and render-path decision service choose the runtime path from the data shape, orientation, rendering config, thresholds, and runtime support.

Viewport data is now mounted as source or overlay bindings. The viewport owns binding order and navigation. Each binding owns its runtime resources and receives view/presentation updates without forcing application code to reach into actors and mappers.

## Camera

Next viewports use semantic `viewState` as durable navigation state instead of treating VTK-style camera fields as the source of truth.

For planar rendering, view state tracks concepts like orientation, slice identity, anchor world/canvas points, scale, rotation, flips, and display area. A computed `ResolvedView` projects that state into whatever the active render path needs: VTK camera fields, CPU canvas transforms, canvas/world conversion, or overlay sampling information.

Legacy camera APIs remain available at the adapter boundary. Existing tools and synchronizers can still ask for an `ICamera`-compatible shape, while the clean Next path avoids persisting renderer-specific camera commands as viewport truth. 3D and WSI keep their runtime-specific camera behavior where that is the right ownership model.

## Loading

Loading now starts from logical data ids instead of direct actor or viewport-type-specific setup.

A data id is registered with metadata such as image ids, volume id, acquisition orientation, or semantic reference information. The viewport asks its data provider to resolve that id, the render-path resolver selects the matching runtime implementation, and the render path returns a binding. Calls like `setDataList`, `setData`, and `addData` then operate on mounted data ids.

This separates four things that were previously tangled together:

- loading source data
- deciding CPU vs GPU and image vs volume-slice render paths
- mounting runtime resources
- updating view state and per-data presentation

Per-data presentation now lives on the binding: VOI, opacity, colormap, blend mode, interpolation, visibility, and related settings can be updated independently for the source and each overlay.

## Segmentation

Segmentation needed to move with the viewport work because labelmaps are the hardest overlay case. The old paths had separate behavior for stack labelmap images, volume labelmap actors, overlap handling, and special volume-viewport image-mapper rendering.

This PR introduces a normalized labelmap model with explicit labelmap layers and segment bindings. A segment can be associated with a labelmap layer and label value, which gives us a clearer foundation for stack-backed labelmaps, volume-backed labelmaps, overlapping segments, and future multi-layer behavior.

Labelmap display now builds a render plan for the viewport. It can use the legacy stack/volume actor paths where needed, or mount compatible labelmaps as slice-rendered overlay data for Next/volume workflows when `useSliceRendering` is enabled. The important architectural change is that segmentation overlays are treated as viewport data bindings instead of independent actor side effects whenever the viewport can support that path.

The PR also updates brush/edit helpers, labelmap update listeners, actor styling, overlap utilities, examples, and screenshot coverage so segmentation can exercise both legacy compatibility and the Next rendering paths.

## Docs And Tests

Added documentation for the Next Viewport concepts, API, camera model, data bindings/loading, render paths, migration notes, and viewport accessor migration.

Coverage includes focused unit tests, browser/Playwright coverage, compatibility-mode screenshots, and dedicated Next examples for core viewport behavior and segmentation workflows.

## Checklist


packages/core/src/RenderingEngine/helpers/cpuFallback/colors/colormap.ts
```diff
@@ -341,3 +394,196 @@ export function getColormap(
...
+
+  const [lower, upper] = transferFunction.getRange();
+  const colors = Array.from({ length: 256 }, (_unused, index) => {
+    const t = index / 255;
+    const x = lower + t * (upper - lower);
+    const rgb: number[] = [];
+
+    transferFunction.getColor(x, rgb);
+
+    return [
```
wayfarer3130: This ends up not being evenly spaced at the start/end over [0,1] because the two endpoints 0 and 255 only receive 1/(255+255)th of the range, while the others receive 1/255
I would suggest creating a mapping function that does this to Point4 evenly spaced, using
x => Math.min(255, Math.floor(x*256))
The Math.min deals with the fact that the upper bound is 255 and not 256
The x*256 deals with the fact that there are 256 distinct equal space values.

packages/core/src/RenderingEngine/helpers/cpuFallback/colors/index.ts
```diff
@@ -1,4 +1,13 @@
-import { getColormap, getColormapsList } from './colormap';
+import {
```
wayfarer3130: Can you use the export * from './colormap' format since I think we will end up having more bits in colormap.

packages/core/src/RenderingEngine/helpers/supportsVolumeCompatibilityApi.ts
```diff
@@ -0,0 +1,17 @@
+import type { IVolumeViewport } from '../../types';
+
+type VolumeCompatibilityMethod = 'setVolumes' | 'addVolumes';
+
+function supportsVolumeCompatibilityApi<
```
wayfarer3130: Could you use the typescript 5.5 is method for this, something like:
// The type guard function
function isCat(pet: Pet): pet is Cat {
  return pet.species === "cat";
}

So
function isVolumeCompatible(viewport: unknown): viewport is ICompatibleVolumeVIewport {
}

rather than passing in the method specifically?  That way you can do:

if( !isVolumeCompatible(viewport)) {
   return ...
}
// viewport is now known to be an instance of ICompatibleVolumeViewport

packages/core/src/RenderingEngine/GenericViewport/ECG/ECGViewport.ts
wayfarer3130, line 1: For ECG, Video, WSI could we just move to the new viewports wholesale and delete the old ones?  I don't think those ones are worth having old/new given the relatively low usage of them.

packages/core/src/RenderingEngine/ViewportV2/Planar/PlanarViewportV2.ts
wayfarer3130, line 1: Can we call it PlanarViewport not PlanarViewportV2 since V1 doesn't have a Planar Viewport?

packages/core/src/RenderingEngine/GenericViewport/GenericViewport.ts
wayfarer3130, line 1: There are quite a few compatibility methods in this - is it possible to split the compatibility layer into something on top of PlanarViewport and 3dViewport?  That way it is obvious what is done for compatibility and what is done as a new base functionality.  It would then be really nice to have a new IBaseViewport defined in @cornerstonejs/metadata as an interface that has just the base functionality that is expected.
It is a bit of a pain creating that, although it can be done dynamically by defining a function that returns a class definition.

packages/core/src/RenderingEngine/ViewportV2/ViewportV2.ts
```diff
@@ -0,0 +1,528 @@
...
+        ...(this.camera.frame || {}),
+        anchorCanvas: [anchorCanvas[0], anchorCanvas[1]],
+      },
+    } as Partial<TCamera>);
+  }
+
+  /**
+   * Uses the current binding's render-path transform.
+   */
+  canvasToWorld(canvasPos: Point2): Point3 {
```
wayfarer3130: Can the canvas to/from world be moved into a function on the camera?  That would allow storing the camera object for later use, which is extremely valuable.  If it had index2world and canvas2world and unitCanvas2World (ie an unscaled canvas), then you get a lot of value from having a rich camera object.   It can still be very cheap to construct - in fact, you can construct it cheaper than the existing camera objects because you only need to fill in the base source data for the camera position/information, and the remaining data is generated on request.

packages/core/src/RenderingEngine/GenericViewport/Planar/planarRenderCamera.ts
wayfarer3130, line 1: If you construct a camera object instead of having standalone functions, then the derivations here can be a function of the camera.  There should probably be two cameras here - a slice index one for non-volumes, and a volume camera for mpr, which can then be extended to a 3d camera.
It also means there is an obvious spot for time frame/3+d information - ideally that could go in a shared base camera implementation.  
Then, if there is a camera implementation, there is an obvious way to test it for various conditions, and I think have jest tests for the cameras would help enormously in isolating that bit of code so that we are reasonably sure it is stable.  

It removes a bunch of methods from the base viewport since scale/zoom/pan are attributes/methods of the camera and not the viewport.

packages/core/src/RenderingEngine/GenericViewport/ECG/ECGViewport.ts
sedghi: I know they are not used much, but better to keep them around maybe for some time

sedghi:
@claude review

sedghi:
@claude review

packages/core/examples/nextStackAPI/index.ts
```diff
@@ -0,0 +1,281 @@
...
+
+const flipHorizontalInfo = document.createElement('div');
+info.appendChild(flipHorizontalInfo);
+
+const flipVerticalInfo = document.createElement('div');
+info.appendChild(flipVerticalInfo);
+
+element.addEventListener(Events.CAMERA_MODIFIED, () => {
+  const renderingEngine = getRenderingEngine(renderingEngineId);
+  const viewport = renderingEngine.getViewport(viewportId) as PlanarViewport;
```
wayfarer3130, line 61: Can you add PlanarViewport to the return type, or else have a getNextViewport (need a better name)?

packages/core/src/RenderingEngine/CanvasActor/CanvasProperties.ts
```diff
@@ -39,11 +68,48 @@ export default class CanvasProperties {
     // No-op - requires outline to be implemented first
   }
 
-  public getColor(index: number) {
+  public modified() {
+    this.invalidateColorCache();
+  }
+
+  public getColorBytes(index: number): [number, number, number, number] {
```
wayfarer3130, line 75: Can you add an optional destination/offset to make it a single call to add that colour to a given array?

packages/core/src/RenderingEngine/CanvasActor/index.ts
```diff
@@ -120,70 +118,65 @@ export default class CanvasActor {
...
-          dirtyX2 = Math.max(x, dirtyX2);
-          dirtyY2 = Math.max(y, dirtyY2);
-          const rgb = this.canvasProperties.getColor(segmentIndex);
-          imageArray[destOffset] = rgb[0] * 255;
-          imageArray[destOffset + 1] = rgb[1] * 255;
-          imageArray[destOffset + 2] = rgb[2] * 255;
-          imageArray[destOffset + 3] = 127;
-          // imageArray.fill(55, offset, offset + 4);
+          hasVisiblePixels = true;
+          const rgba = this.canvasProperties.getColorBytes(segmentIndex);
```
wayfarer3130, line 159: Can you use the suggest destination here, so that this is just:
this.canvasProperties.getColorBytes(segmentIndex, imageArray, destOffset)
It makes the code a little easier to read.

packages/core/src/RenderingEngine/ViewportNext/ECG/CanvasECGRenderPath.ts
```diff
@@ -0,0 +1,351 @@
...
+    rendering.currentDataPresentation = props as
+      | ECGDataPresentation
+      | undefined;
+  }
+
+  private updateCamera(rendering: ECGCanvasRendering, camera: unknown): void {
+    rendering.currentCamera = camera as ECGCamera;
+  }
+
+  private canvasToWorld(
```
wayfarer3130, line 86: Unused

packages/core/src/RenderingEngine/ViewportNext/ECG/CanvasECGRenderPath.ts
```diff
@@ -0,0 +1,351 @@
...
+          waveform.numberOfSamples - 1,
+          (subCanvasPos[0] * waveform.numberOfSamples) / metrics.ecgWidth
+        )
+      ),
+      (layout.baseline - subCanvasPos[1]) / metrics.channelScale,
+      z,
+    ];
+  }
+
+  private worldToCanvas(
```
wayfarer3130, line 132: Also unused

packages/core/src/RenderingEngine/ViewportNext/ECG/DefaultECGDataProvider.ts
```diff
@@ -0,0 +1,17 @@
+import type { DataProvider, LoadedData } from '../ViewportArchitectureTypes';
+import type { ECGWaveformPayload } from './ECGViewportNextTypes';
+import { loadECGWaveform } from '../../../utilities/ECGUtilities';
+import { getViewportNextSourceDataId } from '../viewportNextDataSetAccess';
+
+export class DefaultECGDataProvider implements DataProvider {
```
wayfarer3130, line 6: This will probably not be required once the new metadata stuff is available.

packages/core/src/types/IActor.ts
```diff
@@ -10,6 +14,21 @@ export type ImageActor = vtkImageSlice;
 
 export type ICanvasActor = CanvasActor;
 
+export interface ActorMapperProxy {
+  actor: VolumeActor | ImageActor | ICanvasActor;
+  mapper:
```
wayfarer3130, line 19: If the types are sufficiently distinct to not overlap generally, is it worth having a shared mapper instance?
If they overlap enough to have a lot of shared methods, then maybe create a combined interface/setup from them?

packages/core/src/types/IActor.ts
```diff
@@ -10,6 +14,21 @@ export type ImageActor = vtkImageSlice;
...
+    | vtkImageMapper
+    | vtkImageResliceMapper
+    | vtkVolumeMapper
+    | CanvasMapper;
+  renderMode:
```
wayfarer3130, line 24: Can you extract a string type from this - it seems like it would be useful to have it as it's own export type.
eg RenderModeType = vtkImage | vtkVolume etc

packages/core/src/types/IActor.ts
wayfarer3130: Another way to handle it is to declare 4 or 5 different types, where the renderMode is still a specific string, but once you have checked the value of the string, the type of mapper would be well defined.  Maybe that doesn't work well with interfaces?

packages/core/src/RenderingEngine/ViewportNext/Planar/PlanarViewport.ts
```diff
@@ -0,0 +1,2186 @@
...
+  PlanarViewportInput,
+} from './PlanarViewportTypes';
+import { resolvePlanarViewportRenderMode } from './resolvePlanarViewportRenderMode';
+
+defaultRenderPathResolver.register(new CpuImageSlicePath());
+defaultRenderPathResolver.register(new CpuVolumeSlicePath());
+defaultRenderPathResolver.register(new VtkImageMapperPath());
+defaultRenderPathResolver.register(new VtkVolumeSlicePath());
+
+type PlanarReferenceContext = {
```
wayfarer3130, line 91: Can you make sure you export the types?

packages/core/src/RenderingEngine/ViewportNext/ViewportNext.ts
```diff
@@ -0,0 +1,653 @@
...
+ * Concrete render paths are expected to own:
+ * - runtime add/remove lifecycle
+ * - camera interpretation for that render path
+ * - per-dataset render-state application
+ * - render-path-specific coordinate transforms
+ *
+ * This split keeps migration from legacy viewports incremental without
+ * centralizing render-mode-specific behavior in the controller.
+ */
+abstract class ViewportNext<
```
wayfarer3130, line 52: Could we call this GenericViewport rather than ViewportNext?

packages/core/src/RenderingEngine/ViewportNext/ViewportNext.ts
```diff
@@ -0,0 +1,653 @@
...
+
+  // ====================================================================
+  // Public API -- data
+  // ====================================================================
+
+  /**
+   * Loads a logical dataset through the viewport data provider and adds it
+   * through the render-path resolver.
+   */
+  async setData(dataId: DataId, options: DataAddOptions): Promise<RenderingId> {
```
wayfarer3130, line 91: This should be addData since it extends
If you want a standard setData, it should remove all the data and then call addData internally.

packages/core/src/RenderingEngine/ViewportNext/ViewportNext.ts
```diff
@@ -0,0 +1,653 @@
...
+   * Applies a spatial reference to the current viewport state.
+   */
+  setViewReference(_viewReference: ViewReference): void {
+    // Subclasses can implement.
+  }
+
+  /**
+   * Returns whether a spatial reference is compatible with this viewport.
+   */
+  isReferenceViewable(
```
wayfarer3130, line 206: It would be helpful to have the base implementation of this deal with the standard behaviour:
1. Frame of reference (you already have this)
2. data id (if it is none of the data id's and it isn't only by FOR, then it can be rejected)
3. Position indication as either image ref or planar restriction
4. Time point restriction

This should be the standard behaviour for view references for every type of viewport AFAIK.  

ECG => uses just data id as specifier, so follows above rules

Video => Uses data id + time point indicator

Planar on non-volumetric => Uses data id + image ref
Planar on volumetric => Uses FOR + data id (optional) + plane restriction + time point restriction

3d => Uses FOR + data id (optional) + plane restriction

So, it shouldn't be needed to make the isReferenceViewable customized.

packages/core/src/RenderingEngine/ViewportNext/viewportNextDataSetAccess.ts
```diff
@@ -0,0 +1,176 @@
...
+export interface ViewportNextWSIDataSet {
+  kind: 'wsi';
+  imageIds: string[];
+  options: {
+    miniNavigationOverlay?: boolean;
+    webClient: WSIClientLike;
+  };
+}
+
+export type ViewportNextRegisteredData =
```
wayfarer3130, line 34: Didn't we decide this was going to be DisplaySet?
Even if you register it as a straight array of images, it would be very nice to have it be the same type of object as DisplaySet

packages/core/src/RenderingEngine/ViewportNext/ViewportCameraTypes.ts
```diff
@@ -0,0 +1,14 @@
+export type ViewAnchor = [number, number];
+
+export type CameraScaleMode = 'fit';
+
+export interface CameraFrame<TAnchorWorld = unknown> {
+  anchorWorld?: TAnchorWorld;
+  anchorCanvas?: ViewAnchor;
+  scale?: number;
```
wayfarer3130, line 8: Does this need two values?  One for x fit and one for y fit?  Otherwise you can't specify fit to width versus fit to height automatically to match overall.
It also means that a camera scale mode of 'fitAspect' wouldn't work.

packages/core/src/RenderingEngine/ViewportNext/ViewportArchitectureTypes.ts
```diff
@@ -0,0 +1,199 @@
...
+export type ViewportId = string;
+export type DataId = string;
+export type RenderingId = string;
+export type ViewportRenderContextType =
+  | 'planar'
+  | 'video'
+  | 'ecg'
+  | 'wsi'
+  | '3d'
+  | string;
```
wayfarer3130, line 20: If you use  | (string & {});   then typescript doesn't collapse ViewportRenderContextType into just string.  That makes it auto-complete etc, but still allow custom values.
You can make it even safer with:
type KnownViewportRenderContext =
  | 'planar'
  | 'video'
  | 'ecg'
  | 'wsi'
  | '3d';

type ExtensionViewportRenderContext = string & {
  readonly __brand: 'ExtensionViewportRenderContext';
};

type ViewportRenderContextType =
  | KnownViewportRenderContext
  | ExtensionViewportRenderContext;

Then extension authors must opt in deliberately:

const ext = 'my-custom-mode' as ExtensionViewportRenderContext;

packages/core/src/RenderingEngine/ViewportNext/WSI/WSIViewportNext.ts
```diff
@@ -0,0 +1,636 @@
...
+  canvasToIndexForWSI,
+  computeWSITransforms,
+  indexToWorldWSIMetadata,
+  worldToIndexWSIMetadata,
+} from './wsiTransformUtils';
+import WSIComputedCamera from './WSIComputedCamera';
+
+defaultRenderPathResolver.register(new DicomMicroscopyPath());
+
+class WSIViewportNext extends ViewportNext<
```
wayfarer3130, line 45: Can you add the export line here 
Also, this is already in the ViewportNext path, and can be imported as a Next value - can we just call it WSIViewport
I'd also like to suggest WsiViewport instead of WSIViewport under the convention that longer than 2 character acronyms get lower camel cased.

packages/core/src/RenderingEngine/ViewportNext/WSI/WSIViewportNext.ts
```diff
@@ -0,0 +1,636 @@
...
+  }
+
+  getUseCustomRenderingPipeline(): boolean {
+    return true;
+  }
+
+  constructor(args: WSIViewportNextInput) {
+    super();
+    this.id = args.id;
+    this.element = args.element;
```
wayfarer3130, line 71: Can the id and element assignment get handled in super?  I think every viewport assigns id and element, and if not, they probably should.

packages/core/src/RenderingEngine/ViewportNext/WSI/DicomMicroscopyRenderPath.ts
```diff
@@ -0,0 +1,212 @@
...
+    }
+    if (wsiCamera.centerIndex) {
+      view.setCenter(wsiCamera.centerIndex);
+    }
+    if (typeof wsiCamera.rotation === 'number') {
+      view.setRotation(wsiCamera.rotation);
+    }
+  }
+
+  private canvasToWorld(
```
wayfarer3130, line 152: Not used, nor worldToCanvas

packages/core/src/RenderingEngine/ViewportNext/ECG/ECGViewportNext.ts
```diff
@@ -0,0 +1,502 @@
...
+  createDefaultECGCamera,
+  normalizeECGCamera,
+} from './ecgViewportCamera';
+import ECGComputedCamera from './ECGComputedCamera';
+
+const ECG_AMPLITUDE_INDEX_SIZE = 65536;
+
+defaultRenderPathResolver.register(new CanvasECGPath());
+
+class ECGViewportNext extends ViewportNext<
```
wayfarer3130, line 37: Since the include path has Next in it, can we use the base names - if it needs to be included separately for a while, it is possible to include as '...Next'
Also, The new name should be EcgViewport

packages/core/examples/nextStackAPI/index.ts
sedghi: i'm doing this now

`renderingEngine.getViewport<PlanarViewport>(viewportId)`

packages/core/src/RenderingEngine/ViewportNext/viewportNextDataSetAccess.ts
sedghi: it is now in todo.md will do at the end

packages/core/src/RenderingEngine/ViewportNext/Planar/planarSliceBasis.ts
```diff
@@ -0,0 +1,752 @@
...
+ * of its index-space dimensions. Falls back to averaging the bounding-box
+ * corners if vtkImageData is unavailable.
+ */
+function getGeometricImageVolumeCenter(imageVolume: IImageVolume): Point3 {
+  const imageData = imageVolume.imageData;
+
+  if (imageData) {
+    const [dx, dy, dz] = imageData.getDimensions();
+
+    return imageData.indexToWorld([
```
wayfarer3130: This center is wrong - consider dx,dy,dz being 1,2,3 then the center calculations will be:
(1-1)/2 = 0
(2-1)/2 = 0.5
(3-1)/2 = 1

But the center of a 1 pixel wide value is 0.5, the center of a 2 pixel wide value is index 1, and the center of a 3 pixel wide region is 1.5
Note that the pixel regions for dx,dy,dz are  [0,dx)  [0,dy) and [0,dz)  so that the d value is excluded, but everything below it is within the image area, since the width of the image areas are just <1

This may cause image differences in rendering compared to baseline, but I think that should be fixed rather than leaving the bug in place.

packages/core/src/RenderingEngine/ViewportNext/Planar/planarSliceBasis.ts
```diff
@@ -0,0 +1,752 @@
...
+function getViewportCompatibleImageVolumeCenter(
+  imageVolume: IImageVolume
+): Point3 {
+  const imageData = imageVolume.imageData;
+
+  if (imageData) {
+    const [dx, dy, dz] = imageData.getDimensions();
+
+    return imageData.indexToWorld([
+      Math.floor(dx / 2),
```
wayfarer3130: This is closer, but unless indexToWorld requires integer values, they should be dx/2 etc.  It only matters for odd sized images, but it ends up ofsetting them half a pixel, and that causes issues with 1:1 pixel rendering.
I'd really like to see these two functions get unified in this PR and address the image comparison issues separately.  Using this one consistently without the floor will affect VERY few tests.

packages/tools/src/stateManagement/segmentation/labelmapModel/normalizeLabelmapSegmentationData.ts
```diff
@@ -0,0 +1,129 @@
...
+    type: getPrimaryLabelmapType(labelmapData),
+    volumeId: labelmapData.volumeId,
+    referencedVolumeId: labelmapData.referencedVolumeId,
+    referencedImageIds: labelmapData.referencedImageIds,
+    imageIds: labelmapData.imageIds,
+    labelToSegmentIndex: {},
+  };
+}
+
+function ensureLabelmapState(
```
wayfarer3130, line 95: This function definitely needs documentation at the top.  I'm not convinced this one shouldn't be called after updates to the segmentation, or potentially modify the segmentation object to ensure it's internal state is kept up to date when it is changed?  Certainly doing it in the render path doesn't seem like a good idea.

packages/tools/src/stateManagement/segmentation/labelmapModel/normalizeLabelmapSegmentationData.ts
```diff
@@ -0,0 +1,129 @@
...
+    labelmapData.labelmaps = {
+      [labelmapId]: createPrimaryLabelmapLayer(
+        segmentation,
+        labelmapData,
+        labelmapId
+      ),
+    };
+  }
+
+  if (!labelmapData.segmentBindings) {
```
wayfarer3130: Can you use ||= for updating default values if you aren't otherwise changing it?

packages/tools/src/stateManagement/segmentation/labelmapModel/normalizeLabelmapSegmentationData.ts
```diff
@@ -0,0 +1,129 @@
...
+
+  if (!labelmapData) {
+    return;
+  }
+
+  if (
+    !labelmapData.labelmaps ||
+    Object.keys(labelmapData.labelmaps).length === 0
+  ) {
+    const labelmapId = getPrimaryLabelmapId(segmentation.segmentationId);
```
wayfarer3130: If the primary labelmapId is always defined, you could just compute the string here and then assign it.
If not, you could store it, and then check for that value, eg something like:
if (!labelmapData.labelMaps[labelmapData.primaryLabelmapId]) {
....
}

so that you avoid doing the Object.keys within a render path.
Then, if the outer test fails, you can get the Object.keys zero element, and if that is absent, then generate the primary labelmap for it.  That way you store hte primary labelmap information so this check becomes very quick.

packages/tools/src/stateManagement/segmentation/labelmapModel/normalizeLabelmapSegmentationData.ts
```diff
@@ -0,0 +1,129 @@
...
+        labelmapId
+      ),
+    };
+  }
+
+  if (!labelmapData.segmentBindings) {
+    labelmapData.segmentBindings = {};
+  }
+
+  if (!labelmapData.sourceRepresentationName) {
```
wayfarer3130: ||= SOURCE_REPRESENTAITON_NAME instead of if

packages/tools/src/stateManagement/segmentation/labelmapModel/normalizeLabelmapSegmentationData.ts
```diff
@@ -0,0 +1,129 @@
...
+    labelmapData.sourceRepresentationName = SOURCE_REPRESENTATION_NAME;
+  }
+
+  const labelmapIds = Object.keys(labelmapData.labelmaps);
```
wayfarer3130: This bit of code is almost exactly what you need, except you store fallbackLabelmapId on the object itself so that rendering multiple times/setups is fast.

packages/tools/src/stateManagement/segmentation/labelmapModel/normalizeLabelmapSegmentationData.ts
```diff
@@ -0,0 +1,129 @@
...
+    labelmapIds[0] ?? getPrimaryLabelmapId(segmentation.segmentationId);
+  labelmapData.labelmaps[fallbackLabelmapId] ??= createPrimaryLabelmapLayer(
+    segmentation,
+    labelmapData,
+    fallbackLabelmapId
+  );
+
+  getSegmentOrder(segmentation).forEach((segmentIndex) => {
+    if (!labelmapData.segmentBindings[segmentIndex]) {
+      labelmapData.segmentBindings[segmentIndex] = {
```
wayfarer3130: Again, this one can be ||=

packages/tools/src/stateManagement/segmentation/labelmapModel/normalizeLabelmapSegmentationData.ts
```diff
@@ -0,0 +1,129 @@
...
+        labelmapId: fallbackLabelmapId,
+        labelValue: segmentIndex,
+      };
+    }
+  });
+
+  Object.values(labelmapData.labelmaps).forEach((layer) => {
+    layer.labelToSegmentIndex = {};
```
wayfarer3130, line 128: This bit of code updates the labelToSegmentIndex repeatedly.  It is certainly possible to write a proxy that keeps this up to date, and then replace segmentBindings with that proxy so that it is kept up to date automatically.  Then, it would be heavyweight to compute labelToSegmentIndex the first time, as you would have to replace segmentBindings, but subsequently would be light weight.
Keeping the backwards compatible API is quite good to have - I just wonder if there is an overall solution that is more forwards looking and would allow updates to the old API/data if needed, but would record changes using set state methods?

packages/core/src/RenderingEngine/ViewportNext/Planar/planarSliceBasis.ts
```diff
@@ -0,0 +1,752 @@
...
+  const geometry = getOrthogonalVolumeSliceGeometry({
+    dimensions: imageVolume.dimensions,
+    direction: imageVolume.direction,
+    spacing: imageVolume.spacing,
+    viewPlaneNormal,
+    viewUp,
+  });
+
+  if (geometry) {
+    const columns = Math.max(geometry.columns - 1, 1);
```
wayfarer3130: The Math.max here was a fix to prevent 1x1 images failing to render in a way that didn't update the images, but the real correct change is to fix this to match the CPU version, and then remove the GPU version.  

If you need for a time, you could have "incorrect" versions of them, with comments about when one gets used (eg for backwards compatibility), but otherwise default to the correct one.  Then, the new viewport rendering would never use the incorrect comparisons, while only the "compatibility" tests would use the incorrect version.

packages/core/src/RenderingEngine/GenericViewport/Planar/planarRenderCamera.ts
```diff
@@ -0,0 +1,701 @@
...
+}
+
+/**
+ * Sets the clipping range on a vtkRenderer's camera to cover the full
+ * volume depth. For parallel projection the range is symmetric around zero;
+ * for perspective it starts at the minimum slab thickness.
+ *
+ * @param renderer - The VTK renderer whose camera clipping range is updated.
+ */
+export function setPlanarVolumeCameraClippingRange(
```
wayfarer3130, line 634: This is identical to the version in vtkvolume3drenderpath, except for the source rendered.  Can you extract the function and call  it on the appropriate camera?

packages/core/src/RenderingEngine/ViewportNext/Planar/PlanarViewport.ts
```diff
@@ -0,0 +1,1925 @@
...
+
+    const baseOrientation = getBaseReformatOrientation(orientation);
+    const cameraVectors = getCameraVectors(
+      {
+        getActors: () => this.getActors(),
+        getCamera: () => this.getCameraForEvent(),
+        getCurrentImageId: () => this.getCurrentImageId(),
+        getImageIds: () => this.getImageIds(),
+        type: ViewportType.ORTHOGRAPHIC,
+      } as unknown as IBaseVolumeViewport,
```
wayfarer3130: Could you extract a minimal type for this argument in getCameraVectors so you don't have to cast to unknown?  Or maybe add a child method of it so that you can simply pass in this as the 0 argument, and use that for this function, and the original one with the type===orthographic check in it generally?

packages/tools/src/stateManagement/segmentation/labelmapModel/labelmapImageReferenceResolver.ts
```diff
@@ -0,0 +1,402 @@
...
+      return;
+    }
+
+    const stackViewport = enabledElement.viewport as Types.IStackViewport;
+
+    this.updateLabelmapSegmentationReferences(
+      segmentationId,
+      stackViewport,
+      labelmapImageIds,
+      (stackViewport, segmentationId, labelmapImageIds) => {
```
wayfarer3130, line 389: This method is an O(m * n) in stack image ids and labelmap image ids.  That is ok if it is only used at load time, but I couldn't actually find uses of this specific set of calls - claude thinks this is dead code.  If so, deprecate it and just mark it as being  backwards compatible for the old API, or delete it as you think appropriate.

If not, the method can be re-written to be O(m+n) because it is invariant on index since index isn't used in the path as far as I can tell for as overlay.  I wasn't sure about that because I'm not totally sure why the method is there or what it ends up getting used for.

packages/core/src/RenderingEngine/ViewportNext/Volume3D/VtkVolume3DRenderPath.ts
```diff
@@ -0,0 +1,344 @@
...
+    );
+  } else {
+    activeCamera.setClippingRange(
+      RENDERING_DEFAULTS.MINIMUM_SLAB_THICKNESS,
+      RENDERING_DEFAULTS.MAXIMUM_RAY_DISTANCE
+    );
+  }
+}
+
+function buildVolumeImageData(imageVolume): IImageData | undefined {
```
wayfarer3130: Could imageVolume be typed?  It seems like it should be typeable.

packages/tools/src/stateManagement/segmentation/labelmapModel/labelmapImageReferenceResolver.ts
```diff
@@ -0,0 +1,402 @@
...
+            .filter(Boolean)
+        )
+      );
+    }
+
+    if ((labelmapData as LabelmapSegmentationDataStack).imageIds) {
+      labelmapImageIds = (labelmapData as LabelmapSegmentationDataStack)
+        .imageIds;
+    } else if (
+      !labelmapImageIds &&
```
wayfarer3130: This !labelmapImageIds is an empty condition since the only earlier assignment after declaration is on line 54 above.  Did you mean to add this check after line 55 in some way, or should it just get removed?

packages/core/src/RenderingEngine/GenericViewport/Planar/PlanarViewport.ts
wayfarer3130, line 1: This is a fairly long class, and I think at least some of the logic could be shared with other types of viewports that are related but not identical by use of collaborator methods/functions that are allowed deep/protected access  but are external to the class.  That way related classes can be handled using the collaborators without needing to have parent classes.  This could be as simple as declaring some of the functions externally and assigning them to the standard function names.

It also allows for some things like zoom/pan calculations to be handled externally, and/or switched out by just choosing which external item to use.  That would allow "legacy" compatibility for things like the center calculations where it could be specifically labelled as being incorrect/invalid and deprecated.

packages/core/src/RenderingEngine/ViewportNext/Planar/planarSliceBasis.ts
sedghi: I agree for the rendered physical extent, but I think the center should stay `(d - 1) / 2`. In vtk (dicom too i think), index `0` is the center of the first voxel. The rendered domain is half-voxel padded, `[-0.5, d - 0.5]`, so its width is `d` but its center is `(d  - 1) / 2`.  So the fix I think we want is: use `d` for fit/bounds, but keep `(d - 1) / 2` for `indexToWorld` center.

It matches our discussion with vtk couple of years ago 

https://github.com/Kitware/vtk-js/pull/1993 
https://github.com/Kitware/vtk-js/pull/2089

wayfarer3130: Learn something new every day - you are absolutely correct here.
PLEASE reference 
https://dicom.nema.org/medical/dicom/current/output/html/part03.html#sect_C.7.6.2.1.1
PS3.3 C.7.6.2.1.1 Image Position and Image Orientation
Image Position (Patient) (0020,0032) specifies the x, y, and z coordinates of the upper left hand corner of the image; it is the center of the first voxel transmitted. 

and add the comment that index [i,j,k] is the CENTER of the voxel so that (d-1)/2 is the correct overall center.

Then, we should use this for both CPU and VTK rendering.  It might require a corresponding fix for CPU rendering world to canvas.

packages/core/src/RenderingEngine/ViewportNext/Planar/planarSliceBasis.ts
wayfarer3130: Lets replace this one with the correct one up above - the indexToWorld function should be identical, and the centers should be identical, but that might require a fix to world to canvas.

wayfarer3130:
I had claude look through the review - will make notes prefixed with BW> in some of these:
# Review: PR #2666 — `cornerstone3d redo viewports`

**Author:** sedghi · **Base:** `beta` · **State:** open · **Scope:** 966 files, +62,638 / −4,593 · **Companion:** OHIF `feat/use-beta-5.0-cs3d`

> Disclaimer: at this size no single review catches everything. I focused on the new viewport/camera architecture, OHIF impact, and the items explicitly flagged in the request (long-term API stability + performance). I read the architecture, camera/view-state, adapters, and CPU/VTK render-path code closely; the rest of the diff (codemods, examples, screenshots, segmentation strategies, tests) I only skimmed.

---

## 1. What this PR does

Introduces a "Generic Viewport" / `ViewportNext` family that re-layers the rendering stack:

```
DataProvider  →  RenderPathResolver  →  RenderPath  →  MountedData / Binding
                                                ↘  ResolvedView (camera projection)
```

- New viewport classes: `PlanarViewport` (replaces stack + volume-slice), `ECGViewport`, `VideoViewport`, `VolumeViewport3DV2`, `WSIViewport` — all under `packages/core/src/RenderingEngine/ViewportNext/`.
- New `ViewportType` enum members: `PLANAR_NEXT`, `VIDEO_NEXT`, `ECG_NEXT`, `WHOLE_SLIDE_NEXT`, `VOLUME_3D_NEXT`.
- Camera split into three layers: semantic `PlanarViewState` → geometric `PlanarSliceBasis` → renderer `ICamera`, plumbed through `ResolvedViewportView`.
- Legacy adapters (e.g. `PlanarViewportLegacyAdapter`, `PlanarLegacyCompatibilityController`) so the legacy `STACK` / `ORTHOGRAPHIC` / `VIDEO` / `ECG` / `WHOLE_SLIDE` / `VOLUME_3D` types can be backed by Next when `rendering.useViewportNext` is on.
- Segmentation reworked around a labelmap render plan and "binding" model; brush/edit helpers, overlap utilities, and label rendering all touched.
- Adds `packages/docs/docs/concepts/cornerstone-core/next-viewport/` and a migration guide.

The high-level layering is **good and overdue**. The criticism below is about edges, not the direction.

---

## 2. Long-term stability of the new viewport / camera API

### 2.1 The author has explicitly listed unfinished renames

`tests/todo.md` (new file) reads, verbatim:

```
[] - rename viewport next -> generic viewport
[] - rename the data id to displaySets
[] - rename 3dViewport to viewport3D
[] why we have render mode in options
```

BW> These need updating before commit.

### 2.2 The new public surface is enormous and exports implementation detail

`packages/core/src/RenderingEngine/ViewportNext/index.ts` re-exports ~70 symbols. Many are render-path internals: `PlanarViewportRenderContext`, `PlanarPayload`, `PlanarEffectiveRenderMode`, `PlanarRegisteredDataSet`, `PlanarDataLoadOptions`, `CpuImageSlicePath`, `VtkImageMapperPath`, etc. These appear in user-facing function signatures (e.g. the `DataProvider` interface forces consumers to traffic in `PlanarPayload`).

Once these are exported they are effectively frozen. Recommendations:
- Mark with `@internal` JSDoc and exclude from the type rollup, OR
- Hide them behind a narrower public surface (e.g. only the viewport classes, the `RenderPath` / `DataProvider` interfaces, and a small set of `*Input` / `*Presentation` types).

**BW> I definitely have some concerns about the size of the API surface.  We should try reducing the size of the API surface at least some.  I also wonder if the camera representation should still be made a bit more abstract/re-useable without needing to tweak the camera round trip stuff.  Sometimes for things like figuring out the zoom, to actually use, there are several ways to get it, and I don't quite always know how to declare that in an extensible fashion.  It would be very nice to not require the current scale values, but be able to have a pluggable mechanism that works across several types of image areas to create new types of scaling.  The same with positioning.  The current design feels close, but not QUITE right yet.**  

### 2.3 `ICamera` no longer round-trips on `PlanarViewport`

The most consequential stability concern. The legacy contract downstream tools rely on is:

```ts
const c = vp.getCamera(); vp.setCamera(c);     // identity
```

For `PlanarViewport` / `PlanarViewportLegacyAdapter` this is no longer true:

- **`getViewState()` returns aliased internal arrays.** `packages/core/src/RenderingEngine/ViewportNext/Planar/PlanarViewport.ts` L784-L789 returns `{ ...this.viewState, displayArea: clone(displayArea) }`. Everything else — `anchorCanvas`, `anchorWorld`, `scale`, `slice.sliceWorldPoint` — is shared by reference. A consumer that mutates `state.scale[0] = 2` mutates the viewport.
  **BW>** High area of concern
- **Bogus fallback camera.** `PlanarViewport.ts` L1728-L1745 returns hard-coded values (`position: [0,0,1]`, `viewUp: [0,-1,0]`) when no data is mounted. These ride out as `previousCamera` in `CAMERA_MODIFIED` events. The first event after data loads will report a bogus delta.
- **`setViewPresentation` does two writes.** `PlanarViewport.ts` L884-L930 calls `setViewState` and then `setPan`, which is itself a second `setViewState`. Two events / two render requests per logical write. Worse, pan is **scaled by `nextScale`** while `getViewPresentation` divides by `currentScale` (L862-L865). If the consumer round-trips through get/set with `zoom` changing in between, the pan lands at a different point.
  **BW>** This is kind of intrinsic in the fact that pan is scaled by the scale values so I'm not quite sure how to deal with that.  IN many ways, pan is an artificial construct - really useful for users to understand, but not obvious what it always means.  
- **`projectAnchorWorldToCurrentPlane` is not invertible by design.** `planarRenderCamera.ts` L179-L198 — a zoom-to-cursor anchor placed on slice N gets projected onto slice M when the user scrolls. Going back to slice N does *not* restore the original anchor. Documented in the file comment but deserves to be in the migration guide too — cine/sync that stores and replays camera state across slices will see drift.
- **`resetCamera({ resetPan, resetZoom })`** in `PlanarViewport.ts` L1176-L1193 silently does not reset orientation, flips, or slice. Legacy `StackViewport.resetCamera` resets to the default acquisition orientation. Behavior divergence to call out.
- **`setPan` fallback** in `PlanarViewport.ts` L725-L749, when there is no resolved view, uses `currentPan = getPan()` which itself returns `[0,0]` in the fallback path — so the "delta" is `nextPan - 0`, wrong if the viewport already has a non-default anchor.

These are not theoretical. OHIF stores `getViewPresentation()` in a Zustand store and replays it on layout change — see §3.

### 2.4 Camera projection helpers exist but are not exported

`resolvePlanarICamera`, `derivePlanarPresentation`, `applyPlanarICameraToRenderer` are in `planarRenderCamera.ts` but not in the top-level barrel. A downstream consumer that wants to compute a camera from a view-state (e.g. a custom synchronizer) has no stable entry point. Either commit to exporting them or make them genuinely internal — right now they're advertised in the type system (`PlanarResolvedICamera` is exported) without being constructible.

**BW>** These should be exported as helper modules, probably indicating that they are a little less stable than the core behaviour.

**BW>** Would it be possible to make an external handler/plugin to allow differentiating between camera types in a generic fashion?  Something like being able to get the camera of a given type given the current parameters, or the underlying/base camera type.  That would allow pluging in the transforms earlier by keyword/type and being able to transform bidirectionally on different viewport types things like the stack viewport camera types versus the 3d camera types.  That might be a way to get different types of pan values.  Not quite sure this works, but if it does, I think it would resolve some of the export size differences since there would be mroe smaller exports, but they would allow for better re-use.

### 2.5 `ResolvedViewportView` snapshots aren't cached, but tools call them in tight loops

`PlanarViewport.getResolvedView()` (`PlanarViewport.ts` L795-L810) calls `resolvePlanarViewportView` and constructs a fresh `PlanarStackResolvedView` or `PlanarVolumeResolvedView` *on every call*. The class caches `sliceBasis` and `presentation` per-instance, but the instance is thrown away the moment the caller returns.

`canvasToWorld`, `worldToCanvas`, `getZoom`, `getPan`, `getRotation`, `getScale`, `getViewPresentation`, `getCameraForEvent` all call `getResolvedView()` — frequently more than once per call (`getViewPresentation` calls `getZoom`, `getScale`, `getPan`, `getRotation` separately and each rebuilds). For a mouse-drag tool that calls `canvasToWorld` twice plus `worldToCanvas` once per move event, you're doing 3+ `resolvePlanarViewportView` constructions per move, each of which builds a `PlanarSliceBasis` from scratch (gl-matrix allocations, orientation math).

This is the kind of regression that doesn't show up in screenshot tests but shows up as scroll/pan/window-level latency on mid-range hardware. Cache the resolved view on the viewport, invalidate on `setViewState` / `resize` / binding change.

**BW>** Definitely a concern in mouse move etc.  You will need to test on slower hardware.

### 2.6 Two-layer state machine in the legacy adapter

`PlanarLegacyCompatibilityController` (~830 lines) maintains its own `properties`, `globalDefaultProperties`, `perImageIdDefaultProperties` maps keyed by `dataId` and `imageId`. This is parallel to data state owned by `PlanarMountedData` and view state owned by `PlanarViewport`. No single source of truth, and the divergence cases are real:

- Mixing legacy `setStack()` with modern `addData()` is undefined.
- `perImageIdDefaultProperties` grows unbounded as users scroll a long stack (cleared on `destroy()` so not a permanent leak, but real memory pressure during long sessions).
- Per-imageId caches are tempting for window/level memory, but unless they participate in cache eviction, they tie viewport memory to study size.

Recommend either documenting "do not mix legacy and modern APIs on the same viewport" loudly, or unifying them.

---

## 3. OHIF impact

OHIF (`extensions/cornerstone*`) currently has:

- **21+ `instanceof StackViewport | VolumeViewport | VolumeViewport3D` checks** across `commandsModule.ts`, `CornerstoneViewportService.ts`, `useViewportRendering.tsx`, `CornerstoneViewportDownloadForm.tsx`, several overlay components.
- **~57 `viewport.type === ViewportType.STACK | ORTHOGRAPHIC | VOLUME_3D` checks** in 16 files.
- Deep dependence on `getCamera()` / `setCamera()` (74 call sites), `getViewReference()` / `setViewReference()`, `getViewPresentation()` / `setViewPresentation()`. The presentation API is *canonical* in OHIF — `usePositionPresentationStore` / `useLutPresentationStore` / `useSegmentationPresentationStore` persist these objects across layout changes.
- Segmentation logic that branches `ViewportType.VOLUME_3D ? Surface : Labelmap` (e.g. `SegmentationService.ts:1560-1565`, `createHydrateSegmentationSynchronizer.ts:91-95`), and reaches into `representationData[Labelmap].imageIds` to feed DICOM SEG export.

When `useViewportNext` is on, OHIF instantiates `PlanarViewport` (or a `*LegacyAdapter` thereof) which:

- Does **not** extend `StackViewport` or `VolumeViewport` — every `instanceof` becomes false. The legacy adapter is a separate class.
- Reports `type = PLANAR_NEXT` (or whatever `*_NEXT` value the adapter forwards). Every `viewport.type === ViewportType.STACK` becomes false unless explicit mapping is added in OHIF.
- Exposes `getViewPresentation` / `setViewPresentation` but with the round-trip caveats in §2.3 — the OHIF presentation-store replay path is exactly the case that suffers most.

**Cornerstone3D itself has 51 `instanceof` checks across 36 internal files** (grep on `(StackViewport|VolumeViewport|VolumeViewport3D)`). These are addressed inside this PR via the adapter classes and `viewportTypeToViewportClass.ts` rewrites. OHIF's 21+ external checks are *not* addressed — they will need a companion OHIF PR before `useViewportNext` flips on by default.

**Suggested actions for cs3d → OHIF compatibility:**

1. Document a "viewport capability" pattern. The PR introduces `viewportHasZoom`, `viewportHasPan`, `viewportHasFrameOfReferenceUID`, `viewportHasCanvasWorldTransform` guard functions — good! Push OHIF to use these instead of `instanceof`. The guard functions should be the canonical recommendation in the migration guide.
2. Provide a "behaves-like-stack" / "behaves-like-volume" predicate so OHIF can map `PLANAR_NEXT` to its existing branches without inventing its own logic.
3. Lock down the `getViewPresentation` / `setViewPresentation` round-trip contract with a test that does `set(get())` on a viewport with non-trivial pan + zoom + rotation + flip and asserts identity. Right now it doesn't.

---

## 4. Concrete bugs / smells

### Likely bugs

- **Aliased `viewState` arrays leak from `getViewState()`** — `PlanarViewport.ts` L784-L789. Only `displayArea` is cloned; `scale`, `anchorCanvas`, `anchorWorld`, `slice.sliceWorldPoint` are shared by reference.
- **Bogus fallback camera** in `getCameraForEvent()` — `PlanarViewport.ts` L1735-L1745 — when no resolved view. Those values ride into `CAMERA_MODIFIED` payloads as `previousCamera`.
- **`setPan` fallback** uses `currentPan = [0,0]` instead of the actual anchor-derived pan — `PlanarViewport.ts` L735.
- **`setViewPresentation` fires two state writes** — `PlanarViewport.ts` L925-L929 → two `CAMERA_MODIFIED` events, two render requests.
- **`renderImageObject` collisions** — `PlanarViewport.ts` L405-L416 registers metadata under `image.imageId` with no de-duplication; back-to-back calls overwrite silently.
- **`displayArea.scaleMode` precedence is subtle** — `planarRenderCamera.ts` L127-L129 and `planarViewState.ts` L68-L70. `displayArea.scaleMode` silently overrides `viewState.scaleMode`. Setting either one alone is fine; setting both can produce surprising results.
- **No validation that user-supplied `OrientationVectors` are orthonormal** in `clonePlanarOrientation`. Bad input renders silently wrong.

### Smells

- **Constructor throws after partial DOM mutation.** `this.element.style.position/overflow/background` are written (`PlanarViewport.ts` L141-L143) before the renderer check (L177-L181). Leaves a host element in a half-initialized state if construction fails.
- **Many `as unknown as ICamera` / `as unknown as vec3` casts** — fine in isolation but indicate the type system is not actually checking the shape. `PlanarResolvedICamera` extends `ICamera` with `presentationScale`, `scale`, `scaleMode`, `flipHorizontal`, `flipVertical` — either `ICamera` should be updated, or the resolved type should not be cast back.
  **BW>** For the vec3, suggest defining a new point type that is valid input for both vec3 methods/results and valid for passing as a Point
- **`time.txt` (4640 lines)** is checked in at repo root. Looks like a Playwright JSON dump from the author's machine (path `/Users/alireza/...`). Strip before merge.
- **`tests/todo.md`** ships TODOs in the repo. Move to the PR description or an issue.
- **`PlanarViewport.type === PLANAR_NEXT`** is hard-coded (L109) but the legacy adapters report `STACK` / `ORTHOGRAPHIC` / etc. This duality is what every consumer's `viewport.type === ...` branch will hit; document the rule plainly.
- **`createHydrateSegmentationSynchronizer` uses an OHIF-flavored Surface/Labelmap branch** that is mirrored inside cs3d's labelmap render plan. The PR adds significant complexity here; would want explicit test coverage for stack-labelmap with the lazy brush path (added in `LazyBrushEditController.ts`), particularly the failure modes documented there.

### Performance hotspots worth profiling before merge

- **CPU slice samplers allocate fresh typed arrays per render frame** — `PlanarCPUVolumeSampler.ts` L550, L755 — `new SliceArrayConstructor(width * height * components)` on every sample. For 512² × 1 component that's ~1 MB / frame; scrolling produces a steady stream into GC. Pool by `(w, h, components)` shape.
- **Per-pixel scalar loop with `Math.min/Math.max` calls inline** — `PlanarCPUVolumeSampler.ts` L585-L617. Already has a deferred path (`DEFERRED_VIEWPORT_RESAMPLE_DELAY_MS`); promote it for large viewports or offload to a worker.
- **`drawImageSync` redraws full canvas** — `CpuVolumeSliceRenderPath.ts` L466 — regardless of whether anything but pan/zoom changed. The `renderingInvalidated` flag is plumbed but doesn't gate the actual pixel write.
- **`getResolvedView()` rebuild storm** discussed in §2.5.

VTK actor lifecycle and event-listener cleanup *do* look correct (`removeStreamingSubscriptions?.()`, `removeEventListener` on `IMAGE_VOLUME_MODIFIED`, `bindings.clear()` in destroy). No leak found there.

---

## 5. Tests

- The diff adds a large amount of test infrastructure: Playwright `next-screenshot-server.py`, `git-image-diff-server.py`, `checkForCanvasSnapshot.ts`, `compatibilityMode.ts`, `expectViewportNextRuntime.ts`. Good.
- New `nextStackApi.browser.test.ts` (Vitest browser). Good.
- However: I did not see an explicit `set(get())` identity test for `getCamera/setCamera` or `getViewPresentation/setViewPresentation` on `PlanarViewport`. Given §2.3, this is the single test I would add. A property-based fuzzer over `pan × zoom × rotation × flip × scaleMode` would catch the rounding/aliasing/scaling issues above.
- 51 internal `instanceof` checks; the diff updates many, but `viewportTypeToViewportClass.ts` is the only place I see that systematically routes new types. A grep audit for remaining `instanceof StackViewport` against Next adapters would be worthwhile.

---

## 6. Concrete recommendations before merging

1. **Settle the renames** in `tests/todo.md` — or commit publicly that the current names are final. Don't ship `ViewportNext`, `dataId`, and `VolumeViewport3DV2` as the stable names if you intend to change them.
2. **Strip `time.txt` and `tests/todo.md`** from the diff.
3. **Add a `getCamera` / `setCamera` and `getViewPresentation` / `setViewPresentation` round-trip test** on `PlanarViewport` with non-trivial pan + zoom + rotation + flip + scaleMode. Document the cross-slice anchor-projection caveat explicitly in the migration guide.
4. **Deep-clone what `getViewState()` returns** (`PlanarViewport.ts` L784-L789). Cheap fix, prevents quiet state corruption.
5. **Cache `getResolvedView()` on the viewport instance** with invalidation tied to `viewState` / `bindings` / `resize`. This is the single biggest perf win for tool interaction; right now every tool that touches `canvasToWorld` per mouse-move rebuilds the slice basis several times.
6. **Pool typed arrays in `PlanarCPUVolumeSampler`** (see hotspots above).
7. **Batch the two state writes** in `setViewPresentation` into one `setViewState({ ...nextCamera, anchorCanvas })`.
8. **Fix the `setPan` no-resolved-view path** and the bogus fallback `ICamera` in `getCameraForEvent`. Or have those methods throw / return undefined instead of pretending to work pre-load.
9. **Mark internal types `@internal`** (`PlanarPayload`, `PlanarViewportRenderContext`, `PlanarEffectiveRenderMode`, `PlanarRegisteredDataSet`, `*Path` class names) to give yourself room to evolve without a breaking change.
10. **OHIF coordination:** ship a small "viewport-capability check" migration recipe alongside the `feat/use-beta-5.0-cs3d` PR — OHIF's ~21 `instanceof` checks and ~57 `viewport.type === STACK` checks are the long pole for `useViewportNext` ever becoming the default. The capability guard functions (`viewportHasZoom`, etc.) already exist in this PR; just need them threaded into OHIF.

---

## 7. Verdict

The architectural shape (DataProvider → RenderPathResolver → RenderPath → Binding, with semantic view state separated from renderer camera) is the right move and worth doing. The implementation is mostly thorough, the legacy adapter approach is the right migration strategy, and the docs/tests scaffolding is encouraging.

What I would not do is merge this as a *stable* surface. The author's own TODO list says the central names are not final; the round-trip semantics of the camera API have real holes; OHIF will need a paired PR before `useViewportNext` becomes the default. Land it on `beta` as is — explicitly behind `useViewportNext` — and burn down items 1–10 above before flipping the default or releasing as 5.0.

wayfarer3130 (approved):
There are still things needed to finish in beta before we merge beta to main, but I really like the overall direction here.   Lets go with what is here now, as soon as the ohif-downstream tests are passing, and then continue work on beta for a couple of changes, and then work on a few things in main over the next several months.

PR #2674, opened 2026-03-25
fix(segmentation): replace unbounded render queue with coalescing Set

pedrokohler:
### Context

`SegmentationRenderingEngine._pendingRenderQueue` is a `string[][]` that grows
without bound. Every scroll event on a StackViewport with segmentations appends a
new viewport-ID array to the queue. The queue is drained one entry per animation
frame via `shift()`, but entries arrive far faster than they are consumed.

After 310 scroll events on a viewport with 13 labelmap segmentations, the queue
held **4,077 entries** — each an array of 13+ viewport IDs with associated
closures and promise chains — consuming significant memory and CPU.

### Changes & Results

Replace `_pendingRenderQueue: string[][] = []` with
`_pendingRender: Set<string> = new Set()`.

- Duplicate viewport IDs are coalesced into a single entry in the Set.
- When the animation frame fires, all pending IDs are flushed in one batch and the
  Set is cleared.
- Queue size is now bounded to `O(viewports)` instead of `O(scroll events)`.

**Before:** queue grows linearly with scroll events (4,077 entries after 310 scrolls).
**After:** pending set stays bounded at the number of unique viewport IDs (e.g. 13).

### Testing

1. Open a StackViewport with 10+ labelmap segmentations.
2. Scroll rapidly through 300+ slices.
3. Observe stable memory — no unbounded growth in the render queue.

#### PR

#### Code

      etc.)

      additions or removals.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/packages/tools/src/stateManagement/segmentation/SegmentationRenderingEngine.ts
+++ b/packages/tools/src/stateManagement/segmentation/SegmentationRenderingEngine.ts
@@ -42,7 +42,7 @@ const planarContourToolName = PlanarFreehandContourSegmentationTool.toolName;
  */
 class SegmentationRenderingEngine {
   private _needsRender: Set<string> = new Set();
-  private _pendingRenderQueue: string[][] = [];
+  private _pendingRender: Set<string> = new Set();
   private _animationFrameSet = false;
   private _animationFrameHandle: number | null = null;
   public hasBeenDestroyed: boolean;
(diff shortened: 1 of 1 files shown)


sedghi:
@pedrokohler  do you need this?

pedrokohler:
I think not. Will close.

PR #2675, opened 2026-03-25
fix(labelmap): use content-based comparison for segmentsHidden Set

pedrokohler:
### Context

`_needsTransferFunctionUpdate()` in `labelmapDisplay.ts` compares
`oldSegmentsHidden !== segmentsHidden` using **reference equality**. Because
`segmentsHidden` is a new `Set` instance on every render call, this comparison
always returns `true` — even when the contents are identical. This forces
`forceOpacityUpdate = true` unconditionally, triggering unnecessary `.slice()` and
`new Set()` allocations on every render cycle for every segmentation.

With 13 segmentations across ~180 render calls per scroll, this produces hundreds
of redundant transfer-function updates per interaction.

### Changes & Results

Replace the reference comparison with a content-based comparison:

1. Check `size` equality first (fast path).
2. If sizes match, check `[...segmentsHidden].some(idx => !oldSegmentsHidden.has(idx))`.

**Before:** transfer function updated on every render (~180 times per scroll).
**After:** transfer function updated only when segments are actually hidden/shown (~15 times total in a typical session).

### Testing

1. Open a StackViewport with multiple labelmap segmentations.
2. Scroll through slices without toggling segment visibility.
3. Place a breakpoint or log in `_needsTransferFunctionUpdate` — `forceOpacityUpdate`
   should be `false` when no visibility changes occurred.

#### PR

#### Code

      etc.)

      additions or removals.

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **Performance Improvements**
  * Reduced unnecessary labelmap rendering updates when nothing has changed, which can make repeated rendering smoother.
* **Bug Fixes**
  * Improved consistency of labelmap color, opacity, and outline updates so visual settings are only reapplied when needed.

### DIFF
--- a/packages/tools/src/tools/displayTools/Labelmap/labelmapActorStyle.ts
+++ b/packages/tools/src/tools/displayTools/Labelmap/labelmapActorStyle.ts
@@ -31,6 +31,17 @@ const actorTransferFunctions = new WeakMap<
   LabelmapTransferFunctions
 >();
 
+// Signature of the last style applied to each actor. setLabelmapColorAndOpacity
+// runs on every segmentation render (e.g. every scroll), but the labelmap
+// transfer functions only change when colors, per-segment style, segment
+// visibility, outline config, or active/visibility state change. Caching the
+// applied signature lets us skip the expensive vtk transfer-function rebuild
+// and actor/mapper invalidation when nothing relevant changed.
+const lastAppliedLabelmapStyle = new WeakMap<
+  vtkVolume | vtkImageSlice,
+  string
+>();
+
 const MAX_NUMBER_COLORS = 255;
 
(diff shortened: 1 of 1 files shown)


pedrokohler:
@sedghi this is a performance optimization, it reduces allocation churn / CPU overhead during scrolling.

sedghi:
Thanks, i let QA test it

pedrokohler:
@sedghi it's been almost 2 months since QA was asked to review. Could we merge?

sedghi:
@james-hanks

pedrokohler:
@sedghi 😢

sedghi:
can you rebase

sedghi (approved):

PR #2679, opened 2026-03-29
fix: stack viewport colormap preserve and 4d local improved support

sedghi:
1. In stack viewport if we have different types of data colormap change reset on slice changes

2. 4d for local drag and drop does not have good support

Ask alireza for data

## Summary by CodeRabbit

* **New Features**
  * Enhanced viewport property preservation to restore colormap, sharpening, and smoothing settings
  * Added support for frame parameters in multiple URL formats

* **Bug Fixes**
  * Fixed viewport properties preservation when re-initializing with different data types
  * Improved multiframe image data type handling

* **Tests**
  * Added test validating viewport property preservation during stack re-initialization
  * Enhanced multiframe frame parameter handling test coverage


jbocce (changes requested):
Overall the changes seem fine. I tested and the colormap example no longer resets while navigating the stack. However, the two (formerly) non-readable dicom files are still non-readable.

Something to consider beyond the changes you made...

in retrieveMultiframeDataset.ts, generateMultiframeWADOURIs still appends &frame=${i} unconditionally. If the base URI does not already contain query params, that may produce malformed IDs. Might be worth normalizing with ? vs & there as well.

jbocce (approved):
The code is approved. However the OHIF downstream validation is failing.

PR #2693, opened 2026-04-10
fix(viewport): restore default visual properties on reset in VolumeViewport3D and VolumeViewport

GhadeerAlbattarni:
### Context
This is fixes an issue in ohif #5922

When a user activates the 3D tool (which applies a volume rendering preset such as CT-Bone), zooms in/out, and then clicks Reset, the image reverts to black and white instead of restoring the original 3D color rendering. The same underlying issue may exist in `VolumeViewport`. 

### Root cause 

Both `VolumeViewport3D.resetProperties` and `VolumeViewport._resetProperties` unconditionally called `setTransferFunctionNodes` with `initialTransferFunctionNodes`, which wipes the RGB color transfer function back to the default grayscale ramp captured at volume load time, before any preset or colormap was applied.
`setTransferFunctionNodes` only resets the RGB color function. It does not touch the opacity, gradient opacity, or lighting properties set by a preset. This left the viewport in an inconsistent state: preset opacity/lighting still active, but color rendering in grayscale.

### Changes & Results

Replace the unconditional `setTransferFunctionNodes` call with a check against the configured default properties (`getDefaultProperties`):
- If the defaults include a `preset` or `colormap`, apply them via`setProperties(properties, volumeId, true)`. This routes through `applyPreset`, which fully restores the color transfer function, scalar opacity, gradient opacity, interpolation, and lighting — the complete visual state — in a single call. `suppressEvents` is set to `true` to avoid duplicate or redundant events since `resetProperties` fires its own `VOI_MODIFIED` at the end.
- If no defaults are configured, fall back to the original `setTransferFunctionNodes` behaviour (no regression).
The fix is applied to both `VolumeViewport3D.resetProperties` and `VolumeViewport._resetProperties` for consistency. For `VolumeViewport` (MPR), the fallback path means plain CT viewports with no configured defaults behave exactly as before.

**Before**

https://github.com/user-attachments/assets/645b36fc-d329-4860-a0c5-f6361aa0a5dd

**After**

https://github.com/user-attachments/assets/370001f3-dbb1-4a23-b8c5-7f8c43c25e84

### Testing

1. Open a study and activate the 3D tool
2. Zoom in/out on the viewport
3. Click Reset
4. verify the image correctly restores the 3D color preset

- Also you can apply a different a preset (e.g, CT-AAA2), then click reset. 
Verify that not only the color has been reset to default but also the opacity, shading, and lighting all set to default. 

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] OS: macOS 10.15.4
- [x] Node version: v22.13.0


packages/core/src/RenderingEngine/VolumeViewport3D.ts
```diff
@@ -158,14 +158,23 @@ class VolumeViewport3D extends BaseVolumeViewport {
...
-        .getRGBTransferFunction(0);
-
-      setTransferFunctionNodes(
-        transferFunction,
-        this.initialTransferFunctionNodes
-      );
+      // Check if there are default properties (e.g., a 3D preset or colormap)
+      // to restore. This ensures "Reset" goes back to the configured look
+      // rather than the raw initial grayscale state.
+      const properties = this.getDefaultProperties(volumeId);
```
sedghi: not sure why we are repeating the same thing twice, can you investigate, i remember 3d viewport was extending the volume viewport. worst case we need a refactor

GhadeerAlbattarni: I'll check

GhadeerAlbattarni: Thanks @sedghi for your feedback. Both VolumeViewport3D and VolumeViewport extend BaseVolumeViewport directly. So I added a new method on BaseVolumeViewport so that both subclasses can use it.

packages/core/src/RenderingEngine/BaseVolumeViewport.ts
```diff
@@ -1145,6 +1149,40 @@ abstract class BaseVolumeViewport extends Viewport {
...
+   * Restores the default visual transfer function for a volume actor on reset.
+   * If the viewport has a configured default preset or colormap, it re-applies
+   * them via setProperties to restore the full visual state (color TF, opacity,
+   * lighting). Otherwise, it falls back to the initial transfer function nodes
+   * captured at volume load time.
+   *
+   * @param volumeActor - The actor entry for the volume
+   * @param volumeId - The id of the volume
+   */
+  protected _restoreDefaultVisualProperties(
```
jbocce: How about renaming this to `_restoreVolumeRenderingDefaults `?

jbocce (approved):
Just one small naming suggestion. I have tested locally and it looks good. @sedghi do you have anything to add to this PR?

jbocce (approved):

PR #2699, opened 2026-04-14
fix(core): defer resize when animationFrame is pending

sjsanc:
Hey folks! I've been using this library in a project for a short while and wanted to contribute. I picked something easy to reproduce. Let me know if I'm missing anything.

### Context

Fixes #2667

When resize() is called while a render is already pending, _resizeVTKViewports hits an early return and drops the resize entirely. In a multi-viewport setup this causes canvas dimensions to go stale and annotations to render at wrong positions.

### Changes & Results

Instead of dropping the resize, we now cache the args in _pendingResizeArgs. Once _renderFlaggedViewports completes and clears the flag, it flushes the pending resize.

### Testing

Added a Karma test to RenderingEngineAPI_test.js covering this case.

#### PR

#### Code

#### Tested Environment

- [x] OS: Linux (Arch)
- [x] Node version: 25.9.0
- [x] Browser: Chromium 147.0.7727.55

### DIFF
--- a/packages/core/src/RenderingEngine/ContextPoolRenderingEngine.ts
+++ b/packages/core/src/RenderingEngine/ContextPoolRenderingEngine.ts
@@ -33,6 +33,11 @@ import type { VtkOffscreenMultiRenderWindow } from '../types';
  */
 class ContextPoolRenderingEngine extends BaseRenderingEngine {
   private contextPool: WebGLContextPool;
+  private _pendingResizeArgs: {
+    vtkDrivenViewports: (IStackViewport | IVolumeViewport)[];
+    keepCamera: boolean;
+    immediate: boolean;
+  } | null = null;
 
   constructor(id?: string) {
     super(id);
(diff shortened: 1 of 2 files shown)


sedghi (changes requested):
In `packages/core/src/RenderingEngine/ContextPoolRenderingEngine.ts:339`, the early return in `_renderFlaggedViewports` when `viewportsToRender.length === 0` sets `_animationFrameSet` to false. This happens before the new `_pendingResizeArgs` flush block, which runs after the main render. If a flagged viewport gets disabled between scheduling and firing the requestAnimationFrame, `_needsRender` ends up empty. This means any resize deferred during that frame never gets flushed. Since `_animationFrameSet` is now false, the resize just sits there until some unrelated later render finally flushes it with stale arguments. This edge case basically brings back the exact dropped resize bug this PR is trying to fix.

Also, in `packages/core/src/RenderingEngine/ContextPoolRenderingEngine.ts:244`, `_pendingResizeArgs` isn't cleared in `_clearAnimationFrame`. While it's pretty low impact, a stale pending resize could survive a teardown or destroy, which leaves the lifecycle state kind of inconsistent.

PR #2707, opened 2026-04-21
fix(dicom-image-loader): use Float32Array for non-integer rescale slopes to prevent pixel data corruption

### LINKED ISSUE #2706: [Bug] DICOM images with float RescaleSlope (e.g. DTI FA maps) render as all black in  StackViewport
### Describe the Bug

 Hi, I'm trying to display DTI FA (Fractional Anisotropy) DICOM images using StackViewport, but the viewport shows only black. These images display correctly in other viewers like RadiAnt.

After debugging, I noticed that all pixel values become 0 after image loading.

 getPixelDataTypeFromMinMax selects Uint8Array for float-scaled data

  During preScale in decodeImageFrameWorker.js, _handlePreScaleSetup computes the scaled min/max:

  scaledMin = 0 * 0.001 = 0.0
  scaledMax = 1000 * 0.001 = 1.0

  Then getPixelDataTypeFromMinMax(0.0, 1.0) runs:
```

  if (Number.isInteger(min) && Number.isInteger(max)) {  // Number.isInteger(0.0) → true,
  Number.isInteger(1.0) → true
    if (max <= 255) {
      pixelDataType = Uint8Array;  // ← chosen here
    }
  }

```
  Since JavaScript treats 0.0 and 1.0 as integers, Uint8Array is selected. But the actual pixel values after rescaling are floats like 0.013, 0.5 0.999 — these all truncate to 0 in Uint8Array.

The legacy Cornerstone had a similar float RescaleSlope issue: https://github.com/cornerstonejs/cornerstone/issues/302

I may be wrong about some of the details above — please correct me if I've misunderstood anything. Any guidance would be greatly appreciated. Thank you!

### Steps to Reproduce

This is the dicom info I tried to use:
Modality: MR
BitsAllocated / BitsStored: 16 / 12
Raw pixel range: 0 – 1000 
RescaleSlope: 0.001   
RescaleIntercept: 0.0 
WindowCenter / WindowWidth: 0.5 / 1.0

### The current behavior

Image

### The expected behavior

I want to show MRI fa dicom with floating rescaleSlope just like other usual dicom.

### System Information

I used version below:
"@cornerstonejs/core": "^4.15.33",
"@cornerstonejs/dicom-image-loader": "^4.15.33",
"@cornerstonejs/nifti-volume-loader": "^4.15.33",
"@cornerstonejs/tools": "^4.15.33",

eugenest2557:
### Context

Fixes #2706

DICOM images with non-integer `RescaleSlope` (e.g. DTI FA maps with `RescaleSlope=0.001`) render as completely black in StackViewport. These images display correctly in other viewers such as RadiAnt and OsiriX.

The legacy Cornerstone had a similar issue ([cornerstonejs/cornerstone#302](https://github.com/cornerstonejs/cornerstone/issues/302), fixed in [PR #303](https://github.com/cornerstonejs/cornerstone/pull/303)).

Two independent bugs are involved:

1. **`getPixelDataTypeFromMinMax(0.0, 1.0)` selects `Uint8Array`** because `Number.isInteger(1.0) === true` in JavaScript. All fractional scaled values (0.013, 0.5, 0.999) are truncated to 0. This function is called twice (web worker + main thread `setPixelDataType`), destroying the data at both stages.

2. **`toLowHighRange` with `WindowWidth=1`** produces `lower === upper === 0` via the LINEAR formula `(WW-1) = 0`, causing division by zero in the VOI shader and a binary black/white image.

### Changes & Results

1. **`_handlePreScaleSetup` in `decodeImageFrameWorker.js`**: Detect non-integer scaling parameters and force `Float32Array` before `getPixelDataTypeFromMinMax` is called.

2. **`setPixelDataType` in `setPixelDataType.ts`**: Skip re-typing if `pixelData` is already `Float32Array`, preventing the main thread from downgrading it to `Uint8Array`.

3. **`toLowHighRange` in `windowLevel.ts`**: When `WindowWidth <= 1` with LINEAR function, fall back to LINEAR_EXACT (`lower = WC - WW/2`, `upper = WC + WW/2`) to produce a valid range.

All three fixes are needed — removing any one results in either a black screen or a binary black/white image.

### Testing

- Load a DTI FA DICOM with `RescaleSlope=0.001` in StackViewport — should display grayscale instead of black
- Load a standard CT DICOM (integer `RescaleSlope=1`) — should be unaffected
- Load a DICOM with `WindowWidth > 1` — `toLowHighRange` should use existing LINEAR formula unchanged

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] OS: macOS
- [x] Node version: 20
- [x] Browser: Chrome

I may be wrong about some of the details — please correct me if I've misunderstood anything. Any feedback would be greatly appreciated. Thank you!

## Summary by CodeRabbit

* **New Features**
  * Added support for an additional dose-based scaling option for RTDOSE images.
  * Improved scaling detection so image data is handled more accurately across PET, RTDOSE, and standard rescale cases.

* **Bug Fixes**
  * Fixed pixel type handling for cases where scaling produces non-integer values, reducing unexpected truncation.
  * Improved rendering of very small window widths for certain display settings.
  * Prevented already floating-point pixel data from being downgraded during processing.

### DIFF
--- a/packages/core/src/types/ScalingParameters.ts
+++ b/packages/core/src/types/ScalingParameters.ts
@@ -11,6 +11,8 @@ interface ScalingParameters {
   suvlbm?: number;
   /** SUV body surface area */
   suvbsa?: number;
+  /** dose grid scaling factor applied for RTDOSE modality */
+  doseGridScaling?: number;
 }
 
 interface PTScaling {
--- a/packages/core/src/utilities/generateVolumePropsFromImageIds.ts
+++ b/packages/core/src/utilities/generateVolumePropsFromImageIds.ts
@@ -129,6 +129,10 @@ function _determineDataType(
 
   switch (BitsAllocated) {
     case 8:
+      // Fall back to Float32 for non-integer scaling, like the 16-bit case. See #2706.
+      if (canRenderFloat && floatAfterScale) {
+        return 'Float32Array';
+      }
       return 'Uint8Array';
 
     case 16:
--- a/packages/core/src/utilities/getScalingMode.ts
+++ b/packages/core/src/utilities/getScalingMode.ts
@@ -1,0 +1,30 @@
+import type { ScalingParameters } from '../types';
+
+export type ScalingMode = 'PT_SUV' | 'RTDOSE' | 'RESCALE' | 'NONE';
+
+/**
+ * Which modality-specific scaling transform is applied to the pixel data.
+ * Mirrored in dicom-image-loader (its worker can't import core runtime); keep
+ * both in sync.
+ */
+export function getScalingMode(
+  scalingParameters: ScalingParameters
+): ScalingMode {
+  const { rescaleSlope, rescaleIntercept, modality, doseGridScaling, suvbw } =
+    scalingParameters;
+
+  if (modality === 'PT' && typeof suvbw === 'number' && !isNaN(suvbw)) {
+    return 'PT_SUV';
+  }
+  if (
+    modality === 'RTDOSE' &&
+    typeof doseGridScaling === 'number' &&
+    !isNaN(doseGridScaling)
+  ) {
+    return 'RTDOSE';
+  }
+  if (typeof rescaleSlope === 'number' && typeof rescaleIntercept === 'number') {
+    return 'RESCALE';
+  }
+  return 'NONE';
+}
--- a/packages/core/src/utilities/hasFloatScalingParameters.ts
+++ b/packages/core/src/utilities/hasFloatScalingParameters.ts
@@ -1,15 +1,28 @@
 import type { ScalingParameters } from '../types';
+import { getScalingMode } from './getScalingMode';
 
 /**
- * Checks if the scaling parameters contain a float rescale value.
- * @param scalingParameters - The scaling parameters to check.
- * @returns True if the scaling parameters contain a float rescale value, false otherwise.
+ * True if the applied scaling can produce non-integer output (needs Float32).
+ * Only inspects parameters used by the current modality. See #2706.
  */
 export const hasFloatScalingParameters = (
   scalingParameters: ScalingParameters
 ): boolean => {
-  const hasFloatRescale = Object.values(scalingParameters).some(
-    (value) => typeof value === 'number' && !Number.isInteger(value)
-  );
-  return hasFloatRescale;
+  const { rescaleSlope, rescaleIntercept, doseGridScaling, suvbw } =
+    scalingParameters;
+  const isFloat = (value: unknown): boolean =>
+    typeof value === 'number' && !Number.isInteger(value);
+
+  switch (getScalingMode(scalingParameters)) {
+    case 'PT_SUV':
+      return (
+        isFloat(suvbw) || isFloat(rescaleSlope) || isFloat(rescaleIntercept)
+      );
+    case 'RTDOSE':
+      return isFloat(doseGridScaling);
+    case 'RESCALE':
+      return isFloat(rescaleSlope) || isFloat(rescaleIntercept);
+    default:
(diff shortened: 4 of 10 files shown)


sedghi (changes requested):
The three-part fix correctly addresses the reported StackViewport corruption for non-integer RescaleSlope by forcing a Float32Array in the pre-scale path, preventing the main thread from downgrading it, and avoiding a division by zero when WW=1 where lower equals upper. This core scenario works because comlink structured-clone preserves the Float32Array across the worker boundary, meaning the instanceof guard in setPixelDataType works as expected. The remaining issues are about incomplete coverage in adjacent paths rather than the main flow.

minor packages/dicomImageLoader/src/decodeImageFrameWorker.js:93
The Float32 fix only runs in _handlePreScaleSetup. When options.targetBuffer.type is provided for volume loading, postProcessDecodedPixels takes the _handleTargetBuffer branch instead. Here, validatePixelDataType(scaledMin, scaledMax, Uint8Array) returns true for a scaled range like [0,1] because both 0 and 1 are integers. This leaves invalidType as false, so the data is copied into a Uint8Array target and truncates the fractional values. As a result, loading the same DTI FA map into a volume viewport still causes corruption.

minor packages/dicomImageLoader/src/decodeImageFrameWorker.js:249
hasFloatRescale is recomputed inside _handlePreScaleSetup using the exact same Object.values(...).some(...) expression already computed at lines 79-83 in the caller, which was not passed down. This causes a redundant scan of the scaling parameters.

minor packages/core/src/utilities/windowLevel.ts:61
The WW<=1 guard only rewrites voiLUTFunction when it is LINEAR. However, SAMPLED_SIGMOID goes through the identical LINEAR-formula branch at lines 64-67 and will still produce a division by zero where lower equals upper when WW<=1. The guard should also cover SAMPLED_SIGMOID to match the branch it protects.

nit packages/dicomImageLoader/src/decodeImageFrameWorker.js:250
hasFloatRescale triggers on any non-integer numeric scaling parameter, including ones not actually used by _calculateScaledMinMax for the current modality, such as a stray suvbw on a non-PT image. This can unnecessarily force a Float32Array and double the memory use compared to Uint16 for 16-bit data that did not need it.

eugenest2557:
Thank you very much for reviewing this so carefully, and for laying out each point so clearly — it was really helpful. I've addressed all four:

- 1 (volume truncation): You're right that the volume path was still affected. The target-buffer branch in the worker now also forces Float32 for non-integer scaling (`hasFloatRescale && !isFloatTarget` in the invalidType check), and `generateVolumePropsFromImageIds` now returns Float32Array for the 8-bit case when `floatAfterScale` is true (the 16-bit case already did), so the DTI FA map no longer truncates in a volume viewport.
- 2 (redundant scan): Fixed — `hasFloatRescale` is now computed once in `postProcessDecodedPixels` and passed into `_handlePreScaleSetup`, so the duplicate scan is gone.
- 3 (SAMPLED_SIGMOID): You're right, thank you — the WW<=1 guard now also rewrites SAMPLED_SIGMOID to LINEAR_EXACT, matching the branch it protects.
- 4 (over-triggering): Thank you for pointing this out. Float detection is now modality-aware and only inspects the parameters actually applied for the current modality, so a stray non-integer suvbw on a non-PT image (or doseGridScaling on a non-RTDOSE image) no longer forces Float32.

While addressing these I also consolidated the modality branch selection into a single `getScalingMode` helper shared by `scaleArray` and `_calculateScaledMinMax`, and reused it in `hasFloatScalingParameters` so the volume path gets the same modality-aware behavior.

Please let me know if anything still looks off — happy to make further changes. Thank you again for the thorough review!

Issue #2646, opened 2026-03-05
[Bug] Closed Freehand ROI stat value changes when switching from Stack Viewport to Volume Viewport


Devu-trenser:
### Describe the Bug

The maximum intensity value of a Closed Freehand ROI changes when switching between Stack Viewport and Volume Viewport in the viewer.

When the same ROI annotation is viewed in different viewport types, the reported max value differs, even though the ROI shape and location remain unchanged.

This behavior is more visible after changes introduced in
PR #2415.

https://github.com/user-attachments/assets/1081eae5-bac9-4621-86b2-999f25485d9b

### Steps to Reproduce

1. Open a study in the viewer. Test Data: https://viewer.ohif.org/viewer?StudyInstanceUIDs=1.3.6.1.4.1.5962.99.1.2968617883.1314880426.1493322302363.3.0
2. Use the Planar Freehand ROI tool to draw a closed freehand ROI on an image in the Stack Viewport.
3. Observe the statistics values (especially max).
4. Switch the MPR viewport (volume) and move the text-box.
5. Observe the ROI statistics again.

### The current behavior

The max value of the ROI changes after switching from Stack Viewport to Volume Viewport, even though the ROI annotation remains the same.

### The expected behavior

The ROI statistics (max, min, mean, etc.) should remain consistent when switching between Stack Viewport and Volume Viewport, since the ROI annotation and underlying image data are unchanged.

### System Information

System:
OS: Linux 6.11 Ubuntu 24.10 24.10 (Oracular Oriole)
Node: 22.16.0 - ~/.nvm/versions/node/v22.16.0/bin/node
Chrome: 139.0.7258.154

Devu-trenser:
While investigating the ROI statistics mismatch between stack and volume viewports, we traced the issue to the conversion of ROI points [from world coordinates to image index coordinates](https://github.com/cornerstonejs/cornerstone3D/blob/main/packages/tools/src/tools/annotation/PlanarFreehandROITool.ts#L908).

Because this transformation involves floating-point calculations, the resulting index values can sometimes contain small precision offsets. For example, when an ROI lies exactly on a slice plane, the Z index may be returned as a value slightly below the expected integer (e.g., 33.9999 instead of 34).

`const worldPosIndex = indexPoints[j].map(Math.floor);`

[...]

PR #2717, opened 2026-05-03
fix(annotation): prevent closed freehand ROI stat mismatch between stack and volume viewports

Devu-trenser:
### Context
Fixes issue: https://github.com/cornerstonejs/cornerstone3D/issues/2646

Fixes the issue where Closed Freehand ROI statistics (especially max) changed when switching between Stack Viewport and Volume Viewport.

The root cause was floating-point precision drift during world-to-index coordinate conversion.
The previous implementation used Math.floor for Z index calculation, which could incorrectly map planar annotations to the previous slice in volume viewports.

This caused ROI statistics to be computed from slightly different voxel regions between stack and volume rendering modes, even though the annotation itself remained visually unchanged.

### Changes & Results

- Updated Z index calculation from Math.floor to Math.round
- Ensured planar annotations stay constrained to the nearest slice
- Prevented unintended slice shifts caused by floating-point precision issues
- Fixed inconsistency in ROI statistics between Stack and Volume viewports

Before

- Closed Freehand ROI max value changed after switching viewport types
- ROI could be evaluated on an adjacent slice in volume mode

After

- ROI statistics remain consistent across Stack and Volume viewports
- Annotation mapping stays aligned with the intended slice

https://github.com/user-attachments/assets/666e6a28-b8f5-4c85-af5a-86b40cc42630

### Testing

- Open the OHIF viewer with the  study: https://viewer.ohif.org/viewer?StudyInstanceUIDs=1.3.6.1.4.1.5962.99.1.2968617883.1314880426.1493322302363.3.0
- Use the Planar Freehand ROI tool to draw a closed ROI in the Stack Viewport
- Observe ROI statistics (max, min, mean)
- Switch to the MPR/Volume viewport
- Move or refresh the annotation textbox
- Verify that ROI statistics remain unchanged between viewport types

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [X] "OS:  Ubuntu 24.04
- [X] "Node version:  version: 22
- [X] "Browser: Chrome 139
  

## Release Notes

* **Improvements**
  * Improved accuracy of planar freehand ROI annotation statistics calculations through optimized index boundary handling and enhanced voxel-level precision.
  * Refined index snapping logic to better maintain measurement accuracy and prevent coordinate drift in ROI operations, ensuring more reliable annotations and measurements.


sen-trenser:
@sedghi Could you please take a look at this PR ?
Thanks!

packages/tools/src/tools/annotation/PlanarFreehandROITool.ts
```diff
@@ -915,15 +915,30 @@ class PlanarFreehandROITool extends ContourSegmentationBaseTool {
...
+      // Each slice represents a voxel slab with a finite thickness, and the voxel
+      // index corresponds to the center of that slab. When converting world
+      // coordinates back to index space, floating-point precision can produce
+      // slightly offset fractional indices. Rounding ensures the annotation
+      // maps to the nearest slice instead of unintentionally shifting to an adjacent
+      // slice.
+      const zi = Math.round(worldPosIndex[2]);
+
+      kMin = Math.min(kMin, zi);
+      kMax = Math.max(kMax, zi);
```
sedghi: I am not sure about your solution, this _might_ fix in one plane, but in other planes where aquisitino is sag or cor, the issue still exists since you are taking the round on the z no?

packages/tools/src/tools/annotation/PlanarFreehandROITool.ts
```diff
@@ -915,15 +915,30 @@ class PlanarFreehandROITool extends ContourSegmentationBaseTool {
...
-      jMax = Math.max(jMax, worldPosIndex[1]);
-
-      kMin = Math.min(kMin, worldPosIndex[2]);
-      kMax = Math.max(kMax, worldPosIndex[2]);
+      const worldPosIndex = indexPoints[j];
+
+      // For X and Y we compute the ROI bounds in the image grid. Using floor ensures
+      // the bounding box starts at the containing pixel index (top-left pixel in the
+      // image grid).
+      const xi = Math.floor(worldPosIndex[0]);
```
wayfarer3130: Create the min/max value without rounding or flooring.  Then post-process the iMin/iMax etc with this function for each i,j,k

delta = max-min
if( delta <= 1 ) return min/max are identically Math.round(delta)

return min is Math.floor(min) and max is Math.ceil(max)

That should work for planar ijk.  It doesn't quite work for oblique.

wayfarer3130:
@sedghi @sen-trenser - I can't reproduce this on viewer-dev

packages/tools/src/tools/annotation/PlanarFreehandROITool.ts
Devu-trenser: @wayfarer3130  Thanks for the suggestion, updated the bounds normalization logic for i/j/k values.

Devu-trenser:
> I can't reproduce this on viewer-dev

This can be reproduced in viewer-dev while switching between Stack and Volume viewports for a Closed Freehand ROI.

Test data:  https://viewer-dev.ohif.org/viewer?StudyInstanceUIDs=1.3.6.1.4.1.5962.99.1.2968617883.1314880426.1493322302363.3.0

Attaching a video for reference.

https://github.com/user-attachments/assets/beb2e53b-f1fb-421a-ab21-8bee4a3d5dee

sen-trenser:
@wayfarer3130 Could you please take a look at the above comments from Devu and check the changes?
Thanks!

wayfarer3130 (approved):

Devu-trenser:
@sedghi  Could you please merge this PR?

cc: @wayfarer3130 @sen-trenser

Issue #2633, opened 2026-02-25
[Bug] Inconsistency in the background color of the fusion viewports in TMTV mode


nithin-trenser:
### Describe the Bug

Inconsistency in the background color of the fusion viewports in TMTV mode. The background color of axial and sagittal fusion viewport is different when compared with other viewports. 

The failure of the initial HSV fusion causes color inconsistencies when switching between LUTs across different viewports.

https://github.com/user-attachments/assets/be09c059-4e4c-4a9a-b780-a0803ebaa37d

### Steps to Reproduce

- Launch the PET-CT study in the TMTV mode.
- Observe the background color of the fusion viewport.
- Apply color LUT for CT and PT in fusion viewports.
- Observe the fusion viewports.

### The current behavior

The initial TMTV viewport load exhibits inconsistent HSV fusion behavior. While the default HSV fusion fails to apply correctly in the axial and sagittal views, the coronal view displays the HSV data inconsistently.

### The expected behavior

The default HSV fusion must render consistently and accurately across all fusion viewports (fusionAxial, fusionSagittal, and fusionCoronal).

### System Information

  System:
    OS: Windows 11 10.0.22631
    CPU: (8) x64 11th Gen Intel(R) Core(TM) i5-1135G7 @ 2.40GHz
    Memory: 6.44 GB / 31.74 GB
  Binaries:
    Node: 20.17.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.22 - C:\Users\1497\AppData\Roaming\npm\yarn.CMD
    npm: 11.1.0 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Chrome: 145.0.7632.116
    Edge: Chromium (143.0.3650.139)
    Internet Explorer: 11.0.22621.3527

nithin-trenser:
We will be sharing a fix for this soon. Thank you for your patience.

PR #2719, opened 2026-05-05
fix: tmtv backgound color inconsistency

nithin-trenser:
### Context
Fixes issue #2633 

Inconsistency in the background color of the fusion viewports in TMTV mode. The background color of axial and sagittal fusion viewport is different when compared with other viewports.

The failure of the initial HSV fusion causes color inconsistencies when switching between LUTs across different viewports.

The getColormap method in BaseVolumeViewport.ts was always reading opacity and threshold values directly from the VTK volume actor using getMaxOpacity() and getThresholdValue(), regardless of the source of the values.

This caused two issues:

- Hanging Protocol Values Overwritten: When hanging protocol provided opacity as an array (e.g., [{value: 0, opacity: 0}, {value: 0.1, opacity: 0.8}, {value: 1, opacity: 0.9}]), these were being replaced by actor-derived single number values, losing the protocol-defined opacity mapping.
- Threshold Slider Incompatibility: The threshold slider feature (introduced in v3.11/3.12) expects single number values, but the method was reading from actor even when protocol values should be preserved.

### Changes & Results

- Modified `getColormap` to conditionally retrieve opacity/threshold values based on their type:
- Single Number (Slider Case): Read current value from actor using `getMaxOpacity() / getThresholdValue()` to reflect real-time slider position
- Array (Hanging Protocol Case): Use stored values from `viewportProperties.colormap` to preserve protocol-defined mappings
- Null/Undefined: Preserve stored values including explicit null for threshold

Before

- The background color of the viewports are inconsistent in TMTV mode.

After

- The background color of the viewports are consistent in the TMTV mode.

https://github.com/user-attachments/assets/6b78e6ec-8dc3-4e3d-8e5b-ee7d7ff49c2b

### Testing

- Open the OHIF viewer with the study: https://viewer.ohif.org/tmtv?StudyInstanceUIDs=1.3.6.1.4.1.14519.5.2.1.297577087050970310787702792940607009472
- Launch the PET-CT study in the TMTV mode.
- Observe the background color of the fusion viewport.
- Apply color LUT for CT and PT in fusion viewports.
- Observe the fusion viewports.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [X] "OS: Ubuntu 24.04
- [X] "Node version: 22
- [X] "Browser: Chrome 147.0.7727.138
  

## Summary by CodeRabbit

* **New Features**
  * Added per-value opacity mapping for precise control over volume transparency based on individual data values.
  * Added support for explicit `null` threshold values to clear opacity thresholds.

* **Improvements**
  * Enhanced opacity state synchronization across viewport interactions and colormap updates to preserve both scalar and per-value opacity settings.
  * Improved threshold and opacity handling to work seamlessly with colormap applications.


sen-trenser:
@wayfarer3130 Could you please review this PR and provide yout feedbacks?
Thanks!

packages/core/src/RenderingEngine/BaseVolumeViewport.ts
```diff
@@ -1387,11 +1387,34 @@ abstract class BaseVolumeViewport extends Viewport {
...
-    matchedColormap.opacity = opacity;
+    if (
+      storedColormap?.opacity !== undefined &&
+      typeof storedColormap.opacity === 'number'
+    ) {
+      matchedColormap.opacity = getMaxOpacity(volumeActor);
+    } else if (storedColormap?.opacity !== undefined) {
+      matchedColormap.opacity = Array.isArray(storedColormap.opacity)
+        ? storedColormap.opacity.map((item) => ({ ...item }))
+        : storedColormap.opacity;
```
wayfarer3130: This is a bit odd to have another internal condition here - when do you get it NOT an array and not a number?

packages/core/src/RenderingEngine/BaseVolumeViewport.ts
```diff
@@ -1387,11 +1387,34 @@ abstract class BaseVolumeViewport extends Viewport {
...
+    // (e.g. [{value: 0, opacity: 0}, ...]), it was replaced by a single number from the actor,
+    // and TMTV and similar flows broke. The threshold slider (added in v3.11/3.12) needs the
+    // current value from the actor when the user moves the slider. So we branch: if the stored
+    // value is a number (slider case), read from the actor; otherwise use the stored value so
+    // hanging protocol arrays and explicit null are preserved.
+    const storedColormap = this.viewportProperties.colormap;
 
-    matchedColormap.threshold = threshold;
-    matchedColormap.opacity = opacity;
+    if (
```
wayfarer3130: Short-circuit the return if storedColormap isn't define here, since you access it multiple times below.
Then, extract opacity and threshold  from it and use those.

wayfarer3130: Done — short-circuit and return early when storedColormap is undefined (falling back to the actor-derived opacity/threshold, the original behavior), then destructure opacity and threshold once and use those below.

packages/core/src/RenderingEngine/BaseVolumeViewport.ts
wayfarer3130: Good catch — that branch was dead. `ColormapPublic.opacity` is typed `OpacityMapping[] | number`, so once it's defined and not a number it's always an array. Dropped the inner ternary and just map the array.

wayfarer3130 (approved):

wayfarer3130 (approved):

PR #2722, opened 2026-05-05
[codex] Fix image target id parsing

sedghi:
## Summary

Fix delimiter-unsafe parsing of stack image target IDs. Target IDs use the `imageId:<imageId>` format, but several call sites used `split('imageId:')[1]`, which truncates valid image IDs when the underlying URL or query string also contains `imageId:`.

This change adds a shared `BaseTool` helper that removes only the fixed prefix with `slice`, updates the runtime call sites that parsed image target IDs, and adds regression coverage for an image ID containing the delimiter.

## Impact

Valid stack image IDs containing the literal `imageId:` are preserved for cached stats, probe index calculation, magnification, and rectangle ROI threshold annotations. This avoids incorrect viewport/image lookup and stale or missing measurement recalculation for otherwise valid images.

## Validation

- Confirmed no remaining `split('imageId:')` call sites under `packages/tools/src` or `packages/core/src`.
- Could not run the Jest test suite in this local environment because the repo has no root `node_modules`, `yarn`/`npm`/`npx` are not on `PATH`, and the pre-commit hook also depends on `yarn`.
- Attempted Prettier through a temporary npm bootstrap; repo-configured Prettier could not load `@prettier/plugin-oxc` without the repo dependency tree installed.

### DIFF
--- a/packages/tools/src/tools/MagnifyTool.ts
+++ b/packages/tools/src/tools/MagnifyTool.ts
@@ -49,7 +49,7 @@ class MagnifyTool extends BaseTool {
     let referencedImageId;
 
     if (viewport instanceof StackViewport) {
-      referencedImageId = targetId.split('imageId:')[1];
+      referencedImageId = this.getImageIdFromTargetId(targetId);
     }
 
     return referencedImageId;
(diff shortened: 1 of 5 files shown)


jbocce (changes requested):
The code changes look fine. However, the PR is still in draft and there are code conflicts and failing checks.

Issue #2645, opened 2026-03-05
[Bug] PlanarFreeHandRoiTool: Statistics Text Not Displayed


Belbin-GK:
### Describe the Bug

After drawing an annotation using the PlanarFreeHandRoiTool, the statistics text (such as area, mean, etc.) is not displayed next to the annotation when switching from 2D MPR view to Frame view. The ROI is created and rendered correctly on the viewport, but the associated statistics label is missing.

This issue was raised as part of this [PR](https://github.com/cornerstonejs/cornerstone3D/pull/2419).

https://github.com/user-attachments/assets/0c740941-6413-4754-8b40-bd051636e64d

### Steps to Reproduce

1. **Draw** a measurement using the **PlanarFreeHandRoiTool** in the **2D MPR view**.
2. **Switch** the layout to **Frame view**.
3. In the right-side panel, locate and click the created measurement.
4. **Observe** the **annotation** displayed on the viewport.

### The current behavior

The measurement is displayed on the viewport without the statistics text (e.g., area, mean)

### The expected behavior

The measurement should be displayed on the viewport with the statistics text (e.g., area, mean, etc.) when the measurement is selected from the right-side panel after switching from 2D MPR view to Frame view.

### System Information

  **System**:
        OS: Linux 6.6 Ubuntu 22.04.5 LTS 22.04.5 LTS (Jammy Jellyfish)
        CPU: (12) x64 13th Gen Intel(R) Core(TM) i5-1334U
        Memory: 8.59 GB / 15.47 GB
        Container: Yes
        Shell: 5.1.16 - /bin/bash
  **Binaries**:
    Node: 22.20.0 
    Yarn: 1.22.22 
    npm: 10.9.3 
    pnpm: 10.17.1
  **Browsers**:
    Chrome : 145.0.7632.160

PR #2723, opened 2026-05-06
Fix: FreeHandRoiTool Statistics Text Not Displayed

nithin-trenser:
### Context
Fix issue: #2645 

After drawing an annotation using the PlanarFreeHandRoiTool, the statistics text (such as area, mean, etc.) is not displayed next to the annotation when switching from 2D MPR view to Frame view. The ROI is created and rendered correctly on the viewport, but the associated statistics label is missing.

### Changes & Results
Added an additional condition to ensure statistics are recalculated when cached stats for the current `targetId` are missing. The check now triggers when the annotation is invalidated or when the cached statistics for the specified target are not available, preventing missing measurement values.

https://github.com/user-attachments/assets/f2efe0b8-50a2-4d04-8d27-3335a01b7481

### Testing
- Draw a measurement using the PlanarFreeHandRoiTool in the 2D MPR view.
- Switch the layout to Frame view.
- In the right-side panel, locate and click the created measurement.
- Observe the annotation displayed on the viewport.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] OS: Ubuntu 24.04
- [x] Node version: 22
- [x] Browser: Chrome 147.0.7727.138
  

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved freehand ROI contour statistics accuracy by clamping computed bounds to valid image extents to prevent out-of-range coordinate effects.
* **Performance / Reliability**
  * Refined when ROI contour statistics are recalculated, using cached-stat availability and invalidation signals to avoid unnecessary recomputation while keeping results up to date.


sen-trenser:
@sedghi Could you please take a look at this PR and provide your feedback?
Thanks!

packages/tools/src/tools/annotation/PlanarFreehandROITool.ts
```diff
@@ -700,7 +700,10 @@ class PlanarFreehandROITool extends ContourSegmentationBaseTool {
       return;
     }
 
-    if (annotation.invalidated) {
+    const { data, invalidated } = annotation;
+    const cachedStats = data && data.cachedStats;
```
wayfarer3130: data?.cachedStats please

packages/tools/src/tools/annotation/PlanarFreehandROITool.ts
```diff
@@ -915,7 +919,12 @@ class PlanarFreehandROITool extends ContourSegmentationBaseTool {
     let kMax = Number.MIN_SAFE_INTEGER;
 
     for (let j = 0; j < points.length; j++) {
-      const worldPosIndex = indexPoints[j].map(Math.floor);
+      const worldPosIndex = indexPoints[j].map((v) => Math.floor(v + 1e-3));
```
wayfarer3130: This isn't clear why you are doing this, nor that it is valid for things like WSI images.

wayfarer3130 (changes requested):
Can you strip out the changes for worldPosIndex, or if they are really needed, then give a good explanation and make sure it really works reliably for data like WSI.  
If you need ceiling/floor differences, then calculating both ceiling and floor rather than trying to do some not understandable logic that might not work reliably.

wayfarer3130 (approved):

PR #2729, opened 2026-05-11
fix(segmentation): fully remove segmentations from viewport on delete after reload 

GhadeerAlbattarni:
### Context

Fixes an issue on OHIF https://github.com/OHIF/Viewers/issues/5706 

When a user loads a SEG series, deletes the segmentations, then loads SEG series again and tries to delete it a second time, the segmentation is not fully removed from the viewport. The segments stay visible on the currently displayed slice. 

Reloading SEG series more times makes the segments colors darker as stale actors pile up.

### Root Cause
Two bugs working together:

1. `removeSegmentation` never cleaned up `_stackLabelmapImageIdReferenceMap` or `_labelmapImageIdReferenceMap`. On reload, `_labelmapImageIdReferenceMap` accumulated both old and new labelmap frame IDs, causing `addLabelmapToElement` to register multiple actors for the same slice.
2. `removeLabelmapFromElement` only removed the **first** matching actor (`getLabelmapActorUID` → `filteredActors[0]`), leaving the rest on screen.

### Changes 

- **`removeSegmentation`**: clear both reference maps for the segmentation before removing it from state, so a reload always starts clean
- **`removeLabelmapFromElement`**: switch from `getLabelmapActorUID` (singular) to `getLabelmapActorEntries` (plural) and remove all matching actors

### Result
All segments are fully removed from the viewport when user clicks on delete on the second load. 

https://github.com/user-attachments/assets/37e5465c-9c63-42a6-8f0f-153afb2da1d7

### Testing

- Open a study in a segmentation mode (e.g., StudyInstanceUIDs=1.3.6.1.4.1.14519.5.2.1.256467663913010332776401703474716742458)
- Load SEG series, clicks yes to hydrate segmentations
- From left panel, clicks on `...`, then clicks Delete 
- Verify All segments removed
- Load the SEG series again, hydrate, and delete them again 
- Verify the colors of segments on the first slice stay same (not darker)
- Verify All segments removed from the viewport 

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] OS: macOS 10.15.4
- [x] Node version: v22.13.0

### DIFF
--- a/packages/tools/src/stateManagement/segmentation/SegmentationStateManager.ts
+++ b/packages/tools/src/stateManagement/segmentation/SegmentationStateManager.ts
@@ -64,6 +64,12 @@ export default class SegmentationStateManager {
   private _labelmapImageIdReferenceMap = new Map<string, string[]>();
 
   /**
+   * An index mapping each segmentationId to the set of keys it owns in
+   * _labelmapImageIdReferenceMap
+   */
+  private _labelmapKeysBySegmentationId = new Map<string, Set<string>>();
+
+  /**
    * Creates an instance of SegmentationStateManager.
    * @param {string} [uid] - Optional unique identifier for the manager.
    */
@@ -112,6 +118,7 @@ export default class SegmentationStateManager {
   resetState(): void {
     this._stackLabelmapImageIdReferenceMap.clear();
     this._labelmapImageIdReferenceMap.clear();
+    this._labelmapKeysBySegmentationId.clear();
     this.state = Object.freeze(
       csUtils.deepClone(initialDefaultState) as SegmentationState
     );
@@ -206,10 +213,13 @@ export default class SegmentationStateManager {
   removeSegmentation(segmentationId: string): void {
     // Clean up image ID reference maps to prevent stale entries accumulating on reload
     this._stackLabelmapImageIdReferenceMap.delete(segmentationId);
-    for (const key of this._labelmapImageIdReferenceMap.keys()) {
-      if (key.startsWith(`${segmentationId}-`)) {
+
+    const keys = this._labelmapKeysBySegmentationId.get(segmentationId);
+    if (keys) {
+      for (const key of keys) {
         this._labelmapImageIdReferenceMap.delete(key);
       }
+      this._labelmapKeysBySegmentationId.delete(segmentationId);
     }
 
     this.updateState((state) => {
@@ -888,12 +898,20 @@ export default class SegmentationStateManager {
 
     if (!this._labelmapImageIdReferenceMap.has(key)) {
       this._labelmapImageIdReferenceMap.set(key, [labelmapImageId]);
-      return;
+    } else {
+      const currentValues = this._labelmapImageIdReferenceMap.get(key);
+      const newValues = Array.from(
+        new Set([...currentValues, labelmapImageId])
+      );
+      this._labelmapImageIdReferenceMap.set(key, newValues);
     }
 
-    const currentValues = this._labelmapImageIdReferenceMap.get(key);
-    const newValues = Array.from(new Set([...currentValues, labelmapImageId]));
-    this._labelmapImageIdReferenceMap.set(key, newValues);
+    let keys = this._labelmapKeysBySegmentationId.get(segmentationId);
+    if (!keys) {
+      keys = new Set();
+      this._labelmapKeysBySegmentationId.set(segmentationId, keys);
+    }
+    keys.add(key);
   }
 
   _setActiveSegmentation(
--- a/packages/tools/src/stateManagement/segmentation/SegmentationStateManager.ts
+++ b/packages/tools/src/stateManagement/segmentation/SegmentationStateManager.ts
@@ -204,6 +204,14 @@ export default class SegmentationStateManager {
    * @param {string} segmentationId - The ID of the segmentation to remove.
    */
   removeSegmentation(segmentationId: string): void {
+    // Clean up image ID reference maps to prevent stale entries accumulating on reload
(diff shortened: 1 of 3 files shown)


jbocce:
@claude review

jbocce:
@claude review

packages/tools/src/stateManagement/segmentation/SegmentationStateManager.ts
```diff
@@ -204,6 +204,14 @@ export default class SegmentationStateManager {
    * @param {string} segmentationId - The ID of the segmentation to remove.
    */
   removeSegmentation(segmentationId: string): void {
+    // Clean up image ID reference maps to prevent stale entries accumulating on reload
+    this._stackLabelmapImageIdReferenceMap.delete(segmentationId);
+    for (const key of this._labelmapImageIdReferenceMap.keys()) {
+      if (key.startsWith(`${segmentationId}-`)) {
```
jbocce: This prefix match seems a bit risky. Under what circumstances does it occur?

jbocce (changes requested):
The changes look good in general. I am concerned with that prefix matching when removing segmentations. 

Also, it might be worth investigating contours and seeing if they also have similar, incorrect behaviour.

packages/tools/src/stateManagement/segmentation/SegmentationStateManager.ts
GhadeerAlbattarni: In the same file, the key for `_labelmapImageIdReferenceMap` is generated by `_generateMapKey` which does: 
  ```
private _generateMapKey({ segmentationId, referenceImageId }) {
    return `${segmentationId}-${referenceImageId}`;
  }
```

 `_stackLabelmapImageIdReferenceMap` is a `Map<segmentationId, Map<...>>`, so a single `.delete(segmentationId)` removes everything for that segmentation

`_labelmapImageIdReferenceMap` is `Map<string, string[]>` and the `segmentationId` is embedded in each key (by generateMapKey). so we have to iterate and delete all keys that start with `${segmentationId}-`

Comment (jbocce) [reply packages/tools/src/stateManagement/segmentation/SegmentationStateManager.ts]: Yes, but consider two segmentation ids: `a` and `a-b`. When deleting all the keys for `a` we will unfortunately also delete all the keys for `a-b`.

Here is one suggestion from Claude...

Keep the existing flat Map<string, string[]> and the existing _generateMapKey. Just maintain a secondary index so deletion is exact and O(1) in segmentation count.
```
  private _labelmapImageIdReferenceMap = new Map<string, string[]>();
  // New:
  private _labelmapKeysBySegmentationId = new Map<string, Set<string>>();
```
  In _updateLabelmapImageIdReferenceMap (SegmentationStateManager.ts:882):
```
  _updateLabelmapImageIdReferenceMap({
    segmentationId,
    referenceImageId,
    labelmapImageId,
  }) {
    const key = this._generateMapKey({ segmentationId, referenceImageId });

    if (!this._labelmapImageIdReferenceMap.has(key)) {
      this._labelmapImageIdReferenceMap.set(key, [labelmapImageId]);
    } else {
      const currentValues = this._labelmapImageIdReferenceMap.get(key);
      const newValues = Array.from(new Set([...currentValues, labelmapImageId]));
      this._labelmapImageIdReferenceMap.set(key, newValues);
    }

    let keys = this._labelmapKeysBySegmentationId.get(segmentationId);
    if (!keys) {
      keys = new Set();
      this._labelmapKeysBySegmentationId.set(segmentationId, keys);
    }
    keys.add(key);
  }
```
  Then in removeSegmentation (SegmentationStateManager.ts:206):
```
  removeSegmentation(segmentationId: string): void {
    this._stackLabelmapImageIdReferenceMap.delete(segmentationId);

    const keys = this._labelmapKeysBySegmentationId.get(segmentationId);
    if (keys) {
      for (const key of keys) {
        this._labelmapImageIdReferenceMap.delete(key);
      }
      this._labelmapKeysBySegmentationId.delete(segmentationId);
    }

    // ...existing updateState() / triggerSegmentationRemoved() unchanged
  }
```
  Don't forget to clear it in resetState (SegmentationStateManager.ts:112):
```
  this._labelmapKeysBySegmentationId.clear();
```

**Comment by GhadeerAlbattarni** [comment]
@jbocce, I tested RTSRTUCT deletion locally and it seems that it doesn't have a similar issue (local testing) 
However, in the code there's a function `removeContourFromElement` similar to `removeLabelmapFromElement` but it never called inside `contourDisplay.removeRepresentation`.

File: packages/tools/src/stateManagement/segmentation/SegmentationStateManager.ts (hunk 1 above)
Comment (GhadeerAlbattarni) [reply packages/tools/src/stateManagement/segmentation/SegmentationStateManager.ts]: Sounds good. Thanks for your help with this!

File: packages/tools/src/tools/displayTools/Contour/contourDisplay.ts [hunk 2]
```diff
@@ -38,6 +38,13 @@ function removeRepresentation(
     return;
   }
 
+  const viewportProcessed = processedViewportSegmentations.get(viewportId);
+  if (viewportProcessed) {
+    viewportProcessed.delete(segmentationId);
+  }
+
+  removeContourFromElement(viewportId, segmentationId);
+
```
GhadeerAlbattarni: @jbocce This is a proposed fix for contour path, it will be similar to `removeRepresentation` in `labelmapDisplay.ts`

`removeContourFromElement` is imported is this file but never used.

jbocce (approved):
Changes look great. Thanks for this.

PR #2742, opened 2026-05-25
fix: clipper2 intersection definition

wayfarer3130:
### Context

Intersections with holes are not working consistently.   This PR builds on top of changes from:
https://github.com/cornerstonejs/cornerstone3D/pull/2739
to see if we can figure out a reasonable solution for this.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  

## Release Notes

* **New Features**
  * Stroke-based contour editing with add and remove operations
  * Support for complex contours with holes
  * Enhanced polygon boolean operations (Union, Difference, Intersection, XOR)

* **Documentation**
  * Added detailed behavioral specifications for contour editing and topology handling

* **Tests**
  * Added comprehensive test coverage for boolean operations and contour interactions

* **Chores**
  * Added geometry processing library dependency


wayfarer3130:
@dan-rukas and @sedghi 
CC: @GhadeerAlbattarni and @jbocce 

I was looking at Ghadeer's changes for union/intersection and running into some usability and performance issues with them, and I realized that the area isn't sufficiently well defined to really know what all we are doing, so with clod, I wrote a semantic behaviour document around the contours:
https://github.com/cornerstonejs/cornerstone3D/blob/da64bcc24d955c9cf1dae5c6c505db498f8e3e41/packages/tools/src/utilities/contourSegmentation/SEMANTICS.md

As well, I had it change to use clipper2-ts instead of a custom implementation to try to address some of the performance concerns.  

I think the semantics changes are mostly reasonable - it basically says that only the source/basis contour gets affected, and only when there is an overlap to start with.    I'm not completely comfortable with the definition on that yet, since I think that a union with a contour that is entirely within a hole SHOULD get included.  I'm going to update that.

There is also some question about the user interface in OHIF - I think the union/intersect/difference should be three buttons with a cancel operation, and the whole dialog should have a help button explaining that it applies to the CURRENT view only, and there should be a progress dialog/indication that something has been applied.

Then, the current self-intersect behaviour is very confusion - I have updated the definitions for that to be derived from the union and intersection definitions, and added specific add/remove behaviour that is independent of holes/location of start.

jbocce:
> @dan-rukas and @sedghi CC: @GhadeerAlbattarni and @jbocce
> 
> I was looking at Ghadeer's changes for union/intersection and running into some usability and performance issues with them, and I realized that the area isn't sufficiently well defined to really know what all we are doing, so with clod, I wrote a semantic behaviour document around the contours: https://github.com/cornerstonejs/cornerstone3D/blob/da64bcc24d955c9cf1dae5c6c505db498f8e3e41/packages/tools/src/utilities/contourSegmentation/SEMANTICS.md
> 
> As well, I had it change to use clipper2-ts instead of a custom implementation to try to address some of the performance concerns.
> 
> I think the semantics changes are mostly reasonable - it basically says that only the source/basis contour gets affected, and only when there is an overlap to start with. I'm not completely comfortable with the definition on that yet, since I think that a union with a contour that is entirely within a hole SHOULD get included. I'm going to update that.
> 
> There is also some question about the user interface in OHIF - I think the union/intersect/difference should be three buttons with a cancel operation, and the whole dialog should have a help button explaining that it applies to the CURRENT view only, and there should be a progress dialog/indication that something has been applied.
> 
> Then, the current self-intersect behaviour is very confusion - I have updated the definitions for that to be derived from the union and intersection definitions, and added specific add/remove behaviour that is independent of holes/location of start.

@dan-rukas, @sedghi and @wayfarer3130 
CC: @GhadeerAlbattarni

Thanks for this info @wayfarer3130. It sounds to me then that we should discuss this further, but it very much sounds like for 3.13 we should stay the course as far as contour logical operations are concerned and revisit this in the next release.

packages/tools/src/utilities/contourSegmentation/SEMANTICS.md
```diff
@@ -0,0 +1,792 @@
+# Contour Segmentation Operation Semantics
```
sedghi, line 1: should we add it to the packages/docs. we don't usually carry .md with code

packages/tools/package.json
```diff
@@ -104,6 +104,7 @@
   },
   "dependencies": {
     "@types/offscreencanvas": "2019.7.3",
+    "clipper2-ts": "2.0.1",
```
sedghi, line 107: Do we really need this library? Can't you just ask Claude to implement the specific things we need? The BSL-1.0 license is weird, so we can't really pull it into our MIT-licensed codebase.

I highly doubt Claude would have any trouble writing the utilities we actually need from this dependency, which would save us from having to carry and maintain it.

wayfarer3130: Good catch to flag the license, but I think this is a name collision. clipper2-ts is under BSL-1.0 = the Boost Software License 1.0, not the Business Source License (BUSL-1.1) — the two get confused all the time because of the shared "BSL".

The Boost license is OSI-approved, FSF-certified GPL-compatible, and actually more permissive than MIT (it's basically MIT minus the requirement to reproduce the notice in binary distributions). It's fully compatible with our MIT codebase — the only obligation is keeping the license notice in source distributions, which the node_modules copy already satisfies.

On reimplementing it ourselves: what's behind that import isn't a couple of helpers — it's a full polygon boolean engine (union/difference/intersection/xor) with hole-topology resolution and self-intersection decomposition. Clipper2 is a ~20-year-matured Vatti-algorithm implementation, and the hard part is the degenerate cases (collinear edges, coincident/near-touching vertices, numerical precision) that hand-rolled clippers consistently get wrong. I'd rather not own every geometry bug in that surface forever.

Worth noting the dependency is already cleanly contained: only clipperBooleanOps.ts imports it, exposing just applyBoolean and splitSelfIntersections to the rest of the codebase — so the seam is in the right place if we ever do want to swap it out.

So my suggestion is to keep it, on the basis that the license concern was based on the wrong "BSL".

sedghi: ok sounds good

sedghi (approved):

Issue #2630, opened 2026-02-24
[Bug] Area of Closed Freehand ROI is incorrect


leoqd:
### Describe the Bug

Areas of RectangleROI and PlanarFreehandROI tools doesn´t match.

<img width="2560" height="1227" alt="Image" src="https://github.com/user-attachments/assets/395fcb63-d614-4ac3-8d56-720ee804cfc7" />

### Steps to Reproduce

1. Go to Study [https://viewer.ohif.org/viewer?StudyInstanceUIDs=1.3.6.1.4.1.14519.5.2.1.4792.2001.232252967813565730694525674696](url)
2. Draw a rectangle using RectangleROI tool.
3. Draw a PlanarFreehandROI annotation trying to follow the rectangle borders.

### The current behavior

When drawing a RectangleROI and a PlanarFreehandROI over the same region (manually following the rectangle borders), the calculated area values differ significantly.

In the example shown:

- RectangleROI reports ~1.14 mm²
- PlanarFreehandROI reports ~0.000428 mm²

The difference is several orders of magnitude, suggesting an inconsistency in area calculation between the two tools.

### The expected behavior

For equivalent regions, both RectangleROI and PlanarFreehandROI should produce comparable area measurements (within a small tolerance due to manual drawing differences).

### System Information

  System:
    OS: Windows 11 10.0.22631
    CPU: (14) x64 Intel(R) Core(TM) Ultra 7 155U
    Memory: 852.38 MB / 15.51 GB
  Binaries:
    Node: 22.13.1
    Yarn: 1.22.22
    npm: 10.9.2
  Browsers:
    Chrome: 145.0.7632.76
    Edge: Chromium (140.0.3485.54)
    Internet Explorer: 11.0.22621.3527

nithin-trenser:
Identified an inconsistency in the ellipse area calculation. it is currently producing incorrect results.


<img width="1918" height="942" alt="Image" src="https://github.com/user-attachments/assets/b9dc07f6-4ed1-4b6f-8a02-5255d947797c" />

Area Calculation: Ellipse

Major Axis: $94.2$ mm

Minor Axis: $29.2$ mm

Steps:

- Semi-major axis ($a$): $94.2 / 2 = 47.1$ mm
- Semi-minor axis ($b$): $29.2 / 2 = 14.6$ mm
- Area Formula: $A = \pi \times a \times b$
[...]

PR #2619, opened 2026-02-18
feat(PlanarFreeHandRoiTool) : implementing PlanarFreeHandRoiTool on oblique plane

Celian-abd:
### Context
Right now, if we try to draw an annotation using the PlanarFreeHandRoiTool on a non orthogonal plane, it will show the following error : 'No support yet for oblique plane planar contours'.

### Changes & Results
- Changed getSubPixelSpacingAndXYDirections.ts to add the oblique case.

### Testing
- Use the crosshair to rotate
- Draw an annotation on a non orthogonal viewport

#### PR

- [] My Pull Request title is descriptive, accurate and follows the
  semantic-release format and guidelines.

#### Code

- [] My code has been well-documented (function documentation, inline comments,
  etc.)

#### Public Documentation Updates

- [] The documentation page has been updated as necessary for any public API
  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  

[blog]: https://circleci.com/blog/triggering-trusted-ci-jobs-on-untrusted-forks/
[script]: https://github.com/jklukas/git-push-fork-to-upstream-branch

Issue #2136, opened 2025-06-12
[Bug] PlanarFreehandROITool[max,min,mean] It cannot correctly display the maximum and minimum values ​​in some MR images.


Steven-DD:
### Describe the Bug

As the title says, it cannot calculate the maximum and minimum values ​​in some MR type images, but if you use the rectangle tool or the circle tool, you can see the maximum and minimum values.

### Steps to Reproduce

Select the tool and use

### The current behavior

1.You can see that there is no correct maximum and minimum value when using it. The perimeter is because of the previous bug, so I blocked it.

2.Another problem is that if I drag its text box, it will also trigger a calculation, which should be unnecessary, because frequent calculations of values ​​that have not changed consume performance.

![Image](https://github.com/user-attachments/assets/c7b28cce-6ac8-457c-98c6-a290dc372b6f)

![Image](https://github.com/user-attachments/assets/b33bf68f-90a1-4f18-adeb-7fd69e6c42ca)

### The expected behavior

I hope to correctly obtain the maximum and minimum values ​​of the image.

### OS

win11 24H2

### Node version

16.20.2

### Browser

e.g 137.0.3296.68 (x64)

sedghi:
if the issue happens in some data, we need the anonymized data to investigate

PR #2744, opened 2026-05-26
fix (tools): freehand roi in volume viewport for oblique data

### LINKED ISSUE #976: [Bug] PlanarFreehand ROI does not calculating density values for some DICOM data
### Describe the Bug

We used PlanarFreehand ROI tool and drawn annotation for some DICOM data with different orientations, **density values are not calculated**. Displayed the density values `Area` and `Max` and not displayed the `Mean`, `Std Dev` etc values. In this `Max` density value is shown like `-Infinity`.

When calculating density values(`Mean`, `Std Dev`, `Max` etc), the logic uses canvas points from image world position without considering the image orientation.

Issue can be reproducible in different scenarios
1. After rotating(manual or 90 degree rotation) a image.
2. After flipping(horizontally or vertically) a image.
3. For some DICOM data which have a different orientation and user is not rotated and not flipped, where world coordinates needs to be flipped horizontally to display in canvas by default.
In the above cases  density values are not calculating.

This seems to be coming from  `_calculateCachedStats` method inside PlanarFreehandROITool. Here the points are passed in the order of world coordinates, but the processing logic assumes the points to be in incrementing X coordinates in canvas.

Since the logic assumes the points to be in incrementing X coordinates in canvas, it tries to cache intersections with annotation for each row. When the image is rotated 90/270 degree, the x value changes with each pixel and **causes a huge performance impact** by re-calculating all intersections of the row for all pixels within the bounding box!.

Please find the attached video revealing both performance and stats value calculating issue.

https://github.com/cornerstonejs/cornerstone3D/assets/62233901/8ff848c2-2577-4128-b07f-3fe015acc70b

### Steps to Reproduce

1. Load some DICOM data with different orientation.
2. For some data no need to rotate, For some data rotate or flip.
3. Draw annotation using PlanarFreehandROI tool.
4. Click on toggle calculate stats button.
5. stats values(`Mean`, `Std Dev`, etc) are not calculated.

**Steps to Reproduce Performance Issue**
  1. Rotate the data in 90 degree(https://www.cornerstonejs.org/live-examples/stackannotationtools). 
  2. Draw a large annotation using planarfreehand roi.
  3. Performance issue will get.
  
**Steps to Reproduce Stats value not calculated**
  1. Rotate the data in 90 degree(https://www.cornerstonejs.org/live-examples/stackannotationtools). 
  2. Flip horizontally.
  3. Draw a large annotation using planarfreehand roi.
  4. Stats values are not calculated(`Mean`, `Std Dev` etc).
  
### The current behavior

When annotation is drawn using PlanarFreehandROI for some DICOM data with different orientations, and calculating the stats, the stats are not calculated(`Mean`, `Std Dev` etc), and shows `area` only

### The expected behavior

When annotation is drawn using PlanarFreehandROI for all data(with different orientation) and calculate the stats, it should correctly calculate stats values  irrespective of image or viewport orientation.

### OS

Windows 10 Pro

### Node version

16.17.0

### Browser

120.0.6099.130

### LINKED ISSUE #2386: [Bug] Closed freehand max show as "-Infinity" on oblique data

### LINKED ISSUE #2646: [Bug] Closed Freehand ROI stat value changes when switching from Stack Viewport to Volume Viewport

### LINKED ISSUE #5889: [Bug] Freehand ROI not renders in Volume Viewport for Oblique data
### Describe the Bug

This issue occurs only with oblique datasets. When attempting to draw an annotation using the Freehand ROI tool in volume viewports (MPR layout), the drawing fails. After this failure occurs, other annotation tools also become unresponsive.

https://github.com/user-attachments/assets/121f4d58-e6cf-41b6-89da-d788ff016cb7

### Steps to Reproduce

1. Launch the study https://viewer-dev.ohif.org/viewer?StudyInstanceUIDs=1.3.6.1.4.1.14519.5.2.1.9823.1001.134394060407147891170882809392
2. Switch to 2D MPR layout
3. Draw freehand ROI 

### The current behavior

Rendering fails with below issue. 

```
TypeError: Cannot read properties of undefined (reading 'annotation')
    at D.renderAnnotationInstance (2075.bundle.2f9f4cb00dc13e7af727.js:1:14225)
    at D.renderAnnotation (4287.bundle.29d74b91f8ac8564d753.js:1:33882)
    at https://viewer-dev.ohif.org/9205.bundle.5cff9ac7c81278237d3a.js:1:59858
    at Array.forEach (<anonymous>)
    at https://viewer-dev.ohif.org/9205.bundle.5cff9ac7c81278237d3a.js:1:59813
    at u (2075.bundle.2f9f4cb00dc13e7af727.js:1:1267)
    at Object._triggerRender (9205.bundle.5cff9ac7c81278237d3a.js:1:59795)
    at _renderFlaggedViewports (9205.bundle.5cff9ac7c81278237d3a.js:1:58420)
```

### The expected behavior

Annotation should render properly

### System Information

System:
Memory: 4.29 GB / 7.63 GB
Container: Yes
Shell: 5.1.16 - /bin/bash
Binaries:
Node: 22.19.0 - ~/.nvm/versions/node/v22.19.0/bin/node
Yarn: 1.22.22 - ~/.nvm/versions/node/v22.19.0/bin/yarn
npm: 11.6.0 - ~/.nvm/versions/node/v22.19.0/bin/npm

arul-trenser:
> [!NOTE]
> **Maintainer add-on work, added after the original contribution.** @arul-trenser wrote the
> description below, and that description gives the bug and the fix as first submitted.
> @wayfarer3130 added the commits `f2331654c` and `78625853b` on top afterwards. The first
> commit reworks the sampling onto the shared voxel slab iterator, and the second commit is a
> cleanup and defect pass. Neither commit is the work of the contributor, and the description
> below covers neither commit.
>
> A third commit, `1481b801a`, merges `main` and moves the code to the API names that
> #2893 introduced.
>
> - Rework onto the slab iterator: https://github.com/cornerstonejs/cornerstone3D/pull/2744#issuecomment-5512361270
> - Follow-up cleanup and defects: https://github.com/cornerstonejs/cornerstone3D/pull/2744#issuecomment-5515237263
> - Merge and rename: see "Maintainer add-on" below.

> [!IMPORTANT]
> GitHub closed this pull request automatically on 10 September, because #2893 merged and
> the `feat/roi-voxel-iterator` base branch was deleted. No person rejected the work. The
> base is now `main`, and the pull request is open again.

### Context
This PR resolves an issue where the freehand ROI tool was not functioning correctly within volume viewports displaying oblique data. The PR is incorporated by [FlyWheel.io](https://flywheel.io/).

- Fixes: https://github.com/OHIF/Viewers/issues/5889

### Changes & Results
⚠️⚠️ This change adapts the implementation from [cornerstonejs/cornerstone3D#2619](https://github.com/cornerstonejs/cornerstone3D/pull/2619) with minor modifications:

- Removed _isObliqueView check: Eliminated the restrictive conditional check for oblique views.
- Standardized Stats Calculation: Modified the logic to consistently calculate ROI statistics using the robust approach previously reserved only for oblique views, ensuring consistency across viewports.

Before:

https://github.com/user-attachments/assets/b0eab58c-a22b-42c9-aaef-c926ddad1e91

After:

https://github.com/user-attachments/assets/664a9eb7-0ece-44be-840e-9aa8e4eaf157

---

## Maintainer add-on: the shared voxel iterator

This section covers the commits that @wayfarer3130 added. This section does not cover the
contribution of @arul-trenser above.

### What the tool does now

The tool no longer walks a bounding box and tests each voxel in canvas coordinates. The tool
builds a shape, and the shared iterator walks the voxels:

```ts
const shape = createPolylineShape({
  volume,           // { dimensions, direction, spacing, origin }
  planePoint,       // points[0], the annotation plane anchor
  viewPlaneNormal,  // unit length
  polyline: points, // the outline in world coordinates
});

return sampleVoxelsInShape({
  volume,
  planePoint,
  viewPlaneNormal,
  referencePlaneThickness,
  bounds: this.getContourIndexBounds(...),
  getShapeRuns: shape.getRuns,
  voxelManager,
  onSample: this.configuration.statsCalculator.statsCallback,
  storePointData: this.configuration.storePointData,
});
```

`createPolylineShape` supplies the in-plane test and the exact runs. `sampleVoxelsInShape`
reads a value for each voxel, and it calls the statistics accumulator. Neither function reads
a viewport. The same annotation over the same volume therefore selects the same voxels at any
zoom, any pan, any canvas size and any slab thickness.

A tool with a different outline replaces `createPolylineShape` with `createEllipseShape` or
`createRectangleShape`, and the tool changes nothing else. Circle ROI, Elliptical ROI and
Rectangle ROI can each move to this loop, and none of the three must copy the loop again.

Rule M and Rule D define the voxel set. Issue #2889 states both rules, and
`packages/docs/docs/concepts/cornerstone-tools/annotation/voxel-statistics.md` explains them.

### What this pull request adds to core

- `csUtils.voxelSlab.sampleVoxelsInShape` -- the accumulation half of an area measurement.
  The function reads the value of each voxel, it copies the arrays that the iterator reuses,
  it calls `onSample`, and it stores the samples only when the caller asks for them.
- `csUtils.getEffectiveSpacingAlongDirection` -- the formula of @arul-trenser, moved to core
  beside `getSpacingInNormalDirection`. Its documentation states how the three measures of
  "spacing along a direction" differ, so that no fourth measure appears.

### What this pull request removes

- `getIntersectionIterator` and `getScanlineIntersections`, with their test file. Nothing in
  `src` calls either function after the rework, and both duplicate
  `getLineSegmentIntersectionsCoordinates` and `createPolylineShape.getRuns`.
  `math/polyline/index.ts` is now byte-identical with `main`.
- `snapIndexBounds`, because the bounding box snapping that this pull request deletes was its
  only consumer. The removal makes `utilities.boundingBox` smaller.

### Three defects that the review found

- **The index bounds dropped voxels.** A fixed dilation of one voxel was too small, because
  the slab reaches `(T + T_v) / 2` mm along the normal. The loss was 4.5% of the voxels on a
  0.7 x 0.7 x 5 mm oblique contour, and 20% at `T` = 20 mm. The dilation is now
  `ceil(halfWidth * |a_i . n| / s_i) + 1` for each axis.
- **`imageData.getOrigin()` threw on a CPU-rendered stack.** `StackViewport.getImageDataCPU`
  builds an inner `imageData` that has no `getOrigin`. Every closed freehand ROI on a CPU
  fallback viewport threw inside the render loop. The geometry now comes from the enclosing
  `IImageData`, where the GPU path and the CPU path both carry the four fields as plain
  properties.
- **A contour of fewer than three points threw.** The guard tested the length only. A bad
  hydration, or a click that registered one point, escaped an exception into the render loop,
  and it left the statistics accumulator half filled. The guard now requires three points.

### One sampler for every area annotation tool

`fc0ad122c` and `7fc9e9431` move Circle ROI, Rectangle ROI and Elliptical ROI onto the same
path that Planar Freehand ROI uses. All four area annotation tools share it now.
`utilities/sampleAreaAnnotationVoxels` holds the whole of it: the plane, the thickness, the
index bounds and the accumulation. Each tool supplies its shape, and nothing else:

```ts
// Planar Freehand ROI
createShape: ({ volume, planePoint, viewPlaneNormal }) =>
  createPolylineShape({ volume, planePoint, viewPlaneNormal, polyline: points }),

// Circle ROI
createShape: ({ volume, planePoint, viewPlaneNormal }) =>
  createCircleShape({ volume, planePoint, viewPlaneNormal,
    centerWorld: points[0], radius: vec3.distance(points[0], points[1]) }),

// Rectangle ROI
createShape: ({ volume, planePoint, viewPlaneNormal }) =>
  createRectangleShape({ volume, planePoint, viewPlaneNormal, centerWorld,
    majorAxis, majorHalfLength, minorHalfLength }),

// Elliptical ROI
createShape: ({ volume, planePoint, viewPlaneNormal }) =>
  createEllipseShape({ volume, planePoint, viewPlaneNormal, centerWorld,
    majorAxis, majorRadius, minorRadius }),
```

Three tools change behaviour, and each change corrects a defect:

- **Rectangle ROI applied no shape test at all.** The tool walked the index bounding box of
  the two corner handles, and it counted every voxel in that box. The traversal is exact
  only while the rectangle stays aligned with the index axes. A rotated rectangle, or any
  rectangle on an oblique plane, over-counted its corners.
- **Circle ROI tested an axis-aligned ellipsoid in world coordinates.** A circle that a user
  draws on an oblique plane is a circle in that plane, and an ellipsoid aligned with the
  world axes does not describe it.
- **Elliptical ROI tested the same axis-aligned ellipsoid**, and it derived that ellipsoid
  from canvas corners, so the voxel set moved with the zoom.

`e22f2bd91` also adds the three tools to the dropdown of the PET-CT example, which is where
the testing notes below start. That example already had the crosshairs, and the crosshairs
are how a user reaches an oblique plane.

All three tools also stop reading the display. The voxel set is a property of the annotation
and of the data now, so the mean, the maximum and the count hold still under a zoom and a
pan.

The sampler returns an empty result instead of an exception in two cases: a shape of too few
points, and a `createShape` that reports a degenerate shape, such as a rectangle of no width.
An exception inside the render loop stops the whole viewport.

Each tool also loses the code that the change leaves dead: the index bounding box, the
world-axis ellipsoid, and the imports that only those used.

### The merge, and the renames

`1481b801a` merges `main`. #2893 merged as a squash, so git shares no history with the
commits that this branch was built on, and every file of the voxel slab work conflicted.
`main` holds the reviewed form of that work, so `main` wins for all of it, and this branch
keeps only its own additions.

The API names of #2893 reach this branch:

| before | after |
| --- | --- |
| `sampleVoxelsInSlab` | `sampleVoxelsInShape` |
| `iterateVoxelsInSlab`, `collectVoxelsInSlab` | `iterateVoxelsInShape`, `collectVoxelsInShape` |
| `createContourShape` | `createPolylineShape` |
| `annotationThickness` | `referencePlaneThickness` |
| `normal` | `viewPlaneNormal` |

One behaviour changes with the renames. The polyline shape is planar now, the shape takes no
`depth`, and `getRequiredThickness()` always returns 0. `T` therefore reaches the slab from
the tool. The tool resolves `T` once with `resolveReferencePlaneThickness`, and the tool gives
the same value to the iterator and to the index bounds. The older code gave the unresolved
`getRequiredThickness()` to `getMembershipHalfWidth`, which produced a half width of
`T_v / 2` when the annotation recorded no thickness, and which clipped the bounds by half a
voxel.

### Issues that this pull request closes

Each issue below reports a defect in the canvas-space sampler that this pull request deletes.
The reporter of each issue named the cause, and the cause is gone.

Fixes #976
Fixes #2386
Fixes #2646

- **#976 — PlanarFreehand ROI does not calculate density values for some DICOM data.** The
  reporter identified the cause exactly: `_calculateCachedStats` expects the polyline points
  in order of increasing canvas X, and the row cache of the intersections
  (`curRow`, `intersections`, `intersectionCounter`) depends on that order. A rotation, a
  flip, or an image orientation that the viewport mirrors breaks the order. The tool then
  reports `-Infinity` for the maximum, and it reports no mean and no standard deviation. The
  same issue reports a large loss of speed at 90° and at 270°, because the tool calculates the
  intersections of a row again for each voxel of the row. This pull request deletes the row
  cache and the canvas test. The tool selects voxels by exact runs in index space, and the
  selection does not read the canvas. Both the wrong values and the loss of speed go.
- **#2386 — Closed freehand maximum shows `-Infinity` on oblique data.** This issue is the
  oblique form of #976, and the same deletion corrects it. Rule M keeps at least one layer
  whenever `T >= T_v` (invariant I4 of #2889), so a contour over the volume always selects
  voxels.
- **#2646 — Closed freehand ROI statistics change between a stack viewport and a volume
  viewport.** The reporter identified the cause: `indexPoints[j].map(Math.floor)` floors the
  Z index, so an annotation on a plane that the conversion returns as `k = 33.9999` measures
  slice 33. This pull request deletes that block, together with `snapIndexBounds`, the
  dilation of 1%, and the `Math.floor` and `Math.ceil` calls on the bounds. Rule M is a
  predicate in world coordinates, anchored on the plane of the annotation, and the `T_v`
  dilation covers a plane that misses the voxel centres. The stack target and the volume
  target therefore select the same voxels.

The description above already links OHIF/Viewers#5889, which the change to
`getSubPixelSpacingAndXYDirections` corrects: the function threw
`No support yet for oblique plane planar contours`, and the exception stopped the draw loop.
GitHub closes a linked issue only in the same repository, so that link closes nothing. The
OHIF issue is closed already.

### Issues that this pull request does not close

- **#2889 — the parent feature request.** This pull request moves **one** tool onto the
  iterator. `CircleROITool`, `EllipticalROITool`, `RectangleROITool` and the StartEndThreshold
  tools each keep their own traversal, and each one is still wrong in an oblique view.
  `CircleROITool` and `EllipticalROITool` build an ellipsoid that the **world** axes align,
  from the bounding box of two corners, so an oblique view tests a solid that is neither the
  drawn shape nor a shape on the annotation plane. `RectangleROITool` passes no `isInObject`
  at all, so the tool measures the bounding box in index space and not the rectangle. #2889
  stays open until all of the tools use the iterator.
- **#2645 — the statistics text does not appear after a change from the 2D MPR layout to the
  frame layout.** This defect is in the lifecycle of the display and not in the selection of
  the voxels. `_calculateStatsIfActive` decides when the tool calculates the statistics again,
  and this pull request changes no line of that method. The pull request changes which voxels
  the tool measures, and it does not change when the tool measures them.
- **#2630 — the area of a closed freehand ROI is incorrect.** The tool calculates the area as
  `polyline.getArea(canvasCoordinates) * deltaInX * deltaInY`, and this pull request does not
  change that expression. The report also uses an axial series in a stack viewport, which
  never reaches the oblique branch that this pull request adds. The follow-up comment on the
  issue reports a wrong area for `EllipticalROITool`, which is a different tool that this pull
  request does not touch.
- **#2136 — the freehand tool shows no maximum and no minimum on some MR images.** The first
  half of this report matches the cause of #976, so this pull request probably corrects that
  half. The report names no dataset, so no person can confirm the correction. The report also
  states a second problem: a drag of the text box starts a calculation again. That second
  problem is in the same lifecycle as #2645, and this pull request does not address it. The
  issue therefore stays open until someone confirms the first half and resolves the second.

---

#### The original report

- Launch the study https://viewer-dev.ohif.org/viewer?StudyInstanceUIDs=1.3.6.1.4.1.14519.5.2.1.9823.1001.134394060407147891170882809392
- Switch to 2D MPR layout
- Draw freehand ROI

#### Automated tests

Run `npx jest` from the root of the repository. Two suites cover this work directly:

- `packages/tools/test/planarFreehandVoxelSampling.jest.js` -- 7 tests over the tool. The
  suite covers one layer for a thin annotation, two thin layers for one thick annotation
  slice, the same voxels at two voxel sizes along the normal, voxel centres in world
  coordinates, statistics without stored point data, a degenerate contour, and a volume that
  carries no voxels for the contour.
- `packages/core/test/voxelsInShapeSampling.jest.js` -- 5 tests over `sampleVoxelsInShape`.
  The suite covers a value for each visited voxel, `onSample` with stored point data and
  without it, a volume that holds no value for a voxel, copies instead of the arrays that the
  iterator reuses, and a missing voxel manager.
- `packages/tools/test/CircleROIVoxelSampling.jest.js` -- 2 tests over Circle ROI. The suite
  pins the disc of voxel centres, and it pins the same selection at two canvas scales.
- `packages/tools/test/RectangleROIArea.jest.js` -- 3 tests over Rectangle ROI. The suite
  keeps its area assertion, it pins the 15 voxel centres inside a 27 x 16 mm rectangle, where
  the older traversal counted the whole bounding box, and it covers a rectangle of no width.
- `packages/tools/test/EllipticalROIVoxelSampling.jest.js` -- 3 tests over Elliptical ROI.
  The suite pins the ellipse of voxel centres, it pins the selection through anisotropic
  spacing, and it covers an ellipse of no height.

The whole suite is 129 files and 2019 tests, and all of the tests pass. `tsc --noEmit` is
clean for `core` and for `tools`. Build `core` first with
`pnpm --filter @cornerstonejs/core run build:esm`, because `tools` type checks against
`core/dist`.

#### Manual tests

The statistics must be a property of the annotation and of the data, and never of the
display. Each check below states what to look at, and what a failure looks like.

Run the PET-CT example. `e22f2bd91` adds the other three area annotation tools to its
dropdown, so all four are reachable there now:

```
pnpm run example petCt
```

The example is the one that reaches an oblique plane, and it is the one that puts two
modalities of different slice thickness beside each other. Use it like this:

1. Choose **Crosshairs** in the tool dropdown.
2. Hover a reference line, and drag the **circle** handle, the one further from the centre,
   to rotate the plane. The three views are oblique after this.
3. Drag the **square** handle, the one closest to the centre, to change the slab thickness.
   An annotation records the slab thickness of its viewport when the user creates it.
4. Choose an ROI tool in the dropdown, and draw on any of the nine viewports.

The CT series and the PET series have different slice thickness, so the fusion row shows the
cross-modality case that Rule D covers: one annotation on a thick PET slice can correctly
appear on two thin CT slices.

`pnpm run example planarFreehandROITool` reaches the freehand tool on its own, in a volume
viewport, which is where the original report starts. `pnpm run example stackAnnotationTools`
reaches all four on a stack, for a comparison against the volume path.

Do each check below with the Planar Freehand ROI tool, the Circle ROI tool, the Rectangle ROI
tool and the Elliptical ROI tool, because all four share the sampler now.

1. **The statistics hold still under zoom and pan.** Draw an ROI, and write down the mean,
   the maximum and the area. Zoom in, zoom out, and then pan. The three numbers must not
   change. Before this pull request the numbers moved for the freehand tool, because the tool
   sampled in canvas coordinates.
2. **A rotated rectangle stops over-counting.** Draw a rectangle on an oblique plane, and
   compare the voxel count against the area. The older code counted every voxel of the index
   bounding box, so the count was too high at the corners.
3. **A circle stays a circle on an oblique plane.** Draw a circle on an oblique plane. The
   older code tested an ellipsoid aligned with the world axes, which selects the wrong voxels
   as soon as the plane tilts.
4. **An ellipse follows anisotropic spacing.** Draw an ellipse on a series with a large slice
   spacing, and then rotate to an oblique plane. The older code tested the same world-axis
   ellipsoid, and it derived that ellipsoid from the canvas corners.
5. **An oblique plane keeps its voxels.** Rotate the MPR to an oblique angle, and draw the
   same ROI. The area must stay close to the area of the same region in the acquisition
   plane. A large decrease shows that the traversal misses voxels of the integer lattice.
6. **A thick annotation covers more than one layer.** Increase the slab thickness of the
   viewport, and then draw the ROI. The voxel count must increase, because the annotation
   records the slab thickness at creation, and because the slab then covers more layers.
7. **A half-slice position covers two layers.** Move the MPR to a position midway between two
   slices, and then draw the ROI. Rule M selects both layers there, and the mean is a mean
   over the two layers.
8. **Anisotropic data behaves correctly.** Do the checks above again on a series that has a
   large slice spacing, for example 0.7 x 0.7 x 5 mm. The correction to the index bounds
   applies here: a thick annotation lost 4.5% of its voxels before the correction.
9. **A CPU-rendered viewport does not throw.** Draw a closed freehand ROI on a viewport that
   uses the CPU fallback. Before the correction, `imageData.getOrigin()` threw inside the
   render loop, and the exception stopped the viewport. Look at the console for an exception.
10. **A degenerate contour does not throw.** Click one time with the freehand tool, and do not
   drag. The tool must report nothing, and the tool must not throw.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] "OS:  Windows 11
- [x] "Node version:  22.19.0
- [x] "Browser: 141.0.7390.123
  

## Release Notes

* **Bug Fixes**
  * Improved planar freehand ROI measurement for oblique contours by removing axis-alignment failures and using effective voxel spacing.

* **Improvements**
  * Refined ROI voxel/statistics sampling by rasterizing ROI polygons in canvas space, deriving an appropriate sampling step, and sampling voxels from canvas coordinates (with deduplication).

* **New Features**
  * Added reusable utilities for canvas-space ROI polygon rasterization and canvas-driven voxel sampling, exposed for external use.

* **Tests**
  * Added Jest coverage for the ROI polygon-to-pixel rasterization iterator, including edge cases and fractional sampling.

## Downstream validation

OHIF_REF: fix/rectangle-roi-exact-voxel-set-stats

The area annotation tools now select the voxels whose centres lie inside the
shape. The older code walked the index bounding box and applied no shape test,
so it also counted a border of voxels outside the shape.

Two statistics of the rectangle ROI test of OHIF change. The mean moves from
`80.4 HU` to `79.8 HU`. The standard deviation moves from `38.2 HU` to
`38.1 HU`. The area, the maximum and the minimum keep their values.

The branch `fix/rectangle-roi-exact-voxel-set-stats` of `OHIF/Viewers` holds
the new values, and the `OHIF_REF` line above sends the downstream job to that
branch. A maintainer must merge that branch into `master` before this pull
request goes into `main`. If the branch stays separate, the downstream job
reads `master` again, and the job fails again.


arul-trenser:
@wayfarer3130  This PR replaces #2619. While we are able to draw the freehand ROI in the oblique view, the stats calculation for non-acquisition viewports in MPR mode was incorrect. This PR fixes those issues as well. Kindly review the changes.

Thanks!

packages/tools/src/utilities/math/polyline/getSubPixelSpacingAndXYDirections.ts
```diff
@@ -100,4 +118,51 @@ const getSubPixelSpacingAndXYDirections = (
...
+ * when `d` is aligned with axis i, d·j = d·k = 0, so s_eff = s_i.
+ *
+ * @param direction - A normalized world-space direction vector.
+ * @param iVector - The volume's I direction (row).
+ * @param jVector - The volume's J direction (column).
+ * @param kVector - The volume's K direction (slice).
+ * @param volumeSpacing - The [i, j, k] voxel spacings.
+ * @returns The effective spacing along `direction`.
+ */
+function _computeEffectiveSpacing(
```
wayfarer3130: This method should have it's own file, as it is quite useful for other types of operations.

packages/tools/src/tools/annotation/PlanarFreehandROITool.ts
```diff
@@ -935,74 +916,116 @@ class PlanarFreehandROITool extends ContourSegmentationBaseTool {
...
+          // Nearest-neighbor: round to closest voxel
+          const ijkPoint: Types.Point3 = [
+            Math.round(indexPoint[0]),
+            Math.round(indexPoint[1]),
+            Math.round(indexPoint[2]),
+          ];
+
+          // Bounds check
+          if (
+            ijkPoint[0] < 0 ||
```
wayfarer3130: This should get a function alongside this as a stand-alone function.

packages/tools/src/tools/annotation/PlanarFreehandROITool.ts
```diff
@@ -935,74 +916,116 @@ class PlanarFreehandROITool extends ContourSegmentationBaseTool {
...
+
+    for (let cy = startY; cy <= endY; cy++) {
+      // Compute all intersections of the polygon with this scanline row
+      const intersections = getLineSegmentIntersectionsCoordinates(
+        canvasCoordinates,
+        [startX - 1, cy] as Types.Point2,
+        [canvasMaxXPadded, cy] as Types.Point2
       );
+
+      if (!intersections || intersections.length === 0) {
```
wayfarer3130: This entire logic seems like it should be repeated for every volume type, just the actual method used to create the AABB intersections would differ depending on the actual object.  That would then handle oblique ellipse etc.

packages/tools/src/tools/annotation/PlanarFreehandROITool.ts
```diff
@@ -935,74 +916,116 @@ class PlanarFreehandROITool extends ContourSegmentationBaseTool {
...
+        continue;
+      }
+
+      // Sort intersections by X coordinate
+      intersections.sort((a, b) => a[0] - b[0]);
```
wayfarer3130: Why does this need sorting?

packages/tools/src/tools/annotation/PlanarFreehandROITool.ts
```diff
@@ -935,74 +916,116 @@ class PlanarFreehandROITool extends ContourSegmentationBaseTool {
...
+
+    const startX = Math.floor(canvasMinX);
+    const endX = Math.ceil(canvasMaxX);
+    const startY = Math.floor(canvasMinY);
+    const endY = Math.ceil(canvasMaxY);
+
+    const dimensions = imageData.getDimensions();
+    const canvasMaxXPadded = endX + 1;
+
+    for (let cy = startY; cy <= endY; cy++) {
```
wayfarer3130: What about an overall getIntersectionIterator - that would be a way this could be fully generalized, and as far as I can tell you don't actually care about ordering.  That could also be used for other things like drawing or converting to edge annotations for CPU rendering.

I'd also then like to see some unit tests on the rest of the code since you can just pass it a getIntersectionIterator result and test different cases.

packages/tools/src/tools/annotation/PlanarFreehandROITool.ts
wayfarer3130: Actually, it would be the overall intersections iterator that would need to be generated.  You could then put it onto the stats calculator or the voxel manager as helper methods.  That way the storage of thigns like the set of points iterated over could be standardized, and decided whether it is required or not.

wayfarer3130 (changes requested):
Generally really like this change, but would like it to be made a general solution for any area measurement, in terms of the intersectionIterator

packages/tools/src/tools/annotation/PlanarFreehandROITool.ts
```diff
@@ -935,74 +916,116 @@ class PlanarFreehandROITool extends ContourSegmentationBaseTool {
...
+            continue;
+          }
+
+          const value = voxelManager.getAtIJKPoint(ijkPoint);
+
+          if (value === undefined || value === null) {
+            continue;
+          }
+
+          const key = `${ijkPoint[0]}_${ijkPoint[1]}_${ijkPoint[2]}`;
```
wayfarer3130: Under what circumstances can you add a point a second time?  I wonder if that is a bug in the intersection generation or in the xor/setup of the planar freeform.

packages/tools/src/utilities/math/polyline/getSubPixelSpacingAndXYDirections.ts
arul-trenser: Moved as utilities

packages/tools/src/tools/annotation/PlanarFreehandROITool.ts
arul-trenser: Separated

packages/tools/src/tools/annotation/PlanarFreehandROITool.ts
arul-trenser: Separated

packages/tools/src/tools/annotation/PlanarFreehandROITool.ts
arul-trenser: @wayfarer3130 The loop iterates from left to right `for (let cx = xEnter; cx <= xExit; cx++)`, which requires `xEnter <= xExit`.

If the scanline hits a right edge first, intersections can return unsorted (e.g., [150, 42]). This sets xEnter = 150 and xExit = 42, causing the loop condition to fail immediately and skip rendering.

Sorting by the X-coordinate ensures xEnter is always the leftmost point.

packages/tools/src/tools/annotation/PlanarFreehandROITool.ts
arul-trenser: Moved and written unit tests

packages/tools/src/tools/annotation/PlanarFreehandROITool.ts
arul-trenser: The duplicate points aren't caused by a bug in logic, but by a resolution mismatch. When we sample the high-resolution canvas space (especially when zoomed in), multiple canvas pixels naturally round down to the exact same voxel coordinate. De-duplicating these points ensures we only count each unique voxel once.

arul-trenser:
@wayfarer3130 I've updated the code based on your feedback. Could you please take another look when you have a moment? 

Thanks!

sedghi (dismissed):
[P2] Honor storePointData before caching sampled voxels — /Users/alireza/open-source/clean/cornerstone3D/packages/
    tools/src/tools/annotation/PlanarFreehandROITool.ts:945-951
    With the default storePointData: false, the previous voxelManager.forEach path returned an empty pointsInShape array
    because it passed returnPoints: this.configuration.storePointData; this new call always accepts the full array returned
    by sampleVoxelsFromCanvas and stores every sampled voxel in cachedStats. For large freehand ROIs this can bloat
    annotation state and memory even when point storage is explicitly disabled, so the option should be threaded through or
    the assignment skipped when false.

  - [P2] Use the rounded voxel center for pointLPS — /Users/alireza/open-source/clean/cornerstone3D/packages/tools/src/
    utilities/sampleVoxelsFromCanvas.ts:114-117
    When a canvas sample rounds to a nearby ijkPoint (common for oblique and sub-pixel sampling), the voxel value comes
    from that rounded IJK but pointLPS is left as the original canvas-plane world coordinate. This makes maxLPS, minLPS,
    center, and stored pointsInShape coordinates inconsistent with the sampled voxel; convert the rounded IJK back to world
    space for the sample instead.

sedghi:
Another one 

The rewrite replaces voxel-driven scanline traversal with a canvas-space rasterizer (getIntersectionIterator) plus back-projection sampling (sampleVoxelsFromCanvas), which is a sound approach for oblique MPR and is backed by new jest tests. However it removes an existing defensive guard on voxelManager, so the stats path can now throw where it previously produced empty stats, and the new sampler has per-sample allocation and a fractional-step behavior that can be a hot-path cost on large/zoomed-out ROIs. The oblique spacing support in getSubPixelSpacingAndXYDirections is a reasonable replacement for the previous throw.

major packages/tools/src/tools/annotation/PlanarFreehandROITool.ts:986 The original code wrapped the voxel iteration in `if (voxelManager) { ... }`; the diff removes that guard and unconditionally calls `sampleVoxelsFromCanvas({ ..., voxelManager, ... })`, which dereferences `voxelManager.getAtIJKPoint(...)`. If a target image can lack a voxelManager (the removed guard implies it can, and this exact tool has documented crashes on PT/Fusion targets per PR 2616), stats calculation now throws instead of yielding empty stats.
minor packages/tools/src/utilities/sampleVoxelsFromCanvas.ts:132 Per-sample allocation on a stats hot path: each iterated canvas point builds a `sample` object and a template-literal Set key `${i},${j},${k}` for dedup. Combined with getCanvasVoxelSamplingStep returning a fractional step (1/ceil(maxDist)) when zoomed out, the iteration count scales with canvasArea/step^2 and can produce millions of string allocations for a large or heavily zoomed-out ROI.
minor packages/tools/src/utilities/math/polyline/getIntersectionIterator.ts:46 The scanline pairing `for (let i = 0; i + 1 < intersections.length; i += 2)` silently drops the last intersection when a row has an odd count (a polygon vertex landing exactly on the scanline, or a tangent). That row's final span is skipped, missing voxels. Classic even-odd rasterization robustness gap; low frequency with float canvas coords but unhandled.
nit packages/tools/src/utilities/sampleVoxelsFromCanvas.ts:118 `pointLPS` stored/passed to statsCallback is the canvas-ray world point (viewport.canvasToWorld of the sample pixel), not the world coordinate of the rounded voxel center `ijkPoint`. For stored point data or any stats using pointLPS this is a sub-voxel positional inaccuracy.

Belbin-GK:
@sedghi I have updated the code based on your feedback. Could you please take a look when you have a moment?

Thanks!

wayfarer3130:
Thanks for this, and I am sorry about the slow response. The bug that you fix is real, and it
is worth a fix.

We do not want to merge the canvas-space approach, and the reason is not specific to this PR.
A map from the displayed canvas pixels back to the voxels cannot be correct in general. One
canvas pixel can stand for several voxels along the view plane normal, and a larger canvas
never recovers them: a circle at 2 mm thickness over two 1 mm frames has two voxels behind
each pixel, and a sample can see one voxel only. The world-to-index conversion also rounds.
Together these two effects let `max` and `mean` omit voxels without a message, and they make
the statistics depend on the zoom and on the canvas resolution.

The cause is more general. Cornerstone3D never defined which voxels an ROI contains, so each
tool made its own choice. We now state that definition in
https://github.com/cornerstonejs/cornerstone3D/issues/2889, with a shared oblique-capable
voxel iterator that implements the definition in index space.

The plan has two steps:

1. We merge the iterator, the specification and the tests first.
2. We then adapt this PR onto the iterator. We will push to your branch, so the change stays
   yours.

We keep this PR open until then. Your scanline intersection work can survive into the new
implementation; the part that changes is what the code rasterizes over. Please review #2889
and add your comments, especially if any specified behaviour breaks the Flywheel use case
that started this PR.

wayfarer3130:
Thanks @arul-trenser. This PR identified a real bug: the old oblique scanline traversal did produce invalid statistics. Your diagnosis of *why* the voxel-driven path failed in an oblique view was also right. I retargeted this PR and reworked the sampling, and I give the reasoning in full here rather than push over your work without an explanation.

## What changed

**The base is now `feat/roi-voxel-iterator` (#2893)**, and not `main`. That PR adds the shared voxel iterator that #2889 specifies, and this PR is its first consumer. At the time of this comment the diff is 8 files, all in `packages/tools`. A later commit moves part of the code into `packages/core`; see the follow-up comment.

`iterateVoxelsInSlab` and `createContourShape` replace the canvas-space rasterizer. I **keep** your `viewUp` normalization in `getSubPixelSpacingAndXYDirections`, and I keep `computeEffectiveVoxelSpacing`. Both are correct, and the normalization is an independent bug fix: an unnormalized `viewUp` can fail the `1 - dot < EPSILON` alignment test even for an orthogonal view.

## Why I did not tune the canvas approach

Three problems lie out of reach of a change to the step size. Those three moved the work onto a different primitive instead of a fix to the sampler.

**A nearest-neighbour sample cannot cover an integer lattice under rotation.** At a 45° in-plane oblique angle, samples along `(0.707, 0.707)` in IJK round to `(0,0), (1,1), (2,2)…` and never visit `(1,0)` or `(0,1)`. That is about half the voxels, and no pitch reaches them. The sub-pixel phase of the camera decides which half the code skips, so a half-pixel pan changed the reported maximum. `getCanvasVoxelSamplingStep` also derived its step from `transformWorldToIndex`, which rounds to an integer, so a second quantization sat on top: a true 1.6 vox/px reads as 1.0.

**The sampled set depended on the display.** The set was a function of the zoom, the pan, the canvas size and `devicePixelRatio`, so the same annotation over the same data reported a different mean, because the display state differed when `_calculateCachedStats` ran.

**Every sample lay on the focal plane, so the result was always one voxel thick.** This problem showed me that the shape was wrong, and not the tuning. Draw a freehand annotation on an NM series with 1 mm slices, and fuse it with a CT series at 0.5 mm in the same orientation: the correct CT maximum must then examine **two** CT voxels for each in-plane location. No value of `canvasStep` gives that, because supersampling adds in-plane density on the same plane. Note that `getCanvasVoxelSamplingStep` *did* receive the `imageData` of each target and did adapt per target, which made the resolution mismatch look handled while the error through the plane stayed.

## How the iterator is used

Rule M of #2889: a voxel belongs to an area annotation when its centre lies within `(T + T_v) / 2` of the annotation plane along the normal, **and** its projection onto that plane falls inside the 2D shape. `T` is the thickness of the annotation, and `T_v` is the voxel thickness along the normal. The viewport slab never appears, so the statistics cannot change because a user zoomed or increased a slab.

`sampleVoxelsInContour` is the whole consumer side:

```ts
const shape = createContourShape({
  volume,       // { dimensions, direction, spacing, origin } from imageData
  planePoint,   // points[0] - the contour's own points define the plane depth
  normal,       // annotation.metadata.viewPlaneNormal, normalized
  polyline: points,
  depth: annotation.metadata.planeRestriction?.thickness,   // T
});

for (const { ijk, center } of iterateVoxelsInSlab({
  volume,
  planePoint,
  normal,
  annotationThickness: shape.getRequiredThickness(),
  bounds: this.getContourIndexBounds(points, imageData),
  getShapeRuns: shape.getRuns,
})) {
  // accumulate
}
```

Nothing there reads the viewport. **A tool with a different shape replaces `createContourShape` with `createEllipseShape` or `createRectangleShape`, and it changes nothing else.** That is the adoption path for `CircleROITool`, `EllipticalROITool`, `RectangleROITool` and the StartEndThreshold tools, and it is why this PR is of value as the worked example.

Two details matter. `ijk` and `center` are **reused between iterations**, so copy either one before you retain it. When `planeRestriction.thickness` is absent — stack annotations, and everything drawn before that field existed — `T` defaults to one voxel along the normal, so existing annotations behave as before.

## This is also faster in a non-oblique view

The cost is proportional to the voxels selected plus the rows touched, and not to the canvas area. An ROI at 8× magnification costs what it costs at fit-to-window. The canvas path paid roughly 64×, plus one `canvasToWorld` for each sample, and each of those calls changed the vtk camera state. The depth half of Rule M has a closed form, `depth(p) = p · g + c₀`, which is linear in the integer indices, and the contour gives exact in-plane runs. The code therefore tests no voxel on its own.

## One defect that precedes this PR

The early return of `sampleVoxelsFromCanvas` on `!storePointData` returned before any `statsCallback` ran, and not only before the point storage. `storePointData` defaults to `false` on `PlanarFreehandROITool`, and the example never overrides it, so every freehand ROI reported `mean` and `stdDev` as `NaN`, `min` as `Infinity`, `max` as `-Infinity`, and `count` as `0`. I state this because the defect would have shipped as a silent regression. The replacement always accumulates the statistics, and only the returned point list stays conditional. `planarFreehandVoxelSampling.jest.js` covers it.

## Two decisions that you can overrule

- **I delete `sampleVoxelsFromCanvas.ts`** rather than leave it unused. The file is the rejected approach, the package index exported it, and it carried the `NaN` defect. I can restore it if you prefer to keep it.
- **I keep `getIntersectionIterator` and `getScanlineIntersections`**, with their tests, because both are generic 2D polyline rasterizers that do not depend on the voxel selection. One separate point: the `startX`/`endX` clip of `getScanlineIntersections` can leave an odd count of crossings, and the `i += 2` pairing of `getIntersectionIterator` then gives an inverted fill. No current caller reaches that state, but both functions are new public API, so a follow-up has value if they stay. **The follow-up comment reverses this second decision.**

## Verification

`packages/tools`: 51 suites, 752 passed, 2 skipped. `tsc --noEmit` is clean on both packages, and `oxlint` is clean on the changed files. The new tool test covers the NM/CT thickness case directly: `T = 1` against 0.5 mm voxels selects exactly 2 layers and 18 voxels, where `T = 0.2` selects 1 layer and 9 voxels, with an identical in-plane selection.

#2893 covers the exactness of the iterator against a brute-force reference implementation of Rule M, at oblique angles from 1° to 89°, with anisotropic spacing and rotated direction matrices.

**#2893 must merge first.** I am happy to discuss any part of this, especially if you disagree with the removal of the canvas sampler.

wayfarer3130:
## Follow-up pass on the add-on work (`78625853b`)

This is a second pass over what I pushed in `f2331654c`. I reviewed that change against the rest of the tree, and not against the original bug. Nothing here touches the diagnosis or the fix that motivated this PR: this is cleanup of my own add-on commit, plus three defects that the review found. I split the pass out so the two stay separable, and so the reversal below is visible instead of hidden in a force-push.

### I reversed one of my own decisions

In the last comment I said that **`getIntersectionIterator` and `getScanlineIntersections` stay**, because both are generic 2D polyline rasterizers that do not depend on the voxel selection. That decision was wrong. After the rewrite nothing in `src` calls either function, and both duplicate machinery that already exists: `getLineSegmentIntersectionsCoordinates` gives the crossings, and `createContourShape.getRuns` gives the scanline walk with exact runs and even-odd holes. Two exported functions with no caller are not neutral, and I had just flagged one of the two as defective. They are new public API, and somebody will build on them.

I remove both, with their test file. The odd-crossing defect of `startX`/`endX` that I flagged for a follow-up is therefore moot, and not outstanding.

`math/polyline/index.ts` is now byte-identical with `main`.

`snapIndexBounds` goes too, because its only consumer was the bbox snapping that this PR deletes. That removal does shrink `utilities.boundingBox`. The helper shipped one release ago for exactly this code path, so I state the removal here rather than let someone discover it.

### Code that reimplemented something we already have

- **`getContourIndexBounds`** used its own min/max loop over the index points. It now calls `getBoundingBoxAroundShapeIJK`.
- **`computeEffectiveVoxelSpacing`** stays, because it is @arul-trenser's formula and it is correct. The function moves to core as **`getEffectiveSpacingAlongDirection`**, beside `getSpacingInNormalDirection`, with the same `(volume, direction)` signature as its two siblings. The function measures volume geometry, and not polyline math, and it was the third measure of "spacing along a direction" in the tree. Its doc now states how the three differ, so that no fourth appears:

  | | formula | answers |
  |---|---|---|
  | `getSpacingInNormalDirection` | L2, `sqrt(Σ (d·aᵢ·sᵢ)²)` | how far the camera dollies before it sees new voxels |
  | `getVoxelThicknessAlongNormal` | L1, `Σ \|d·aᵢ\|·sᵢ` | how far one voxel *reaches* along the direction |
  | `getEffectiveSpacingAlongDirection` | harmonic, `1/sqrt(Σ (d·aᵢ/sᵢ)²)` | how far to step to cross one voxel |

  All three agree when `d` is parallel to a voxel axis, and they differ for an oblique direction.

### The accumulation loop moved to core

`sampleVoxelsInContour` held nothing contour-specific below `createContourShape`, and its own docstring said so. The function is now **`csUtils.voxelSlab.sampleVoxelsInSlab`**, which shortens the migration path that I described last time. Circle, Ellipse and Rectangle ROI replace the shape factory, and they get the value reads, the copy-on-retain, the stats callback and the conditional point storage at no cost. They do not copy the loop, and its defects, three more times.

The consumer side is now the whole of it:

```ts
const shape = createContourShape({ volume, planePoint, normal, polyline: points, depth: T });

return sampleVoxelsInSlab({
  volume, planePoint, normal,
  annotationThickness: shape.getRequiredThickness(),
  bounds: this.getContourIndexBounds(points, volume, imageData, normal, halfWidth),
  getShapeRuns: shape.getRuns,
  voxelManager,
  onSample: statsCallback,
  storePointData,
});
```

`voxelSlabSampling.jest.js` covers the loop directly in core, so the reading contract no longer depends on a test in tools.

### Three defects

**The index bounds dropped voxels.** A fixed ±1 voxel dilated the bounds, but the slab reaches `(T + T_v)/2` mm along the normal, which is many voxels for a thick annotation. I measured a loss of 4.5% of the voxels on a 0.7×0.7×5 mm oblique contour, and a loss of 20% at `T` = 20 mm. The dilation is now `ceil(halfWidth·|aᵢ·n|/sᵢ) + 1` for each axis.

**`imageData.getOrigin()` threw on a CPU-rendered stack.** `StackViewport.getImageDataCPU` builds an inner `imageData` with `getDirection`, `getDimensions`, `getSpacing`, `worldToIndex` and `indexToWorld`, and with no `getOrigin`, which matches the `CPUImageData` interface. Every closed freehand ROI on a CPU fallback viewport would have thrown inside the render loop. The geometry now comes from the enclosing `IImageData` instead of a fallback, because both the GPU path and the CPU path carry `dimensions`, `direction`, `spacing` and `origin` there as plain fields. That removes the mismatch between a getter and a property as a class of defect, and not only this one instance.

**A contour with fewer than three points threw.** `createContourShape` rejects a ring with fewer than three points, and the guard tested `!points?.length` only. A degenerate contour — bad hydration, or a click that registered one point — therefore escaped an exception into the render loop, and it left the shared stats accumulator half-filled. The guard now requires three points.

### One change of behaviour worth stating

The half-voxel dilation of Rule M means that an annotation plane exactly midway between two voxel centres **selects both layers**, where the old `snapIndexBounds` path always picked one. Both voxels overlap the slab by equal amounts, so no principled way to choose exists, and a choice would make the count depend on a rounding tie. A mean over two layers is not the same number as a mean over one, and MPR at a half-slice position is the ordinary way to reach that state. `getMembershipHalfWidth` now documents the behaviour, so that nobody discovers it in a diff of statistics.

### Verification

`packages/core` and `packages/tools`: 89 suites, 1416 passed, 4 skipped. `tsc --noEmit` is clean on both, and `oxlint` and prettier are clean on the changed files.

wayfarer3130 (approved):
The approval is based on the current state, but requires the underlying PR for the voxel iterator to be completed.

wayfarer3130:
Reopening and retartgetting at origin/main

wayfarer3130:
Hi @arul-trenser — an explanation of what happened here, because none of it was visible from your side.

**This pull request closed by accident.** #2893 merged on 10 September, and the `feat/roi-voxel-iterator` branch that this pull request targeted was deleted with it. GitHub closes any pull request whose base branch is deleted. No person rejected your work. The base is `main` now, and the pull request is open again.

**The API changed under you.** #2893 renamed the voxel iterator after review. `createContourShape` is `createPolylineShape`, `iterateVoxelsInSlab` is `iterateVoxelsInShape`, `annotationThickness` is `referencePlaneThickness`, and the shape options take `viewPlaneNormal` instead of `normal`. I merged `main` and moved the code to the new names, so you do not have to.

**Your formula moved to core.** `computeEffectiveVoxelSpacing` is now `csUtils.getEffectiveSpacingAlongDirection`, beside `getSpacingInNormalDirection`. The function measures volume geometry rather than polyline mathematics, and it was the third measure of "spacing along a direction" in the tree. Its documentation now states how the three differ.

**Circle ROI, Rectangle ROI and Elliptical ROI use your path now.** One function holds the plane, the thickness, the index bounds and the accumulation, and each tool supplies only its shape. That work found two more defects: Rectangle ROI applied no shape test at all, and Circle ROI and Elliptical ROI tested an ellipsoid aligned with the world axes.

The PET-CT example offers all four tools in its dropdown now, because that example has the crosshairs, and the crosshairs are how a user reaches an oblique plane.

The description carries the full detail, including the testing notes. Thank you for the original fix. It is what started all of this.

wayfarer3130 (approved):

wayfarer3130:
@sedghi thank you for the review. The implementation changed after it. The
canvas rasterizer is gone: `sampleVoxelsFromCanvas.ts` and
`getIntersectionIterator.ts` are both deleted, and every area annotation tool
now calls the shared voxel iterator of
[`utilities.voxelSlab`](https://github.com/cornerstonejs/cornerstone3D/blob/main/packages/docs/docs/concepts/cornerstone-tools/annotation/voxel-statistics.md).
The iterator works in index space, reads no viewport, and emits each voxel one
time. Here is each of your points against the code of today.

**[P2] Honor `storePointData`.** Correct, and resolved.
`sampleVoxelsInShape` adds a sample to the returned array only when
`storePointData` is true, and `PlanarFreehandROITool` passes
`this.configuration.storePointData`, which defaults to false. The statistics
accumulate through `onSample` in both cases, so `cachedStats.pointsInShape` is
an empty array by default.

**[P2] and nit: use the voxel centre for `pointLPS`.** Correct, and resolved.
The sample carries the voxel centre that the iterator computes, and `pointIJK`
names the same voxel, so `maxLPS`, `minLPS` and `pointsInShape` agree with the
value. No back projection of a canvas point exists any more.

**major: the removed `voxelManager` guard.** Correct, and resolved.
`sampleVoxelsInShape` returns an empty array when the target has no
`voxelManager`, so a WSI target reports no statistics and throws nothing, as
before. Your point also found a second case: `VideoViewport.getImageData`
returned an object that carries `forEach` alone, so `getAtIJKPoint` threw
inside the render loop. The video viewport now builds a real voxel manager over
the RGBA pixels of the frame on display, and the value of a pixel is its
`[r, g, b]` triple, as a colour image in a stack viewport also reports.

**minor: per-sample allocation.** The dedup `Set` and the template-literal keys
are gone. The iterator emits each voxel one time, so no dedup is necessary, and
the count of iterations is the count of voxels in the ROI. The count no longer
depends on the canvas area, on the zoom, or on a fractional step. One small
object per voxel stays, because `BasicStatsCalculator` keeps the `pointLPS`
array when point storage is on. A reused object would corrupt that array.

**minor: the odd count of intersections.** The old file is gone, but your point
still holds, because `createPolylineShape` pairs consecutive crossings in the
same way. Three rules keep the count even, and each rule now has a test in
`packages/core/test/voxelSlabPolylineRings.jest.js`:

- A side value of zero counts as negative. A vertex whose two edges both leave
  upwards therefore gives two crossings at the same place, and the pair closes
  over the one voxel of the vertex.
- A vertex that touches the row from below gives no crossing at all. A separate
  rule contributes the one voxel of the vertex.
- An edge that lies along the row gives no crossing, because both ends of the
  edge sit on the row. A third rule contributes the extent of the edge.

Each new test fails when I disable the rule that the test covers, so the tests
guard the behaviour and do not pass by accident. `getRuns`, `containsPoint` and
a brute-force reference select the same voxels for every case, at an axial
orientation and at three oblique angles.

wayfarer3130 (approved):

PR #2766, opened 2026-06-20
fix: more generic viewport push for parity with legacy

sedghi:
OHIF_REF: ohifohifnextapi

> Linked downstream test: the OHIF Downstream Validation workflow runs the OHIF branch `ohifohifnextapi` (OHIF/Viewers#6101) against this cornerstone PR.

### Context

This PR teaches the legacy `tools` package and parts of `core` to operate against the new native **GenericViewport** ("next") rendering lane, so existing tools and synchronizers work across both the legacy and native viewport families.

### Changes & Results

**Core**
- New viewport-capability predicates (`isGenericViewport`, `viewportIsInVolumeMode`, `getCurrentMode`) plus a `viewportCapabilities` utility, replacing `instanceof` checks.
- `voiLUTFunction` plumbed through the planar render paths and presentation props.
- Per-mount `forceCPU` option for planar render-path selection.
- Native spacing calibration (CS-12) and a CPU image pre-scaling flag.
- Slice-count and presentation-modified events for native viewports; planar orientation predicate and segmentation-preserving `setDisplaySets`.
- `RenderModesPanel` / `StatsOverlay` now include `viewportType` for debugging context.
- `getImageActor` added to `PlanarViewport`; keep the on-screen canvas visible for `VolumeViewport3D`.

**Core — VOI / slice-index parity (recent)**
- Emit `VOI_MODIFIED` when a display set is mounted (its default VOI), mirroring legacy stack/volume viewports so OHIF's window-level readout, colorbar, and VOI synchronizers update on a display-set change.
- Prescaled PT (SUV) images default to a `0–5` VOI on the native image path, matching legacy `StackViewport` — previously the raw DICOM window center/width was used, rendering PET far too wide.
- Slice-index is derived from the camera (single source of truth, like legacy `VolumeViewport.getSliceIndex`); the scrollbar index is reconciled with the rendered slice on volume mount; and the acquisition orientation centers the volume on MPR mount instead of pinning to the first slice.

**Tools**
- Lane-agnostic camera helpers (`getViewportICamera` / `setViewportCamera` / `resetViewportCamera`) and `isGenericViewport`-gated branches so the following support native (next) viewports: Crosshairs, Magnify, AdvancedMagnify, ReferenceLines, ReferenceCursors, cine `playClip`, 3D VR tools (Trackball/VolumeRotate/VolumeCropping), overlay/region 2D tools, and contour segmentation rendering (CS-8).
- VOI and image-slice synchronizers, camera-position sync, and stack-prefetch updated to the capability-based checks.

**Examples**
- New `genericViewportShowcase` example demonstrating the native viewport with interactive crosshairs.

### Testing

- `viewportContentMode` jest coverage added.
- Open the `genericViewportShowcase` example and exercise the supported tools (crosshairs jump, magnify, reference lines, cine) on native viewports.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [ ] "OS: 
- [ ] "Node version: 
- [ ] "Browser:
  

## Summary by CodeRabbit

* **New Features**
  * Added VOI LUT configuration for planar rendering, including sampled sigmoid.
  * Added planar dataset option to force CPU rendering (`forceCPU`).
  * Expanded Generic (“next”) viewport support across rendering, camera/view-state sync, VOI updates, cine playback, and core tools.
* **Improvements**
  * Improved planar CPU rendering: pre-scaled image handling, user calibration overrides, VOI/LUT propagation, and more accurate new-image signaling.
  * Updated planar colormap opacity to better respect absolute `threshold`; refined render-mode decisions and stats UI (viewport type).
* **Bug Fixes**
  * Improved generic viewport resize sizing and avoided redundant legacy camera reset behavior.
  * Fixed PT/RTDOSE scaling identity short-circuits and improved tool geometry using shared camera helpers.


packages/core/src/RenderingEngine/GenericViewport/Planar/planarImageEvents.ts
```diff
@@ -53,8 +54,19 @@ export function triggerPlanarVolumeNewImage(
...
 
-  triggerPlanarNewImage(ctx);
+  // A volume-backed slice change must also emit VOLUME_NEW_IMAGE so that
+  // volume-aware consumers (volume scroll indicators, MPR slice synchronizers,
+  // segmentation slice tracking) react. The native volume-slice path previously
+  // emitted only STACK_NEW_IMAGE, which those consumers do not listen for, so
+  // every slice change on a PLANAR_NEXT volume viewport was missed.
+  triggerEvent(ctx.viewport.element, Events.VOLUME_NEW_IMAGE, {
+    imageIndex: params.imageIdIndex ?? 0,
+    numberOfSlices: params.maxImageIdIndex + 1,
```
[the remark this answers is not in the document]
sedghi: Addressed in a5492df33a. VOLUME_NEW_IMAGE is now emitted only when the slice index is known (guarded on typeof params.imageIdIndex === number), so an unknown or oblique index no longer reports slice 0.

packages/core/src/RenderingEngine/GenericViewport/Planar/PlanarViewport.ts
```diff
@@ -577,6 +624,97 @@ class PlanarViewport extends GenericViewport<
...
+        });
+      }
+    }
+
+    if (colormapTouched && props.colormap) {
+      triggerEvent(this.element, Events.COLORMAP_MODIFIED, {
+        viewportId: this.id,
+        colormap: props.colormap,
+        volumeId,
+      });
```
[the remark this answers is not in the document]
sedghi: Addressed in a5492df33a. The events now use the modified display sets own volume id via getDataSet(displaySetId)?.volumeId, falling back to getVolumeId() only when the modified set is the active source.

packages/core/src/RenderingEngine/GenericViewport/Planar/PlanarViewport.ts
```diff
@@ -610,6 +748,57 @@ class PlanarViewport extends GenericViewport<
...
+   * native equivalent of the legacy stack/volume `getNumberOfSlices()` and is
+   * what `csUtils.jumpToSlice`, scroll, and image-slice synchronizers rely on
+   * for a `PLANAR_NEXT` viewport (which is not a `StackViewport`, so it takes
+   * the slice-count navigation branch).
+   *
+   * @returns The total slice count for the active source data.
+   */
+  getNumberOfSlices(): number {
+    return Math.max(this.getImageIds().length, this.getMaxImageIdIndex() + 1);
+  }
```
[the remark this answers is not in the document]
sedghi: Addressed in a5492df33a. getNumberOfSlices now returns 0 when no images are mounted (early return before the Math.max fallback).

packages/core/src/utilities/isPTPrescaledWithSUV.ts
```diff
@@ -1,7 +1,7 @@
 import type { IImage } from '../types';
 
 const isPTPrescaledWithSUV = (image: IImage) => {
-  return image.preScale.scaled && image.preScale.scalingParameters.suvbw;
+  return image.preScale?.scaled && image.preScale?.scalingParameters?.suvbw;
```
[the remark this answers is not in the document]
sedghi: Addressed in a5492df33a. Wrapped the result in Boolean() and annotated the return type as boolean.

packages/tools/src/tools/MagnifyTool.ts
```diff
@@ -112,12 +126,31 @@ class MagnifyTool extends BaseTool {
...
+          getDisplaySetPresentation?: (
+            dataId?: string
+          ) => Record<string, unknown>;
+        })
+      : undefined;
+    const viewportProperties = nativeSource
+      ? (nativeSource.getDisplaySetPresentation?.(
+          nativeSource.getSourceDataId?.()
+        ) ?? {})
+      : viewport.getProperties();
```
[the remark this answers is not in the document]
sedghi: Addressed in a5492df33a. The casing is bridged in the shared getNativeSourceProperties helper (maps voiLUTFunction to VOILUTFunction), so both the MagnifyTool and AdvancedMagnifyTool loupes honor a sigmoid source VOI.

packages/tools/src/tools/ReferenceCursors.ts
```diff
@@ -519,6 +520,49 @@ class ReferenceCursors extends AnnotationDisplayTool {
           renderingEngine.renderViewport(viewport.id);
         }
       }
+    } else if (utilities.isGenericViewport(viewport)) {
+      // Native PLANAR_NEXT (stack or volume mode) is neither StackViewport nor
+      // VolumeViewport. Move the rendering plane to the cursor along the view-plane
+      // normal via the view reference (snaps to the nearest slice); it has no setCamera.
+      // (The instanceof checks above narrow `viewport` to never here, so cast back.)
+      const nativeViewport = viewport as unknown as Types.IViewport & {
```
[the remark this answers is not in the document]
sedghi: Addressed in a5492df33a. updateViewportImage now accepts Types.IViewport and the as-unknown cast in the generic branch was removed.

packages/tools/src/tools/VolumeCroppingControlTool.ts
```diff
@@ -682,7 +684,7 @@ class VolumeCroppingControlTool extends AnnotationTool {
     let orientation = null;
     if (enabledElement.viewport && enabledElement.viewport.getCamera) {
       orientation = getOrientationFromNormal(
-        enabledElement.viewport.getCamera().viewPlaneNormal
+        getViewportICamera(enabledElement.viewport).viewPlaneNormal
       );
```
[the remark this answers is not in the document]
sedghi: Addressed in a5492df33a. Removed the legacy getCamera gate; orientation is now derived through the lane-agnostic getViewportICamera and guarded on viewPlaneNormal, so it resolves on native (generic) viewports too.

packages/tools/examples/genericViewportShowcase/index.ts
```diff
@@ -0,0 +1,549 @@
...
+);
+
+const content = document.getElementById('content');
+
+const grid = document.createElement('div');
+grid.style.display = 'flex';
+grid.style.flexWrap = 'wrap';
+grid.style.gap = '8px';
+grid.style.marginTop = '8px';
+content.appendChild(grid);
```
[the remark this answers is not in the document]
sedghi: Not applying this. Dereferencing the content element directly is the established convention across every example in this repo (the examples host template always provides the content container). Adding a guard or throw here would diverge from the other 180+ examples without practical benefit for a demo page.

packages/tools/examples/genericViewportShowcase/index.ts
```diff
@@ -0,0 +1,549 @@
...
+  labelText: 'Acquisition tool: ',
+  onSelectedValueChange: (selectedValue) => {
+    const toolGroup = ToolGroupManager.getToolGroup(acqToolGroupId);
+    toolGroup.setToolPassive(currentAcqPrimaryTool);
+    toolGroup.setToolActive(selectedValue as string, {
+      bindings: [{ mouseButton: MouseBindings.Primary }],
+    });
+    currentAcqPrimaryTool = selectedValue as string;
+    console.log(`${LOG} acquisition primary tool -> ${selectedValue}`);
+  },
```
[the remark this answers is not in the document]
sedghi: Leaving as-is. None of the repo examples guard toolbar handlers against clicks before run() completes; the fire-and-forget pattern is the established convention for these demo files, and the pre-init window is negligible since the toolbar renders with the page and run() resolves quickly.

packages/tools/examples/genericViewportShowcase/index.ts
```diff
@@ -0,0 +1,549 @@
...
+  console.log(`${LOG} 3D trackball rotate active on VR viewport`);
+
+  // Cine prefetch so stack scrolling/cine stays smooth.
+  csToolsUtilities.stackPrefetch.enable(acqViewport.element);
+
+  renderAll();
+  console.log(`${LOG} ready`);
+}
+
+run();
```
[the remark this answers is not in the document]
sedghi: Leaving as-is to match repo convention: 181 examples call run() without a catch and none use .catch(). Keeping this consistent with the rest of the examples.

packages/core/src/RenderingEngine/GenericViewport/Planar/planarImageEvents.ts
```diff
@@ -53,8 +54,25 @@ export function triggerPlanarVolumeNewImage(
       imageId: params.imageIds[params.imageIdIndex],
       imageIdIndex: params.imageIdIndex,
     });
-    return;
+  } else {
+    triggerPlanarNewImage(ctx);
   }
 
-  triggerPlanarNewImage(ctx);
+  // A volume-backed slice change must also emit VOLUME_NEW_IMAGE so that
```
wayfarer3130, line 61: Probably we should deprecate both volume and stack new image in favour of viewport new image, but keep on firing the old ones for a while.

packages/tools/src/tools/WindowLevelRegionTool.ts
```diff
@@ -355,15 +356,44 @@ class WindowLevelRegionTool extends AnnotationTool {
...
-    viewport.setProperties({ voiRange });
+    if (presentationViewport) {
+      if (sourceDataId) {
+        presentationViewport.setDisplaySetPresentation(sourceDataId, {
+          voiRange,
+        });
+      }
+    } else {
+      viewport.setProperties({ voiRange });
+    }
```
wayfarer3130, line 409: **Correctness — window-level write silently dropped on a presentation viewport with no source data id.**

The computed `voiRange` is applied only when `presentationViewport && sourceDataId`. The legacy `viewport.setProperties({ voiRange })` fallback is in the `else` that is reached **only when `presentationViewport` is undefined**. So if a native viewport supports display-set presentation but `getSourceDataId()` returns `undefined` (region committed before/while the source binding is mounted, or only overlay bindings present), neither branch runs — the VOI is discarded, `viewport.render()` runs, and the drag does nothing with no error.

Consider applying to the active/first binding (or at least warning) when `sourceDataId` is undefined, rather than silently no-op'ing.

packages/core/src/RenderingEngine/GenericViewport/Planar/VtkImageMapperRenderPath.ts
```diff
@@ -400,10 +430,49 @@ async function updateRenderedImage(args: {
...
+  if (
+    scalars.getDataType() !==
+    image.voxelManager.getScalarData().constructor.name
+  ) {
+    return false;
+  }
+
+  const incomingComponents = image.color ? 3 : 1;
+
+  return scalars.getNumberOfComponents() === incomingComponents;
```
wayfarer3130, line 489: **Correctness (perf) — `canReuseImageDataForImage` assumes 3 components for every color image, so RGBA images never reuse their texture.**

`image.color ? 3 : 1` yields 3 for a 4-component RGBA image, while the existing `vtkImageData` reports 4 components, so this equality always fails and a fresh `vtkImageData`/texture is allocated on every scroll — reintroducing the one-frame black flash this function exists to prevent for RGBA ultrasound / secondary-capture stacks.

Derive the component count from the image (e.g. `image.rgba ? 4 : image.color ? 3 : 1`, or from the actual pixel-data layout) instead of hard-coding 3 for all color.

(The `getDataType()` vs `constructor.name` compare a few lines up is fine — VTK's data-type strings match the TypedArray constructor names.)

packages/tools/src/utilities/voi/windowlevel/extractWindowLevelRegionToolData.ts
```diff
@@ -56,4 +62,34 @@ function extractImageDataStack(viewport: Types.IStackViewport) {
   };
 }
 
+function extractImageDataGeneric(viewport: Types.IGenericViewport) {
+  const imageId = viewport.getCurrentImageId();
+
+  if (!imageId) {
+    throw new Error('Viewport not supported');
+  }
```
wayfarer3130, line 80: **Correctness — generic extractor throws on an oblique/reformatted native plane.**

`extractImageDataGeneric` keys off `getCurrentImageId()`, which is `undefined` for an oblique/orthogonal reslice on a volume-mode `PLANAR_NEXT` viewport (the comment above even scopes this to "stack / vtkImage mode (and axially aligned MPR)"). On those planes this throws `Viewport not supported`, aborting the WindowLevelRegion tool's stats — exactly the MPR planes where the legacy volume extractor still worked by reading `imageData` directly.

Consider falling back to the volume-slice path (like `extractImageDataVolume` / `getCurrentVolumeViewportSlice`) when the viewport is in volume mode and no single image id resolves.

packages/core/src/utilities/VoxelManager.ts
```diff
@@ -430,7 +430,10 @@ export default class VoxelManager<T> {
     if (this._getScalarData) {
       const scalarData = this._getScalarData();
       if (storeScalarData) {
-        console.log('Not transient, should store value', scalarData);
+        // Retain the expanded scalar data so subsequent calls return it
+        // directly instead of re-expanding (e.g. re-running the RLE decode)
+        // on every access. Callers opt in because this trades memory for speed.
+        this.scalarData = scalarData as PixelDataTypedArray;
```
wayfarer3130, line 461: **Correctness (latent) — `storeScalarData` turns a no-op placeholder into a real cache with no invalidation.**

The old `console.log('Not transient, should store value', …)` is replaced with `this.scalarData = scalarData`, so once a caller passes `storeScalarData = true` on a lazy/derived source (e.g. an RLE-decoded labelmap), later `getScalarData()` calls return the cached buffer and never re-run `_getScalarData()`. No voxel setter invalidates `this.scalarData`, so a subsequent segmentation edit would be invisible to stats/rendering.

No in-repo caller passes `true` today, so this is a trap for the next caller rather than a live bug — worth either invalidating `this.scalarData` on writes or a `// TODO: invalidate on edit` note here.

packages/core/src/RenderingEngine/GenericViewport/Planar/PlanarViewport.ts
```diff
@@ -1536,9 +1956,16 @@ class PlanarViewport extends GenericViewport<
...
+    // otherwise centers the volume, matching legacy and the reformatted
+    // (sagittal/coronal) orientations. Previously a defaulted index of 0 (e.g.
+    // the MPR hanging protocol, which supplies no initial slice) pinned the
+    // acquisition viewport to the first slice instead of the center, so the
+    // scrollbar sat at the top. `|| undefined` maps the default 0 to "center"
+    // while preserving a real carried slice (> 0).
     const initialImageIdIndex =
       orientation === OrientationAxis.ACQUISITION
-        ? planarData.initialImageIdIndex
+        ? planarData.initialImageIdIndex || undefined
```
wayfarer3130: **Correctness (edge) — `|| undefined` discards a legitimate slice 0 on the data path.**

For `ACQUISITION` orientation, `planarData.initialImageIdIndex || undefined` maps a real carried index of `0` (e.g. a saved presentation parked on the first slice) to "center the volume," so the viewport opens mid-stack instead of at slice 0. The comment documents the intent (map the hanging-protocol default 0 to center), but it can't distinguish "defaulted 0" from "explicitly 0."

The cleaner fix is to leave `initialImageIdIndex` `undefined` upstream when no slice is requested, rather than defaulting it to 0 and then re-interpreting falsy 0 here. (Note: this is the *data* path; the analogous *event* path `imageIndex ?? 0` was already fixed in `a5492df`.)

packages/core/src/RenderingEngine/GenericViewport/Planar/PlanarViewport.ts
```diff
@@ -401,6 +425,31 @@ class PlanarViewport extends GenericViewport<
...
+      this.notifyDataPresentationModified(dataId, { voiRange });
+    }
+
+    if (role === 'source') {
+      // Emit an initial CAMERA_MODIFIED for the source binding only, mirroring
+      // legacy StackViewport's first-image triggerCameraEvent. Overlays and
+      // orientation markers paint their first frame off this event. Trigger the
+      // event directly rather than routing through modified() to avoid a
+      // redundant render (addLoadedData already rendered).
+      this.triggerCameraModifiedEvent(this.getCameraForEvent());
```
wayfarer3130, line 451: **Correctness — mount-time `VOI_MODIFIED`/`CAMERA_MODIFIED` can reset or echo across synchronizer groups.**

`addLoadedData` now fires `notifyDataPresentationModified` (→ `VOI_MODIFIED`) per binding and `triggerCameraModifiedEvent` for the source on **every** mount. For a `PLANAR_NEXT` viewport in a camera or VOI sync group, mounting/replacing one display set drives the sync callback, which can copy this freshly-centered camera / default VOI onto the **other** viewports in the group (overwriting a slice/zoom the user already set), or feed a `VOI_MODIFIED` echo back into the synchronizer.

The per-display-set `volumeId` scoping here is already correct (fixed in `a5492df`); the remaining concern is the unconditional mount-time emission itself. Consider suppressing these during the mount/replace path, or guarding the synchronizer against self-originated mount events.

wayfarer3130 (commented):
Automated review pass (correctness-focused). These are findings **not** already covered by the existing CodeRabbit review/`a5492df` fixes. Each was verified against the current head. Cleanup/duplication nits (parallelScale reconstruction across the magnify tools, dataId→volumeId actor resolution shared between `getViewportModality`/`isViewportPreScaled`) are omitted here to keep signal high.

wayfarer3130:
Generally I'm happy with the PR, but some repeated code bothered me, and also some of hte prescaling handling.  I worked with claude to generate an analysis of that which is below:

How much of the PR is prescaling
It's a small fraction by line count but a recurring, disproportionately fiddly motif. The directly prescale-of-pixel-values cluster is roughly ~150 lines across ~7 files (~5% of the +3122):

[CpuImageSliceRenderPath.ts](vscode-webview://0ojil4esr7c805nn2pf1dkdm93jglh77udcgtdrjo1lc8e4fnbfo/packages/core/src/RenderingEngine/GenericViewport/Planar/CpuImageSliceRenderPath.ts) — markCpuImagePreScaled (sets image.isPreScaled = preScale?.scaled) called at two render sites, so the CPUFallback skips the modality LUT.
[planarImageRendering.ts](vscode-webview://0ojil4esr7c805nn2pf1dkdm93jglh77udcgtdrjo1lc8e4fnbfo/packages/core/src/RenderingEngine/helpers/planarImageRendering.ts) getDefaultImageVOIRange — isPTPrescaledWithSUV(image) → hardcode {lower:0, upper:5}.
[isPTPrescaledWithSUV.ts](vscode-webview://0ojil4esr7c805nn2pf1dkdm93jglh77udcgtdrjo1lc8e4fnbfo/packages/core/src/utilities/isPTPrescaledWithSUV.ts) — null-safety hardening of the predicate.
[getScalingParameters.ts](vscode-webview://0ojil4esr7c805nn2pf1dkdm93jglh77udcgtdrjo1lc8e4fnbfo/packages/dicomImageLoader/src/imageLoader/getScalingParameters.ts) — normalize rescale slope/intercept defaults before the identity short-circuit.
[WindowLevelTool.ts](vscode-webview://0ojil4esr7c805nn2pf1dkdm93jglh77udcgtdrjo1lc8e4fnbfo/packages/tools/src/tools/WindowLevelTool.ts#L66) — re-detect "prescaled PT SUV" (two different ways, because native viewports lack .modality/.preScale.scalingParameters).
[isViewportPreScaled.ts](vscode-webview://0ojil4esr7c805nn2pf1dkdm93jglh77udcgtdrjo1lc8e4fnbfo/packages/tools/src/utilities/viewport/isViewportPreScaled.ts) and [getViewportModality.ts](vscode-webview://0ojil4esr7c805nn2pf1dkdm93jglh77udcgtdrjo1lc8e4fnbfo/packages/core/src/utilities/getViewportModality.ts) — per-family branches to recover prescale/modality state.
The tell is that the same question — "is this a prescaled PT/SUV image, and what's its modality?" — is re-derived in five places with different code, because the native PlanarViewport doesn't expose the StackViewport surface (.modality, .preScale.scalingParameters, .isPreScaled) the legacy consumers were written against. That's the friction you're sensing, and it's the right thing to want to remove.

Would a single Uint16 raw→display LUT simplify it?
Partly yes for rendering, but it can't be the universal mechanism, and it doesn't delete the modality concept — it relocates it. Three grounded points:

1. The CPU path already does this when not prescaled. [generateLut.ts:42](vscode-webview://0ojil4esr7c805nn2pf1dkdm93jglh77udcgtdrjo1lc8e4fnbfo/packages/core/src/RenderingEngine/helpers/cpuFallback/rendering/generateLut.ts) builds exactly a composite lut[stored] = vlutfn(mlutfn(stored)) (modality ∘ VOI) over the integer pixel range when isPreScaled is false. So your proposal is essentially "make the non-prescaled composite-LUT path the only path." Doing that would let you drop markCpuImagePreScaled, the Float32 storage selection, and the float-fallback toggling of isPreScaled in [BaseStreamingImageVolume.ts:369](vscode-webview://0ojil4esr7c805nn2pf1dkdm93jglh77udcgtdrjo1lc8e4fnbfo/packages/core/src/cache/classes/BaseStreamingImageVolume.ts) — a real loader/volume-type simplification.

2. It actually fixes the PT float problem rather than creating it. Today prescaling bakes SUV into a Float32Array, and floats can't index a LUT, so [renderGrayscaleImage.ts:71](vscode-webview://0ojil4esr7c805nn2pf1dkdm93jglh77udcgtdrjo1lc8e4fnbfo/packages/core/src/RenderingEngine/helpers/cpuFallback/rendering/renderGrayscaleImage.ts) needs a special per-pixel function VOI for PT. If pixels stay raw integer counts, a 65536-entry composite LUT renders PT fine and that special case disappears. This is the strongest argument for your instinct.

3. But it breaks on two hard edges and one semantic one:

GPU path isn't an indexed LUT. Volumes and the new PlanarViewport VTK image mapper use a vtkColorTransferFunction over a continuous scalar range, not a 64K-entry index array. You'd fold the affine modality transform into the TF domain instead — fine for affine, but it's a different mechanism, so "one Uint16 LUT everywhere" isn't achievable; you'd have two composing strategies.
Genuinely-float source data (RTDOSE, parametric maps, float rescale slopes) still can't index an integer LUT. Rare, but needs the function path to survive.
Tools need modality units, not display units. ROI mean/stddev, ProbeTool, and threshold segmentation read voxel values expecting HU/SUV ([ProbeTool.ts:657](vscode-webview://0ojil4esr7c805nn2pf1dkdm93jglh77udcgtdrjo1lc8e4fnbfo/packages/tools/src/tools/annotation/ProbeTool.ts), [getStatistics.ts:62](vscode-webview://0ojil4esr7c805nn2pf1dkdm93jglh77udcgtdrjo1lc8e4fnbfo/packages/tools/src/utilities/segmentation/getStatistics.ts)). A pixel→display LUT throws that domain away. Prescaling exists partly so tools read modality values straight off the voxels. Remove it and every such tool must apply slope/intercept/SUV itself.
The verdict / what would actually simplify
A raw-pixels + composite-LUT model is genuinely cleaner for the render path and removes the float-baking machinery — keep that idea. But the modality transform can't be deleted, only moved: the cheap affine (slope·x + intercept, ·suvbw) has to stay first-class and queryable for the GPU TF and for tools. So you don't get to "delete prescaling and the concept goes away."

The bigger win — the one that kills most of this PR's special-casing regardless of bake-vs-LUT — is a single scaling descriptor on the image/viewport: { rescaleSlope, rescaleIntercept, suvbw, modality, units } exposed uniformly by both StackViewport and PlanarViewport. Rendering reads it to build the composite LUT or TF; tools read it to convert raw→modality. That alone collapses markCpuImagePreScaled, the WindowLevelTool dual-detection, the isViewportPreScaled/getViewportModality per-family branches, and the scattered isPTPrescaledWithSUV checks into one source of truth — which is ~80% of the prescale pain in this PR. Whether you then also drop baking in favor of always-LUT is a separable, smaller decision.

If it's useful I can sketch what that descriptor + the two render-side consumers (CPU composite LUT, GPU TF folding) would look like, scoped to where this PR already touches.

packages/core/src/RenderingEngine/GenericViewport/Planar/planarImageEvents.ts
sedghi: Agreed, that's the right long-term direction: a single VIEWPORT_NEW_IMAGE that supersedes STACK_NEW_IMAGE / VOLUME_NEW_IMAGE, with both legacy events kept firing through a deprecation window so existing consumers (the OHIF scrollbar, synchronizers, stack prefetch, etc.) keep working. I'd rather land that as its own focused PR than fold it into this one since it touches event consumers across core, tools, and downstream. Tracking it as a follow-up.

packages/tools/src/tools/WindowLevelRegionTool.ts
sedghi: Addressed in 2900638bc. When the presentation (native) viewport has no resolvable source data id, the computed voiRange is now applied via the single-argument setDisplaySetPresentation (which targets the current binding) instead of being silently dropped; it is a no-op only when the viewport has no bound data at all.

packages/core/src/RenderingEngine/GenericViewport/Planar/VtkImageMapperRenderPath.ts
sedghi: Leaving this as `image.color ? 3 : 1` (comment clarified in c9391878a). In this render path a color image's vtkImageData is always built with 3 components: `getImageDataMetadata` derives `numberOfComponents` from the photometric interpretation (RGB/YBR -> 3), and `updateVTKImageDataWithCornerstoneImage` down-converts an RGBA (4-component) source to RGB in place. So the post-transform component count is 3 for color and 1 otherwise, and `color ? 3 : 1` matches what the imageData is actually built with.

Using `image.rgba` (or a literal 4) would mis-report a freshly loaded RGBA frame (rgba still true, `numberOfComponents` unset) as 4, never match the 3-component imageData, and force a fresh vtkImageData on every scroll - reintroducing the one-frame black flash this reuse path exists to prevent. Verified on a 94-frame RGB ultrasound: the scalars are 3-component and reuse stays on the fast path.

packages/tools/src/utilities/voi/windowlevel/extractWindowLevelRegionToolData.ts
sedghi: Addressed in 2900638bc. `extractImageDataGeneric` no longer throws when the current native slice has no backing image (reformatted/oblique volume plane); it warns and returns undefined, and the caller (`applyWindowLevelRegion`) guards on that and skips. Full reformatted-slice region window-leveling for native volume viewports needs slice-view plumbing on the generic viewport and is tracked as a follow-up.

packages/core/src/utilities/VoxelManager.ts
sedghi: Addressed in f76bdd082. A scalar expansion cached via `getScalarData(true)` is now flagged and invalidated on `setAtIJK`/`setAtIndex` writes, so a later voxel edit can't be masked by a stale cache. A real backing store (constructor / `setScalarData` / `_updateScalarData`-backed managers) is left untouched, so it stays a single boolean check on the write path.

packages/core/src/RenderingEngine/GenericViewport/Planar/PlanarViewport.ts
sedghi: Addressed in c9391878a. `DefaultPlanarDataProvider` no longer defaults `initialImageIdIndex` to 0; 'no slice requested' propagates as `undefined` through `PlanarPayload` (render-path seeds default the placeholder index with `?? 0`), so the volume acquisition view centers when nothing is requested while an explicit index - including 0 - is honored instead of being collapsed to center.

packages/core/src/RenderingEngine/GenericViewport/Planar/PlanarViewport.ts
sedghi: I believe this matches legacy behavior rather than introducing a new reset/echo: legacy `StackViewport.setStack` emits `VOI_MODIFIED` on mount (via `setVOI`) and the first-image `CAMERA_MODIFIED`, and the native emission goes through `notifyDataPresentationModified`, which honors `suppressEvents` the same way. The VOI/camera synchronizers already have echo-suppression, and the per-binding `volumeId` scoping was fixed in a5492df.

So I'd prefer not to blanket-suppress the mount emission, since OHIF relies on it to refresh the W/L readout, colorbar, and VOI/camera sync on a display-set change. I'll verify the specific synced-group scenario (load a series into one pane of a synced group and confirm it doesn't reset the others' slice/VOI any differently than legacy) and narrow the emission only if there's a real divergence. Happy to take a concrete repro if you have one.

Issue #2763, opened 2026-06-14
ESM-only packages ship dist/esm/*.js without "type": "module" (no root field, no nested package.json) → Node resolves output as CJS


benfsmith:
### Summary

The published ESM-only packages emit `dist/esm/**/*.js` that use ESM syntax (`import`/`export`), but none of the packages declare their output as ESM — there is **no `"type": "module"` in the package root `package.json` and no nested `dist/esm/package.json` containing `{"type":"module"}`**. Because the `.js` files carry no module-type signal, Node's resolver classifies them as **CommonJS**, which `publint` reports as `FILE_INVALID_FORMAT` (`actualFormat: "ESM"`, `expectFormat: "CJS"`).

### Affected packages

Confirmed on the published `3.33.5` and on `main` (`5.0.8`):

| Package | `type` field | `module` | nested `dist/esm/package.json` |
|---|---|---|---|
| `@cornerstonejs/core` | _absent_ | `./dist/esm/index.js` | _missing_ |
| `@cornerstonejs/dicom-image-loader` | _absent_ | `./dist/esm/index.js` | _missing_ |

The `exports` map points every subpath at `./dist/esm/*.js`, and each file uses ESM syntax (e.g. `dist/esm/Settings.js` → `export default class Settings {`), but the package is not declared as a module.

### Why it matters

- **Node ≥ 22/24 tolerates this via syntax detection; Node ≥ 25 tightened the loader** and rejects the shape at link time with errors like `SyntaxError: The requested module 'X' does not provide an export named 'Y'`. Consumers on newer Node who import these packages on a Node code path (SSR, tests, tooling, any non-bundled require/import) hit this.
- `publint` / `@arethetypeswrong` and similar packaging linters flag it as a format mismatch, which trips strict packaging audits in downstream monorepos.

Bundler-only consumers (webpack/Vite/Next) are unaffected because the bundler resolves the ESM directly — but the package is unusable/incorrectly-typed on a raw Node loader path.

### Reproduction

```bash
npx publint @cornerstonejs/core
# → FILE_INVALID_FORMAT for dist/esm/*.js (actualFormat ESM, expectFormat CJS)
```

Or, on Node ≥ 25, `import` any subpath on a Node (non-bundled) path.

### Suggested fix

Either is a small, localized change; both packages build with plain `tsc` into `dist/esm`, so option (2) is the most surgical and matches the common dual-package convention:

[...]

PR #2773, opened 2026-06-26
feat(packaging): native-Node-resolvable ESM (extensions, type:module, exports) + CI

sedghi:
## Summary

The published `@cornerstonejs/*` packages emit ESM into `dist/esm/**` but didn't fully declare it: no `type:module`, slightly inconsistent `exports`, and **extensionless relative imports**. Bundlers (webpack/Vite/Next — OHIF and most viewers) tolerate all of this, but a consumer loading the files on a **native Node code path** (SSR, server, Node test runners, packaging linters) cannot resolve them — Node 22–24 limp along via syntax detection, **Node ≥25 hard-fails at link time**, and `publint`/`@arethetypeswrong` flag `FILE_INVALID_FORMAT`. Reported in #2763.

This PR makes the published packages correct and self-describing for **both bundlers and native Node**, and adds CI so it can't regress. It is **non-breaking** (the files were already ESM; this makes Node classify and resolve them correctly) and applies to every published package: `core`, `tools`, `dicom-image-loader`, `nifti-volume-loader`, `polymorphic-segmentation`, `ai`, `labelmap-interpolation`, `adapters`, `metadata`, `utils`.

## Changes

**Packaging metadata**
- Declare ESM via a scoped `dist/esm/package.json` = `{"type":"module"}` (postbuild step), so Node treats the output as ESM **without** forcing ESM onto each package's root-level CommonJS config files (`.webpack/*.js`, `babel.config.js`, `karma.conf.js`).
- Tidy every `exports` map: `types` first, then `import`/`default`; drop the redundant standalone `node` condition; add `"./package.json": "./package.json"`.
- Fix a latent broken export in `adapters`: Rollup `preserveModules` never emitted `version.js` (not reachable from the entry) though `exports["./version"]` pointed at it — `src/version.ts` is now an explicit Rollup input.

**Real extensions (native-Node resolution)**
- Run `tsc-alias --resolve-full-paths` after `tsc` (and after `rollup` for `adapters`) so emitted relative imports in **both `.js` and `.d.ts`** carry explicit `.js` extensions (directory imports → `/index.js`). **Source stays extensionless** — this is a dist-only post-build rewrite, not a source codemod.
- The `dicom-image-loader` decode worker (`new Worker(new URL('./decodeImageFrameWorker.js', import.meta.url))`) and the codec `new URL(..., import.meta.url)` references are emitted **verbatim**, and the worker file ships as a real sibling, so consumer bundlers continue to handle them.

**Tests + CI (so this can't regress)**
- `scripts/validate-esm-packaging.mjs` (`pnpm run validate:packaging`): per package, verifies the `type:module` marker, that every relative import in `.js`/`.d.ts` resolves to a real file, that relative `new URL(...)` assets exist, then runs `publint`, `@arethetypeswrong/cli` (`esm-only` profile), and a native-Node leaf import. Negative-tested: it fails (exit 1) on a reintroduced extensionless import.
- `.github/workflows/validate-packaging.yml`: builds and validates on every PR across **Node 24 and 25** (25 is where #2763 hard-fails).

## Validation

`publint` is clean and `attw` reports `node16 (from ESM): 🟢` for all 10 packages; the full validator passes for all 10 from a clean `rm dist → build → validate`.

## Notes for release

- **Non-breaking → minor.** No public API or `exports` surface change; bundler consumers resolve identically. Lands as a `feat:` (semantic-release cuts the minor).
- **TypeScript `moduleResolution` caveat:** types now use `.js`-extensioned ESM imports. Consumers on the legacy `moduleResolution: "node"` (node10) — which can't map a `.js` specifier to its `.d.ts` and ignores `exports` — may see some deep types resolve as `any`; switch to `"bundler"`, `"node16"`, or `"nodenext"`. Type-only, no runtime impact. Documented in `migration-guides/5x`.
- The large `pnpm-lock.yaml` diff is mostly a **pnpm 11.5.2 reserialization** of the previously stale-format lockfile (no dependency version changes) plus the three new dev dependencies (`tsc-alias`, `publint`, `@arethetypeswrong/cli`).

## Out of scope (separate follow-ups)

Removing the orphan UMD bundles (a breaking change → the eventual major), lazy-loading vtk.js, and DOM-access guards so the heavy package *entries* (`core`/`tools`) also execute in DOM-less Node. Native-Node WASM decode in `dicom-image-loader` additionally needs a small source change (bare-specifier `new URL` → `import.meta.resolve`) and is independent of this packaging work.

Refs #2763

## Summary by CodeRabbit

* **New Features**
  * ESM builds now write a `dist/esm/package.json` marker (`type: "module"`), and multiple packages expose `./package.json` publicly.
  * Public ESM entry points now consistently include `default` alongside `import` and type definitions.
* **Bug Fixes**
  * Improved `./version` subpath ESM resolution by adjusting ESM build outputs and export conditions.
* **Tests / CI**
  * Added a “Validate ESM Packaging” workflow and a `validate:packaging` script to verify ESM correctness.
* **Chores**
  * Reduced Playwright PR concurrency (3 → 2).


packages/adapters/package.json
```diff
@@ -12,54 +12,54 @@
...
     "./version": {
-      "node": "./dist/esm/version.js",
-      "import": "./dist/esm/version.js",
-      "types": "./dist/esm/version.d.ts"
-    }
+      "types": "./dist/esm/version.d.ts",
+      "import": "./dist/esm/version.js"
+    },
+    "./package.json": "./package.json"
   },
```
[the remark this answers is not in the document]
sedghi: Addressed in the latest push (branch force-updated to 3693c537e): every runtime export entry now ends with a `default` condition pointing at the same ESM file, so `require(esm)` resolves on Node >=22.12 and `require("@cornerstonejs/...")` no longer throws `ERR_PACKAGE_PATH_NOT_EXPORTED`. We intentionally did NOT add a `require`->CJS build, since a dual ESM+CJS package would re-introduce the dual-package hazard for the stateful singletons in cs3d (image cache, rendering-engine and metadata registries). `publint` is clean across all published packages; rationale is in the updated PR description.

packages/dicomImageLoader/package.json
```diff
@@ -26,48 +26,57 @@
...
     },
     "./types": {
       "types": "./dist/esm/types/index.d.ts"
     },
     "./types/*": {
       "types": "./dist/esm/types/*.d.ts"
-    }
+    },
+    "./package.json": "./package.json"
   },
```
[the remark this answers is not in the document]
sedghi: Addressed in the latest push (branch force-updated to 3693c537e): every runtime export entry now ends with a `default` condition pointing at the same ESM file, so `require(esm)` resolves on Node >=22.12 and `require("@cornerstonejs/...")` no longer throws `ERR_PACKAGE_PATH_NOT_EXPORTED`. We intentionally did NOT add a `require`->CJS build, since a dual ESM+CJS package would re-introduce the dual-package hazard for the stateful singletons in cs3d (image cache, rendering-engine and metadata registries). `publint` is clean across all published packages; rationale is in the updated PR description.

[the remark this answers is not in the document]
sedghi: To confirm explicitly: ESM-only is intentional, and this is **not** a breaking change for CommonJS consumers — because no `require()`-able CJS build was ever published, so there is nothing being dropped.

The currently published `@cornerstonejs/dicom-image-loader@5.0.14` (and `origin/main`) ship:

```jsonc
"exports": { ".": { "import": "./dist/esm/index.js", "types": "./dist/esm/index.d.ts" } }
// no "main", no "require"/"node"/"default" condition, no dist/cjs
```

So `require('@cornerstonejs/dicom-image-loader')` already throws `ERR_PACKAGE_PATH_NOT_EXPORTED` on the released package today.

This PR's `default` condition therefore **improves** CommonJS interop rather than removing it: on Node ≥22.12, `require()` now resolves the ESM file via `require(esm)`, where it previously failed outright. We're intentionally **not** adding a `require`→CJS build — a dual ESM+CJS package would re-introduce the dual-package hazard for the stateful singletons in cs3d (image cache, rendering-engine and metadata-provider registries).

The broader ESM-only stance is the intended long-term direction; consumer-facing migration guidance (`require()` needs Node ≥22.12, otherwise use `import`) will accompany the next major release. Thanks for confirming the types-first / `./package.json` / `./types` changes look right.

package.json
```diff
@@ -58,6 +59,7 @@
     "webpack:watch": "pnpm -r --filter=!docs run webpack:watch"
   },
   "devDependencies": {
+    "@arethetypeswrong/cli": "^0.18.4",
```
jbocce: exact versioning please

package.json
```diff
@@ -144,6 +146,7 @@
     "postcss-loader": "7.3.4",
     "postcss-preset-env": "8.5.1",
     "prettier": "3.6.2",
+    "publint": "^0.3.21",
```
jbocce: exact versioning please

package.json
```diff
@@ -158,6 +161,7 @@
     "stylelint-config-recommended": "12.0.0",
     "ts-loader": "9.5.1",
     "ts-node": "10.9.2",
+    "tsc-alias": "^1.8.17",
```
jbocce: you know...

scripts/validate-esm-packaging.mjs
```diff
@@ -0,0 +1,394 @@
+// Validates that every published @cornerstonejs/* package ships correct,
+// natively-resolvable ESM, so bundler AND native-Node consumers can always
+// load it. Guards the packaging work from https://github.com/cornerstonejs/cornerstone3D/issues/2763
+// from ever regressing. Run AFTER building (dist/esm must exist).
+//
+// Per package it checks:
+//   1. ESM marker     - dist/esm/package.json declares {"type":"module"}
```
jbocce: indentation is off

jbocce (changes requested):
Please use exact versioning in the package.json files.

tests/stackAnnotationTiled.spec.ts
```diff
@@ -22,7 +22,7 @@ test.describe('Stack Annotation Tools - Tiled', async () => {
       0,
       // Absorb sub-pixel font drift on the "138 mm" label across CI
       // environments; a missing/mis-drawn annotation differs by far more.
-      { maxDiffPixelRatio: 0.003 }
+      { maxDiffPixelRatio: 0.003, timeoutMs: 20000 }
```
jbocce, line 25: I am surprised you needed this. In general I have found that the timeout for `visitExample` needs to be increased for these types of tests. FYI.

jbocce (approved):

PR #2780, opened 2026-07-02
feat: add ClickSegmentTool for click-to-segment lesion segmentation

sedghi:
### Context

Segmenting a PET lesion today means picking a strategy and tuning parameters
(disk radius, mean/percent thresholds, deltas) before you get a result you can
trust. This PR adds **`ClickSegmentTool`** — a configuration-free,
click-to-segment lesion tool built for PET-style hot lesions but
modality-agnostic. There is nothing to configure: no disk radii, no
strategies, no deltas.

The tool deliberately does not promise to segment *anything* you click: it
only accepts clicks on a coherent, contrast-distinct, lesion-scale region, and
the hover cursor tells you up front whether a click will land.

### Changes & Results

**New tool — `ClickSegmentTool`** (exported publicly from `@cornerstonejs/tools`):

- **Hover previews segmentability.** A `+` cursor means a click here produces
  a meaningful segment; a blocked cursor means it will not (flat area, noise
  speck, or an unbounded region), and a pending cursor means the region is
  still being evaluated. Clicks on blocked spots are no-ops. Verdicts attach to
  *regions*, not pixels, so the cursor does not flicker as you move within one
  structure.
- **Click derives a one-sided intensity threshold dynamically** — everything at
  least as intense as the clicked structure and connected to it — so the
  hottest core of a lesion is always included and there are no interior holes.
- **`expand()` / `shrink()` step deterministically** along the growth curve
  measured at click time: each step clears the previous result and refills at
  the next/previous stable threshold, so the segment visibly grows or shrinks
  every step rather than silently no-op'ing.
- **Lesion-ness is judged by shape, not size.** The would-be segment must be
  one coherent, self-contained entity (compactness against its 3D bounding box,
  plus a through-slice self-containment check). Large solid lesions pass;
  sprawling webs (chained bones, noise) are rejected. There is no size or
  extent cap.
- **Undo/redo.** Each click and each expand/shrink step is recorded as one
  `LabelmapMemo` on the segmentation history stack, so it steps through
  click operations the same way it does through brush strokes.

**Supporting utilities** (under `packages/tools/src/utilities/segmentation/`):

- `runFloodFillSegmentation` — orchestrates the fill: dynamic band resolution,
  lazy 3D flood, commit, external/internal island removal, and optional
  history recording via a `historyVoxelManager` option.
- `floodFillSliceLazy` — 3D BFS flood fill with per-slice visit masks and
  on-demand slice loading, with budget and shape-gate hooks.
- `adaptiveRegionIntensityRange` — the config-free adaptive strategy (VOI-mapped
  display bytes, seed snapping, one-sided band, quiet-run tolerance selection).
- `commitSliceMasksToLabelmap`, `createEnsureSliceLoadedForVolume`,
  `floodFillIslandRemoval`, `neighborhoodStats`, `getViewportVoiMappingForVolume`.
- Core: `viewportVoiIntensityMapping` utilities to map between display VOI and
  raw scalar intensity.

**Demo & helpers:**

- New `clickSegment` example (PET-only single viewport, hover-cursor legend,
  Shrink/Expand/Clear, Undo/Redo).
- `validateAndSortVolumeIds` helper (spatial slice ordering/validation) and
  `addDropdownToToolbar` support for async (Promise) option sources.

### Testing

1. `yarn run example clickSegment` and open the example.
2. Hover over the PET image: a `+` cursor appears over lesions, a blocked
   cursor over background/flat areas.
3. Click a lesion — it segments in one click with no interior holes.
4. Use **Expand** / **Shrink** and confirm the segment visibly grows/shrinks
   each step; **Undo** / **Redo** (or Ctrl/Cmd+Z, Shift+Z, Ctrl+Y) step through
   each operation.

Unit tests (Jest) added and passing:
`adaptiveRegionIntensityRange`, `floodFillSliceLazy`, `floodFillIslandRemoval`,
`commitSliceMasksToLabelmap`.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] "OS: macOS"
- [x] "Node version: 22.x"
- [x] "Browser: Chrome"

## Summary by CodeRabbit

* **New Features**
  * Added a click-to-segment tool with hover feedback, expand/shrink controls, refresh, and undo/redo support.
  * Added adaptive flood-fill segmentation for more precise region selection.
  * Added support for loading and applying slice-based segmentation updates more reliably.
  * Added new volume validation and VOI/intensity mapping helpers for better segmentation behavior.
  * Added a new demo example for the click-to-segment workflow.


wayfarer3130 (changes requested):
I've looked over it generally and am happy with it, but did find a few things from clod:
Code correctness & risks
growCutLog.setLevel('info') in core logger.ts is a global side effect at import time. [logger.ts]. This forces info-level logging for the growCut logger regardless of the host app's configuration, and runFloodFillSegmentation emits many log.info calls per click/expand plus ungated console.time/console.timeEnd/console.warn/console.info. In a library this is real console/perf noise in production. Recommend defaulting to warn (or leaving level to the consumer) and gating the console.time* timing labels behind the existing ENABLE_VERBOSE_FLOOD_FILL_LOGS flag.

floodFillIslandRemoval.ts (519 lines) is a fork of the existing islandRemoval.ts (383 lines). Same SegmentationEnum, same IslandRemoval class and algorithm, with added features (getInternalFilledPoints, usingPreviewLayer, no-preview warning, verbose logging). Two near-identical island-removal implementations in the same directory is a maintenance hazard — divergent bug fixes are almost guaranteed. Strongly prefer extending the existing class (or adding the new hooks to it) over forking. The base tool still imports the old one (../../utilities/segmentation/islandRemoval), so both will live on indefinitely.

Oblique-view path throws from preMouseDownCallback. super.preMouseDownCallback throws 'Oblique view is not supported yet' and this override doesn't wrap it, producing a rejected promise on click. Hover already blocks oblique via _isOrthogonalView → BLOCKED_CURSOR, so in practice the click is gated, but a click without a fresh probe (e.g. probe stale/expired) could still reach this throw. Consider catching it and calling notifySegmentationError for a clean UX.

validateAndSortVolumeIds is added and exported but unused by the new example, which sorts with csUtils.sortImageIdsAndGetSpacing directly. Either wire it into the example (it's the more robust validator) or drop it from this PR — as-is it's dead weight with no exercising caller.

Performance
Hover is heavier than it looks. Each throttled hover flush runs probeAdaptiveRegion (a ~193×193 window sweep + bucket-queue flood) once for the pointer plus up to 4 more times in neighborsAgree — ~5 full window analyses synchronously on the main thread per flush, before the async 3D dry-run. On large/fine-spaced volumes this is a lot of redundant per-hover compute. Consider caching the center probe result and/or sampling neighbors more cheaply (the neighbor votes don't need the full growth-curve sweep).

sedghi:
@wayfarer3130 Thanks for the careful pass — all five points were valid and are addressed in the latest push (also merged latest main, no conflicts):

- **growCutLog level / console noise**: removed the import-time `setLevel('info')` — the level is now left to the consumer like the other cs3d loggers (a comment on `growCutLog` documents how to opt in). The `console.time`/`timeEnd` timing labels in `runFloodFillSegmentation` are gated behind the existing `ENABLE_VERBOSE_FLOOD_FILL_LOGS` flag, and the bare `console.warn` calls now go through the logger, so the default production experience is silent.

- **floodFillIslandRemoval fork**: reworked as a subclass of the existing `IslandRemoval` instead of a copy. The base class gained two minimal hooks (`fillInternalEdge`/`maxInternalRemove` became `protected`, and `removeInternalIslands` calls an overridable no-op `onInternalPointFilled(point)` for each hole voxel it paints). The subclass now only carries its real deltas: the delta-safe `removeExternalIslands` (preview-layer guard + source-value check), `getInternalFilledPoints()`, and the verbose logging. Net effect: the algorithm lives in one place again (~520 lines down to ~200, all of them additive behavior).

- **Oblique click throw**: `ClickSegmentTool.preMouseDownCallback` now wraps the base setup in try/catch; a click that reaches the oblique guard (e.g. with a stale/expired hover probe) surfaces through `notifySegmentationError` and returns `false` instead of producing a rejected promise.

- **validateAndSortVolumeIds**: wired into the clickSegment example as suggested — it replaces the direct `sortImageIdsAndGetSpacing` call and warns with the specific reason when the series is not a clean volume.

- **Hover cost**: fixed at the source rather than by caching. The adaptive probe's bucket-queue sweep already visits every window pixel, so it now records the tolerance level at which each of the click's 4 in-plane neighbors joins the region (`clickNeighborJoinLevels` on the probe result). `neighborsAgree` reads its votes off that with the same consensus semantics (2-of-4, edge leniency), so each hover flush runs one window analysis instead of five; the remaining per-hover cost is the single probe plus the budgeted async 3D dry-run.

All segmentation utility jest suites pass against the refactor (including the internal-fill tracking test now running through the subclass + base hook path).

wayfarer3130 (approved):
Thanks for the fixes.

PR #2782, opened 2026-07-03
feat(rendering): unify render backend selection with live GPU/CPU switching

sedghi:
### Context

GenericViewport's planar render-path selection was spread across several ad-hoc
signals: a `gpuTier` config, per-mount `forceCPU`, byte-size `cpuThresholds`,
and scattered WebGL/texture probes in `init.ts`. There was no first-class way to
express "render this on GPU / CPU / auto", no way to switch backends at runtime
without a page reload, and no signal for applications to react to GPU
degradation (context loss).

### Changes & Results

**Render backend model**
- New `RenderBackend` enum (`gpu` | `cpu` | `auto`) with `renderBackend`
  configurable globally (`rendering.planar.renderBackend`) and per display set
  (mount option). Replaces `gpuTier`, `forceCPU`, and `cpuThresholds`.
- `getRenderBackend()` / `getEffectiveRenderBackend(override?)` /
  `setRenderBackend()` centralize the precedence ladder (per-mount pin > global
  pin > `auto` resolved from capability detection). `setUseCPURendering` /
  `resetUseCPURendering` are kept but deprecated in favor of `setRenderBackend`.

**Capability detection**
- New `utilities/renderingCapabilities` probes WebGL/WebGL2 availability,
  `MAX_TEXTURE_SIZE`, the unmasked renderer string, software-rasterizer
  detection, and per-format texture support (norm16/float/halfFloat draw +
  readback). Results are memoized in-process and cached in `localStorage`,
  keyed by renderer string, WebGL2 availability, and a probe version.

**Live backend switching**
- `setRenderBackend()` live-swaps every mounted GenericViewport's render paths
  in place — viewport ids, mounted data, cameras, presentation state, and tool
  annotations are preserved; only the render paths rebuild. Reversible in both
  directions at runtime.
- New degradation events on the eventTarget: `RENDER_BACKEND_CHANGED`,
  `WEBGL_CONTEXT_LOST` / `WEBGL_CONTEXT_RESTORED`, `RENDER_PATH_ERROR`, and
  `RENDERING_PIPELINE_CHANGED`. Cornerstone never switches backends on its own;
  applications listen and call `setRenderBackend('cpu')` themselves.

**Segmentation rehydration across a switch**
- Labelmap representations now re-reconcile after a live backend switch: the
  rebuilt viewport actors are re-styled via a `RENDERING_PIPELINE_CHANGED`
  listener, and the labelmap render plans force a remount when a surviving
  actor has an incompatible shape (image-mapper vs volume), so stack, MPR
  volume, and PET labelmaps all survive gpu↔cpu switches.

### Testing

- New example `genericRenderBackendSwitch` (and `genericPetRenderBackendGrid`)
  live-switch the backend of stack, MPR, and PET viewports; verified all three
  labelmaps and annotations survive gpu→cpu→gpu round-trips and CPU slice
  scrolling.
- Unit tests: `planarRenderBackend.jest.js`, `renderingCapabilities.jest.js`,
  labelmap render-plan specs; e2e `genericRenderBackendSwitch.spec.ts`.

## Summary by CodeRabbit

* **New Features**
  * Added live render-backend switching (auto/GPU/CPU) with per-viewport backend pinning and smooth pipeline updates.
  * Enhanced demos with multi-viewport CT/MPR/PET, mock segmentations/annotations, richer status/event reporting, and WebGL context-lost simulation.
  * Introduced rendering capability detection APIs to drive planar backend selection.
* **Bug Fixes**
  * Improved stability of backend render-path switching and segmentation refresh after pipeline changes.
  * Strengthened labelmap remounting by honoring actor-entry compatibility across backend switches.
* **Tests**
  * Added/expanded automated coverage for backend switching, capability detection, render-path/pipeline behavior, and CPU fallback correctness.

### DIFF
--- a/packages/core/src/utilities/textureSupport.ts
+++ b/packages/core/src/utilities/textureSupport.ts
@@ -27,15 +27,6 @@ interface FormatProbe {
   glDataType: (gl: WebGL2RenderingContext, ext?) => number;
 }
 
-const NO_SUPPORT: TextureFormatSupport = {
-  norm16: false,
-  norm16Linear: false,
-  float: false,
-  floatLinear: false,
-  halfFloat: false,
-  halfFloatLinear: false,
-};
-
 /**
  * Creates the single offscreen context and point-sprite program shared by all
  * format probes. Context creation and shader compilation dominate the probe
@@ -160,10 +151,18 @@ function probeFormat(
   }
 }
 
-export function getSupportedTextureFormats(): TextureFormatSupport {
+/**
+ * Probes each texture format through a real draw + readback. Returns `null`
+ * when the probes could not run at all (probe context creation failed — e.g.
+ * the browser's live-context limit was hit — or the context was lost
+ * mid-probe), so callers can tell "probing failed this run" apart from "the
+ * GPU genuinely does not support these formats" and avoid persisting the
+ * former.
+ */
+export function getSupportedTextureFormats(): TextureFormatSupport | null {
   const gl = createProbeContext();
   if (!gl) {
-    return { ...NO_SUPPORT };
+    return null;
   }
 
   const norm16TexData = new Int16Array([
@@ -215,10 +214,14 @@ export function getSupportedTextureFormats(): TextureFormatSupport {
     }),
   };
 
+  // A context lost during the probes yields spurious all-false results
+  // (readPixels returns zeros); report failure instead of fake answers.
+  const contextLost = gl.isContextLost();
+
   const webglLoseContext = gl.getExtension('WEBGL_lose_context');
   if (webglLoseContext) {
     webglLoseContext.loseContext();
   }
 
-  return result;
+  return contextLost ? null : result;
 }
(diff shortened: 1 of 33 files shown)


IbrahimCSAE:
🔥🔥🔥🔥🔥🔥🔥

packages/core/src/utilities/textureSupport.ts
```diff
@@ -131,30 +159,30 @@ export function getSupportedTextureFormats() {
       internalFormat: (gl, ext) => ext.R16_SNORM_EXT,
       glDataType: (gl) => gl.SHORT,
     }),
-    // float: main({
```
wayfarer3130, line 134: Can you quantify how much extra startup time this imposes, and exactly how often?  I'd rather see this get stored/re-used as needed rather than run every time and impose a startup cost.

wayfarer3130 (approved):
Just the one comment about the startup time.  Otherwise I think this is a solid improvement to the stability of CS3D viewports.

packages/core/src/utilities/textureSupport.ts
sedghi: Here is the analysis

| | Detection cost | Contexts | Formats probed |
|:---|:---|:---|:---|
| main, every load | 14.6 ms | 3 | 2 |
| branch, first visit (cold) | 13.3 ms | 2 | 6 |
| branch, every load after (warm) | 5.9 ms | 1 | 0 (localStorage hit) |

PR #2784, opened 2026-07-05
fix(segmentation): labelmap edge projection over MIP for generic planar viewports

sedghi:
### Context

Labelmap edge projection over MIP only worked on legacy volume viewports; generic planar viewports had no working labelmap-over-MIP path at all. This PR adds that capability, with a new `labelmapMIPGeneric` example.

The generic-viewport slab label outline relies on the slab-projected label outline added to the vtk.js `ImageResliceMapper` in Kitware/vtk-js#3541 (which also carries a fix that keeps mapper fragment blocks compatible with render-pass postambles - early returns skipped the OIT weight write and composited the outlines white - and a slab-loop exit that keeps huge slab thicknesses proportional to the in-volume march). **Until that vtk.js change is released, this PR is blocked**; running the example requires a linked vtk.js build containing that PR.

The editing-performance and legacy-path fixes that do not depend on vtk.js were split out into #2797, which can merge independently.

### Changes & Results

**Generic planar viewports (new capability + example `labelmapMIPGeneric`)**

- The labelmap mounts as a separate overlay display set and now inherits the source display set's `slabThickness`, so labels project across the same slab as the MIP instead of reflecting only the current slice.
- Segmentation overlays keep the transfer functions assigned by the segmentation styling; previously the data-presentation update reapplied a VOI grayscale ramp over the labelmap colors.

**Editing performance in the generic render path**

- The generic volume-slice render path defers volume-modified repaints of slab-projecting overlays to a 240 ms trailing debounce (a slab-projecting overlay re-marches the whole slab per fragment, far too expensive to repeat for every brush-stroke event), and no longer invalidates the full 3D texture per modification - brush edits upload only the modified slices through the existing streaming-texture partial update path.
- #2797 complements this on the tools side by deferring the segmentation-render trigger itself for projection-heavy viewports.

Measured on the example (PET volume, Apple Silicon): brush strokes cost ~1.3 ms of main thread per pointer move, the projection viewport refreshes ~240 ms after the stroke ends, and rotating the generic MIP viewport with a whole-volume slab (50000) plus outline overlay renders at ~30 ms/frame.

### Testing

- `pnpm run example labelmapMIPGeneric` (requires linked vtk.js with Kitware/vtk-js#3541): click "Add Labelmap Representation" - the MIP viewport should show outlines in the segment colors (not white). Draw on the left viewport with the brush: drawing stays responsive, and the MIP outline updates shortly after the stroke ends. Rotate the MIP viewport with the mouse wheel. Delete and re-add the representation.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] "OS: macOS (Darwin 25.5.0)"
- [x] "Node version: 22.22.2"
- [x] "Browser: Chrome for Testing (latest)"


sedghi:
@salimkanoun also speed improvements

salimkanoun:
Amazing, thanks!

PR #2796, opened 2026-07-10
feat(webgpu): experimental WebGPU render backend for GenericViewport stacks and MPR

sedghi:
### Context

Groundwork for evaluating WebGPU as the rendering backend: an opt-in `webgpu` render backend for planar GenericViewports covering image stacks and volume-slice (MPR) rendering through the vtk.js WebGPU view API.

Depends on vtk.js fixes: Kitware/vtk-js#3547 (ImageMapper projection + imageData texture invalidation) and Kitware/vtk-js#3548 (translucent-pass depth for coplanar overlays, needed for labelmap segmentation). Both merged upstream and shipped in vtk.js v36.4.2; this branch now pins `@kitware/vtk.js` 36.4.2, so no linked build is required.

### Changes & Results

- `registerWebGPURenderBackend()` (opt-in, throws without `navigator.gpu`): registers the `webgpu` backend with self-rendering image (`webgpuImage`) and volume-slice (`webgpuVolume`) render paths; each viewport gets a ref-counted vtk.js WebGPU render window whose frames are blitted into the `cpu` surface canvas after the device work completes. Volume rendering materializes one shared, ref-counted mapper-input scalar array per volume (cornerstone volumes are image-backed and own no contiguous array).
- Planar shared code and the labelmap display resolve render-mode kind/surface through the render-backend registry instead of hardcoded `ActorRenderMode` literals (image and volume kinds), so extension render modes participate in view resolution, view references, scrolling, labelmap styling, and the labelmap image-mapper plan. `isImageRenderMode` / `isVolumeRenderMode` are exported from core. The `addImages` overlay gate keys on the CanvasActor `cpuVolume` mode rather than the composited surface.
- Perf: brush `setValue` resolves labelmap bindings once per stroke instead of per voxel; labelmap slice extraction walks flat voxel indices incrementally (original per-voxel IJK loop kept as fallback; obliques are unaffected — axis-aligned extraction only) and reuses slice buffers.
- New examples: `genericWebGPUStack` (stack + tools + annotations + segmentation editing on a fake labelmap) and `genericWebGPUOrtho` (sagittal/coronal CT MPR + mock ellipsoid labelmap + tools), both with live webgpu/gpu/cpu switching, a `?renderBackend=` URL parameter, and a debug panel reporting the effective backend, mounted render mode, and WebGPU adapter. New Generic Viewports section in the example index.

Verified on the webgpu backend: stack and MPR rendering with pixel parity to WebGL, wheel scroll with exact pixel round-trip, zoom/pan/VOI/invert/rotation, annotation tools, labelmap segmentation rendering and editing with segment colors matching the gpu and cpu backends, and live backend switching without errors. JS heap holds a single shared volume copy regardless of viewport count (verified via heap snapshots).

### Testing

- vtk.js 36.4.2 (pinned) contains the fixes above; a plain `pnpm install` is enough, no linked build needed.
- Run the `genericWebGPUStack` and `genericWebGPUOrtho` examples in Chrome; the debug panels should show `webgpuImage` / `webgpuVolume` and the WebGPU adapter. Use the backend buttons (or `?renderBackend=gpu|cpu`) to compare against gpu/cpu.

#### PR

#### Code

  etc.)

  additions or removals. (Experimental opt-in API; not documented on the docs site yet by design.)

#### Tested Environment

- [x] "OS: macOS 15"
- [x] "Node version: 22.22.2"
- [x] "Browser: Chrome (WebGPU, Apple Metal adapter)"

## Summary by CodeRabbit

- **New Features**
  - Added experimental WebGPU render paths for planar image stacks and WebGPU volume-slice (MPR) viewports, including per-viewport WebGPU window management and backend availability/debug controls.
  - Added interactive WebGPU demo examples (generic stack/orthographic MPR, PET MIP) with live render-backend switching and debug overlays/toolbars.
- **Performance**
  - Reduced per-slice allocations via cached slice pixel buffers and faster voxel extraction.
  - Improved segmentation/labelmap interaction by caching per-stroke lookups.
- **Compatibility / Bug Fixes**
  - Broadened image/volume render-mode detection across planar and labelmap/segmentation workflows.
  - Fixed CPU-surface sizing/guarding, improved view-reference rescaling on reorientation, and tightened 3D clipping behavior.


daker:
Nice 👍🏼

wayfarer3130:
Superseded by #2894 — a request for a **vtk-wasm** render backend, delivered as a separable rendering module covering planar, MPR and 3D.

The rejection here is about direction, not quality. vtk.js's WebGPU work is not going to be completed; the supported route to WebGPU is `vtk-wasm`, with vtk.js expected to be deprecated in its favour. Extending the vtk.js WebGPU view API therefore builds on a path that will not be maintained.

Most of what this branch established carries forward into #2894 and is credited there:

- the **registry-driven render-mode resolution** — replacing hardcoded `ActorRenderMode` literals with `isImageRenderMode` / `isVolumeRenderMode` / `getRenderSurfaceForRenderMode` across planar view resolution, view references, scrolling and labelmap styling. That refactor is what makes *any* extension backend possible, and it is good independently of which backend consumes it;
- the **examples with live backend switching**, `?renderBackend=`, and the debug panel;
- the shape of the extension seam, and the specific places it turned out to be incomplete.

**Several changes on this branch are not WebGPU-specific and should land on their own rather than being lost with it:**

- **iOS Safari half-float volume prescaling** — `getFilterableFloatTexturePrecision()`, `setUseHalfFloatForFloatData()` on `vtkStreamingOpenGLTexture`, and the two mapper call sites. This fixes R16F-filterable / R32F-non-filterable devices, where modality prescaling is currently disabled outright. Independent of WebGPU.
- **3D volume clipping-range banding** — the `resetCameraClippingRange()` call in `VtkVolume3DRenderPath`; without it the ray sampling is quantized by the ±1e6 fallback range and the render shows visible banding and streaks. That is a live defect in the WebGL 3D path.
- **Reorientation scale preservation** — `getReorientationScale()` in `PlanarViewReferenceController`, stopping a rotating view reference from pulsing in and out as the fit parallel scale is recomputed per plane.
- **Performance** — per-stroke brush binding resolution in `setValue`, incremental flat-index labelmap slice extraction, and slice pixel buffer reuse.

Could these be split out as their own PRs?

wayfarer3130:
## Revised position — not rejected, if it moves out of core

@sedghi I've gone back through the full diff and I'm withdrawing the rejection. I want to take a different route than dropping this.

**The offer:** package the WebGPU backend as its own module — `@cornerstonejs/rendering-webgpu-vtkjs` or similar — marked `@deprecated` from day one, not installed by default, not in the core dependency graph. On those terms I'm happy to carry it. A module nobody installs costs nothing to keep, and it stops being a maintenance liability the moment vtk-wasm (#2894) lands, without anyone having to delete working code.

**What changed my mind:** of 43 changed files, only four core/tools edits are actually WebGPU-specific. Everything else is either the generic registry refactor — `ActorRenderMode` literals giving way to `isImageRenderMode` / `isVolumeRenderMode` / `getRenderSurfaceForRenderMode` — or the four independent fixes in my previous comment. The branch is far less invasive than its line count suggests, and the registry refactor is work we want regardless of which backend consumes it.

**But "entirely outside core" is not achievable, and I don't want to pretend otherwise.** Three things have to come from core, and I'd rather state them as the terms up front than have you discover them mid-rework:

### 1. Core has to export a render-path SDK

The two WebGPU paths import nine internal modules. None of them are exported from `packages/core/src/index.ts` today — I checked all nine:

- `GenericViewport/Planar/buildPlanarActorEntry`
- `RenderingEngine/helpers/planarImageRendering` — `applyPlanarImagePresentation`, `createVTKImageDataFromImage`, `getDefaultImageVOIRange`, `updateVTKImageDataGeometryFromImage`
- `Planar/planarRenderCamera` — `applyPlanarICameraToActor`, `applyPlanarICameraToRenderer`
- `Planar/planarRenderPathProjection` — `resolvePlanarRenderPathProjection`, `resolvePlanarRenderPathCurrentImageIdIndex`, `getPlanarRenderPathActiveSourceICamera`
- `Planar/planarImageEvents` — `triggerPlanarNewImage`, `triggerPlanarVolumeNewImage`
- `Planar/planarVolumePresentation` — `applyPlanarVolumePresentation`
- `Planar/CpuImageSliceRenderPath` — `buildPlanarImageData`
- `RenderingEngine/helpers/setDefaultVolumeVOI`
- `RenderingEngine/renderingEngineCache`

Plus the type contracts: `ViewportArchitectureTypes` (`RenderPath`, `RenderPathDefinition`, `RenderPathAttachment`, `LoadedData`, `DataAddOptions`), `PlanarViewportTypes`, `planarRuntimeTypes`. And `isImageRenderMode` / `isVolumeRenderMode`, which this branch already adds to `index.ts` and which `@cornerstonejs/tools` needs too.

Roughly 15 named exports and a types entry point. **This is the reason I'd rather do it this way than drop the branch:** that surface is exactly the seam the vtk-wasm work in #2894 needs. Defining it against a real second backend that already exists is a far better test of it than designing it in the abstract and discovering the gaps later.

### 2. Backend-owned surfaces, so the two workarounds can go

Two of the four WebGPU-specific edits exist only because `RenderSurface` is a closed `'vtk' | 'cpu'` union. The backend can only declare `surface: 'cpu'` and blit, yet it still hosts vtk actors — so both `PlanarViewport.supportsImageOverlays` and the tools-side `isGpuVolumeSliceRenderMode` had to special-case `renderMode !== ActorRenderMode.CPU_VOLUME` rather than ask the surface. The branch's own comment says it plainly:

> this must not key on the composited surface — extension backends such as webgpu blit into the cpu surface canvas while still hosting vtk actors

A registry capability for "hosts vtk actors" (or backend-declared surfaces, which is the fuller fix and is already on the #2894 list) makes both special cases disappear. I don't want either literal landing in core.

### 3. The registry augmentation belongs in the plugin

This branch adds `webgpu` to core's `RenderBackendRegistry` and `RenderBackendConstants` interfaces. That contradicts the documented design of that file, which has extensions augment those interfaces from their own `.d.ts`. Move it to the plugin package, along with the four `registerWebGPURenderBackend` / `isWebGPURenderingAvailable` / `getWebGPUViewportDebugInfo` / `setWebGPUViewportBackground` exports in `index.ts` — those become the plugin's public entry point.

### One packaging detail

Core pins `"@kitware/vtk.js": "36.4.1"` exactly, and this branch needs 36.4.2 for the two upstream fixes (Kitware/vtk-js#3547, #3548). A single build can't resolve two versions, so core's pin takes the patch bump regardless of where the plugin lives. That's fine — it's a patch release containing fixes we want.

### Suggested split

1. The four non-WebGPU fixes as their own PRs (previous comment) — iOS half-float, 3D clipping banding, reorientation scale, the brush/slice performance work. These shouldn't be gated on any of the rest.
2. Registry completion in core: the SDK exports, the actor-hosting capability, and the surface question.
3. The registry refactor (`ActorRenderMode` literals → registry queries) in core and tools — generic, wanted either way.
4. This branch reduced to `@cornerstonejs/rendering-webgpu-vtkjs`, `@deprecated`, opt-in install, registered through the add-on pattern we already use for `polySeg`.

Happy to discuss the ordering, and to take the core-side pieces (2 and 3) if that's easier than reworking the whole branch yourself. Removing the rejection marker from the title now.

daker:
@wayfarer3130 Could you clarify how VTK.js is expected to be deprecated in favor of VTK.wasm? Is there an official statement from Kitware confirming this direction? as of today it's still maintained and WebGPU support is in parity with WebGL https://github.com/Kitware/vtk-js/issues/3473

I also have some concerns about VTK.wasm as the primary path. While exposing VTK C++ has benefits, it makes debugging, patching, and contributing fixes more difficult(debug build with DWARF). We will have to deal with all their complex CPP tooling and build systems, even small fixes may need to go through the VTK development and release cycle. With VTK.js, we can debug directly in the browser, land fixes quickly or do postinstall patchs. VTK.wasm is a 22Mb(78Mb uncompressed) file that you have to load before doing anything.

Do we gain any performance improvement compared to VTK.wasm compared to VTK.js WebGL or even WebGPU version ?

Some NPM stats:

VTK.js ESM : 100K downloads/week
VTK.js UMD: 5K downloads/week
VTK.wasm: 1.6K downloads/week

@sedghi

PR #2797, opened 2026-07-10
perf(segmentation): defer projection-heavy renders and merge only modified slices during labelmap edits

sedghi:
### Context

Editing a segmentation that is displayed in a projection-heavy viewport was effectively unusable. Every `SEGMENTATION_DATA_MODIFIED` event (brush strokes fire many per second) re-rendered every viewport showing the segmentation, and for viewports that re-march a slab per fragment (legacy volume viewports in `LABELMAP_EDGE_PROJECTION_BLEND`, or generic planar viewports with a slab presentation) each of those renders is expensive enough to freeze the UI for seconds. On top of that, the legacy independent-components path (`addVolumesAsIndependentComponents`, used by the labelmap-over-MIP example) re-merged the whole segmentation volume on the CPU with a per-voxel `getAtIndex` loop on every event, and had several lifecycle bugs around removing and re-adding the representation.

These changes are independent of the new generic-viewport labelmap-over-MIP capability (which depends on an unreleased vtk.js change) and can merge on their own; the generic feature is split out into #2784.

### Changes & Results

**Deferred rendering for projection-heavy viewports**

- New `triggerSegmentationRenderForModified`: on `SEGMENTATION_DATA_MODIFIED`, cheap viewports render live while projection-heavy viewports (legacy edge-projection blend, or generic viewports with slab presentations) are re-rendered on a 240 ms trailing debounce, once the edit stream goes quiet. Projection viewports do not need to update in real time while drawing.

**Legacy independent-components path (`addVolumesAsIndependentComponents`)**

- The data-modified merge now touches only `modifiedSlicesToUse` with direct per-slice typed-array copies instead of a full-volume `getAtIndex` loop, keeping the CPU cost proportional to the edit instead of the volume.
- The combined actor is re-added under the render plan's `representationUID`; the plan's reconcile step otherwise tore down the combined actor (which is also the viewport's base volume actor) on the next render.
- Removing the representation now restores the shared `vtkImageData` scalars (the injected CPU array misconfigured the shared streaming mapper) and the label-outline property state (the plain MIP otherwise kept rendering through the label-outline shader).
- The debounced data-modified listener is removed with `removeEventListenerDebounced`; the plain `removeEventListener` leaked the debounce wrapper, which kept merging into the detached image data after every delete/re-add cycle.

Measured on the `labelmapMIP` example (PET volume, Apple Silicon): a brush stroke costs ~1.3 ms of main thread per pointer move, and the projection viewport refreshes ~240 ms after the stroke ends.

### Testing

- `pnpm run example labelmapMIP` (stock vtk.js): click "Add Labelmap Representation", draw on the left viewport with the brush - drawing stays responsive and the MIP viewport updates shortly after the stroke ends. Delete and re-add the representation: after deleting, the MIP viewport should render the plain MIP without posterized/white artifacts, and repeated delete/re-add cycles should not slow editing down.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] "OS: macOS (Darwin 25.5.0)"
- [x] "Node version: 22.22.2"
- [x] "Browser: Chrome for Testing (latest)"

## Summary by CodeRabbit

* **Performance**
  * Improved responsiveness during rapid segmentation edits, including brush-stroke updates.
  * Reduced unnecessary rendering for projection-heavy views by deferring expensive updates until editing activity slows.

* **Bug Fixes**
  * Segmentation display updates now refresh only the affected image slices when possible.
  * Improved restoration of volume and label-outline display settings when segmentation overlays are removed.
  * Prevented stale update handling after segmentation representations are closed.

### DIFF
--- a/packages/tools/src/eventListeners/segmentation/segmentationDataModifiedEventListener.ts
+++ b/packages/tools/src/eventListeners/segmentation/segmentationDataModifiedEventListener.ts
@@ -1,5 +1,5 @@
 import type { SegmentationDataModifiedEventType } from '../../types/EventTypes';
-import { triggerSegmentationRenderBySegmentationId } from '../../stateManagement/segmentation/SegmentationRenderingEngine';
+import { triggerSegmentationRenderForModified } from '../../stateManagement/segmentation/SegmentationRenderingEngine';
 import onLabelmapSegmentationDataModified from './labelmap/onLabelmapSegmentationDataModified';
 import { getSegmentation } from '../../stateManagement/segmentation/getSegmentation';
 
@@ -16,7 +16,10 @@ const onSegmentationDataModified = function (
     onLabelmapSegmentationDataModified(evt);
   }
 
-  triggerSegmentationRenderBySegmentationId(segmentationId);
+  // Data modifications stream in at brush-stroke frequency; this trigger
+  // renders cheap viewports live and defers projection-heavy ones (labelmap
+  // over MIP) until the stream goes quiet.
+  triggerSegmentationRenderForModified(segmentationId);
 };
 
 export default onSegmentationDataModified;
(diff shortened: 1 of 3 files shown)


sedghi:
known flaky test at ohif 
 1 failed
    [chromium] › tests/Scoord3dProbe.spec.ts:39:5 › should hydrate SCOORD3D probe measurements correctly 
  5 skipped
  168 passed (2.5m)

wayfarer3130 (approved):

PR #2815, opened 2026-07-17
feat(tools): touch interaction support with dispatcher fallback to mouse callbacks

sedghi:
## Context

An audit of all viewport tools across the clinical modes found that touch failures were architectural, not per-tool: tools entering through `preMouseDownCallback` / `mouseClickCallback` were unreachable by touch because the touch pipeline never calls those handlers, hover-dependent affordances have no touch equivalent, and in-draw precision targets stayed at mouse sizes. This PR addresses the tool-framework side of that audit in one pass.

## Foundation: dispatcher fallback

- New `eventDispatchers/shared/getTouchCallbackWithMouseFallback.ts`: touch dispatchers resolve the explicit touch callback when a tool implements it, otherwise fall back to the mouse counterpart (`touchDragCallback -> mouseDragCallback`, `preTouchStartCallback -> preMouseDownCallback`, `postTouchStartCallback -> postMouseDownCallback`, `touchTapCallback -> mouseClickCallback`).
- The fallback (only the fallback) is gated on the tool declaring `'Touch'` in `supportedInteractionTypes` - the first runtime consumer of that field. Explicit touch callbacks keep working regardless of declaration.
- A payload adapter additively synthesizes the mouse-detail fields touch events lack (`mouseButton`, `startPoints`/`lastPoints`/`deltaPoints` on tap details).
- `touchTap.ts` is now a dedicated handler (first-Active-tool-wins semantics preserved); `touchStartActivate` gained the same try/catch as its mouse twin; the `TOUCH_TAP` listener leak in `touchToolEventDispatcher.disable` is fixed.
- Deferred-tap guard in `touchStartListener`: a pending `TOUCH_TAP` timer no longer fires into a new in-progress gesture (it used to end brush/scissors strokes mid-drag).
- Declaration corrections: Crosshairs `['Mouse'] -> ['Mouse','Touch']`; hover-only SegmentSelect/SegmentLabel narrowed to `['Mouse']`; VolumeCropping and RectangleROIStartEndThreshold gained declarations.
- Companion safety edits for tools the fallback newly reaches: TOUCH listener triples in RegionSegment/WholeBodySegment draw loops, TrackballRotate `touchend` cleanup beside its `mouseup`-only sample-distance restore.

## Engine

- The rendering engine sets `touch-action: none` on every enabled viewport element (restored on disable) via `RenderingEngine/helpers/elementTouchAction.ts`, so browsers no longer consume viewport gestures (scroll, pinch zoom, double-tap zoom). Apps no longer need per-app workarounds.

## Per-tool touch work

- **Brush family**: TOUCH listeners in the draw loop; strokes paint by finger. When a second finger joins mid-stroke (pinch zoom), the stroke cancels and rolls back its voxel writes via the memo system without entering undo history.
- **Scissors x3**: same multi-finger cancellation - the rubber band is dropped instead of resizing at the mean touch point and applying on release.
- **Spline/Livewire**: double-tap closes the contour; the first-point close target widens on touch (`touchCloseCurveDistance`); Livewire's live path preview tracks the finger during a drag.
- **Probe/DragProbe/Magnify**: render offset above the finger on touch so the finger does not occlude the readout/loupe (`touchCanvasOffset`, tunable, `{x:0,y:0}` disables).
- **AdvancedMagnify**: touches inside the loupe no longer leak to the underlying viewport; long-press on the loupe opens the zoom-factor picker (mouse binding unchanged); on touch, dragging anywhere on the glass moves the loupe (ring handles still resize) and interior taps no longer stack a second magnifier.
- **VolumeRotate**: additive drag-to-rotate for both mouse and touch (`rotateDragDegreesPerPixel`); wheel behavior unchanged.
- **ClickSegment**: on touch, the probe runs at the tap point instead of requiring hover-arming; the desktop hover flow is untouched.
- **Crosshairs**: `isPointNearTool` honors the dispatcher-provided proximity (36px on touch instead of the hardcoded 6); the `mobile` config block defaults on for coarse-pointer platforms (same `(any-pointer:coarse)` check the examples use; `configuration.mobile.enabled = false` opts out).
- **ArrowAnnotate**: double-tap label re-edit uses the touch proximity.

## Fixes found during device testing

- `ToolGroup.clone()` crashed on tools added to a group without ever receiving a mode (undefined `toolOptions` read) - hit by AdvancedMagnify's tool-group cloning outside OHIF-style setups.
- `getCurrentVolumeViewportSlice` error message named Livewire but the utility is shared with WindowLevelRegion; generalized.

## Example

- New `touchAllTools` example: sagittal volume + stack viewport, every touch-capable tool selectable onto Primary (1-finger), brush size/segment index controls, an on-page log console for device debugging, and mobile-safe init (`isMobile`, capability-gated `preferSizeOverAccuracy`). The example runner gained an env-gated LAN mode (`CS3D_ALLOW_LAN=1`) for phone testing.

## Behavior notes for reviewers

- Tap routing: a tool with only `mouseClickCallback` that declares `'Touch'` (e.g. MIPJumpToClick) now receives taps; in groups where such a tool precedes another tap-consumer in add order, tap delivery can change.
- SegmentSelect/SegmentLabel default declarations narrowed to `['Mouse']` (they are hover-only; consumers can still pass their own `supportedInteractionTypes`).
- Mouse behavior is unchanged throughout; new interactions were added to mouse only where they are strictly additive (VolumeRotate drag).

## Testing

- `tsc --noEmit` clean on core and tools; oxlint clean (no new warnings).
- Verified interactively with synthetic touch sequences (Chromium touch emulation) and on-device iOS WebKit (Simulator via safaridriver + iPhone hardware): brush paint/cancel-on-pinch rollback, scissors cancel, spline/livewire double-tap close, probe/magnify offsets, AdvancedMagnify move/resize/press-picker, Crosshairs handle visibility, MIPJumpToClick tap.

## Summary by CodeRabbit

* **New Features**
  * Expanded mobile and touch support across viewing tools, annotations, segmentation, rotation, and magnification.
  * Added the “Touch: All Tools” demonstration and consistent touch interaction tuning.
* **Bug Fixes**
  * Improved touch gesture routing, multi-touch handling, tap detection, annotation creation, and interaction cleanup.
  * Viewport touch behavior is now managed consistently and restored when viewports reset.
* **Documentation**
  * Added migration guidance for viewport touch-action behavior.


wayfarer3130:
@james-hanks comments
I can't seem to use the brush on stack viewport
[1:08 PM]Also as Alireza said it would be good to use pan and zoom with the finger shortcuts while there are annotation/active tool on stack viewport

dan-rukas:
Here are my notes and testing. Everything actually looks and feels very good. I also made a separate link of touch testing notes across devices, but the important details are below with notes from my Claude session.

**Device testing of #2815 — touchAllTools example @ `1cf20f7da`, iPad Pro (M1) and iPhone 15 Pro, iOS 26.6 (latest),  Safari, finger touch + Apple Pencil 2 throughout.** Full per-tool results are in our test catalog: https://touch-testing.netlify.app/

### Bugs / gaps found

1. **AdvancedMagnify zoom-factor picker: reselecting the current value leaves the pill stuck on the canvas.** Pick a *different* factor → closes normally; pick the *same* one → the pill never dismisses. Looks like dismissal rides a value-change event. (The picker predates this PR, but the long-press route added here makes it much more reachable — happy to check the mouse route if useful.)

2. **Multi-finger gestures are unreliable while capture-on-touch-start tools are active.** With a measurement tool (or Magnify) selected, 2-finger zoom and 3-finger scroll intermittently do nothing or fire the 1-finger tool instead — the gesture is classified by the first finger (`touchStartListener.ts` early-returns on the second), so `numTouchPoints` bindings lose a millisecond landing race. Notably, the *same pinch works reliably* under Brush/Scissors — the cancel-on-second-finger pattern this PR added there releases the gesture correctly. Suggestion: extend that cancel pattern to annotation creation. (The capture logic looks pre-existing rather than introduced here, but this PR is what makes 1-finger-tool + 2-finger-zoom the advertised convention, so it seemed worth flagging on it.)

3. **iOS long-press always triggers text selection over the viewports (at least on the demo page).** `touch-action: none` stops scrolling and pinch/double-tap zoom (verified working end-to-end), but it doesn't cover text selection or the iOS text-magnifier loupe — those need `-webkit-user-select: none` + `-webkit-touch-callout: none`, which nothing currently sets. In practice every hold-and-press interaction (Magnify, AdvancedMagnify's long-press picker) can fight the OS, and the trigger threshold is unpredictable. Suggestion: extend the `elementTouchAction` store/restore helper to these two properties — same rationale as the migration note's argument for `touch-action` itself ("no useful behavior to preserve").

4. **ArrowAnnotate's new double-tap label re-edit didn't fire in our testing (but may not be needed)** Creation + label dialog work; double-tapping the arrow afterwards (repeatedly, on-target) never reopened the dialog on device. We didn't root-cause it (tap-chain race vs. proximity vs. example wiring). Separately, a design question: double-tap may be a crowded gesture for this — it's also the spline/livewire close convention — and label re-edit might belong in app chrome instead.

5. *(Pre-existing, noting only for completeness)*: with Zoom active on Primary, a 1-finger drag pans and never zooms — the zoom branch in `_pinchCallback` is gated on >1 touch point while the `pan: true` fall-through runs unconditionally. Same wiring exists before this PR.

### What we verified working (briefly)

The core of the PR held up well on hardware: the `touch-action` contract (no page hijack anywhere, all session), VolumeRotate's new drag, Brush's drag-time cursor), AdvancedMagnify's touch shielding (interior touches no longer leak to the viewport — the old bug is gone) and long-press picker, spline double-tap close (very good with Pencil), freehand drawing (best-in-class with both inputs), and Sculptor is genuinely usable via the fallback. PaintFill's feared pinch-flood-fill did **not** reproduce for us.

### Not tested (so you know the coverage edges)

Android and hybrid-laptop devices (none available); Crosshairs and ClickSegment (not in the example); ONNX/SAM tools (need an app context); TrackballRotate (needs a volume3d viewport); Pencil *hover* on livewire (requires an M2+ iPad Pro — our hardware can't hover; if hover drives the path preview between taps, livewire's touch story changes a lot, and we'd love to hear if anyone can try it).

wayfarer3130:
Thanks @dan-rukas — quite useful comments, and the per-device catalog is much appreciated. Items 1, 2 and 3 are now addressed; responses to all five below.

### 2. Multi-finger gestures unreliable while capture-on-touch-start tools are active — **fixed**

You diagnosed this exactly, and your suggested fix is the one we landed. `_activateDraw` / `_activateModify` set `state.isInteractingWithTool = true`, and every touch dispatcher early-returns on that flag — so the gesture never re-resolves against the new finger count. The `numTouchPoints: 2` binding was always correct; the event simply never reached it. That also explains why Brush/Scissors were reliable: they never set the flag, so their dispatcher was never blocked. Their `currentPointsList` checks are about rolling back voxel writes, which is a separate concern.

Rather than add a bespoke check to each of the ~29 tools that set the flag, the release happens once in the dispatcher, ahead of the guard:

```js
// touchDrag.ts
const bypassInteractionGuard = releaseToolForMultiTouchGesture(evt);

if (state.isInteractingWithTool && !bypassInteractionGuard) {
  return;
}
```

`releaseToolForMultiTouchGesture` calls `AnnotationTool.cancel(element)` across the tool group. That works because `cancel` is **abstract** on `AnnotationTool`, so every annotation tool already implements it (it backs the Escape key), and each guards on its own draw state — so calling it broadly is safe and needs no bookkeeping about which tool is mid-interaction. Cancelling runs `_deactivateDraw`, which clears the flag, and Zoom picks the gesture up.

Spline and Livewire opt out via a new `BaseTool.handlesMultiTouchGestures`: cancelling them would discard already-placed points. For those the guard is bypassed rather than cleared, so the pinch still reaches Zoom while the contour survives and their existing `_touchDragDuringDrawCallback` ignores the extra finger.

One known gap: `cancel()` is gated on `isDrawing`, which `_dragCallback` sets but `toolSelectedCallback` does not — so when *moving an existing* annotation the release no-ops on the first drag event and takes effect on the second. Self-correcting, but being tightened.

### 3. iOS long-press triggers text selection over the viewports — **fixed**

Routed through the same helper, as you suggested. `elementTouchAction` is now a property table rather than a single hardcoded property, covering `touch-action`, `user-select`, `-webkit-user-select` and `-webkit-touch-callout`, each with its own store/restore attribute. Both selection properties are set because iOS Safari still needs the prefixed form.

Two related fixes while in there: the prior-value capture is guarded per-property, so a repeat call cannot overwrite the stored originals with `none`; and restore now removes the declaration when there was no prior inline value, instead of leaving an empty string.

### 1. AdvancedMagnify picker sticks when reselecting the current value — **fixed**

Your read was right — dismissal rode the `change` event, which does not fire when the value is unchanged. Two more commit paths now exist, funnelled through a one-shot guard because the callback tears the dropdown down and a second invocation would throw:

- `click` on an `OPTION`, which handles the same-value case. Clicks on the listbox chrome are ignored so scrolling the list cannot dismiss it.
- `blur`, giving click-away dismissal (previously absent) and covering platforms that render `<select>` as a native modal picker, where neither of the other paths fires reliably on an unchanged value.

The desktop path is straightforward; the iOS behaviour is reasoned rather than observed, so a re-test on your hardware would be very welcome.

### 4. ArrowAnnotate double-tap re-edit — **acknowledged, and your design question is the more important half**

Not root-caused yet. But I agree double-tap is crowded: it already means "close contour" for spline/livewire, and overloading it invites exactly the kind of ambiguity that is miserable to debug on device. Label re-edit probably belongs in app chrome. Open to dropping the double-tap route rather than fixing it, if others agree.

### 5. Zoom on Primary: 1-finger drag pans, never zooms

Confirmed pre-existing — the zoom branch in `_pinchCallback` is gated on `>1` touch points while the `pan: true` fall-through is unconditional. Out of scope here; worth its own issue.

---

### Filling in part of your Android coverage gap

We tested on an Amazon Fire HD 10 (9th gen, Fire OS 7.3.3.1, Silk / Chromium 136, Mali-G72, `devicePixelRatio` 1.5) and hit two things worth recording, neither touch-specific:

**Mali advertises `EXT_texture_norm16` but cannot render through it.** `getExtension` returns non-null while core's validated `getRenderingCapabilities()` probe reports `norm16: false`. Any capability gate written against extension *presence* fails open, 16-bit CT textures break, and the viewport looks broken while 8-bit labelmaps still draw correctly — a genuinely confusing failure mode. The `touchAllTools` example now gates on the validated profile instead.

**The CPU rendering path mis-scales interaction when `devicePixelRatio !== 1`.** Filed separately as #2849. `worldToCanvasCPU` / `canvasToWorldCPU` work in device pixels while the GPU paths and the tools event layer use CSS pixels; they agree only at DPR 1. Panning moves the image by exactly `devicePixelRatio` times the pointer distance. Reproducible on desktop with `?cpu=1` and 150% browser zoom — no device needed. Relevant to your notes because HiDPI Android devices that fall back to CPU rendering will hit it, and it is easy to misread as a touch bug.

wayfarer3130 (approved):

dan-rukas:
Tested changes and everything looks really good. 
- 2 finger gesture now works as expected on measurements
- iOS long press no longer activates iOS text selection on viewports at all
- I was still able to reproduce the "select same zoom" in AdvancedMagnify, but I don't think this bug will be encountered often or at all by real users.

Overall everything works and feels great - thank you!

PR #2825, opened 2026-07-30
fix(synchronization): Slice Synchronization Fails for Series with Mismatched FOR

### LINKED ISSUE #2762: [Bug] Slice Synchronization Fails for Series with Mismatched FOR
### Describe the Bug

The Image Slice Sync feature fails to synchronize scrolling between two viewports when the loaded series have different Frame of Reference (FOR), either same or different Image Position Patient (IPP).

### Steps to Reproduce

- Launch the OHIF Viewer using this specific study link: [OHIF Viewer Dev](https://viewer-dev.ohif.org/viewer?StudyInstanceUIDs=1.3.6.1.4.1.25403.345050719074.3824.20170125095438.5)
- Change the layout to a 2-viewport configuration (1x2 or 2x1).
- Load the series as follows:
  - Viewport 1: Body 3.0 CE
  - Viewport 2: Body 5.0 Lung I+ CE
- Enable the Image Slice Sync tool.
- Scroll through the images in Viewport 1.

https://github.com/user-attachments/assets/cfdb15a7-cc50-4228-80b1-9ea9b711a22b

### The current behavior

The slice sync is not working properly as the calculation of spatial registration is not proper.

### The expected behavior

Slice sync should work properly

### System Information

System:
Memory: 5.09 GB / 7.63 GB
Container: Yes
Shell: 5.1.16 - /bin/bash
Binaries:
Node: 22.22.0
Yarn: 1.22.22
npm: 10.9.4
bun: 1.3.9
Browsers:
Chrome: 146.0.7680.80

arul-trenser:
### Context

When two viewports have different `FrameOfReferenceUID`, scroll synchronization falls back to `calculateViewportsSpatialRegistration()` to align them using `imagePositionPatient`

This fallback was calling `viewport.getSliceIndex()` and passing the returned slice number as the `imageId` to `metaData.get(imagePlaneModule, imageId)`. Since a slice index is not a valid `imageId`, the metadata lookup returned undefined, so the spatial registration was silently skipped, breaking scroll sync whenever viewports belonged to different frames of reference. Sync only appeared to work when `FrameOfReferenceUID` was identical (or empty/undefined on both sides), because that path bypasses the fallback entirely.

The PR is incorporated by [FlyWheel.io](https://flywheel.io/).

Fixes: https://github.com/cornerstonejs/cornerstone3D/issues/2762

### Changes & Results

**File changed:** `calculateViewportsSpatialRegistration.ts`

- Replaced `viewport.getSliceIndex()` with `viewport.getImageIds()[0]` to get the imageId of each viewport.
- Passed this `imageId` directly to `metaData.get('imagePlaneModule', imageId)`, instead of the slice index.

**Result:** `imagePlaneModule` metadata now resolves correctly for both viewports, so the translation between `imagePositionPatient` values is calculated properly, and scroll synchronization works again for viewports with mismatched `FrameOfReferenceUID`.

**Before:**

https://github.com/user-attachments/assets/ad17110b-d296-4a36-b4fc-e7c04d3329b2

**After:**

https://github.com/user-attachments/assets/fb726e7c-63fb-4a91-9378-e6f3bcba4609

### Testing

- Launch the OHIF Viewer using this study link: [OHIF Viewer Dev](https://viewer-dev.ohif.org/viewer?StudyInstanceUIDs=1.3.6.1.4.1.25403.345050719074.3824.20170125095438.5)
- Change the layout to a 2-viewport configuration (1x2 or 2x1).
- Load the following series into each viewport:
- Viewport 1: Body 3.0 CE
- Viewport 2: Body 5.0 Lung I+ CE
- Enable the Image Slice Sync tool.
- Scroll through the images in Viewport 1.
- Verify that Viewport 2 scrolls in sync with Viewport 1, matching the corresponding slice position.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] "OS:  Windows 11
- [x] "Node version:  22.19.0
- [x] "Browser: 150.0.7871.187
  

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved spatial registration across viewports by using the active image identifiers directly.
  * Increased reliability when calculating image alignment and translation between viewports.

### DIFF
--- a/packages/core/src/utilities/calculateViewportsSpatialRegistration.ts
+++ b/packages/core/src/utilities/calculateViewportsSpatialRegistration.ts
@@ -27,11 +27,11 @@ function calculateViewportsSpatialRegistration(
   viewport1: IStackViewport | IVolumeViewport,
   viewport2: IStackViewport | IVolumeViewport
 ): void {
-  const imageId1 = viewport1.getSliceIndex();
-  const imageId2 = viewport2.getSliceIndex();
+  const imageId1 = viewport1.getImageIds()[0];
+  const imageId2 = viewport2.getImageIds()[0];
 
-  const imagePlaneModule1 = get('imagePlaneModule', imageId1.toString());
-  const imagePlaneModule2 = get('imagePlaneModule', imageId2.toString());
+  const imagePlaneModule1 = get('imagePlaneModule', imageId1);
+  const imagePlaneModule2 = get('imagePlaneModule', imageId2);
 
   if (!imagePlaneModule1 || !imagePlaneModule2) {
     console.log('Viewport spatial registration requires image plane module');


sen-trenser:
@sedghi Could you please take a look at this PR?
Thanks!

wayfarer3130 (commented):
Thank you for the fix. The old code was broken, and you found the correct place. It gave a slice index to `metaData.get('imagePlaneModule', ...)`, the lookup never matched, and the function always returned at the guard.

I ask for two changes.

## 1. Use `getViewReference()`

I put the details in a comment on the code.

## 2. Write the specification of the behaviour

This repository holds specifications in `packages/docs/docs/specs/`, and `packages/docs/docs/specs/tools/control/Crosshairs.md` shows the format. A specification of the slice synchronization is necessary, because no document records what the behaviour must be. Please write one.

These are the three rules as I understand them:

1. The two viewports have the same frame of reference, and the same view plane normal or a similar one. The target viewport moves the same distance in world coordinates as the source viewport.
2. One viewport is not a volume, or the two volumes have different view plane normals. The target viewport moves the same number of slices as the source viewport.
3. The two volumes have the same view plane normal or a similar one, but different frames of reference. Use rule 1, and take the alignment of the user at the first synchronization as the origin.

Please confirm these three rules, or correct them, and then write them in the specification. Give the specification a `## User requirements` section and an `## Implementation requirements` section, and keep the two sections separate. A user requirement changes in two cases only: the requirement describes the behaviour wrongly, or the intended behaviour itself changes. An implementation requirement changes freely. The `Crosshairs.md` file shows this change rule in its Scope section.

### Two limitations for the specification

Rule 2 needs a mechanism that the code does not have now. The synchronizer stores a translation matrix only, and a matrix cannot express a change of the slice index. An offset of the index gives the same result. The code stores the offset one time at the first synchronization, and then sets `targetIndex = sourceIndex + offset`. This is work for a separate pull request, and the specification must record the limitation.

`spatialRegistrationMetadataProvider` stores the matrix under the pair of viewport IDs, and no code removes the entry. A viewport that shows a new series keeps the matrix of the old series. The specification must record this limitation also.

packages/core/src/utilities/calculateViewportsSpatialRegistration.ts
```diff
@@ -27,11 +27,11 @@ function calculateViewportsSpatialRegistration(
...
+  const imageId1 = viewport1.getImageIds()[0];
+  const imageId2 = viewport2.getImageIds()[0];
 
-  const imagePlaneModule1 = get('imagePlaneModule', imageId1.toString());
-  const imagePlaneModule2 = get('imagePlaneModule', imageId2.toString());
+  const imagePlaneModule1 = get('imagePlaneModule', imageId1);
+  const imagePlaneModule2 = get('imagePlaneModule', imageId2);
 
   if (!imagePlaneModule1 || !imagePlaneModule2) {
     console.log('Viewport spatial registration requires image plane module');
```
wayfarer3130, line 37: `getImageIds()[0]` gives the first image of the viewport. This function must use the image that the viewport shows now. The choice of the first image creates three problems:

1. The JSDoc on line 24 says "use the current image's metadata". `imageSliceSyncCallback` reads the source position with `getCurrentImageId()`, and its own documentation says that the user scrolls the target viewport to the correct slice before the synchronization starts. With the first image, a comparison of two studies moves the target viewport by the difference between the start positions of the two series.
2. `StackViewport.getImageIds()` returns the order that the application supplies. A volume viewport returns the order that `sortImageIdsAndGetSpacing` produced. When the two orders run in opposite anatomical directions, the translation is wrong by almost the full length of the scan.
3. `BaseVolumeViewport.getImageIds()` throws an error when the viewport has no volume actor, or when the volume left the cache. `getCurrentImageId()` returns `undefined` instead. This utility is public, so a caller that registers two viewports before the volume loads gets an exception in place of the early return.

The change also makes lines 46 to 70 reachable for the first time. Line 47 reads `imageOrientationPatient` without a guard, but line 54 uses `?.` on the same field. A provider that returns an `imagePlaneModule` without `imageOrientationPatient` or `imagePositionPatient` (XA, US, secondary capture) makes line 47, or `vec3.subtract` on lines 63 to 67, throw an error inside the asynchronous callback.

## Please use `getViewReference()`

```ts
const ref1 = viewport1.getViewReference();
const ref2 = viewport2.getViewReference();

const n = ref1.viewPlaneNormal;
const d = vec3.subtract(vec3.create(), ref1.cameraFocalPoint, ref2.cameraFocalPoint);
const translation = vec3.scale(vec3.create(), n, vec3.dot(d, n));
```

`Viewport.getViewReference()` returns `FrameOfReferenceUID`, `cameraFocalPoint`, `viewPlaneNormal`, `viewUp` and `sliceIndex` for every type of viewport. The camera supplies these values, so the function needs no metadata provider. This removes all four problems above, and it always uses the current view.

One caution: `cameraFocalPoint` holds the pan of the user, but `imagePositionPatient` does not. The projection onto the normal, which the code above makes, keeps the pan out of the registration. The projection is also more correct than the present difference of the two `imagePositionPatient` values, because that difference holds the in-plane offset between the origins of the two series.

Two more points:

- The test of the orientation on lines 46 to 58 repeats the `areViewportsCoplanar` test that `imageSliceSyncCallback` makes on line 104. A comparison of `viewPlaneNormal` values replaces the test of `imageOrientationPatient`.
- `_getClosestImageIdIndex` in `imageSliceSyncCallback` reads `imagePositionPatient` for every target image, and it throws an error when a provider returns nothing. The fallback to the change of the slice index belongs there, and not in this function.

No test covers this function. `packages/tools/test/imageSliceSyncCallback.jest.js` replaces the function with a mock. Please add a test.

PR #2832, opened 2026-07-31
fix(Stretch): Image aspect ratio resets to fit viewport when changing layout after calling setAspectRatio with isFitViewportAfterStretch disabled

### LINKED ISSUE #2830: [Bug] Image aspect ratio resets to fit viewport when changing layout after calling setAspectRatio with isFitViewportAfterStretch disabled
### Describe the Bug

When an image's aspect ratio is updated using setAspectRatio with isFitViewportAfterStretch set to false, changing the viewport layout causes the image to auto-fit to the new viewport dimensions instead of preserving the aspect ratio applied by setAspectRatio.

https://github.com/user-attachments/assets/54b56a22-8542-4005-a3ab-85973bedd06d

### Steps to Reproduce

- Launch a study in the viewer.
- Set a custom aspect ratio for the viewport using `setAspectRatio` with `isFitViewportAfterStretch` set to false.
- Switch to any common, custom, or MPR viewport layout.
- Observe the image aspect ratio in the new layout.

### The current behavior

The image automatically fits to the viewport upon changing the layout, ignoring the `isFitViewportAfterStretch: false` setting passed to `setAspectRatio`.

### The expected behavior

The image should maintain the aspect ratio set by `setAspectRatio` and **not** fit to the viewport when switching layouts with `isFitViewportAfterStretch` set to `false`.

### System Information

System:
  OS: Linux 6.6 Ubuntu 22.04.5 LTS (Jammy Jellyfish)
  Memory: 5.09 GB / 7.63 GB
  Container: Yes
  Shell: 5.1.16 - /bin/bash
Binaries:
  Node: 22.22.0
  Yarn: 1.22.22
  npm: 10.9.4
  bun: 1.3.9
Browsers:
  Chrome: 146.0.7680.80

arul-trenser:
### Context

`Viewport.setAspectRatio(aspectRatio, isFitViewportAfterStretch)` stretches the camera to a custom aspect ratio. When `isFitViewportAfterStretch` is false, the function must keep that stretched ratio, and the function must not fit the image back to the viewport borders. But earlier, no code saved the `isFitViewportAfterStretch` flag. `setAspectRatioForVTKCamera` used the flag one time for the calculation, and then the code discarded the flag. The VTK camera object had no memory of this flag.

Because of this, `resetCamera()` did not know that fit-after-stretch was disabled. `resetCamera()` used a fixed value `false` for `isFitViewportAfterStretch` when `resetCamera()` created the camera again. Also, `setViewPresentation()`, which restores the state of the layout, called `setAspectRatio()` without the second argument, and the default value set the behaviour back to fit-to-viewport. So, the custom aspect ratio went back to the fit-viewport behaviour at each change of the layout.

The PR is incorporated by [FlyWheel.io](https://flywheel.io/).

Fixes: https://github.com/cornerstonejs/cornerstone3D/issues/2830

#### The first implementation

The first implementation saved the flag on the camera.

**File changed:**
1. `extendedVtkCamera.ts`
  - Added `getIsFitViewportAfterStretch()` and `setIsFitViewportAfterStretch()` to the custom VTK camera interface and implementation. The flag was saved directly on the camera model itself as `model.isFitViewportAfterStretch`.
2. `vtkSlabCamera.ts`
  - Added matching `getIsFitViewportAfterStretch()` and `setIsFitViewportAfterStretch()` type declarations to the `vtkSlabCamera` interface.
3. `RenderingEngine/Viewport.ts`
  - Added `getIsFitViewportAfterStretch()` on `Viewport`, which reads the flag from `getCamera()` (defaults to `false` if not set).
  - `getCameraNoRotation()` included `isFitViewportAfterStretch: vtkCamera.getIsFitViewportAfterStretch()` in the returned camera state, so the flag stayed available through `getCamera()`.
  - `setCamera()` called `vtkCamera.setIsFitViewportAfterStretch(isFitViewportAfterStretch)` whenever the aspect ratio was applied, and saved the flag on the VTK camera.
  - `resetCamera()` used `previousCamera.isFitViewportAfterStretch` instead of the fixed value `false` when it created the camera again, and kept the stretch mode that was active before.
  - `setViewPresentation()` called `this.setAspectRatio(aspectRatio, this.getIsFitViewportAfterStretch())` instead of skipping the second argument, so the restore of the layout also respected the saved flag.

#### The issues with the first implementation

`isFitViewportAfterStretch` is not a property of a camera. The flag selects between two calculations during one transition from an old aspect ratio to a new aspect ratio. A camera holds an aspect ratio and a `parallelScale`. A camera does not hold the method that gave the camera those two values.

That one decision gave four problems:

- `DEFAULT_VALUES` in `extendedVtkCamera.ts` had no entry for the flag, and `macro.setGet` had no entry for the flag. A camera that no code configured therefore returned `undefined`.
- The code read that same `undefined` as `false` in `getIsFitViewportAfterStretch()`, and as `true` in `setCamera`, in `setAspectRatio` and in `setAspectRatioForVTKCamera`.
- The CPU path did not make a round trip of the flag. `getCameraCPU()` did not return the flag, and `setCameraCPU()` did not store the flag. `getIsFitViewportAfterStretch()` therefore always returned `false` for a CPU stack viewport.
- A `setCamera()` call that had no `aspectRatio` lost the flag, because the code wrote the flag to the VTK camera inside the `if (aspectRatio)` block only.

The first implementation also kept a second effect of the setter. The fit branch of `setAspectRatioForVTKCamera` multiplied the camera `parallelScale` by a fit factor, and the branch multiplied `initialCamera.parallelScale` and `fitToCanvasCamera.parallelScale` by the same factor. Those two values are the baselines of `getZoom()` and `setZoom()`. A setter for the aspect ratio therefore changed the meaning of every zoom value that came after the setter. PR #2443 added that calculation, and `resetCamera` already owns those baselines.

#### The current implementation

No code stores the flag. `setAspectRatio` only stretches the image, and `resetCamera` does the fit.

**Files changed:**
1. `RenderingEngine/Viewport.ts`
  - `resetCamera` calculates `parallelScale` from the stretched width of the image. The calculation uses `targetAspectRatio`, and `resetCamera({ resetZoom: true })` therefore fits a stretched image.
  - `setAspectRatio(value, storeAsInitialCamera)` only sets the aspect ratio. The function does not change `parallelScale`, `initialCamera` or `fitToCanvasCamera`.
  - `setCamera` gives the aspect ratio to the VTK camera, and `setCamera` does nothing else with the aspect ratio.
  - `setViewPresentation` sets the aspect ratio first, and sets the zoom second.
  - `setAspectRatioForVTKCamera` and `getIsFitViewportAfterStretch` are removed.
2. `RenderingEngine/StackViewport.ts`
  - `setAspectRatioForViewport` only sets the aspect ratio.
  - The new function `getFitParallelScaleCPU` holds the fit calculation of the CPU path.
  - `resetCameraCPU` uses that calculation, and `resetCameraCPU` accepts the option `resetAspectRatio`.
3. `extendedVtkCamera.ts` and `vtkSlabCamera.ts`
  - `getIsFitViewportAfterStretch()` and `setIsFitViewportAfterStretch()` are removed.
4. `types/ICamera.ts` and `types/IViewport.ts`
  - `isFitViewportAfterStretch` is removed from `ICamera` and from `ViewPresentation`.
5. `examples/axisBasedImageStretching` and `examples/calibrationTools`
  - The examples call `resetCamera` after `setAspectRatio` when a fit is necessary.

**Breaking change:** the second argument of `setAspectRatio` was `isFitViewportAfterStretch`, and the second argument is now `storeAsInitialCamera`. A caller that gave `true` for a fit must now call `resetCamera` after `setAspectRatio`.

**Result:**
`setAspectRatio` stretches the image, and `setAspectRatio` keeps the zoom. `resetCamera({ resetZoom: true })` fits the stretched image. A view presentation records the stretch and the zoom as two independent values, and a restore of the presentation writes both values without a new fit calculation. The image therefore keeps the stretch across a reset of the camera and across a change of the layout, and the image no longer goes back to the fit-viewport behaviour.

**Before**

https://github.com/user-attachments/assets/eacd05c8-ac6c-4b52-9e85-d18a98a7e617

**After**

https://github.com/user-attachments/assets/41437432-ebf0-45c2-9f77-5097b3ec9f15

### Testing

These steps use the example `packages/tools/examples/axisBasedImageStretching` of this repository. A link with the OHIF viewer is not necessary, and a new function is not necessary.

1. Run `yarn dev` in the root of the repository.
2. Open the example `axisBasedImageStretching` in the browser.
3. Clear the checkbox `Fit to Viewport after stretch`. The checkbox controls the call of `resetCamera` after `viewport.setAspectRatio()`.
4. Select the value `2:1` in the aspect ratio dropdown. The image gets the stretch, and the image keeps the zoom.
5. Select a value in the `Rotation` dropdown. The `Rotation` dropdown calls `viewport.setViewPresentation()`, and a change of the layout calls the same function.
6. Verify that the image keeps the stretch of step 4 and the zoom of step 4.
7. Set the checkbox `Fit to Viewport after stretch`. Select a different aspect ratio, and then select a different rotation. Verify that the image fits the borders of the viewport.

To test the CPU rendering path, add `?useCpu=true` to the URL of the example. Repeat step 3 and step 4, and then scroll the stack. Each scroll of the stack reads the view presentation and writes the view presentation back. Verify that the image keeps the stretch of step 4.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] "OS:  Windows 11
- [x] "Node version:  22.19.0
- [x] "Browser: 150.0.7871.187
  

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved image fitting when applying non-uniform aspect-ratio stretching.
  - Reset Zoom now accounts for horizontal and vertical stretch differences to fit the image correctly within the viewport.
  - Applying an aspect-ratio change preserves the current zoom level instead of automatically refitting the image.
  - Saved view presentations now apply their recorded zoom after aspect-ratio adjustments, ensuring the expected zoom is restored.

### DIFF
--- a/packages/core/src/RenderingEngine/Viewport.ts
+++ b/packages/core/src/RenderingEngine/Viewport.ts
@@ -1266,7 +1266,7 @@ class Viewport {
       viewUp: viewUpToSet,
       clippingRange: clippingRangeToUse,
       aspectRatio: targetAspectRatio,
-      isFitViewportAfterStretch: false,
+      isFitViewportAfterStretch: previousCamera.isFitViewportAfterStretch,
     });
 
     const modifiedCamera = this.getCamera();
@@ -1504,6 +1504,16 @@ class Viewport {
   }
 
   /**
+   * Returns whether to change aspect ratio in anamorphic or not
+   *
+   * @returns change aspect ratio in anamorphic or not
+   */
+  public getIsFitViewportAfterStretch(): boolean {
+    const { isFitViewportAfterStretch } = this.getCamera();
+    return isFitViewportAfterStretch ?? false;
+  }
+
+  /**
    * Because the focalPoint is always in the centre of the viewport,
    * we must do planar computations if the frame (image "slice") is to be preserved.
    * 1. Calculate the intersection of the view plane with the imageData
@@ -1601,6 +1611,7 @@ class Viewport {
       viewAngle: vtkCamera.getViewAngle(),
       flipHorizontal: this.flipHorizontal,
       flipVertical: this.flipVertical,
+      isFitViewportAfterStretch: vtkCamera.getIsFitViewportAfterStretch(),
       aspectRatio: vtkCamera.getAspectRatio(),
     };
   }
@@ -1783,6 +1794,7 @@ class Viewport {
 
     if (aspectRatio) {
       this.setAspectRatioForVTKCamera(aspectRatio, isFitViewportAfterStretch);
+      vtkCamera.setIsFitViewportAfterStretch(isFitViewportAfterStretch);
     }
 
     // update clipping range only if focal point changed of a new actor is added
@@ -2254,7 +2266,7 @@ class Viewport {
       this.setDisplayArea(displayArea);
     }
     this.setZoom(zoom);
-    this.setAspectRatio(aspectRatio);
+    this.setAspectRatio(aspectRatio, this.getIsFitViewportAfterStretch());
     if (pan) {
       const [aspectX, aspectY] = aspectRatio;
       this.setPan([pan[0] * zoom * aspectX, pan[1] * zoom * aspectY] as Point2);
--- a/packages/core/src/RenderingEngine/Viewport.ts
+++ b/packages/core/src/RenderingEngine/Viewport.ts
@@ -1730,7 +1730,7 @@ class Viewport {
       flipVertical,
       clippingRange,
       aspectRatio,
-      isFitViewportAfterStretch,
+      isFitViewportAfterStretch: providedIsFitViewportAfterStretch,
     } = cameraInterface;
 
     // Note: Flip camera should be two separate calls since
@@ -1791,6 +1791,12 @@ class Viewport {
     if (clippingRange !== undefined) {
       vtkCamera.setClippingRange(clippingRange);
     }
+
+    const isFitViewportAfterStretch =
+      providedIsFitViewportAfterStretch ??
+      previousCamera.isFitViewportAfterStretch ??
+      true;
+    updatedCamera.isFitViewportAfterStretch = isFitViewportAfterStretch;
 
     if (aspectRatio) {
       this.setAspectRatioForVTKCamera(aspectRatio, isFitViewportAfterStretch);
@@ -2213,6 +2219,7 @@ class Viewport {
     }
     const currentAspectRatio = this.getAspectRatio();
     target.aspectRatio = currentAspectRatio;
+    target.isFitViewportAfterStretch = this.getIsFitViewportAfterStretch();
     if (pan) {
       const currentPan = this.getPan();
       const [aspectX, aspectY] = currentAspectRatio;
@@ -2261,12 +2268,13 @@ class Viewport {
       rotation,
       flipHorizontal = this.flipHorizontal,
       flipVertical = this.flipVertical,
+      isFitViewportAfterStretch = this.getIsFitViewportAfterStretch(),
     } = viewPres;
     if (displayArea !== this.getDisplayArea()) {
       this.setDisplayArea(displayArea);
     }
     this.setZoom(zoom);
-    this.setAspectRatio(aspectRatio, this.getIsFitViewportAfterStretch());
+    this.setAspectRatio(aspectRatio, isFitViewportAfterStretch);
     if (pan) {
       const [aspectX, aspectY] = aspectRatio;
       this.setPan([pan[0] * zoom * aspectX, pan[1] * zoom * aspectY] as Point2);
--- a/packages/core/src/RenderingEngine/vtkClasses/extendedVtkCamera.ts
+++ b/packages/core/src/RenderingEngine/vtkClasses/extendedVtkCamera.ts
@@ -26,6 +26,7 @@ interface ICameraInitialValues {
   physicalViewUp?: number[];
   physicalViewNorth?: number[];
   aspectRatio?: number[];
+  isFitViewportAfterStretch?: boolean;
 }
 
 declare module '@kitware/vtk.js/Rendering/Core/Camera' {
@@ -44,6 +45,18 @@ declare module '@kitware/vtk.js/Rendering/Core/Camera' {
      * @param aspectRatio - [scaleX, scaleY] stretch factors applied along the canvas X and Y axes
      */
     setAspectRatio(aspectRatio: [x: number, y: number]): boolean;
+
+    /**
+     *  Returns whether fit to viewport after stretch or not
+     *  @defaultValue true
+     */
+    getIsFitViewportAfterStretch(): boolean;
+
+    /**
+     * Set whether fit to viewport after stretch or not
+     * @param isFitViewportAfterStretch - Whether fit to viewport after stretch or not
+     */
+    setIsFitViewportAfterStretch(isFitViewportAfterStretch: boolean): void;
   }
 }
 
@@ -85,6 +98,14 @@ function extendedVtkCamera(publicAPI, model) {
 
   publicAPI.setAspectRatio = (aspectRatio) => {
     model.aspectRatio = aspectRatio;
+  };
+
+  publicAPI.getIsFitViewportAfterStretch = () => {
+    return model.isFitViewportAfterStretch;
+  };
+
+  publicAPI.setIsFitViewportAfterStretch = (isFitViewportAfterStretch) => {
+    model.isFitViewportAfterStretch = isFitViewportAfterStretch;
   };
 }
 
--- a/packages/core/src/RenderingEngine/vtkClasses/vtkSlabCamera.ts
+++ b/packages/core/src/RenderingEngine/vtkClasses/vtkSlabCamera.ts
@@ -759,6 +759,18 @@ export interface vtkSlabCamera extends vtkObject {
    * @param aspectRatio - aspectRatio of the viewport in x and y axis
    */
   setAspectRatio(aspectRatio: [x: number, y: number]): boolean;
+
+  /**
+   *  Returns whether fit to viewport after stretch or not
+   *  @defaultValue true
+   */
+  getIsFitViewportAfterStretch(): boolean;
+
+  /**
+   * Set whether fit to viewport after stretch or not
+   * @param isFitViewportAfterStretch - Whether fit to viewport after stretch or not
+   */
+  setIsFitViewportAfterStretch(isFitViewportAfterStretch: boolean): void;
 }
 
 const DEFAULT_VALUES = {
--- a/packages/core/src/types/IViewport.ts
+++ b/packages/core/src/types/IViewport.ts
@@ -349,6 +349,11 @@ export interface ViewPresentation {
    * The aspect ratio is how the viewport image is stretched and the default is [1,1].
    */
   aspectRatio?: Point2;
+
+  /**
+   * Whether to fit the viewport after stretching the aspect ratio.
+   */
+  isFitViewportAfterStretch?: boolean;
 }
 
 /**


sen-trenser:
@wayfarer3130 Could you please take a look at this PR?
Thanks!

packages/core/src/RenderingEngine/vtkClasses/extendedVtkCamera.ts
```diff
@@ -86,6 +99,14 @@ function extendedVtkCamera(publicAPI, model) {
   publicAPI.setAspectRatio = (aspectRatio) => {
     model.aspectRatio = aspectRatio;
   };
+
+  publicAPI.getIsFitViewportAfterStretch = () => {
+    return model.isFitViewportAfterStretch;
```
wayfarer3130: This getter returns `model.isFitViewportAfterStretch`, but no code gives the model a default value for the flag.

`DEFAULT_VALUES` at line 118 contains `aspectRatio: [1, 1]`, and it does not contain `isFitViewportAfterStretch`. `macro.setGet` at line 127 contains `'aspectRatio'`, and it does not contain `'isFitViewportAfterStretch'`. A camera that no code configured therefore returns `undefined` from this getter. The new JSDoc at line 51 says `@defaultValue true`, but the code gives no default value.

Please add `isFitViewportAfterStretch: true` to `DEFAULT_VALUES`. This one change removes the cause of the two different fallback values in `Viewport.ts`.

packages/core/src/RenderingEngine/Viewport.ts
```diff
@@ -1503,6 +1503,16 @@ class Viewport {
...
   }
 
+  /**
+   * Returns whether to change aspect ratio in anamorphic or not
+   *
+   * @returns change aspect ratio in anamorphic or not
+   */
+  public getIsFitViewportAfterStretch(): boolean {
+    const { isFitViewportAfterStretch } = this.getCamera();
+    return isFitViewportAfterStretch ?? false;
```
wayfarer3130: This getter uses `false` as the fallback value. The other new code uses `true`: `setCamera` at line 1798, `setAspectRatio` at line 1488, `setAspectRatioForVTKCamera` at line 1639, and the JSDoc `@defaultValue true` in `extendedVtkCamera.ts` at line 51. One of the two values is wrong.

The difference has an effect on `setViewPresentation`. Line 2271 reads the default value from this getter, and line 2277 gives the value to `setAspectRatio`. For a camera that no code configured, the value is now `false`. Before this change, `setViewPresentation` called `setAspectRatio(aspectRatio)` and the value was `true`. `setAspectRatioForVTKCamera` then returns at line 1643, and the code does not calculate the compensation of `parallelScale`.

Please use `?? true` here, and also add the default value to `DEFAULT_VALUES` in `extendedVtkCamera.ts`.

packages/core/src/RenderingEngine/Viewport.ts
```diff
@@ -1781,8 +1792,15 @@ class Viewport {
...
 
+    const isFitViewportAfterStretch =
+      providedIsFitViewportAfterStretch ??
+      previousCamera.isFitViewportAfterStretch ??
+      true;
+    updatedCamera.isFitViewportAfterStretch = isFitViewportAfterStretch;
+
     if (aspectRatio) {
       this.setAspectRatioForVTKCamera(aspectRatio, isFitViewportAfterStretch);
+      vtkCamera.setIsFitViewportAfterStretch(isFitViewportAfterStretch);
```
wayfarer3130: This call is inside the `if (aspectRatio)` block, but line 1799 writes `updatedCamera.isFitViewportAfterStretch` for every call.

A call of `setCamera({ isFitViewportAfterStretch: false })` that has no `aspectRatio` therefore does not reach the VTK camera. The flag goes into `updatedCamera`, and `setInitialCamera` can store the flag in `this.initialCamera`, but `getCamera()` reads the VTK camera and returns the old value. The caller sees no effect, and `this.initialCamera` and the live camera hold two different values.

Please move this call out of the `if (aspectRatio)` block, to a position beside line 1799.

packages/core/src/RenderingEngine/Viewport.ts
```diff
@@ -1266,7 +1266,7 @@ class Viewport {
       viewUp: viewUpToSet,
       clippingRange: clippingRangeToUse,
       aspectRatio: targetAspectRatio,
-      isFitViewportAfterStretch: false,
+      isFitViewportAfterStretch: previousCamera.isFitViewportAfterStretch,
```
wayfarer3130: This line replaces a fixed value `false` with `previousCamera.isFitViewportAfterStretch`. For a camera that no code configured, the value is `undefined`. `setCamera` then changes `undefined` to `true` at line 1798, because the second and the third operand of the `??` chain are also `undefined`.

`resetCamera` therefore selects the fit calculation where `resetCamera` selected the direct calculation before. The result is visible when `options.aspectRatio` is not `[1, 1]` and the caller uses `resetCamera({ resetZoom: false })`. `setAspectRatioForVTKCamera` multiplies `parallelScale` by `ratioFactor` at line 1691, and the zoom that the caller wants to keep gets a different value.

A default value of `true` in `DEFAULT_VALUES` makes this line correct, because `previousCamera.isFitViewportAfterStretch` then always holds a real value.

wayfarer3130 (commented):
I merged `origin/main` into this branch, and the merge is clean. I then reviewed the change.

The mechanism is correct. The camera keeps `isFitViewportAfterStretch`, and a layout change keeps the choice of the user. Two problems remain.

### 1. The flag has no default value, and the code uses two different fallback values

`DEFAULT_VALUES` in `extendedVtkCamera.ts` does not contain `isFitViewportAfterStretch`, and `macro.setGet` does not contain the flag. A camera that no code configured returns `undefined`. The code then reads that same `undefined` in two different ways:

- `Viewport.getIsFitViewportAfterStretch()` (line 1513) uses `false`.
- `setCamera` (line 1798), `setAspectRatio` (line 1488), `setAspectRatioForVTKCamera` (line 1639) and the new JSDoc use `true`.

The inline comments give the details.

### 2. The CPU rendering path does not keep the flag

`StackViewport.ts` is not in this diff, but this change has an effect on `StackViewport.ts`:

- `getCameraCPU()` (line 1041) does not return `isFitViewportAfterStretch`.
- `setCameraCPU()` (line 1157) reads the flag and gives the flag to `setAspectRatioForViewport()` at line 1226, but no code writes the flag to `_cpuFallbackEnabledElement.viewport`.

The flag therefore does not make a round trip on the CPU path, and `Viewport.getIsFitViewportAfterStretch()` always returns `false` for a CPU stack viewport.

This result has an effect on each scroll of a stack. `_updateActorToDisplayImageId` reads the view presentation at line 2618 and writes the view presentation back at line 2628. The presentation now contains `isFitViewportAfterStretch: false`, because `getViewPresentation()` (`Viewport.ts` line 2222) reads the getter. `setAspectRatioForViewport` then returns at line 1114, and `viewport.parallelScale` and `viewport.scale` do not get a new value. Before this change, `setViewPresentation` called `setAspectRatio(aspectRatio)` and the flag was `true`. The behaviour on the CPU path is therefore different, and the behaviour does not follow the choice of the user.

### A suggested direction for the correction

1. Add `isFitViewportAfterStretch: true` to `DEFAULT_VALUES` in `extendedVtkCamera.ts`, and add the name to the `macro.setGet` list.
2. Change `getIsFitViewportAfterStretch()` to use `?? true`.
3. Move `vtkCamera.setIsFitViewportAfterStretch(...)` out of the `if (aspectRatio)` block.
4. Return the flag from `getCameraCPU()`, and store the flag in `setCameraCPU()`.

### A request about the test instructions

The Testing section asks the reader to link the library with the viewer, and to add a function that calls `viewport.setAspectRatio([x, y], false)`. Please give instructions that use an example of this repository instead. The example `packages/tools/examples/axisBasedImageStretching` already has the necessary controls:

- a checkbox `Fit to Viewport after stretch`, which sets the second argument of `setAspectRatio` (line 241),
- an aspect ratio dropdown with the values `1:1`, `1:2`, `2:1`, `0.5:1`, `1:0.5` and `3:17` (line 279),
- a `Rotation` dropdown, which calls `vp.setViewPresentation({ rotation })` (line 339). `setViewPresentation` is one of the functions that this PR changes.
- a URL parameter `?useCpu=true` (line 76), which selects the CPU rendering path.

A reader can therefore clear the checkbox, select an aspect ratio, and then select a rotation. The reader can also add `?useCpu=true` and scroll the stack, to test the CPU path of problem 2. Please confirm these steps, and then put the steps in the description of the PR.

wayfarer3130:
### A comment about the design

**This comment replaces problem 1 and steps 1, 2 and 4 of the suggested correction in my review above.** My review asked you to add a default value for `isFitViewportAfterStretch` to `DEFAULT_VALUES`. Please do not do that. The camera must not hold the flag at all, and I give the reason below.

#### The camera must not hold the flag

`isFitViewportAfterStretch` is not a property of a camera. The flag selects between two calculations during one transition from an old aspect ratio to a new aspect ratio. A camera holds an aspect ratio and a `parallelScale`. A camera does not hold the method that gave the camera those two values.

The four problems in my review are all results of that one decision:

- the camera model has no default value for the flag,
- the code reads the same `undefined` as `false` in one place and as `true` in three places,
- the CPU path cannot make a round trip of the flag,
- a `setCamera` call that has no `aspectRatio` loses the flag.

No code stores the flag in the design below, and therefore none of the four problems exists.

#### `setAspectRatio` must only set the value and apply the value

A caller that wants the image to fit the viewport after the stretch calls `resetCamera()` with the correct flags. `resetCamera` at line 1128 of `Viewport.ts` already has `resetPan`, `resetZoom`, `resetToCenter`, `resetAspectRatio` and `storeAsInitialCamera`. `resetCamera` is also the function that already fits an image into a viewport. Two clear calls are better than one setter that has a hidden second effect.

#### The hidden second effect

The fit branch of `setAspectRatioForVTKCamera` does three writes that the caller of a setter does not expect:

- line 1691 multiplies the camera `parallelScale` by `ratioFactor`,
- lines 1695 to 1699 multiply `this.initialCamera.parallelScale` by the same factor,
- lines 1701 to 1705 multiply `this.fitToCanvasCamera.parallelScale` by the same factor.

The last two writes move the zoom baselines. `getZoom()` at line 1420 and `setZoom()` at line 1437 read those baselines. A setter for the aspect ratio therefore changes the meaning of every zoom value that comes after the setter.

`resetCamera` already owns that baseline. Line 1274 calls `this.setFitToCanvasCamera(this.getCamera())`. Two functions therefore write the same baseline, and only one of the two functions is the correct owner.

PR #2443 added this calculation, and this PR does not add the calculation. But this PR makes the effect permanent across a reset and across a layout change. A permanent effect is more difficult to remove later.

#### The effect on `setViewPresentation`

`setViewPresentation` calls `this.setZoom(zoom)` at line 2276, and then calls `this.setAspectRatio(aspectRatio, ...)` at line 2277. The fit branch then multiplies `parallelScale` at line 1691. The restored zoom therefore gets a different value immediately after the restore.

A view presentation records `zoom` and `aspectRatio` as two independent values. A restore must write both values, and neither value must change the other value. `setAspectRatio` without the zoom effect gives that result, and the new field `isFitViewportAfterStretch` in `ViewPresentation` is then not necessary.

#### One necessary change in `resetCamera`

`resetCamera` calculates `parallelScale` at lines 1206 to 1214 from `widthWorld`, `heightWorld` and the canvas size. That calculation does not use the aspect ratio. `resetCamera` therefore does not fit a stretched image today.

`getFitScale` at line 1666 already has the correct formula. The formula multiplies `widthWorld` by the ratio value, and the formula divides the result by the ratio value when the ratio value is less than 1. Please move that formula into the calculation of `parallelScale` in `resetCamera`, and use `targetAspectRatio` from line 1257. `resetCamera({ resetZoom: true })` then fits a stretched image, and `resetCamera({ resetZoom: false })` then keeps the zoom of the user.

#### A summary of the request

1. Remove `getIsFitViewportAfterStretch` and `setIsFitViewportAfterStretch` from `extendedVtkCamera.ts` and from `vtkSlabCamera.ts`.
2. Remove `isFitViewportAfterStretch` from `ViewPresentation` in `IViewport.ts`, from `getViewPresentation` and from `setViewPresentation`.
3. Remove the zoom effect from `setAspectRatio` and from `setAspectRatioForVTKCamera`. Let the two functions set the aspect ratio and apply the aspect ratio.
4. Put the fit calculation in `resetCamera`, and let the flags of `resetCamera` control the calculation.
5. Keep the `StackViewport.ts` CPU path in agreement with the same rule. The problem 2 of my review also disappears with this design.

This design changes the signature of the public function `setAspectRatio`, because the second argument goes away. Please tell me if you know a caller that needs the old behaviour of one single call. Please also tell me if you see a use case that this design does not cover.

wayfarer3130 (approved):

PR #2850, opened 2026-08-07
fix(core): return CSS pixels from the CPU worldToCanvas/canvasToWorld

### LINKED ISSUE #2849: CPU rendering path uses device pixels in worldToCanvas/canvasToWorld, breaking pan and zoom when devicePixelRatio !== 1
### Summary

On the CPU rendering path, `worldToCanvas` / `canvasToWorld` operate in **device pixels**, while the GPU paths and the tools event layer both use **CSS pixels**. The two agree only when `devicePixelRatio === 1`, so on any HiDPI display, zoomed browser, or mobile device the CPU path mis-scales interaction by exactly `devicePixelRatio`.

The most visible symptom is that **panning moves the image faster (or slower) than the pointer**, by a constant factor equal to `devicePixelRatio`. It is not a lag or a drift that settles — the offset is permanent, zero at the canvas centre, and grows linearly with distance from it.

### Steps to reproduce

1. Open any example with a stack viewport, e.g. `stackBasic`.
2. Append `?cpu=1` to force CPU rendering (handled by `applyUrlParameterOverridesToDemoConfig` in `initDemo`).
3. Set browser zoom to **150%** (this makes `window.devicePixelRatio` ≈ 1.5). Reload.
4. Pan the image.

**Expected:** the image tracks the pointer 1:1.
**Actual:** the image moves ~1.5× the pointer distance — the same ratio as `devicePixelRatio`.

At 100% zoom (`devicePixelRatio === 1`) the bug is invisible, which is why it has gone unnoticed. It reproduces on a real device without any browser zoom: an Amazon Fire HD 10 (9th gen, `devicePixelRatio` 1.5, Fire OS 7 / Silk) shows it at default zoom.

Not touch-specific — a mouse drag reproduces it identically. Touch merely made it obvious, because the finger gives an unambiguous reference for how far the content *should* move.

### Root cause

The CPU transform is built from the canvas **backing store**, which `getOrCreateCanvas` sizes in device pixels:

`packages/core/src/RenderingEngine/helpers/getOrCreateCanvas.ts`
```js
const w = Math.round(rect.width * devicePixelRatio);
const h = Math.round(rect.height * devicePixelRatio);
canvas.width = w;
canvas.height = h;
```

`packages/core/src/RenderingEngine/helpers/cpuFallback/rendering/calculateTransform.ts`
```js
// Move to center of canvas  -- canvas.width/height are DEVICE pixels
transform.translate(
  enabledElement.canvas.width / 2,
  enabledElement.canvas.height / 2
);
```

So `pixelToCanvas` / `canvasToPixel` produce and consume **device** pixels, and `worldToCanvasCPU` returns that value unchanged:

`packages/core/src/RenderingEngine/StackViewport.ts` (~L3033)
```js
private worldToCanvasCPU = (worldPos: Point3): Point2 => {
  ...
  const canvasPoint = pixelToCanvas(this._cpuFallbackEnabledElement, indexPoint);
  return canvasPoint;          // <-- device pixels, no DPR division
};
```

Both GPU implementations do the opposite and explicitly normalise to CSS pixels:

`packages/core/src/RenderingEngine/BaseVolumeViewport.ts` (~L2154 `worldToCanvas`, ~L2232 `worldToCanvasContextPool`)
```js
const canvasCoordWithDPR = [
  canvasX / devicePixelRatio,
  canvasY / devicePixelRatio,
];
```

And the tools event layer feeds **CSS** pixels in, with no DPR term at all — identical for mouse and touch:

`packages/tools/src/eventListeners/mouse/getMouseEventPoints.ts` (and the touch twin `getTouchEventPoints.ts`)
```js
function _pagePointsToCanvasPoints(element, pagePoint) {
  const rect = element.getBoundingClientRect();
  return [
    pagePoint[0] - rect.left - window.pageXOffset,
    pagePoint[1] - rect.top - window.pageYOffset,
  ];
}
```

So `canvasToWorldCPU` receives CSS pixels into a transform that expects device pixels, and `worldToCanvasCPU` hands device pixels back to callers that expect CSS pixels. The GPU paths are self-consistent; the CPU path is not.

There is also a comment in `BaseRenderingEngine.addCustomViewport` that explicitly notes DPR is ignored "for ... the Cornerstone CPU rendering pathway", which suggests the CPU path was written against a CSS-pixel-sized canvas at some point and the DPR-aware sizing in `getOrCreateCanvas` was introduced later without updating it.

### Affected

- `worldToCanvasCPU` / `canvasToWorldCPU` in `StackViewport`
- Anything downstream of them at `devicePixelRatio !== 1`: pan, zoom, and annotation hit-testing/placement on the CPU path

### Suggested fix

Divide by `devicePixelRatio` on the way out of `worldToCanvasCPU` and multiply on the way into `canvasToWorldCPU`, matching the GPU implementations, so the public contract is CSS pixels everywhere.

One caution: the same `canvas.width`-derived transform is also used for CPU **drawing** (`drawImageSync` and friends), where device pixels are the correct unit. A fix should convert only at the `worldToCanvas` / `canvasToWorld` boundary rather than changing the transform itself, and should be verified at both `devicePixelRatio === 1` and `!== 1` — the bug is invisible in the former, so a regression there would be easy to miss.

### Environment

- cornerstone3D `main` @ `cf1a1371b`
- Reproduced on: Chrome/Edge desktop at 150% browser zoom; Amazon Fire HD 10 9th gen (Fire OS 7.3.3.1, Silk / Chromium 136, `devicePixelRatio` 1.5)
- Independent of touch vs mouse, and independent of `webGlContextCount` (the tiled and context-pool engines are both bypassed under `?cpu=1`)

wayfarer3130:
The CPU transform built by calculateTransform is anchored on the canvas backing store, which getOrCreateCanvas sizes in device pixels, so pixelToCanvas and canvasToPixel speak device pixels. StackViewport passed their results straight through worldToCanvasCPU and canvasToWorldCPU, making the CPU path the only viewport path whose public coordinates are not CSS pixels: the GPU implementations divide by devicePixelRatio before returning, and the tools event layer derives its canvas points from getBoundingClientRect, which has no DPR term.

The two agree only at devicePixelRatio 1, which is why this went unnoticed. At any other ratio - HiDPI display, zoomed browser, mobile device - interaction is mis-scaled by exactly that factor, most visibly as panning that moves the image devicePixelRatio times further than the pointer. The offset is permanent rather than transient, zero at the canvas centre and growing with distance from it.

Convert at the public boundary only. The conversion deliberately does not live inside pixelToCanvas/canvasToPixel: internal rendering callers pass device pixels to those on purpose, and setToPixelCoordinateSystem applies the same transform to a 2D context where device pixels are the correct unit. The three internal round trips through canvasToPixel are bridged so their results are byte-for-byte unchanged.

resetCameraCPU already passed element.clientWidth/2 - CSS pixels - into canvasToWorldCPU, so it was silently wrong at DPR != 1 and is corrected by this change.

Reproduce before the fix: open a stack example with ?cpu=1, set browser zoom to 150%, and pan.

Fixes #2849

Note that testing this requires the touchAllTools PR to be merged first.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  


wayfarer3130:
@jbocce - I've tested this on a tablet device (amazon fire) and it resolves hte scaling transform issue.

packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -1151,18 +1160,20 @@ class StackViewport extends Viewport {
...
+        cssToDevicePixels(focalPointCanvas)
       );
 
       const prevFocalPointCanvas = this.worldToCanvasCPU(
         previousCamera.focalPoint
       );
       const prevFocalPointPixel = canvasToPixel(
         this._cpuFallbackEnabledElement,
-        prevFocalPointCanvas
+        cssToDevicePixels(prevFocalPointCanvas)
```
jbocce, line 1176: Does the if(parallelScale) branch need to consider DPR as well?

jbocce (approved):
One small comment/question in the code changed, but I am also wondering why aren't other viewports affected by DPR or is it already handled?

PR #2856, opened 2026-08-13
feat(voi): VOI LUT Function and VOI LUT Sequence support

### LINKED ISSUE #938: [Bug] VOI LUT images not rendered correctly
### Describe the Bug

Images with a VOI LUT don't get their levels computed correctly when rendering, so they display incorrectly.

Cornerstone3D version: 1.36.3
This works as expected when using Cornerstone JS 2.6.1 (wado image loader 3.3.2)

Attached is a file which demonstrates the issue.
[hand.dcm.zip](https://github.com/cornerstonejs/cornerstone3D/files/13640395/hand.dcm.zip)

## Correct rendering
hand

## Current rendering
(ignore the green annotation in the top right corner)

image

## Tags relevant to VOI LUT
```
(0028,3010) SQ (Sequence with undefined length #=1)     # u/l, 1 VOILUTSequence
  (fffe,e000) na (Item with undefined length #=6)         # u/l, 1 Item
    (0028,3002) US 16384\0\14                               #   6, 3 LUTDescriptor
    (0028,3003) LO [User Selection]                         #  14, 1 LUTExplanation
    (0028,3006) OW 0000\0000\0000\0000\0000\0000\0000\0000\0000\0000\0000\0000\0000... # 32768, 1 LUTData
    (07a3,0010) LO [ELSCINT1]                               #   8, 1 PrivateCreator
    (07a3,109d) DS [2332]                                   #   4, 1 Unknown Tag & Data
    (07a3,109e) DS [1149]                                   #   4, 1 Unknown Tag & Data
```

### Steps to Reproduce

1. Load an image with a VOI LUT

### The current behavior

The image gets rendered with incorrect levels

### The expected behavior

The image renders with correct levels. This used to work in the "old" cornerstone.

### OS

Linux

### Node version

18.18.2

### Browser

Chromium 119.0.6045.159, Firefox 120.0b9

### LINKED ISSUE #985: [Bug] Image does not rendered in proper way (GPU rendering)

### LINKED ISSUE #1141: [Bug] image display error (Transfer Syntax: 1.2.840.10008.1.2  , Photometric Interpretation: RGB)
### Describe the Bug

The result is colorful, and I don't know how to make it normal.

The problem may have been caused by Transfer Syntax(1.2.840.10008.1.2), as I have seen in other issues
#1017
#1016

### Steps to Reproduce

O704CS( 9E5X%C2I{FBV}TH

### The current behavior

the image

[wado.zip](https://github.com/cornerstonejs/cornerstone3D/files/14507345/wado.zip)

### The expected behavior

7UDCF{%)B3J8%85CXG`9IGU

### OS

win11

### Node version

16.20.2

### Browser

Microsoft Edge  122.0.2365.66

### LINKED ISSUE #1767: [Bug] setDefaultVolumeVOI bug
### Describe the Bug

getVOIFromMetadata logic error in setDefaultVolumeVOI

### Steps to Reproduce

1、`  const volume = await volumeLoader.createAndCacheVolume(ctVolumeId, {
    imageIds: [...imageIds]
  });`
2、options.imageIds of cornerstoneStreamingImageVolumeLoader is in positive sequence, 
```
const [middleImageIndex, lastImageIndex] = [ Math.floor(options.imageIds.length / 2), 
options.imageIds.length - 1, ];
```
the obtained middleImageIndex is 29, and the imageId is 'wadouri:http://~1.2.840.113619 .2.472.3.40120842.710.1685265533.1.30.dcm'
Why is the imageIds obtained in 
`const { imageIds, metadata } = imageVolume; `
in getVOIFromMetadata in reverse order, resulting in 
```
const imageIdIndex = Math.floor(imageIds.length / 2);
const imageId = imageIds[imageIdIndex];
```
The imageId obtained is 'wadouri:http://~1.2.840.113619.2.472.3.40120842.710.1685265533.1.29.dcm', 
the image has not been downloaded yet, 
`const voiLutModule = metaData.get('voiLutModule', imageId);`
voiLutModule is undefined, the voi obtained by getVOIFromMiddleSliceMinMax is incorrect and very different from the image itself. Why is the order of imageIds reversed?

### The current behavior

The obtained windowWidth and windowCenter are incorrect.

### The expected behavior

getVOIFromMetadata returns the correct voi

### OS

windows11

### Node version

16.14.0

### Browser

chrome

### LINKED ISSUE #1806: [Bug] scaled PT image with lut defined on metadata return wrong Volume VOI
### Describe the Bug

I'm not sure at 100% that this is a bug as I'm not sure about the standard for this.
But I got a PT serie where images were scaled AND a `windowWidth` and a `WindowCenter` were defined. Except those values are defined for raw pixel not scaled ones.

This result in a VOI/Lut totally wrong, because pixels are scaled and VOI values are based on raw pixels value.

### Steps to Reproduce

Load a PT image with scaling and VOI defined on metadata.

### The current behavior

Wrong VOI is returned

### The expected behavior

I'm not so sure about what should be done.
It seems to me that in [the function below](https://github.com/cornerstonejs/cornerstone3D/blob/7b04fb76ec5b2a87d6c76d1617d77664ede84d14/packages/core/src/RenderingEngine/helpers/setDefaultVolumeVOI.ts#L25-L34) we should check if volume is scaled first and return according VOI no matter what.

```js
async function setDefaultVolumeVOI(
  volumeActor: VolumeActor,
  imageVolume: IImageVolume
): Promise<void> {
  let voi = getVOIFromMetadata(imageVolume);

  if (!voi && imageVolume.imageIds.length) {
    voi = await getVOIFromMiddleSliceMinMax(imageVolume);
    voi = handlePreScaledVolume(imageVolume, voi);                 <==== // do this as the first thing in the function
  }
```

But maybe correct behavior is to scale the metadata VOI somehow or even something else I don't know.

### OS

Windows 11

### Node version

v22.11.0

### Browser

any

### LINKED ISSUE #2104: [Bug] Getting inappropriate initial VOI values for certain PT data in Volume Viewport
### Describe the Bug

For certain PET (PT) datasets that include `windowWidth` and `windowCenter` in the DICOM metadata, the Volume Viewport applies a very high initial VOI range. This causes the image to appear black (washed out or invisible) .

### Steps to Reproduce

1. Load a PET dataset that:
       - Has Modality: PT
       - Includes `windowWidth` and `windowCenter` in metadata
2. Observe Volume Viewport - Image appears black or poorly contrasted.
3. Load the same data in Stack - Viewport  Image appears with correct VOI.

### The current behavior

The Volume Viewport VOI is derived from metadata. This causes improper mapping and renders the PET image visually useless in the default view.

### The expected behavior

If PET data is pre-scaled (e.g., imageVolume.scaling.PT.suvbw is present):

- The metadata `windowCenter` / `windowWidth` values should be ignored.

- A default VOI range of { lower: 0, upper: 5 } should be applied in the volume viewport, similar to the behavior in the stack viewport.

### OS

Windows 10

### Node version

20.9.0

### Browser

Google Chrome 136.0.7103.113

### LINKED ISSUE #2236: [Bug] Mamography images not rendering properly
### Describe the Bug

When upgrading to cornerstone 3.x I've noticed some images, mostly mamography, dont render properly on first render (they look more gray than before) and changing the window level, even just a little bitm completely breaks the render.

### Steps to Reproduce

https://github.com/user-attachments/assets/612e1d00-cb6f-4d14-a903-d37846f46f83

Image

### The current behavior

What you can see in the video

### The expected behavior

Normal rendering

### OS

Windows 11

### Node version

v22.11.0

### Browser

Chroem 138.0.7204.169, Firefox 141.0

### LINKED ISSUE #2458: [Bug] Clicking VOI buttons for the volumevoisigmoid example do nothing

### LINKED ISSUE #2520: [Bug] VOI LUT Sequence is not applied
### Describe the Bug

I have dicom files (MG) that includes VOI LUT Sequence (0028,3010). Cornerstone is not applying it. All files includes VOI LUT Sequence.

Dicom files:
https://drive.google.com/file/d/1VFflwSAtSXXhs-s9kuXy_Hcvb4IQk250/view?usp=sharing

### Steps to Reproduce

1. Render dicom file with Cornerstone3D

### The current behavior

Rendered with Cornerstone3D V4.

Image

### The expected behavior

Rendered with another viewer which applies VOI LUT Sequence.

Image

### System Information

  System:
    OS: Linux 6.14 Ubuntu 24.04.3 LTS 24.04.3 LTS (Noble Numbat)
    CPU: (16) x64 AMD Ryzen 7 7800X3D 8-Core Processor
    Memory: 39.77 GB / 46.13 GB
    Container: Yes
    Shell: 5.9 - /usr/bin/zsh
  Binaries:
    Node: 22.19.0 - /home/-/.nvm/versions/node/v22.19.0/bin/node
    Yarn: 1.22.22 - /home/-/.nvm/versions/node/v22.19.0/bin/yarn
    npm: 11.6.0 - /home/-/.nvm/versions/node/v22.19.0/bin/npm
    pnpm: 10.15.1 - /home/-/.nvm/versions/node/v22.19.0/bin/pnpm
  Browsers:
    Chrome: 142.0.7444.175
    Firefox: 146.0
    Firefox Developer Edition: 146.0

### LINKED ISSUE #2716: Window Level Incorrect After Upgrade to Cornerstone 4.x (Image Overexposed on Initial Load)
### Describe the Bug

After upgrading Cornerstone packages from **v1.84.4 → v4.22.3**, the initial **window level (WL/WW)** is no longer applied correctly, causing the image to appear overexposed (completely **white**) on first render.

**This issue did not occur in the older version and appears to be a regression**.

### Steps to Reproduce

1. Load a attached DICOM image.
2. Observe the initial rendered image.
3. Notice that the image appears **overexposed**.
4. Adjust window level manually → image becomes normal.
5. This issue occurs during the initial rendering phase.
6. Possible causes:
7. Missing or incorrectly applied DICOM tags (WindowCenter, WindowWidth)
8. Improper VOI (Volume of Interest) initialization
9. Incorrect scaling or default viewport settings

**Example DICOM file for test**

[Sample Dicom Images.zip](https://github.com/user-attachments/files/27298408/Sample.Dicom.Images.zip)

### The current behavior

1. On initial load, the image appears fully white (**overexposed**).
2. **WL/WW** values seem to be set **too high** or not initialized correctly.
3. The issue occurs before any user interaction.
4. After **manually adjusting the window level, the image renders correctly**.

Image
Image

### The expected behavior

1. The image should render with appropriate default WL/WW values derived from DICOM metadata.
2. No manual adjustment should be required for proper visualization.

### System Information

```
{
  "@cornerstonejs/core": "^1.84.4",
  "@cornerstonejs/dicom-image-loader": "^1.84.4",
  "@cornerstonejs/streaming-image-volume-loader": "^1.84.4",
  "@cornerstonejs/tools": "^1.84.4"
}
```
Behavior:

1. Images render **correctly** on initial load
2. Proper WL/WW applied automatically

```
{
  "@cornerstonejs/core": "^4.22.3",
  "@cornerstonejs/dicom-image-loader": "^4.22.3",
  "@cornerstonejs/streaming-image-volume-loader": "^1.86.1",
  "@cornerstonejs/tools": "^4.22.3"
}
```

Behavior:

1. Initial image is completely **white (overexposed)**
2. WL/WW appears too high or not initialized
3. Manual window adjustment fixes the image

### LINKED ISSUE #2733: [Bug] - StackViewport: degenerate initial VOI produces stuck step-function LUT that survives subsequent voi changes
### Describe the Bug

  When a StackViewport's first rendered image has a degenerate window/level (zero or near-zero window width  e.g. WC=0.5, WW=1 on an all-zero padding frame, common at the start of multi-frame US cines), the resulting vtkColorTransferFunction has both RGB nodes placed at the same scalar value, producing a step-function LUT. Subsequent voi changes (auto-derived on scroll, or via setVOI) only call transferFunction.setRange() and never rebuild the LUT, so every frame after the degenerate one renders as bilevel B&W posterization regardless of the apparently-correct voiRange.

Affected Code
- utilities/createLinearRGBTransferFunction.js — addRGBPoint(lower, ...) and addRGBPoint(upper, ...) with lower===upper creates two coincident nodes.
- RenderingEngine/StackViewport.js#setVOIGPU — the branch at if (isSigmoidTFun || !transferFunction || forceRecreateLUTFunction) skips LUT recreation when a transfer function already exists and forceRecreateLUTFunction is false, falling through to transferFunction.setRange(lower, upper). setRange adjusts input clamping but does not relocate existing RGB nodes — they stay at the degenerate scalar position.

### Steps to Reproduce

Open any DICOM multi-frame US cine whose first frame has WindowCenter=0.5, WindowWidth=1 (or any cine where frame 0 is all-zero pixels). Scroll to any frame with real content. viewport.voiRange will look correct (e.g. {lower:0, upper:255}), but the image renders bilevel. Verified in DevTools that vtk_first10 matches cachedImage_first10 (buffer update is fine), but viewport.setVOI(viewport.voiRange, { forceRecreateLUTFunction: true }) immediately fixes the render.

The exam in the sample data which caused me to see this issue is ReMIND-001 which has US and MR images.

### The current behavior

Posturization of the ultrasound image to black and white until reset.

### The expected behavior

I expect it to look normal!

Per CC : 
Suggested fix (which worked for me)

In setVOIGPU, also rebuild the LUT when the previously-applied voiRange had degenerate width (e.g. upper - lower <= 1), or unconditionally rebuild when the range delta crosses a degenerate-to-non-degenerate boundary. Alternatively, make createLinearRGBTransferFunction defensively expand a zero-width range by 1 epsilon before placing nodes.

### System Information

  System:
    OS: Linux 6.17 Linux Mint 22.2 (Zara)
    CPU: (24) x64 12th Gen Intel(R) Core(TM) i9-12900K
    Memory: 14.73 GB / 31.14 GB
    Container: Yes
    Shell: 5.2.21 - /bin/bash
  Binaries:
    Node: 22.22.2 - /home/james/.nvm/versions/node/v22.22.2/bin/node
    Yarn: 1.22.22 - /home/james/.nvm/versions/node/v22.22.2/bin/yarn
    npm: 10.9.7 - /home/james/.nvm/versions/node/v22.22.2/bin/npm
  Browsers:
    Chrome: 148.0.7778.167
    Firefox: 150.0.2
    Firefox Developer Edition: 150.0.2

### LINKED ISSUE #2745: [Bug] wadouri returns undefined WC/WW for Enhanced multi-frame images
## Description
 
For Enhanced multi-frame SOPs (Enhanced CT/MR/PET/US/XA), VOI lives in [FrameVOILUTSequence (0028,9132)](https://dicom.innolitics.com/ciods/breast-tomosynthesis-image/breast-tomosynthesis-image-multi-frame-functional-groups/52009229/00289132) nested inside [SharedFunctionalGroupsSequence](https://dicom.innolitics.com/ciods/breast-tomosynthesis-image/breast-tomosynthesis-image-multi-frame-functional-groups/52009229) / [PerFrameFunctionalGroupsSequence](https://dicom.innolitics.com/ciods/breast-tomosynthesis-image/breast-tomosynthesis-image-multi-frame-functional-groups/52009230) instead of the dataset root for traditional SOP classes. 

The wadors loader flattens this correctly but wadouri does not. The `voiLutModule` returns undefined [WindowCenter](https://dicom.innolitics.com/ciods/breast-tomosynthesis-image/breast-tomosynthesis-image-multi-frame-functional-groups/52009229/00289132/00281050) / [WindowWidth](https://dicom.innolitics.com/ciods/breast-tomosynthesis-image/breast-tomosynthesis-image-multi-frame-functional-groups/52009229/00289132/00281051), viewports fall back to image min/max, and brightness/contrast differs for the same file. 

## Example Images
Reproducible using the Enhanced CT and MR Multi-frame test DICOM images at NEMA:
ftp://medical.nema.org/medical/dicom/Multiframe/

## Suspected Root Cause
[combineFrameInstanceDataset.ts ](https://github.com/cornerstonejs/cornerstone3D/blob/main/packages/dicomImageLoader/src/imageLoader/wadouri/combineFrameInstanceDataset.ts#L38-L54)  The combiner iterates through Shared/Per Frame Sequences and does `shared[it.tag] = it`, but it never flattens the sequence into `items[0].dataSet.elements`. So `FrameVOILUTSequence` lands at the root, but WC/WW stay one level deeper than the VOI provider reads.

For comparison, wadors/[combineFrameInstance.ts](https://github.com/cornerstonejs/cornerstone3D/blob/main/packages/dicomImageLoader/src/imageLoader/wadors/combineFrameInstance.ts#L8-L21) unwraps via it['Value']?.[0], which is why that path works.

### Steps to Reproduce

 Any Enhanced multi-frame DICOM whose VOI is encoded only in `FrameVOILUTSequence` (no legacy top-level `0028,1050`/`0028,1051`) will reproduce.

Run `loadAndCacheImage` on same image using `wadouri:` and `wadors:`

#### MG
Image

#### CT
Image

#### MG
Image

#### CT
Image

### System Information

  System:
    OS: macOS 26.4.1
    CPU: (16) arm64 Apple M3 Max
  Binaries:
    Node: 22.14.0
    Yarn: 1.22.22
    npm: 10.9.2
  Browsers:
    Chrome: 148.0.7778.179
    Firefox: 151.0.1
    Safari: 26.4

### LINKED ISSUE #2844: [Bug] `createImage` truncates `voiLUTFunction` to its first character, causing `Error: Invalid VOI LUT function` on images with (0028,1056)
### Describe the Bug

Any image whose DICOM metadata contains **VOI LUT Function (0028,1056)** ends up with `image.voiLUTFunction` set to a **single character** (`"S"` for `SIGMOID`, `"L"` for `LINEAR`). `toLowHighRange()` then rejects that value and throws:

```
Uncaught (in promise) Error: Invalid VOI LUT function
    at toLowHighRange (windowLevel.js)
    at getVOIRangeFromWindowLevel (getVOIRangeFromWindowLevel.js)
    at StackViewport._getVOIRangeForCurrentImage (StackViewport.js)
    at StackViewport.resetToDefaultProperties (StackViewport.js)
```

The visible result is that the VOI range is never computed, so the viewport falls back to its default range and the image renders blown out / without contrast. In our case this affects every mammography (MG) study; CT/MR are unaffected because they usually omit (0028,1056), which leaves `voiLUTFunction === undefined` and lets `toLowHighRange()` apply its `LINEAR` default.

### Root cause

The metadata providers return `voiLUTFunction` as a **string**, but `createImage` treats it as an **array**.

`packages/dicomImageLoader/src/imageLoader/wadors/metaData/metaDataProvider.ts` (~line 218) — `getValue()` returns a scalar string, unlike `getNumberValues()` used for the neighbouring fields:

```ts
if (type === MetadataModules.VOI_LUT) {
  return {
    windowCenter: getNumberValues(metaData['00281050'], 1),
    windowWidth: getNumberValues(metaData['00281051'], 1),
    voiLUTFunction: getValue(metaData['00281056']),
  };
}
```

The wadouri provider does the same via `dataSet.string('x00281056')`.

`packages/dicomImageLoader/src/imageLoader/createImage.ts` (~line 264):

```ts
voiLUTFunction:
  (voiLutModule.voiLUTFunction?.length &&
    voiLutModule.voiLUTFunction[0]) ||
  voiLutModule.voiLutFunction ||
  undefined,
```

For a string, `.length` is truthy (7 for `"SIGMOID"`) and `[0]` yields the first **character**, so `image.voiLUTFunction === "S"`.

`StackViewport._getValidVOILUTFunction()` masks part of the problem — it silently falls back to `LINEAR` for the viewport's own `VOILUTFunction` property, which already means a SIGMOID image is rendered as linear. But `_getVOIRangeForCurrentImage()` passes `this.csImage.voiLUTFunction` straight through:

```js
_getVOIRangeForCurrentImage() {
    const { windowCenter, windowWidth, voiLUTFunction } = this.csImage;
    return getVOIRangeFromWindowLevel(windowWidth, windowCenter, voiLUTFunction);
}
```

and `toLowHighRange()` throws on anything that is not `LINEAR`, `SIGMOID` or `LINEAR_EXACT`.

### Steps to Reproduce

1. Load a stack via `wadors:` (or `wadouri:`) whose instances contain (0028,1056) — e.g. a mammogram with `VOILUTFunction = SIGMOID`.
2. Render it in a `StackViewport`, then call `resetToDefaultProperties()` (our UI does this from a Reset button).
3. Console shows `Error: Invalid VOI LUT function`; the image renders saturated.

Inspecting the loaded image object shows `voiLUTFunction: "S"`.

### Expected behavior

`image.voiLUTFunction` should be the full string (`"SIGMOID"`), and the VOI range should be computed with the sigmoid transform.

### Versions

- `@cornerstonejs/core` / `@cornerstonejs/dicom-image-loader` **5.6.11** — reproduced.
- **5.6.13** (latest published at the time of writing) and `main` — the offending code is unchanged.

### Suggested fix

Normalize in `createImage.ts` so both shapes are accepted (this is what we are currently applying locally via `patch-package`):

```diff
-        voiLUTFunction:
-          (voiLutModule.voiLUTFunction?.length &&
-            voiLutModule.voiLUTFunction[0]) ||
-          voiLutModule.voiLutFunction ||
-          undefined,
+        voiLUTFunction:
+          (typeof voiLutModule.voiLUTFunction === 'string'
+            ? voiLutModule.voiLUTFunction.trim()
+            : voiLutModule.voiLUTFunction?.length &&
+              voiLutModule.voiLUTFunction[0]) ||
+          voiLutModule.voiLutFunction ||
+          undefined,
```

(The `trim()` covers the even-length space padding of CS-valued attributes.)

Alternatively the providers could be changed to return an array, but the string shape is the one currently produced by both `wadors` and `wadouri`, so normalizing at the consumer seems less disruptive.

Independently of which side is fixed, `_getVOIRangeForCurrentImage()` arguably should route `csImage.voiLUTFunction` through `_getValidVOILUTFunction()` as well, so that malformed metadata degrades to LINEAR instead of throwing an uncaught exception mid-render.

### The current behavior

Images with VOI LUT are displayed with the wrong window/level.

### The expected behavior

Images with VOI LUT should be displayed with the correct window/level

### System Information

none

daker:
### Context

Cornerstone3D used one VOI model: a window (0028,1050 and 0028,1051) with a
linear ramp. Two DICOM attributes were absent from that model.

- VOI LUT Function (0028,1056) selects the shape of the window. The value
  SIGMOID and the value LINEAR_EXACT gave a LINEAR display, and some shapes
  of the attribute made the render throw "Invalid VOI LUT function".
- VOI LUT Sequence (0028,3010) is a full VOI transformation, and it replaces
  the window and the function (PS3.3 C.11.2.1). No GPU path read it. Thus a
  CR, a DX or an MG image that carries its own curve showed a flat linear
  window. These curves are strongly non linear, so the display was
  incorrect and not only different.

Some other defects around the VOI became visible with this work, and this PR
corrects them also.

#### Metadata and the loader

- Normalize the VOI LUT Function across the providers: a string, a single
  element array, a padded value, a value of the incorrect case, and the key
  SAMPLED_SIGMOID. The code indexed the string, which made "S" from
  "SIGMOID".
- Normalize the VOI LUT Sequence across wadouri, naturalized dcmjs and raw
  DICOMweb JSON, and give it to the renderers on `image.voiLUT`. Read the
  descriptor of a LUT with the pixel representation, read the entries as
  unsigned, and read 8 bit entries that the file holds one for each byte.
- Read the window from the Frame VOI LUT macro of an enhanced multiframe
  SOP. The window is one level below the root there, so the viewports used
  the minimum and the maximum of the image.
- Expose the sequence from the wadors provider and from
  @cornerstonejs/metadata.
- Apply no window, no VOI LUT and no rescale to a color image. These
  attributes are for a monochrome image (C.11.2.1.2.2), and a grey window
  on the range 0 to 255 of the samples of a color image gives a very bright
  display.

#### The stack viewport

- Build the transfer function from the VOI LUT Sequence, and stretch the
  curve over the current range. Thus window level reshapes the curve of the
  file and does not replace it, as the sampled sigmoid already did.
- Fall back to LINEAR with one warning for an unknown function. The code
  threw.
- Separate the two nodes of a window of the width 1. Coincident nodes stay
  coincident, because vtk.js rescales the nodes, and each later change of
  the window then divided by zero.
- Keep the prescaled PT range 0 to 5 when the metadata also carries a
  window.
- Keep the sequence when an application sets the VOI LUT Function to the
  value that the image has. An absent (0028,1056) becomes LINEAR, and
  getProperties gives that value to the application. Thus an application
  that read the properties and set them again stopped the curve.
- Keep the colormap. A colormap fills the transfer function with its own
  colors, so it stops the curve of a sequence and the curve of a sigmoid.
  Before, one window level drag on an image that has a sequence made the
  image grey again.

#### The volume viewports

- Read the VOI LUT Function and the VOI LUT Sequence of the file, as the
  stack viewport does. A SIGMOID series got a sigmoid range with a linear
  transfer function, and no volume path read a sequence.
- Take the window from the nearest instance that has one, and keep the
  function with the window. The middle instance can have no metadata.
- Ignore the window and the curve of a prescaled PT volume. Both are in the
  unscaled counts, and the volume holds SUV.
- Store the requested function before the transfer function is made. The
  old order used the previous function, so a request for SIGMOID had no
  effect until the next change of the VOI.
- Normalize the requested function, as the stack viewport does.

#### The generic viewports (PLANAR_NEXT)

- Use one rule for the sequence on all the viewport types: a function that
  is different from the function of the image stops the sequence, and a
  function that is equal to it does not.
- Give the volume paths the function and the sequence, on the GPU and on
  the CPU.
- Map VOILUTFunction onto the presentation of a compatibility viewport. The
  property had no path, so a request for SIGMOID did nothing.

#### The CPU path

- Implement LINEAR (C.11.2.1.2.1), LINEAR_EXACT (C.11.2.1.3.2) and SIGMOID
  (C.11.2.1.3.1), and permit setVOILUTFunction under CPU rendering.
- Treat a window of the width 1 as the threshold that the standard defines.
- Stretch the curve of a sequence over the window, as on the GPU. Window
  level did nothing on the CPU for a file that has a sequence.
- Take the number of bits of the entries from the largest entry with a
  division and not with a shift. A shift by a negative count shifts by 31
  in JavaScript, so a LUT whose largest entry is below 128 gave a black
  image.

#### The API

New:

- `useVOILUTSequence` on the properties of the stack viewports, the volume
  viewports and the generic viewports. The value `false` ignores the curve
  of the file.
- `voiLUTSequenceApplied` on the detail of the VOI_MODIFIED event. A
  colorbar that draws a ramp from the range and the function needs to know
  that the viewport shows a curve.
- The utilities `isRenderableVOILUT`, `getVOILUTSequenceRange`,
  `createVOILUTSampler`, `sampleVOILUT`, `invertVOILUTSample`,
  `getVOILUTOutputScale`, `normalizeVOILUTFunction` and
  `getValidVOILUTFunction`.
- A `voiLutFunction` example. It loads a local DICOM P10 file and shows how
  its VOI resolves on the GPU path and on the CPU path.

Removed:

- The second `getValidVOILUTFunction` in buildMetadata. That one tested for
  a member of the enum, so it made LINEAR from the shapes that the
  providers give. The name stays available from the same path.

#### Other

- The tools that work in a display intensity sample the curve of the
  sequence. A linear ramp did not agree with the pixels on the screen.
- The migration notes give the changes that an application can see.

####  Known limits

- LUT Data that a DICOMweb server gives as a BulkDataURI is not resolved.
  The metadata provider is synchronous, and a bulkdata reference needs a
  request.
- A sequence of more than one item uses the first usable item. C.11.2.1
  permits this choice, but an application cannot select another item and the
  LUT Explanation is not available.
- A Modality LUT Sequence is applied on the CPU path only. On the GPU path
  the values stay the stored values, and the input domain of a VOI LUT
  Sequence is the output of the modality LUT.

Fixes #938
Fixes #985
Fixes #1141
Fixes #1767
Fixes #1806
Fixes #2104
Fixes #2236
Fixes #2458
Fixes #2520
Fixes #2716
Fixes #2733
Fixes #2745
Fixes #2844

### Testing

Run `pnpm run dev`, and then open one of the two examples below. Each example
tests a different part of this PR by hand.

#### 1. `volumeVoiSigmoid` - "Sigmoid VOI Volume"

This example tests the VOI LUT Function on a volume viewport. This PR also
repairs the manual test itself. Issue #2458 reports that a click on either
button does nothing.

Both buttons apply the same window, WW 1000 and WC -300. Only the shape of the
curve changes between the two buttons. This PR adds the **Invert** toggle and the
state label to the example.

| Step | Expected result |
| --- | --- |
| Click **Set Sigmoid VOI**, and then click **Set Linear VOI**. | The render changes on each click. The label shows the function, the window and the state of the inversion. |
| Compare the two renders. | The linear function clips at the window: the air and the lung below -800 are solid black, and the bone above 200 is solid white. The sigmoid function does not clip. The sigmoid function maps the values out to about WC +/- 1.73 * WW, so the lung and the bone keep a gradation. The mean difference is about 11 grey levels of 255. |
| Drag with the left mouse button. | The window in the label follows the drag. Each function keeps its shape on the new window. |
| Click **Invert**. | The render inverts, and the label shows `inverted`. |
| Click **Set Sigmoid VOI**, and then drag with the left mouse button. | The render stays inverted, and the label still shows `inverted`. Before this PR, the first drag lost the inversion, because `setVOI` made a new curve and the new curve held no inversion. |
| Click **Set Linear VOI** and **Set Sigmoid VOI** one after the other. | The inversion stays through each switch of the function. |

#### 2. `voiLutFunction` - "VOI LUT Function and VOI LUT Sequence (local DICOM)"

This example is new in this PR. The example tests the stack path on the GPU and
on the CPU. Drop a local DICOM P10 file on the viewport, or use the file picker.

Test file: `hand.dcm` from issue #938:
https://github.com/cornerstonejs/cornerstone3D/files/13640395/hand.dcm.zip

The file holds these attributes:

- Modality DX, MONOCHROME2, 14 bits stored.
- No Window Center (0028,1050), no Window Width (0028,1051), and no VOI LUT
  Function (0028,1056).
- One VOI LUT Sequence (0028,3010) item: 16384 entries, first value mapped 0,
  and 14 bits for each entry.

The curve of the file is strongly non linear. The curve deviates from a straight
ramp by 75% of the output range. The curve holds almost all of the contrast
between the input value 1353 and the input value 4157, and it maps every value
above 5000 to white.

The panel to the right of the viewport must show this state:

```
windowCenter (0028,1050): -
windowWidth  (0028,1051): -
VOILUTFunction (0028,1056): (absent)
VOI LUT Sequence (0028,3010): 16384 entries, firstValueMapped 0, 14 bits/entry

viewport VOILUTFunction : LINEAR
viewport invert         : no
viewport voiRange       : 0.00 .. 16383.00
voiRange source         : metadata / VOI LUT
applied VOI             : VOI LUT Sequence, stretched over voiRange
```

The hand must show full contrast. Before this PR, the same file rendered dark
and flat: the GPU path found no window, so it put a linear ramp over the full
range of the stored values. Issue #938 holds an image of each render.

| Step | Expected result |
| --- | --- |
| Click **Use the file's window instead**. | The panel says `This file has no Window Center/Width to fall back to.` The render does not change. |
| Select **LINEAR** in the **VOI LUT Function** dropdown. | A request for a function is the opt out from the sequence, so the dark and flat render returns. `applied VOI` becomes `LINEAR over voiRange`. |
| Click **Back to the VOI LUT Sequence**. | The contrast returns, and `applied VOI` names the sequence again. |
| Drag with the left mouse button. | The shape of the curve stays, and only the range of the curve changes. |
| Click **Invert**, and then drag with the left mouse button. | The render stays inverted, and the panel shows `viewport invert : yes`. |
| Toggle **Use CPU rendering**, and then repeat each step above. | The CPU render agrees with the GPU render. Before this PR, a drag on the CPU path changed nothing for a file that holds a sequence. |

Other files test the other defects that this PR corrects. Each file is in the
issue that the row names.

| File | Issue | Expected result |
| --- | --- | --- |
| MG files that hold a VOI LUT Sequence | #2520 | The panel names the sequence, and the render agrees with the render of the viewer in the issue. |
| An MG file that holds SIGMOID in (0028,1056) | #2844 | The load gives no `Error: Invalid VOI LUT function`, and the panel shows `SIGMOID` and not `S`. |
| `Sample Dicom Images.zip` | #2716 | The first render is not overexposed. |
| `DICOM sample Issue with Invertion.zip` | #985 | The CPU render agrees with the GPU render. |
| An enhanced multiframe file | #2745 | The panel shows the window of the Frame VOI LUT macro, and not the minimum and the maximum of the image. |
| A color file, such as `wado.zip` | #1141 | The colors are correct, because the loader applies no window and no rescale to a color image. |

#### Automated tests

- `pnpm run test:unit` runs the new jest suites: `voiLUTFunction`, `getVOILut`,
  `setDefaultVolumeVOI`, `getVoiFromSigmoidRGBTransferFunction`,
  `normalizeVOILUTSequence` and `createImageColorVOI`.
- `pnpm run test:ci` runs the karma suites, which hold the new guard in
  `stackViewport_gpu_render_test.js`.
- `pnpm run test:e2e:ci -- tests/volumeVoiSigmoid.spec.ts` runs the playwright
  guard for the sigmoid example. The guard reads the grey level of each pixel of
  the canvas, and it fails if the linear render and the sigmoid render agree.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] "OS: 
- [x] "Node version: 
- [x] "Browser:
  

## Summary by CodeRabbit

* **New Features**
  * Added VOI LUT Sequence support for GPU and CPU rendering.
  * Added LINEAR_EXACT and SIGMOID VOI LUT functions.
  * Added enhanced multi-frame DICOM VOI metadata support.
  * Added a demonstration for comparing VOI rendering modes.

* **Bug Fixes**
  * Improved VOI range calculation, LUT validation, caching, and fallback behavior.
  * Improved handling of empty, invalid, and zero-width LUT data.
  * Preserved VOI settings when switching rendering modes or replacing images.
  * Improved normalization of VOI LUT metadata and sequence formats.


daker:
@sedghi @wayfarer3130 can you have a look ? It should fix a bunch of reported issues.

sedghi:
We will review very soon

packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -790,6 +805,7 @@ class StackViewport extends Viewport {
     }
 
     if (typeof VOILUTFunction !== 'undefined') {
+      this.voiLUTFunctionSetByUser = true;
```
jbocce: **The opt-out is triggered by presence rather than by intent, and OHIF trips it on viewport mount.**

Any defined `VOILUTFunction` reaching `setProperties` latches this flag, including a plain `LINEAR` that nobody actually chose. Combined with `_getVOILUTSequenceToApply` returning `undefined` forever after, that means an application which round-trips `getProperties()` back into `setProperties()` permanently disables VOI LUT Sequence rendering.

OHIF does exactly that. Its stored LUT presentation keeps `VOILUTFunction` whenever `isComputedVOI` is false (`CornerstoneViewportService._getLutPresentation`), and `isComputedVOI` is `!voiUpdatedWithSetProperties`, which latches true on the first `setProperties({ voiRange })` — i.e. the first window-level drag. That presentation is then replayed through `setProperties` both in `LegacyViewportBackend.mountStack` and in `setLutPresentation`, so:

1. Open a CR/DX/MG carrying a VOI LUT Sequence — renders correctly with the file's curve. ✅
2. Window-level it. OHIF now persists `VOILUTFunction` in the presentation.
3. Change layout / switch hanging protocol / jump to a measurement and back.
4. `setProperties({ voiRange, VOILUTFunction: 'LINEAR', ... })` replays at mount, the flag latches, and the file's VOI LUT is silently dropped for the life of the viewport — no way back short of `resetProperties()`.

The replayed value is not a user choice at all: `getProperties()` returns `this.VOILUTFunction`, which `_getImageDataMetadata` reassigns from the *image's* metadata on every image load. So what comes back is usually the image's own `LINEAR`.

One fix is to make the opt-out mean "different from what the image itself asks for":

```suggestion
      this.voiLUTFunctionSetByUser =
        getValidVOILUTFunction(VOILUTFunction) !==
        getValidVOILUTFunction(this.csImage?.voiLUTFunction);
```

Both sides normalize an absent value to `LINEAR`, so replaying the image's own function is a no-op while a genuine switch to `SIGMOID`/`LINEAR_EXACT` still opts out. The trade-off is that an application can then no longer ask for a plain `LINEAR` window on an image whose own function is already `LINEAR` — if that matters, this wants a companion escape hatch (something like `setProperties({ useVOILUTSequence: false })`) rather than overloading `VOILUTFunction` with two meanings.

See also my other comment for an alternative fix that I think is smaller.

packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -2730,10 +2832,64 @@ class StackViewport extends Viewport {
...
+
+    if (!viewport) {
+      return;
+    }
+
+    viewport.voiLUT = this._getVOILUTSequenceToApply(image ?? this.csImage);
+  }
+
+  private _getVOILUTSequenceToApply(image: IImage = this.csImage) {
+    if (this.voiLUTFunctionSetByUser) {
```
jbocce: **Alternative (and I think smaller) fix for the opt-out problem, plus a stale-flag question.**

The root cause of the OHIF failure in my other comment is that `getProperties()` reports an image-derived `VOILUTFunction` through the same field an application uses to express intent, so a replayed presentation is indistinguishable from a deliberate request.

That can be fixed on the reporting side instead of the request side: have `getProperties()` omit `VOILUTFunction` when `voiLUTFunctionSetByUser` is false, or add an `isComputedVOILUTFunction` sibling to the existing `isComputedVOI`. OHIF's `cleanProperties` already strips `VOILUTFunction` when the VOI is computed, so its intent is exactly "only persist the function if the user picked one" — it just has no way to tell today. Fixing it here makes that work with no downstream change, and protects other integrators who persist and replay `getProperties()`.

Separately, on the lifetime of the flag: `_getImageDataMetadata` reassigns `this.VOILUTFunction` from each newly loaded image, so after a stack scroll the user's chosen function is gone while this flag is still `true` — the opt-out outlives the setting that caused it, and the sequence stays disabled for images the user never made a choice about. Would it make sense to clear the flag in the same place `this.VOILUTFunction` is reassigned?

Minor, while you are in here: the doc comment describing this method sits above `_syncCPUVOILUTSequence` instead, so that method now carries two stacked JSDoc blocks and this one has none. Looks like a reorder artifact.

packages/core/src/RenderingEngine/StackViewport.ts
jbocce: **A VOI LUT Sequence can be switched off by a value nobody chose, and it never comes back.**

`VOILUTFunction` is tag (0028,1056), which qualifies the *window* branch of C.11.2.1 — it says nothing about a VOI LUT Sequence, and a file shipping an explicit curve usually omits it altogether. Cornerstone still reports a value for it: an absent tag normalizes to `LINEAR`, and `getProperties()` hands that back out. Since the test here is presence rather than intent, an application that saves properties and restores them later replays a `LINEAR` nobody chose — and the curve is then gone for the life of the viewport, because `_getVOILUTSequenceToApply` returns `undefined` until `resetProperties()`.

OHIF does exactly that today:

1. Open a CR/DX/MG carrying a VOI LUT Sequence. The file's curve renders correctly. ✅
2. Window-level it. That sets `voiUpdatedWithSetProperties`, so `isComputedVOI` goes false and OHIF starts keeping `VOILUTFunction: 'LINEAR'` in its saved presentation — it only discards it while the VOI is computed.
3. Change layout, or jump to a measurement and come back. OHIF replays the presentation through `setProperties`, both in `LegacyViewportBackend.mountStack` and in `setLutPresentation`.
4. The flag latches, the curve is gone, and nothing short of `resetProperties()` brings it back.

Two ways out, in the order I would pick them:

- **A dedicated property states the choice directly** — something like `setProperties({ useVOILUTSequence: false })`. `VOILUTFunction` then keeps to its own branch and cannot disable a sequence as a side effect, while "ignore the file's curve, give me a plain window" stays expressible.
- **If you would rather not add public API**, only latch when the request differs from the image's own function:

```suggestion
      this.voiLUTFunctionSetByUser =
        getValidVOILUTFunction(VOILUTFunction) !==
        getValidVOILUTFunction(this.csImage?.voiLUTFunction);
```

  Running both the request and the image's own function through `getValidVOILUTFunction` is what makes that work: it turns anything missing or unrecognized into `LINEAR`, so an image with no (0028,1056) tag compares equal to a request for `LINEAR`. Replaying it becomes a no-op, while a real switch to `SIGMOID`/`LINEAR_EXACT` still opts out. The catch is that on an image whose own function is already `LINEAR`, an application can then no longer ask for a plain window in place of the curve.

One more thing about the flag: it outlives the choice that set it. `getImageDataMetadata` overwrites `this.VOILUTFunction` from each newly loaded image, but the flag stays `true`, so after a stack scroll the sequence is still suppressed for images nobody chose anything for. Clearing it where `this.VOILUTFunction` is reassigned would keep the two in step.

packages/core/src/RenderingEngine/StackViewport.ts
jbocce: **Alternative (and I think smaller) fix for the opt-out problem, plus a stale-flag question.**

The root cause of the OHIF failure in my other comment is that `getProperties()` reports an image-derived `VOILUTFunction` through the same field an application uses to express intent, so a replayed presentation is indistinguishable from a deliberate request.

That can be fixed on the reporting side instead of the request side: have `getProperties()` omit `VOILUTFunction` when `voiLUTFunctionSetByUser` is false, or add an `isComputedVOILUTFunction` sibling to the existing `isComputedVOI`. OHIF's `cleanProperties` already strips `VOILUTFunction` when the VOI is computed, so its intent is exactly "only persist the function if the user picked one" — it just has no way to tell today. Fixing it here makes that work with no downstream change, and protects other integrators who persist and replay `getProperties()`.

Separately, on the lifetime of the flag: `_getImageDataMetadata` reassigns `this.VOILUTFunction` from each newly loaded image, so after a stack scroll the user's chosen function is gone while this flag is still `true` — the opt-out outlives the setting that caused it, and the sequence stays disabled for images the user never made a choice about. Would it make sense to clear the flag in the same place `this.VOILUTFunction` is reassigned?

Minor, while you are in here: the doc comment describing this method sits above `_syncCPUVOILUTSequence` instead, so that method now carries two stacked JSDoc blocks and this one has none. Looks like a reorder artifact.

packages/core/src/RenderingEngine/helpers/setDefaultVolumeVOI.ts
```diff
@@ -67,9 +68,22 @@ export async function getDefaultVolumeVOIRange(
...
+  // window. PT window width/center is expressed in the unscaled counts, so
+  // applying it to SUV values produces an enormous range and a black volume.
+  // This override used to run only on the min/max path below, so a PT series
+  // that shipped a window skipped it and the volume viewport disagreed with the
+  // stack viewport, which has always preferred its own PT range. It applies to
+  // a volume with no imageIds too, whose window came from the volume metadata
+  // rather than from an instance: scaling is a property of the volume, not of
+  // how its window was found.
+  if (voi) {
+    voi = handlePreScaledVolume(imageVolume, voi);
```
jbocce: **This is a bigger improvement than the diff makes it look — worth a release note.**

A prescaled PT volume holds SUV values, while the file's Window Center/Width describes the raw counts, so the window is denominated in the wrong units — applied to SUV data it gives an enormous range and a near-black volume. The 0–5 override only ran when no window was found, which means the series that *did* ship a window were exactly the ones being displayed wrongly. Running it on both paths fixes them, and brings volume viewports in line with what stack viewports have always done.

Nice to see the three new `setDefaultVolumeVOI` cases pinning this, including the unscaled PT volume keeping its metadata window.

Only ask: could this get a line in the release notes? PET volumes will visibly change for the better on upgrade, and nobody would guess that from the PR title.

packages/core/src/RenderingEngine/StackViewport.ts
```diff
@@ -2730,10 +2832,64 @@ class StackViewport extends Viewport {
...
+   */
+  /**
+   * Keeps the CPU fallback viewport's VOI LUT Sequence in step with
+   * {@link _getVOILUTSequenceToApply}, which is what decides the same question
+   * on the GPU path: the image's sequence drives the display unless the user
+   * asked for a VOI LUT Function explicitly, in which case it has to be off the
+   * viewport - `getVOILut` prefers `viewport.voiLUT` over the function and the
+   * request would otherwise be silently ignored.
+   */
+  private _syncCPUVOILUTSequence(): void {
```
jbocce, line 2954: Nit: this method has picked up two stacked doc comments. The first one describes `_getVOILUTSequenceToApply` (and reads well there) — looks like it was left behind when the methods were reordered, so that method now has none.

jbocce (changes requested):

packages/core/src/RenderingEngine/StackViewport.ts
daker: i proposed something

packages/core/src/RenderingEngine/helpers/setDefaultVolumeVOI.ts
daker: added a new section in the migration note

packages/core/src/RenderingEngine/helpers/planarImageRendering.ts
```diff
@@ -157,11 +173,28 @@ export function applyPlanarImagePresentation(args: {
...
+    defaultVOILUT,
+    props,
+  } = args;
   const property = actor.getProperty();
   const voiRange = props?.voiRange ?? defaultVOIRange;
+  // An explicitly requested VOI LUT Function opts out of the file's VOI LUT
+  // Sequence; an explicit range does not - the curve is stretched over it, so
+  // window level keeps the shape the file specified
+  const voiLUT =
+    props?.voiLUTFunction === undefined ? defaultVOILUT : undefined;
```
jbocce: This path still uses the old rule: if the presentation carries a `voiLUTFunction` at all, the file's curve is dropped — the same presence-over-intent test the StackViewport fix just replaced. There is no `useVOILUTSequence` here either, so an application on the generic viewports cannot ask for one behaviour or the other.

Nothing in the library writes `voiLUTFunction` into a display set presentation today, so it takes an application doing that to hit it. But the two viewport families now disagree about the same file, and the new doc comment on `_getVOILUTSequenceToApply` describes only the StackViewport rule. Align it here, or track it separately? This might be something to bring up with @wayfarer3130 too.

packages/core/src/utilities/windowLevel.ts
```diff
@@ -1,22 +1,39 @@
...
  * @param low - The low window level.
  * @param high - The high window level.
+ * @param voiLUTFunction - 'LINEAR' (default) | 'LINEAR_EXACT' | 'SIGMOID'
  * @returns a JavaScript object with two properties: windowWidth and windowCenter.
  */
 function toWindowLevel(
   low: number,
-  high: number
+  high: number,
+  voiLUTFunction?: VOILUTFunctionType | string
```
jbocce, line 22: The main caller does not pass this new parameter, so the round trip the docstring promises is not lossless yet.

`WindowLevelTool.getNewRange` converts range → window with `toWindowLevel(lower, upper)` — no function, so LINEAR — and converts back with `toLowHighRange(ww, wc, voiLutFunction)`. For `LINEAR_EXACT` those two formulas differ by half a unit: a 0–100 window returns as 0–101, so every mouse move event in a drag adds 1 to the width and shifts the center by 0.5, on top of the intended delta. `LINEAR` and `SIGMOID` round trip exactly, so only `LINEAR_EXACT` drifts — and this PR is what makes it reachable.

One line: pass the function in the forward direction too.

jbocce (changes requested):

wayfarer3130:
## Review of the VOI LUT Function / VOI LUT Sequence work

Reviewed at branch head `9ccba0ca` — i.e. including the follow-up commits I pushed onto this branch, not just the original two. I re-verified every finding against that head, so the "already addressed" set below is separated out from what is still live.

Test status at `9ccba0ca`: `packages/core` VOI suites 43/43 pass, `packages/dicomImageLoader` VOI suites 91/91 pass.

---

### What each block of changes does

**New shared primitives**

- `utilities/voiLUTFunction.ts` — one normalizer for (0028,1056), handling string / single-element array / padded / wrong-case / `SAMPLED_SIGMOID`. Fixes the `"SIGMOID"[0] === "S"` bug (#2844). `getValidVOILUTFunction` warns once and falls back to LINEAR instead of throwing. The duplicate in `buildMetadata.ts` is deleted and re-exported.
- `utilities/createVOILUTSequenceTransferFunction.ts` — builds a grayscale vtk transfer function from a VOI LUT Sequence, capped at 1024 nodes with a step-preserving refinement pass; plus `isRenderableVOILUT`, `getVOILUTSequenceRange`, `createVOILUTSampler`, `invertVOILUTSample`. Output scale is taken from the largest entry, matching legacy cornerstone's "don't trust `numBitsPerEntry`".
- `windowLevel.ts` — *behavioural*: `toWindowLevel` now takes a function and uses the exact (no ±0.5) formula for LINEAR_EXACT; `toLowHighRange` no longer throws on an unknown value and treats SIGMOID as LINEAR so the window round-trips exactly. The dead `logit` branch is removed.
- `createLinearRGBTransferFunction` — widens a zero-width range by an epsilon so vtk's `setMappingRange` cannot divide by a zero node span (the "all-zero first frame of a US cine poisons every later frame" bug).

**Where the LUT gets decided**

- One rule, implemented three times: `StackViewport._getVOILUTSequenceToApply`, `BaseVolumeViewport._getVOILUTSequenceToApply`, and `planarImageRendering.resolveVOILUTSequenceToApply`. The sequence wins over the window unless (a) `useVOILUTSequence: false`, (b) the app asked for a *different* function than the image's own, or (c) a colormap is set — GPU only, since on the CPU path the colormap composes after the VOI LUT and so is kept.
- *Behavioural*: window/level no longer *replaces* the curve, it re-stretches it over the new range. `_getInitialVOIRange` / `_getVOIRangeForCurrentImage` / `getDefaultImageVOIRange` now return the LUT's own input domain when a sequence applies.
- New public surface: `useVOILUTSequence` property, `voiLUTSequenceApplied` on `VOI_MODIFIED` and on volume `getProperties()`. `StackViewport.getProperties().VOILUTFunction` is now *resolved* (`_getEffectiveVOILUTFunction`) rather than raw.
- `setVOILUTFunction` no longer throws on CPU. `_resetProperties` short-circuits the node-replay / colormap-detection when a sequence is on the actor.

**Volumes**

`setDefaultVolumeVOI` refactored to `getDefaultVolumeVOI` returning range + shape; the VOI search now walks outward from the middle slice instead of only probing it; `windowCenter === 0` is now accepted; prescaled PT always gets 0–5; volumes get sigmoid and sequence curves for the first time.

**Loader**

`normalizeVOILUTSequence.ts` accepts wadouri / dcmjs-naturalized / DICOMweb-JSON-with-InlineBinary shapes; wadors now actually returns `voiLUTSequence`; wadouri reads the Frame VOI LUT macro (0028,9132) per-attribute (fixes #2745); `getLUTs` fixes 65535→65536, always reads LUT Data unsigned, and handles 8-bit-packed entries. *Behavioural*: color images now get no window / VOI / modality LUT and are not prescaled.

**CPU render path**

The VOI LUT Function is now honoured (LINEAR / LINEAR_EXACT / SIGMOID kernels, `w<=1` threshold form), the sequence is stretched over the window rather than index-addressed, and `voiLUTFunction` is added to the LUT cache key, to `lastRenderedViewport`, and to the "identity 256/128" fast-path test.

---

### Still live at `9ccba0ca`

**A. `setVOIGPU`'s early return ignores the LUT's identity — wrong curve on ordinary frame navigation.** `StackViewport.ts:1748-1757`

This is the one that matters most. An ordinary scroll within a stack goes `_updateActorToDisplayImageId` → the `sameImageData && !wasStackInvalidated` fast path → `_setPropertiesFromCache()` → `_getVOIFromCache()` → `_getVOIRangeForCurrentImage()`, which for a sequence returns `getVOILUTSequenceRange(voiLUT)` → `setVOI(voiRange)` with no `forceRecreateLUTFunction` and with `stackInvalidated === false`.

The guard then tests only `voiRange.lower/upper`, `useVOILUTSequence === this.voiLUTSequenceApplied`, `!forceRecreateLUTFunction` and `!stackInvalidated`. Two frames whose sequences share `firstValueMapped` and `numberOfEntries` but differ in LUT Data produce an identical range, so all four conditions hold and the function returns before `_createVOITransferFunction` is ever reached. Frame N renders with frame N-1's curve, silently. Multi-frame files with a per-frame Frame VOI LUT macro are exactly the shape that hits this.

Suggest folding a LUT identity into the guard — the `voiLUT` object reference, or a cheap digest of `(firstValueMapped, numberOfEntries, lut[0], lut[mid], lut[last])`.

**B. `toPlanarDataPresentation` pins the *resolved* function into the shared presentation.** `planarLegacyCompatibility.ts:212-221`

`StackViewport.getProperties().VOILUTFunction` is now resolved rather than raw, so the common `setProperties(getProperties())` presentation save/restore round-trip writes a concrete function into `presentation.voiLUTFunction`. `resolveVOILUTSequenceToApply` (`planarImageRendering.ts:257-278`) then computes `functionIsDifferent === true` for any later frame whose own function differs, drops that frame's sequence, and renders it with the previous frame's function.

Worth noting that `_setPropertiesFromCache` (`StackViewport.ts:1058-1064`) already documents and avoids precisely this hazard by re-asserting the *raw* field; the presentation path has no equivalent guard.

**C. The generic planar CPU *image* path ignores `voiLUTFunction` and `useVOILUTSequence`.** `CpuImageSliceRenderPath.ts:418-433`

It reads `enabledElement.image?.voiLUTFunction` and never assigns `viewport.voiLUT`. The sibling volume path (`PlanarCPUVolumeSampler.ts:413-422`) does both, routing through `resolveVOILUTSequenceToApply`. So on a CPU generic planar *image* viewport, `setProperties({ VOILUTFunction: 'SIGMOID' })` and `useVOILUTSequence: false` are silent no-ops, while the GPU path honours both. Using the image's own function for the range round-trip is defensible, but the app's request should still reach the renderer.

**D. CPU volume window uses the exact algebra but is tagged with the real function.** `PlanarCPUVolumeSampler.ts:410-417`, and the synthetic slice at `1205-1224`

`windowCenter = (lower + upper) / 2`, `windowWidth = upper - lower` is the LINEAR_EXACT relation, but `voiLUTFunction` is then set to the presentation's or the volume's actual function. The CPU renderer un-converts with that function's own formula, and for LINEAR that is the ±0.5 form, so the window comes back shifted by about one level. The stack path was fixed in this PR to route through `toWindowLevel(lower, upper, voiLUTFunction)`; these two sites weren't.

**E. Volume `resetToDefaultProperties` clears the shape flags but may not repaint the actor.** `BaseVolumeViewport.ts:1369-1387`

`voiLUTFunctionSetByUser`, `useVOILUTSequence` and `viewportProperties.VOILUTFunction` are reset unconditionally, but `setVOI` / `setVOILUTFunction` only run when `globalDefaultProperties` actually carries those keys. With no stored defaults, a sigmoid or sequence transfer function stays on the actor while the reported properties say otherwise.

**F. `setStack` does not reset the VOI-decision state.** `StackViewport.ts:2189-2203`

`voiRange`, `invert`, the flips and `interpolationType` are all reset, but `useVOILUTSequence` and `voiLUTSequenceApplied` are not. So an app-level `useVOILUTSequence: false` leaks onto the next display set, and a stale `voiLUTSequenceApplied === true` feeds straight into the guard in (A) on the first frame of the new stack. (`voiLUTFunctionSetByUser` does get cleared, indirectly, via `getImageDataMetadata`.)

**G. Perf: `getVolumeVOIShape` is O(number of instances) and uncached.** `setDefaultVolumeVOI.ts:369-391`

`getVOISourceFromImageIds` walks outward from the middle over every imageId with a `metaData.get(VOI_LUT, …)` per step, and only returns after the whole list when no instance carries a window. Its own doc comment notes it is called on every window-level change, and `PlanarCPUVolumeSampler.createSliceImage` (line 1211) calls it on every resample. For a 500-slice volume with no VOI metadata that is ~500 provider lookups per window-level move and per resample. The outward search is a real improvement in correctness — it just wants a per-volume memo.

**H. Minor: an unbuildable sequence still forces a rebuild on every move.**

The fallback at `StackViewport.ts:1813-1821` correctly stops recording an unbuildable LUT as applied, which fixed the "window level permanently inert" half of this. What remains is that `useVOILUTSequence` stays true while `voiLUTSequenceApplied` is false, so `useVOILUTSequence === this.voiLUTSequenceApplied` never holds and the transfer function is rebuilt on every window-level move. Narrow — it needs `isRenderableVOILUT` to pass while `_createVOITransferFunction` returns nothing.

---

### Already addressed on this branch

For anyone reading against the earlier state of the diff, these were found and are already fixed in the follow-up commits:

- `_syncCPUVOILUTSequence` is now called from `resetProperties` (`:935`) and `setVOILUTFunction` (`:1453`), so the CPU stack path no longer applies frame 1's curve to every later frame.
- `StackViewport.resetProperties` now restores `useVOILUTSequence` / `VOILUTFunction` and clears `voiLUTFunctionSetByUser` (`:918-930`).
- An unbuildable sequence is no longer recorded as applied (`:1813-1821`), so window level is no longer permanently inert on such a viewport.
- A colormap now keeps its own transfer function on the GPU path while the CPU path composes the two (`:1740-1746`).

---

### One design question worth settling before merge

`getImageDataMetadata` clears `voiLUTFunctionSetByUser` and overwrites `this.VOILUTFunction` from each image (`StackViewport.ts:1934-1938`). The comment there argues this is correct, and it is what makes per-frame sequence resolution work. The consequence is that `setProperties({ VOILUTFunction: 'SIGMOID' })` on a stack does **not** survive frame navigation — there is no `voiUpdatedWithSetProperties` equivalent for the function.

That is a defensible reading (the function is per-instance metadata), but it is an asymmetry in the public API: `voiRange` is sticky across frames, `VOILUTFunction` is not. Either choice is fine; it should be documented explicitly, since apps that set a function once and expect it to hold will see it revert on scroll.

---

### Behavioural changes worth a release note

These are intentional rather than bugs, but they will move pixels for existing users:

- Prescaled PT volumes now always get 0–5 even when the metadata carries a usable window (previously only on the min/max fallback branch), and `shouldUseImageIdsForVOI` no longer gates it.
- Volume VOI now accepts `windowCenter === 0` (previously rejected by a truthiness test) and searches neighbouring slices, so series that used to fall back to middle-slice min/max will now render with a metadata window.
- Color images lose their window / VOI / modality LUT and are no longer prescaled, so `_getInitialVOIRange` returns undefined and `setVOIGPU` falls back to the scalar data range.
- `getLUTs` reading LUT Data as unsigned and honouring 8-bit packing changes the rendered curve for any file previously mis-parsed — intended, but it will move pixels.
- `toLowHighRange` no longer throws `'Invalid VOI LUT function'`; anything unrecognised degrades to LINEAR.

---

### Verdict

The centralisation is real and the rule is coherent — one predicate, three call sites, and the doc comments carry the DICOM reasoning. The remaining risk is concentrated in staleness rather than in the rule itself.

(A) is the blocker: it produces a visibly wrong curve on ordinary single-frame navigation, with no error. (B) matters for any app that round-trips presentations. (F) is a cheap fix that also removes one input to (A). (C), (D) and (E) are contained to the generic-viewport CPU paths and to volume reset, and (G)/(H) are performance rather than correctness.

Happy to push fixes for (A) and (F) onto this branch as well if that would help.

daker:
Feel free to fix the remaining items

daker:
@wayfarer3130 i pushed a new version

wayfarer3130 (approved):

PR #2858, opened 2026-08-13
fix(core): keep the camera finite for single-slice volumes

wayfarer3130:
## Context

A viewport showing a single-image series (a scout / topogram) goes black as soon as a second actor is mounted onto it — in our case a segmentation labelmap — and logs:

```
WebGL: INVALID_VALUE: uniformMatrix3fv: no array
```

The slice indicator also reads an out-of-range index, e.g. `536/512`, because the slice count is being measured along the wrong axis.

## Root cause

`snapFocalPointToSlice` locates the current frame with

```js
const fraction = (current - min) / (max - min);
```

A volume one slice thick collapses the slice range to `min === max === current`, so this is `0 / 0` → `NaN`. The clamps below it do not catch it, because `NaN > steps` and `NaN < 0` are both false, so the NaN reaches every component of `newFocalPoint` and `newPosition`.

`setViewReference` hands those straight to `setCamera`, which only sets focal point and position — but vtk.js derives the view plane normal from `position − focalPoint`. With both NaN there is no direction of projection, so the camera falls back to the default axial direction. The view up is never touched and stays on the acquisition plane, leaving a normal and an up that are **parallel**: a basis that spans no plane, hence the unbuildable view matrix and the blank viewport.

## How it is reached

`Viewport.addActors` snapshots the view reference, resets the camera, then replays the reference:

```js
const prevViewRef = this.getViewReference();
this.resetCamera();
this.setViewReference(prevViewRef);      // ← reorients the camera to NaN
this.setViewPresentation(prevViewPresentation);
```

The replayed reference names the plane the viewport is *already* on, so this should be a no-op for orientation. On a single-slice volume it silently moves the camera off the acquisition plane instead. Mounting any second actor is enough to trigger it.

Instrumented trace of the moment it breaks, on a sagittal scout:

```
[cs-camera] addActors:captured   normal=[1,0,0] viewUp=[0,0,1]
[cs-camera] setCamera plane [1,0,0] -> [0,0,-1] | BECAME DEGENERATE
    normal=[0,0,-1] viewUp=[0,0,1]  *** DEGENERATE ***
  setCamera        @ BaseVolumeViewport
  setViewReference @ BaseVolumeViewport
  addActors        @ Viewport
```

## The fix

There is no slice to snap to when the range is degenerate, so return the camera unchanged rather than computing `0 / 0`. `getVolumeViewportScrollInfo` already guards the same zero range for the scroll-step calculation; this is the sibling case that was missed.

Also guards zero and non-finite `spacingInNormalDirection`, which produce the same NaN by a different route.

## Tests

`packages/core/test/snapFocalPointToSlice.jest.js` — 7 cases. Verified that the four degenerate-range cases fail against `main` and pass with the fix, while the two normal scrolling cases are unaffected either way.

Found and verified against a real study in an OHIF-based viewer (contributed by U. Calgary).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved focal-point snapping when scrolling is unavailable, ranges are degenerate, or spacing values are invalid.
  * Prevented non-finite coordinates and preserved valid camera and focal-point positions in these cases.

* **Tests**
  * Added coverage for normal scrolling, zero movement, invalid spacing, degenerate ranges, finite results, and safe result copying.

### DIFF
--- a/packages/core/src/utilities/snapFocalPointToSlice.ts
+++ b/packages/core/src/utilities/snapFocalPointToSlice.ts
@@ -26,6 +26,24 @@ export default function snapFocalPointToSlice(
 ): { newFocalPoint: Point3; newPosition: Point3 } {
   const { min, max, current } = sliceRange;
 
+  // A single-slice volume collapses the range to zero, which makes the fraction
+  // below 0 / 0 = NaN. That NaN reaches newFocalPoint and newPosition, and a camera
+  // whose focal point and position are both NaN has no direction of projection at
+  // all: vtk.js falls back to its default axial direction while the view up stays on
+  // the acquisition plane, leaving a basis that spans no plane and renders nothing.
+  // There is no slice to snap to in that case, so leave the camera alone.
+  // getVolumeViewportScrollInfo guards the same degenerate range.
+  if (
+    max - min === 0 ||
+    !Number.isFinite(spacingInNormalDirection) ||
+    spacingInNormalDirection === 0
+  ) {
+    return {
+      newFocalPoint: [...focalPoint] as Point3,
+      newPosition: [...position] as Point3,
+    };
+  }
+
   // Get the current offset off the camera position so we can add it on at the end.
   const posDiffFromFocalPoint = vec3.create();
 
--- a/packages/core/src/utilities/snapFocalPointToSlice.ts
+++ b/packages/core/src/utilities/snapFocalPointToSlice.ts
@@ -26,18 +26,32 @@ export default function snapFocalPointToSlice(
 ): { newFocalPoint: Point3; newPosition: Point3 } {
   const { min, max, current } = sliceRange;
 
+  // How many steps there are in this direction. Computed before the guard below
+  // because the guard tests it.
+  const steps = Math.round((max - min) / spacingInNormalDirection);
+
   // A single-slice volume collapses the range to zero, which makes the fraction
   // below 0 / 0 = NaN. That NaN reaches newFocalPoint and newPosition, and a camera
   // whose focal point and position are both NaN has no direction of projection at
   // all: vtk.js falls back to its default axial direction while the view up stays on
   // the acquisition plane, leaving a basis that spans no plane and renders nothing.
   // There is no slice to snap to in that case, so leave the camera alone.
-  // getVolumeViewportScrollInfo guards the same degenerate range.
-  if (
-    max - min === 0 ||
-    !Number.isFinite(spacingInNormalDirection) ||
-    spacingInNormalDirection === 0
-  ) {
+  //
+  // The test is on `steps` rather than only on the raw range because getSliceRange
+  // measures that range by rotating the volume's corners until the normal points at
+  // +X and taking min/max of the X components. The rotation is the identity only for
+  // an axis-aligned positive normal; for an antiparallel or oblique one its
+  // off-diagonal terms are ~1e-16, so the coplanar corners of a one-slice volume come
+  // out differing in the last bits and the range is something like 3e-14 rather than
+  // 0. `steps` rounds both shapes of it to 0.
+  //
+  // `!(spacingInNormalDirection > 0)` folds together zero, negative, NaN and
+  // undefined spacing, which produce the same NaN by a different route. It is needed
+  // in its own right, since `steps` is NaN when spacing is and `NaN <= 0` is false.
+  //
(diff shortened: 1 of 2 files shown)


packages/core/src/utilities/snapFocalPointToSlice.ts
```diff
@@ -26,6 +26,24 @@ export default function snapFocalPointToSlice(
...
   const { min, max, current } = sliceRange;
 
+  // A single-slice volume collapses the range to zero, which makes the fraction
+  // below 0 / 0 = NaN. That NaN reaches newFocalPoint and newPosition, and a camera
+  // whose focal point and position are both NaN has no direction of projection at
+  // all: vtk.js falls back to its default axial direction while the view up stays on
+  // the acquisition plane, leaving a basis that spans no plane and renders nothing.
+  // There is no slice to snap to in that case, so leave the camera alone.
+  // getVolumeViewportScrollInfo guards the same degenerate range.
+  if (
```
jbocce: The exact `max - min === 0` check is narrower than it looks, because the numbers it's comparing have already been through a rotation.

`getSliceRange` doesn't measure the range along the view normal directly — it rotates all eight corners so the normal points at `+X`, then takes min/max of the X components. When the normal is axis-aligned and positive that rotation is the identity, the four corners sharing a Z bound keep identical X, and a single-slice volume gives `max - min === 0` exactly, which is the case this guard catches.

But for an antiparallel normal like `[0,0,-1]`, or any oblique one, the rotation matrix carries ~1e-16 off-diagonal terms. Those four coplanar corners come out with X values that differ in the last bits, so the range is something like 3e-14 rather than 0 and the guard doesn't fire.

That isn't a NaN in practice — I followed it through: `steps` rounds to 0, so `floatingStepNumber` is `fraction * 0 = 0` and the function ends up returning the camera unchanged anyway. So this is about the guard covering the case it's aimed at, not a live bug.

Checking `steps` instead of the raw range covers both shapes of it, and matches what `getPlanarVolumeSliceNavigationState` already does a few files over (`planarVolumeRendering.ts:29` → `if (steps <= 0 || range === 0)`). It needs the `steps` line hoisted above the guard:

```js
const steps = Math.round((max - min) / spacingInNormalDirection);

if (steps <= 0 || max - min === 0 || !(spacingInNormalDirection > 0)) {
  return {
    newFocalPoint: [...focalPoint] as Point3,
    newPosition: [...position] as Point3,
  };
}
```

`!(spacingInNormalDirection > 0)` is just the two spacing checks folded together — it catches zero, negative, `NaN`, and `undefined` in one predicate. (Note `steps` is `NaN` when spacing is, and `NaN <= 0` is false, so the spacing check still has to be there in its own right.)

With that, the three degenerate-range guards in the codebase — here, `getVolumeViewportScrollInfo`, and `planarVolumeRendering` — all agree on what "can't scroll" means.

jbocce (changes requested):

packages/core/src/utilities/snapFocalPointToSlice.ts
wayfarer3130: Good catch — you're right that the exact `max - min === 0` test is narrower than it reads, and that the rotation in `getSliceRange` is what makes it so. Switched to your version in 89fe5e42f:

```js
const steps = Math.round((max - min) / spacingInNormalDirection);

if (steps <= 0 || max - min === 0 || !(spacingInNormalDirection > 0)) {
```

`steps` hoisted above the guard, and the two spacing checks folded into the single `> 0` predicate. The three guards — here, `getVolumeViewportScrollInfo`, and `getPlanarVolumeSliceNavigationState` — now agree.

I also added the cases the old guard was missing to the test, so the reason for testing `steps` is written down where it can't drift:

- `{ min: 5, max: 5 + 3e-14, current: 5 }` — the residue a non-identity rotation leaves on a one-slice volume, i.e. exactly the shape the old guard let through.
- `{ min: 0, max: 0.4, current: 0 }` with spacing 1 — a range narrower than one step, which rounds to zero steps just the same.
- negative and `undefined` spacing, which the folded predicate now covers.

11 tests pass, and the full `packages/core` suite (562) is green. As you noted these two aren't live NaN bugs — they return the camera unchanged either way — so they're documenting the guard's intended reach rather than catching a regression.

The comment above the guard now explains the rotation residue and why the spacing check still has to stand on its own (`steps` is NaN when spacing is, and `NaN <= 0` is false).

jbocce (approved):
Just wait for the checks to all pass.

PR #2863, opened 2026-08-14
fix(ai): resolve the ONNX wasm directory against the app, not the route

wayfarer3130:
### Context

`ONNXSegmentationController.getConfig()` points ONNX Runtime at its WebAssembly binaries with

```ts
ort.env.wasm.wasmPaths = 'ort/';
```

That prefix is *document-relative*. The browser resolves it against the current route rather than against the application, so it only lands on the copy of `onnxruntime-web/dist` when the page sits exactly one segment deep — which is why it works for the examples and for `viewer.ohif.org/segmentation`.

### Problem

A viewer served from a deeper route (`/viewer/dicomweb`, `<PUBLIC_URL>viewer/<dataSource>`, …) requests `<route>/ort/ort-wasm-*.wasm`. Nothing is served there: the SPA fallback answers with `index.html`, and compiling that as WebAssembly dies with

```
expected magic word 00 61 73 6d, found 3c 21 64 6f   // "<!do"
```

ONNX then reports `no available backend found`, the SAM controller never finishes loading, and the next viewport render throws on the labelmap it never got. Downstream (OHIF/Manta) this reaches the user as a generic "Something went wrong" on the labelmap assist tool, and every consumer whose viewer is not at a single-segment route has to patch it in application code.

The second half of the problem is that the assignment runs from **every** `ONNXSegmentationController` construction, so an application that sets `ort.env.wasm.wasmPaths` itself (a CDN, a versioned path) has its value clobbered — patching it from the app requires trapping the write with an accessor rather than simply assigning.

### Change

* New `packages/ai/src/utils/getOrtWasmPaths.ts` resolves the prefix to an absolute URL against the base the bundler already uses for the assets it emits — webpack/rspack's public path (`__webpack_public_path__`), falling back to `document.baseURI`, and to the bare `'ort/'` outside a browser.
* `getConfig()` only sets `wasmPaths` when the application has not already set it.

### Why not `new URL(..., import.meta.url)`

Every other wasm binary here is located the way the codecs and workers do it — `new URL('@cornerstonejs/codec-charls/decodewasm', import.meta.url)` — letting the bundler resolve, emit and hash the file. That is not available for ONNX at `onnxruntime-web@1.17`: its `exports` map publishes only the JavaScript entry points, so

```
"./dist/ort-wasm-simd.jsep.wasm" is not exported ... from package onnxruntime-web
```

Applications therefore copy `onnxruntime-web/dist` next to their bundle (this repo's `utils/ExampleRunner/template-config.js` copies it to `<example>/ort`) and point the runtime at the copy. Anchoring that copy to the bundle's public path is the closest equivalent, and it is the same base the emitted asset URLs carry. When `onnxruntime-web` is eventually bumped to ≥ 1.21, its `*.bundle.min.mjs` builds resolve their own `.wasm` through `import.meta.url` and this whole assignment can be deleted.

### Impact

| context | before | after |
| --- | --- | --- |
| examples (`publicPath: 'auto'`, ort copied to `<example>/ort`) | `<example>/ort/` | `<example>/ort/` — unchanged |
| viewer at a one-segment route | `<root>/ort/` | `<root>/ort/` — unchanged |
| viewer at a deeper route | `<route>/ort/` — **404 / index.html** | `<root>/ort/` |
| app-configured `wasmPaths` | overwritten | respected |

No new work at load time: one URL resolution, once, inside `getConfig()`. No change to what is fetched or when.

### Testing

* Resolution verified across the deployment shapes above (root public path + deep route, `PUBLIC_URL=/pacs/` + deep route, `publicPath: 'auto'` as the example runner uses, no bundler with and without a `<base>` tag, and a non-browser context).
* Examples resolve to the identical URL they do today, since the example runner leaves `publicPath` at `'auto'` and copies `onnxruntime-web/dist` beside the bundle.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved loading of ONNX Runtime WebAssembly resources across different deployment and bundling configurations.
  * Preserves existing custom resource paths while automatically resolving paths when needed.

### DIFF
--- a/packages/ai/src/ONNXSegmentationController.ts
+++ b/packages/ai/src/ONNXSegmentationController.ts
@@ -14,6 +14,7 @@ import {
   LabelmapBaseTool,
 } from '@cornerstonejs/tools';
 import { Events as aiEvents } from './enums';
+import getOrtWasmPaths from './utils/getOrtWasmPaths';
 
 const { strategies } = cstSegmentation;
 const { fillInsideCircle } = strategies;
@@ -1641,7 +1642,13 @@ export default class ONNXSegmentationController {
     }
     config.threads = parseInt(String(config.threads));
     config.local = parseInt(config.local);
-    ort.env.wasm.wasmPaths = 'ort/';
+    // Leave a location the application configured alone — it may be serving
+    // the binaries from a CDN or a versioned path. Otherwise resolve the
+    // copy of `onnxruntime-web/dist` against the application rather than
+    // against the current route. See `getOrtWasmPaths`.
+    if (!ort.env.wasm.wasmPaths) {
+      ort.env.wasm.wasmPaths = getOrtWasmPaths();
+    }
     ort.env.wasm.numThreads = config.threads;
     ort.env.wasm.proxy = config.provider == 'wasm';
 
--- a/packages/ai/src/utils/getOrtWasmPaths.ts
+++ b/packages/ai/src/utils/getOrtWasmPaths.ts
@@ -44,6 +44,19 @@ function getBundlePublicPath(): string | undefined {
 }
 
 /**
+ * The base a bundler anchors its emitted asset URLs to. This is the definition
+ * webpack and rspack generate for `__webpack_require__.b`, which is what
+ * `new URL(<specifier>, import.meta.url)` compiles down to — so the runtime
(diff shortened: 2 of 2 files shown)


wayfarer3130:
Superseded by #2864 — same commits, from a branch on the repo rather than a fork.

PR #2864, opened 2026-08-14
fix(ai): resolve the ONNX wasm directory against the app, not the route

wayfarer3130:
### Context

`ONNXSegmentationController.getConfig()` points ONNX Runtime at its WebAssembly binaries with

```ts
ort.env.wasm.wasmPaths = 'ort/';
```

That prefix is *document-relative*. The browser resolves it against the current route rather than against the application, so it only lands on the copy of `onnxruntime-web/dist` when the page sits exactly one segment deep — which is why it works for the examples and for `viewer.ohif.org/segmentation`.

### Problem

A viewer served from a deeper route (`/viewer/dicomweb`, `<PUBLIC_URL>viewer/<dataSource>`, …) requests `<route>/ort/ort-wasm-*.wasm`. Nothing is served there: the SPA fallback answers with `index.html`, and compiling that as WebAssembly dies with

```
expected magic word 00 61 73 6d, found 3c 21 64 6f   // "<!do"
```

ONNX then reports `no available backend found`, the SAM controller never finishes loading, and the next viewport render throws on the labelmap it never got. Downstream (OHIF/Manta) this reaches the user as a generic "Something went wrong" on the labelmap assist tool, and every consumer whose viewer is not at a single-segment route has to patch it in application code.

The second half of the problem is that the assignment runs from **every** `ONNXSegmentationController` construction, so an application that sets `ort.env.wasm.wasmPaths` itself (a CDN, a versioned path) has its value clobbered — patching it from the app requires trapping the write with an accessor rather than simply assigning.

### Change

Locating a wasm binary is a standard operation — the codecs and the ONNX Runtime have the same problem for the same reason — so the resolution lives in `@cornerstonejs/core`, beside the base path it reads, rather than in the one package that happened to need it first.

`utilities.resolveWasmBasePath(defaultDirectory)` answers the whole question, in the order an application declares where it is served from:

1. **the configured wasm directory**, when one is set. An application that already serves its codec binaries out of one place names that directory once — `init({ wasmBasePath })` on `@cornerstonejs/dicom-image-loader` — and the ONNX Runtime binaries load from there too, exactly the way the codec binaries do. No second setting, nothing ONNX-specific to configure.
2. **otherwise the standard directory** its owner copies the binaries into (`ort/` here), resolved against the application's base — `PUBLIC_URL` from `window.PUBLIC_URL`, `window.config.path`, or the build-time `process.env.PUBLIC_URL`, defaulting to `'/'` when nothing declares one.
3. **anchored at the page's *origin***: the protocol and host of `window.location` and nothing more. This is the formula `dicom-microscopy-viewer` has always used to locate its own assets from `PUBLIC_URL`, so deployments that already set `PUBLIC_URL` (and those that do not) keep the location they have today. The route never takes part.

Because the wasm directory was private to the DICOM image loader, the value moves into core as well: `utilities.setWasmBasePath` / `getWasmBasePath`, written by the loader's `setOptions` and read back as the fallback in `createImage`. Setting it either way reaches both the codecs and ONNX; the loader option remains the documented entry point.

Under `resolveWasmBasePath` sits its general half, `utilities.resolveApplicationUrl(path)` with `getPublicUrl` alongside — resolving a path against the application rather than against the current route is not specific to wasm either, and this is the pattern `dicom-microscopy-viewer` and `initDemo` already follow by hand.

What is left in `packages/ai/src/utils/getOrtWasmPaths.ts` is the one thing specific to the ONNX Runtime — the standard `ort/` directory name — and nothing else:

```ts
export default function getOrtWasmPaths(): string {
  return utilities.resolveWasmBasePath(DEFAULT_ORT_WASM_DIRECTORY);
}
```

There is deliberately no third way to name the location. `getConfig()` sets `wasmPaths` only when the application has not already set it, so an application either configures `wasmBasePath`, declares `PUBLIC_URL`, or assigns `ort.env.wasm.wasmPaths` itself.

### Why not `new URL(..., import.meta.url)`

Every other wasm binary here is located the way the codecs and workers do it — `new URL('@cornerstonejs/codec-charls/decodewasm', import.meta.url)` — letting the bundler resolve, emit and hash the file. That is not available for ONNX at `onnxruntime-web@1.17`: its `exports` map publishes only the JavaScript entry points, so

```
"./dist/ort-wasm-simd.jsep.wasm" is not exported ... from package onnxruntime-web
```

Applications therefore copy `onnxruntime-web/dist` somewhere they serve and point the runtime at the copy, which is why the location has to be *declared* rather than resolved. When `onnxruntime-web` is eventually bumped to ≥ 1.21, its `*.bundle.min.mjs` builds resolve their own `.wasm` through `import.meta.url` and this whole assignment can be deleted.

### Impact

The old prefix resolved against the *document's directory* — the route with its last segment dropped as a file name — so it happened to work only while the page sat at exactly the right depth.

| deployment | page URL | before | after |
| --- | --- | --- | --- |
| root app, index or one-segment route | `/`, `/segmentation` | `/ort/` | `/ort/` — unchanged |
| root app, trailing slash | `/segmentation/` | `/segmentation/ort/` — **404 / index.html** | `/ort/` |
| root app, deeper route | `/viewer/dicomweb` | `/viewer/ort/` — **404 / index.html** | `/ort/` |
| subpath + `PUBLIC_URL=/pacs/` | `/pacs/viewer` | `/pacs/ort/` | `/pacs/ort/` — unchanged |
| subpath + `PUBLIC_URL=/pacs/` | `/pacs/viewer/dicomweb` | `/pacs/viewer/ort/` — **404 / index.html** | `/pacs/ort/` |
| `wasmBasePath` set | any | `<route>/ort/` — ignored the setting | `<wasmBasePath>` |
| app assigned `ort.env.wasm.wasmPaths` | any | overwritten | respected |
| examples, docs live-examples | `/`, `/live-examples/x.html` | `<page dir>/ort/` | `<page dir>/ort/` — unchanged |

A **subpath** deployment that declares neither `wasmBasePath` nor `PUBLIC_URL` looks in `/ort/`, where the old code happened to find `/<subpath>/ort/` from a page at the subpath root. That case was never supported rather than regressed: `onnxruntime-web@1.17` exports only its JavaScript entry points, so there is no module-relative base to derive and nothing to fall back to — the location has to be declared. Both ways of declaring it are documented in `packages/docs/docs/getting-started/vue-angular-react-vite.md`.

The examples serve `ort/` beside the page — at the root under the example dev server, under `/live-examples/` on the docs site — so `initDemo` declares the page's own directory as `PUBLIC_URL`, which is exactly what the route-relative prefix resolved to in both places.

No new work at load time: one URL resolution, once, inside `getConfig()`. No change to what is fetched or when.

### Testing

* New `packages/core/test/wasmBasePath.jest.js` — 21 tests over both resolvers: every `PUBLIC_URL` source and precedence, a deep route, a base with and without its trailing slash, absolute paths, full-URL paths and bases, and a configured `wasmBasePath` (absolute, relative, missing its trailing slash, cleared by an empty value).
* `packages/ai` gets a jest project (picked up by the root `projects: packages/*/jest.config.js` glob) covering the three decisions that module still makes.
* Full suites pass: 34 suites / 572 tests in `core`, 397 tests in `dicomImageLoader` including `wasmBasePath.spec.ts`. `tsc --noEmit` is clean on `ai`, `core` and `dicomImageLoader`.
* `packages/ai/tsconfig.json` now excludes `*.test.ts`/`*.spec.ts`, so the new test does not ship in `dist`.

## Summary by CodeRabbit

* **New Features**
  * Added shared WebAssembly path configuration for AI and image-loading features.
  * Improved ONNX Runtime asset discovery across application base URLs, deep routes, and custom deployments.
  * Exposed utilities for configuring and retrieving WebAssembly paths.
  * Added support for application-relative asset resolution in subpath deployments.

* **Bug Fixes**
  * Preserved custom WebAssembly path settings.
  * Added fallback handling for unresolved or non-browser environments.

* **Documentation**
  * Documented WebAssembly path configuration and default ONNX Runtime asset locations.


wayfarer3130:
@TFRadicalImaging - this is the wasm fix, if you could test it in context.

rleisti:
> @TFRadicalImaging - this is the wasm fix, if you could test it in context.

I tested the fix, and confirmed that the labelmap assist works (and the WASM library is loaded) with and without the local workaround.

wayfarer3130:
@jbocce - this PR has been tested now and is working correctly in a non root URL situation.  Can you review?  It is just loading from a different path as required - not quite in the standard way the codecs libraries work because the path needs setting as a path, not as a resolve unfortunately.

packages/docs/docs/getting-started/vue-angular-react-vite.md
```diff
@@ -151,6 +151,10 @@ dicomImageLoaderInit({
 
 A relative `wasmBasePath` resolves against the decode worker's location, and an absolute path or full URL (e.g. a CDN) is used as given. When the option is unset, the default `import.meta.url` resolution applies, which is what unbundled and script-tag usage relies on.
 
+The path is system-wide rather than loader-specific, so it is also where `@cornerstonejs/ai` looks for the ONNX Runtime binaries — copy `onnxruntime-web/dist` into the same directory and there is nothing further to configure. With no `wasmBasePath` set, those binaries are expected in `ort/` under the application's base, which is taken from `PUBLIC_URL` (`window.PUBLIC_URL`, `window.config.path` or the build-time `process.env.PUBLIC_URL`) and defaults to the server root.
```
jbocce, line 154: Small docs inconsistency with the shared setting.

Line 152 says a relative `wasmBasePath` resolves against the decode worker's location. This paragraph then says the same path is system-wide and is also where `@cornerstonejs/ai` looks. But the ONNX side resolves it against the application root via `resolveApplicationUrl`, not against the worker. So a relative value means two different things depending on who reads it.

Example with `init({ wasmBasePath: 'wasm/' })` and a worker emitted at `/assets/decode-worker-abc123.js`:

| Consumer | Resolves against | Result |
| --- | --- | --- |
| Codec decoders (`shared/wasmBasePath.ts` `resolveWasmUrl`) | worker script URL | `/assets/wasm/` |
| ONNX Runtime (`core/utilities/wasmBasePath.ts` `resolveWasmBasePath`) | application root | `/wasm/` |

Absolute paths and full URLs, like the `/assets/cs-wasm/` in the example above, resolve identically in both, so most users will never hit this.

Two possibilities for this PR: 
1. Say here that `wasmBasePath` should be an absolute path or full URL when the AI package is in use, and note that a relative value is worker-relative for the codecs but app-relative for ONNX. 
2. Making the codecs resolve against the app root too would remove the difference, but that changes existing codec behavior and belongs in its own PR.

jbocce (approved):
Approved but please review the one comment.

PR #2865, opened 2026-08-15
feat(Core): set log level using config

### LINKED ISSUE #2848: [Feature Request] remove or guard extensive logging inside loadImageFromNaturalizedMetadata

With the naturalized metadata introduced in 5.0, loading local .dcm files is supposed to use `loadImageFromNaturalizedMetadata` from `packages/dicomImageLoader/src/imageLoader/wadouri/loadImage.ts` (at least that's my understanding, please correct me if I'm wrong). The function contains an unskippable console.log that is executed on every loaded frame, which results in hundreds of logs per loaded series (which use the expected "naturalized" path). I'd really like to conform to the new loader specifications but I did not find an intuitive way to prevent the resulting console bloat.

**Suggestion**
I'd appreciate either removing the log from the natural path or making it optional. Specific lines (`v5.6.14`):
```js
// packages/dicomImageLoader/src/imageLoader/wadouri/loadImage.ts @228
    const start = Date.now();
    console.log(
      '[dicomImageLoader/wadouri] loadImageFromNaturalizedMetadata: start',
      {
        imageId,
        scheme: parsedImageId.scheme,
        url: parsedImageId.url,
        frameIndex,
      }
    );
```

It's a minimal patch that would substantially improve development experience with the new version.

Please call me out if I'm misunderstanding the new metadata + loader path and the function is not supposed to be called for each frame.

daker:
### Context
Fixes #2848

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [] "OS: 
- [] "Node version: 
- [] "Browser:
  

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed unnecessary diagnostic messages during image metadata loading.
  * Removed a misleading warning when compressed frame data is unavailable.
  * Preserved existing error handling and metadata loading behavior.

### DIFF
--- a/packages/adapters/src/adapters/Cornerstone3D/MeasurementReport.ts
+++ b/packages/adapters/src/adapters/Cornerstone3D/MeasurementReport.ts
@@ -32,7 +32,9 @@ import {
 import LabelData from './LabelData';
 import { utilities as cornerstoneUtilities } from '@cornerstonejs/core';
 
-const cs3dLogger = cornerstoneUtilities.logger.cs3dLog.getLogger('adapters.Cornerstone3D.MeasurementReport');
+const cs3dLogger = cornerstoneUtilities.logger.adaptersLog.getLogger(
+  'Cornerstone3D.MeasurementReport'
+);
 
 const { MetadataModules } = Enums;
 
--- a/packages/adapters/src/adapters/Cornerstone3D/MeasurementReport.ts
+++ b/packages/adapters/src/adapters/Cornerstone3D/MeasurementReport.ts
@@ -30,6 +30,9 @@ import {
   TEXT_ANNOTATION_POSITION,
 } from './constants';
 import LabelData from './LabelData';
+import { utilities as cornerstoneUtilities } from '@cornerstonejs/core';
+
+const cs3dLogger = cornerstoneUtilities.logger.cs3dLog.getLogger('adapters.Cornerstone3D.MeasurementReport');
 
 const { MetadataModules } = Enums;
 
@@ -854,13 +857,13 @@ export default class MeasurementReport {
           measurement.TrackingUniqueIdentifier = trackingUniqueIdentifierValue;
           measurement.predecessorImageId = predecessorImageId;
 
-          console.log(`=== ${toolAdapter.toolType} ===`);
-          console.log(measurement);
+          cs3dLogger.info(`=== ${toolAdapter.toolType} ===`);
+          cs3dLogger.info(measurement);
           measurementData[toolAdapter.toolType] ||= [];
           measurementData[toolAdapter.toolType].push(measurement);
         }
       } catch (e) {
-        console.warn('Unable to generate tool state for', measurementGroup, e);
+        cs3dLogger.warn('Unable to generate tool state for', measurementGroup, e);
       }
     });
 
--- a/packages/core/src/types/Cornerstone3DConfig.ts
+++ b/packages/core/src/types/Cornerstone3DConfig.ts
@@ -1,8 +1,6 @@
 import type { RenderingEngineModeType } from '../types';
 import type { RenderBackendValue } from '../enums';
 
-type LogLevel = 'trace' | 'debug' | 'info' | 'warn' | 'error' | 'silent';
-
 interface Cornerstone3DConfig {
   /**
    * Whether the device is mobile or not.
@@ -116,16 +114,6 @@ interface Cornerstone3DConfig {
   };
 
   /**
-   * Configures logging globally or for individual named loggers.
-   */
-  logging?: {
-    /** The default level for all loggers without a specific override. */
-    level?: LogLevel;
-    /** Levels keyed by the full logger name, such as `cs3d.core.init`. */
-    levels?: Record<string, LogLevel>;
-  };
-
-  /**
    * This function returns an imported module for the given module id.
    * It allows replacing broken packing system imports with external importers
    * that perform lazy imports.
@@ -134,4 +122,4 @@ interface Cornerstone3DConfig {
   peerImport?: (moduleId: string) => Promise<any>;
 }
 
-export type { Cornerstone3DConfig as default, LogLevel };
+export type { Cornerstone3DConfig as default };
--- a/packages/core/src/types/Cornerstone3DConfig.ts
+++ b/packages/core/src/types/Cornerstone3DConfig.ts
@@ -1,6 +1,8 @@
 import type { RenderingEngineModeType } from '../types';
 import type { RenderBackendValue } from '../enums';
 
+type LogLevel = 'trace' | 'debug' | 'info' | 'warn' | 'error' | 'silent';
+
 interface Cornerstone3DConfig {
   /**
    * Whether the device is mobile or not.
@@ -114,6 +116,16 @@ interface Cornerstone3DConfig {
   };
 
   /**
+   * Configures logging globally or for individual named loggers.
+   */
+  logging?: {
+    /** The default level for all loggers without a specific override. */
+    level?: LogLevel;
+    /** Levels keyed by the full logger name, such as `cs3d.core.init`. */
+    levels?: Record<string, LogLevel>;
+  };
+
+  /**
    * This function returns an imported module for the given module id.
    * It allows replacing broken packing system imports with external importers
    * that perform lazy imports.
@@ -122,4 +134,4 @@ interface Cornerstone3DConfig {
   peerImport?: (moduleId: string) => Promise<any>;
 }
 
-export type { Cornerstone3DConfig as default };
+export type { Cornerstone3DConfig as default, LogLevel };
--- a/packages/dicomImageLoader/src/imageLoader/wadouri/loadImage.ts
+++ b/packages/dicomImageLoader/src/imageLoader/wadouri/loadImage.ts
@@ -1,7 +1,11 @@
 import type { ByteArray, DataSet } from 'dicom-parser';
 import type { Types } from '@cornerstonejs/core';
 import { Enums, metaData } from '@cornerstonejs/core';
-import { Enums as MetadataEnums, utilities } from '@cornerstonejs/metadata';
+import {
+  Enums as MetadataEnums,
+  utilities,
+  logging,
+} from '@cornerstonejs/metadata';
 import createImage from '../createImage';
 import { xhrRequest } from '../internal/index';
 import dataSetCacheManager from './dataSetCacheManager';
@@ -16,13 +20,14 @@ import parseImageId from './parseImageId';
 
 const { ImageQualityStatus } = Enums;
 
+const log = logging.loaderLog.getLogger('wadouri');
 const { addDicomPart10Instance } = utilities;
 const NATURALIZED = MetadataEnums.MetadataModules.NATURALIZED;
 
 // add a decache callback function to clear out our dataSetCacheManager
 function addDecache(imageLoadObject: Types.IImageLoadObject, imageId: string) {
   imageLoadObject.decache = function () {
-    // console.log('decache');
+    // log.debug('decache');
     const parsedImageId = parseImageId(imageId);
 
     dataSetCacheManager.unload(parsedImageId.url);
@@ -226,7 +231,7 @@ function loadImageFromNaturalizedMetadata(
 
   const promise = (async (): Promise<DICOMLoaderIImage> => {
     const start = Date.now();
-    console.log(
+    log.info(
       '[dicomImageLoader/wadouri] loadImageFromNaturalizedMetadata: start',
       {
         imageId,
@@ -238,7 +243,7 @@ function loadImageFromNaturalizedMetadata(
 
     let natural = metaData.get(NATURALIZED, imageId);
     if (!natural) {
-      console.log(
+      log.info(
         '[dicomImageLoader/wadouri] loadImageFromNaturalizedMetadata: no NATURALIZED metadata, attempting to fetch and populate',
         { imageId }
       );
@@ -267,7 +272,7 @@ function loadImageFromNaturalizedMetadata(
       { frameIndex }
     );
     if (!frameData) {
-      console.warn(
+      log.warn(
         '[dicomImageLoader/wadouri] loadImageFromNaturalizedMetadata: no COMPRESSED_FRAME_DATA for imageId',
         { imageId, frameIndex }
       );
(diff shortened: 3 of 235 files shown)


jbocce:
@daker is there a particular reason why you are removing these particular loggings?

daker:
logging in a hot path can slowdown the rendering, those logs are unskipable

packages/dicomImageLoader/src/imageLoader/wadouri/loadImage.ts
jbocce, line 1: I got this from Claude. See if this works. thanks.
  @cornerstonejs/utils exports a loglevel-based hierarchy (packages/utils/src/utilities/logging/index.ts:27), including a logger already named for this package:

  export const loaderLog = cs3dLog.getLogger('dicomImageLoader') as Logger;

  It's re-exported from @cornerstonejs/metadata — which loadImage.ts:4 already imports — so it costs no new dependency.

  Why this solves the perf complaint

  dcmjs pulls in loglevel@1.9.2, whose replaceLoggingMethods (node_modules/loglevel/lib/loglevel.js:89) rebinds any method below the active level to a noop. Default level is WARN, so loaderLog.debug(...) is a bare no-op call by default. That kills the two things that actually cost you:

  - the synchronous console I/O per frame
  - devtools retaining the logged object references, which blocks GC and grows memory across hundreds of frames — usually the bigger of the two

  What survives is the object-literal allocation for the args, which is noise next to decoding a frame. If you want literally zero, guard with if (loaderLog.getLevel() <= 1), but I wouldn't bother.

  Suggested patch

  import { Enums as MetadataEnums, utilities, logging } from '@cornerstonejs/metadata';

  const log = logging.loaderLog.getLogger('wadouri');

  // ...
  const start = Date.now();
  log.debug('loadImageFromNaturalizedMetadata: start', {
    imageId,
    scheme: parsedImageId.scheme,
    url: parsedImageId.url,
    frameIndex,
  });

jbocce (changes requested):

daker:
@jbocce i replaced most of the console calls to logger and added a new config to let users toggle the logging

jbocce:
> @jbocce i replaced most of the console calls to logger and added a new config to let users toggle the logging

Thanks. Although, I was really only expecting the one file to change 😃, but I think it is fine. @wayfarer3130 are you ok with these changes too?

jbocce (approved):

jbocce:
Please check the failing unit tests.

jbocce:
I'm not sure about the playwright test failures.

daker:
I fixed the tests

daker:
@jbocce I fixed the tests, but I'm seeing the same Playwright test failures on other new PRs, so I don't think they're related to my code changes.

jbocce:
> @jbocce I fixed the tests, but I'm seeing the same Playwright test failures on other new PRs, so I don't think they're related to my code changes.

Yes, it looks like this example is broken and thus the related tests are also broken... https://www.cornerstonejs.org/live-examples/contourrenderingconfiguration.

wayfarer3130:
Please update the description to actually match - this PR seems a lot mroe about moving console logs into loglevel logs so that they can be configured.  
Hot path logs should be retained, but at a level at which they don't normally display - the time taken to run them when turned off should be sufficiently low that it doesn't interfere in all but the absolute critical paths - and usually those ones are better improved by optimizing them overall than by removing the log messages.

packages/core/src/types/Cornerstone3DConfig.ts
```diff
@@ -113,6 +115,16 @@ interface Cornerstone3DConfig {
     statsOverlay?: boolean;
   };
 
+  /**
```
wayfarer3130: I think we should configure loglevel via the existing loglevel interface, since that will be required for managing it once we also have logs on dcmjs and ohif as well as on other components.   That also allows specifying interceptors to notify/send logs to other locations.

packages/core/src/types/Cornerstone3DConfig.ts
```diff
@@ -113,6 +115,16 @@ interface Cornerstone3DConfig {
...
+   * Configures logging globally or for individual named loggers.
+   */
+  logging?: {
+    /** The default level for all loggers without a specific override. */
+    level?: LogLevel;
+    /** Levels keyed by the full logger name, such as `cs3d.core.init`. */
```
wayfarer3130: We should standardize on exactly what we use for the log names - suggestion:
cs3d.<PACKAGE>.<PATH>.{FILEORAREA>
so that it is well defined what the base path looks like.

daker: i did add a documentation section

packages/core/src/types/Cornerstone3DConfig.ts
daker: i proposed something

packages/docs/docs/how-to-guides/configuration.md
```diff
@@ -3,3 +3,105 @@ id: configuration
...
+## Logging
+
+Cornerstone writes its messages with [loglevel](https://github.com/pmurias/loglevel),
+through the same log root as dcmjs. Configure the logs with the loglevel
+interface, not with the `init` configuration. Thus one interface controls the
+logs of Cornerstone, of dcmjs, and of the other components of your application.
+
+```ts
+import { log } from '@cornerstonejs/core';
+
```
jbocce, line 22: This import doesn't work. `@cornerstonejs/core` never exports `log` - it's behind the `utilities.logger` namespace.

Use one of:

```ts
import { logging } from '@cornerstonejs/utils';
logging.log.getLogger('cs3d.dicomImageLoader.wadouri').setLevel('info');
```

or `import { utilities } from '@cornerstonejs/core'` and then `utilities.logger.log`.

Same line appears in the two snippets further down.

packages/docs/docs/how-to-guides/configuration.md
```diff
@@ -3,3 +3,105 @@ id: configuration
...
+Each logger keeps its own level. To find the loggers that exist, use
+`log.getLoggers()`.
+
+```ts
+import { log } from '@cornerstonejs/core';
+
+// One area
+log.getLogger('cs3d.core.RenderingEngine').setLevel('debug');
+
```
jbocce: Please double check this, but I think this doesn't set the level for an area but instead only for the one logger named exactly `cs3d.core.RenderingEngine`.

dcmjs pins loglevel 1.9.2, which keeps every logger in one flat map (`_loggersByName`). Names with dots in them are just strings; there is no parent/child relationship. So `cs3d.core.RenderingEngine.StackViewport` stays at whatever level it had. Again double check this.

Only the root `log.setLevel()` reaches everything (it rebuilds all loggers that have no level of their own).

So either reword this to "one logger", and/or add an example that loops over `log.getLoggers()` and sets every name starting with the prefix.

packages/docs/docs/how-to-guides/configuration.md
```diff
@@ -3,3 +3,105 @@ id: configuration
...
+# Configuration
+
+Cornerstone Core accepts configuration through its `init` function. Pass the
+configuration on the first call to `init`; subsequent calls return immediately
+after Cornerstone has been initialized.
+
+## Logging
+
+Cornerstone writes its messages with [loglevel](https://github.com/pmurias/loglevel),
+through the same log root as dcmjs. Configure the logs with the loglevel
```
jbocce, line 16: Wrong repo - loglevel is at https://github.com/pimterry/loglevel.

packages/polymorphic-segmentation/src/workers/polySegConverters.js
```diff
@@ -7,6 +7,11 @@ import vtkPlane from '@kitware/vtk.js/Common/DataModel/Plane';
 import vtkPolyData from '@kitware/vtk.js/Common/DataModel/PolyData';
 import vtkContourLoopExtraction from '@kitware/vtk.js/Filters/General/ContourLoopExtraction';
 import vtkCutter from '@kitware/vtk.js/Filters/Core/Cutter';
+import { utilities as cornerstoneUtilities } from '@cornerstonejs/core';
+
+const cs3dLogger = cornerstoneUtilities.logger.workerLog.getLogger(
+  'polymorphicSegmentation.workers.polySegConverters'
```
jbocce: `workerLog` is `cs3d.worker`, so this name comes out as `cs3d.worker.polymorphicSegmentation.workers.polySegConverters`.

`worker` isn't a package, so the package ends up in the third slot instead of the second, which doesn't match the `cs3d.<PACKAGE>.<PATH>.<FILEORAREA>` spec in the docs. Same in `packages/tools/src/workers/computeWorker.js` and `packages/core/src/webWorkerManager/webWorkerManager.ts` (which is core, not a worker package).

Suggestion: drop `workerLog` and use the package's own logger, so this becomes `cs3d.polymorphicSegmentation.workers.polySegConverters`. Being in a worker is already visible from the path.

packages/polymorphic-segmentation/src/canComputeRequestedRepresentation.ts
```diff
@@ -1,5 +1,10 @@
 import type { Types as ToolsTypes } from '@cornerstonejs/tools';
 import { Enums, segmentation, utilities } from '@cornerstonejs/tools';
+import { utilities as cornerstoneUtilities } from '@cornerstonejs/core';
+
+const cs3dLogger = cornerstoneUtilities.logger.toolsLog.getLogger(
+  'polymorphicSegmentation.canComputeRequestedRepresentation'
```
jbocce: `polymorphic-segmentation` is its own package, but using `toolsLog` names it `cs3d.tools.polymorphicSegmentation.*`. Someone silencing `cs3d.tools` would silence this too.

Same in `packages/labelmap-interpolation`.

Suggestion: add `polySegLog` / `labelmapInterpolationLog` next to the other package loggers in `packages/utils/src/utilities/logging/index.ts`, and list both packages in the docs' `<PACKAGE>` examples.

packages/adapters/src/adapters/Cornerstone3D/MeasurementReport.ts
```diff
@@ -30,6 +30,9 @@ import {
   TEXT_ANNOTATION_POSITION,
 } from './constants';
 import LabelData from './LabelData';
+import { utilities as cornerstoneUtilities } from '@cornerstonejs/core';
+
+const cs3dLogger = cornerstoneUtilities.logger.cs3dLog.getLogger('adapters.Cornerstone3D.MeasurementReport');
```
jbocce, line 38: The name is right, but adapters is the only package with no package logger - it uses `cs3dLog` and writes `adapters.` into the string by hand. Every other package has one (`coreLog`, `toolsLog`, ...).

Suggestion: add `adaptersLog` and use `adaptersLog.getLogger('Cornerstone3D.MeasurementReport')`, so the package prefix can't drift.

jbocce (changes requested):

zenna9:
Thanks for working on this!

The current console spam is having a significant impact on the DX for applications using the library. I’m seeing several console logs per second during normal series loading, which can quickly result in hundreds of messages flooding the browser console.

The main issue for users is that these messages are emitted via `console.log` directly, so they cannot be disabled through the existing logger configuration.

I completely agree that standardizing the overall logging architecture is worth discussing, especially given the namespace issues raised in this PR. However, would it be possible to take a **phased approach**?

A minimal fix to suppress/remove the unskippable `console.log` calls would immediately unblock users experiencing this issue. The broader logger namespace refactoring could then be addressed separately in a follow-up PR.

This would allow us to solve the immediate user-facing problem without blocking the fix on the larger logging architecture discussion.

Thanks again to the author and maintainers for working through this!

jbocce:
@daker just following up on this. I think there are review comments from Aug 20 I made that have not been addressed yet - I might be wrong. Are you planning on following up on these? I just don't want this PR that touches many files to go stale because of other commits and/or new files that might have come up since. If we could finish this up soon then that would be great. Thanks so much.

daker:
@jbocce yes i will push the fixes

packages/utils/src/utilities/logging/index.ts
```diff
@@ -24,10 +49,19 @@ export const cs3dLog = getRootLogger('cs3d');
...
 export const loaderLog = cs3dLog.getLogger('dicomImageLoader') as Logger;
+export const polySegLog = cs3dLog.getLogger(
+  'polymorphicSegmentation'
+) as Logger;
+export const labelmapInterpolationLog = cs3dLog.getLogger(
+  'labelmapInterpolation'
+) as Logger;
 export const aiLog = cs3dLog.getLogger('ai') as Logger;
 export const examplesLog = cs3dLog.getLogger('examples') as Logger;
 export const workerLog = cs3dLog.getLogger('worker') as Logger;
```
jbocce, line 66: `workerLog` no longer has any callers in `src` after the worker files moved to their package loggers, but it is still exported here and re-exported from core and metadata.

Either drop it, or mark it `@deprecated` so the re-exports do not look like something new code should pick up.

Related: `packages/core/test/dicomImageLoader_wadouri_test.js:93` still does `utilities.logger.workerLog.setLevel('error')` to hide the "Worker type ... is already registered" warning, but that warning now comes from `cs3d.core.webWorkerManager.webWorkerManager`, so the test output is noisy again. Switching that line to `coreLog.getLogger('webWorkerManager', 'webWorkerManager').setLevel('error')` (or dropping it) keeps it quiet.

packages/nifti-volume-loader/src/cornerstoneNiftiImageLoader.ts
```diff
@@ -16,6 +16,11 @@ import * as NiftiReader from 'nifti-reader-js';
 import { Events } from './enums';
 import { modalityScaleNifti } from './helpers';
 import { getOptions } from './internal';
+import { utilities as cornerstoneUtilities } from '@cornerstonejs/core';
+
+const cs3dLogger = cornerstoneUtilities.logger.loaderLog.getLogger(
+  'nifti.cornerstoneNiftiImageLoader'
```
jbocce: This produces `cs3d.dicomImageLoader.nifti.cornerstoneNiftiImageLoader`, but the file lives in `packages/nifti-volume-loader/src/`, and there is no `nifti` folder in dicomImageLoader.

Under the `cs3d.<PACKAGE>.<PATH>.<FILEORAREA>` convention this should be `cs3d.niftiVolumeLoader.cornerstoneNiftiImageLoader`. Same in `createNiftiImageIdsAndCacheMetadata.ts`.

Same shape as the polySeg / labelmapInterpolation change: add a `niftiVolumeLoaderLog = cs3dLog.getLogger('niftiVolumeLoader')` next to the other package loggers, use it here, and add `niftiVolumeLoader` to the `<PACKAGE>` list in the docs.

jbocce (changes requested):
Just a couple more comments

jbocce (approved):
Thanks for this effort.

The one thing I will kindly ask for is going forward, it is much easier to review code that has main/master merged in instead of rebased on top of. So if you could kindly do that in the future then that would be appreciated.

PR #2873, opened 2026-08-19
fix(voi): switch the VOI LUT function on volume viewports

### LINKED ISSUE #2458: [Bug] Clicking VOI buttons for the volumevoisigmoid example do nothing

wayfarer3130:
> ## Closed: folded into #2856
>
> Every fix here now lives on the `voi-lut-support` branch of #2856, which is
> where it belongs. #2856 is what makes a SIGMOID tagged series apply a sigmoid
> on a volume viewport at all — the `setDefaultVolumeVOI` gap listed under
> **Notes** below — so the read back and inversion fixes in this PR are only
> fully exercised once that lands. Landing them separately would have put the
> window drift and the lost inversion into `main` for however long the rebase
> took.
>
> | Here | In #2856 |
> | --- | --- |
> | `616e2d81c` read the sigmoid VOI range back exactly | taken whole, with its four case jest test |
> | `2ccfc2d7b` report the decoded window in `VOI_MODIFIED` | cherry-picked clean |
> | `b9af62ce6` inversion re-applied on a rebuilt curve | hand ported, and widened to cover a VOI LUT Sequence curve |
> | `b9af62ce6` `volumeVoiSigmoid` example and `tests/volumeVoiSigmoid.spec.ts` | taken whole |
>
> **Not carried over:** the `setVOILUTFunction` ordering fix, the
> `forceRecreateLUTFunction` port, and the planar `VOILUTFunction` mapping.
> #2856 arrived at all three independently, and its per volume curve tracking
> also covers the `useVOILUTSequence: false` transition that a flag keyed on a
> change of function misses. The `setProperties` / `resetToDefaultProperties`
> reorder was dropped as well: it was only needed because the range was read
> back through the inexact sigmoid decode, and with that decode exact both
> orders land on the same curve and the same window.
>
> #2856 also picked up one further defect found while merging: `getProperties`
> solved whatever was on the actor for a window whenever the VOI LUT Function
> was SIGMOID. A colormap and a VOI LUT Sequence each replace the curve while
> that attribute stays in effect, so a colormapped SIGMOID series returned a
> range of `NaN` and window level went dead on the viewport.
>
> **Two corrections to the Notes section below.** The claim that this PR is
> "complementary to #2856 and conflict free (no shared files)" was true when
> written and is not any more — #2856 grew to cover the volume path and now
> shares `BaseVolumeViewport.ts`, `planarLegacyCompatibility.ts` and
> `planarVolumePresentation.ts`. And two of the three items listed there as
> still open are fixed by #2856: a SIGMOID tagged series rendering linear on a
> volume viewport, and a colormap being dropped on a switch to sigmoid.
>
> The original description follows unchanged, as the record of the analysis.

---

### Context

Fixes #2458 - clicking either **Set Linear VOI** or **Set Sigmoid VOI** in the [volumevoisigmoid](https://www.cornerstonejs.org/live-examples/volumevoisigmoid) example does nothing, with no console error.

The example was not the only problem. Two ordering defects in `BaseVolumeViewport` meant the switch never reached the transfer function:

- `setVOILUTFunction` called `setVOI` **before** recording the new function, so `setVOI` always built the transfer function for the *previous* mode. The first sigmoid click did nothing; a later click applied it off by one.
- Coming back from a sampled sigmoid only called `setRange` on the existing 1024 node sigmoid shaped function, so the curve stayed sigmoid (the `Todo:` that was in that branch).

And on the generic/planar viewports `VOILUTFunction` was dropped entirely: `toPlanarDataPresentation` never mapped it onto the presentation's `voiLUTFunction`, and `applyPlanarVolumePresentation` never passed it to the transfer function builder.

Separately, the example could not have shown the difference even once fixed: with an unconstrained window the two curves nearly coincide, which is what the issue thread observed.

### Changes & Results

Core:

- Record the VOI LUT function before `setVOI` reads it back, and rebuild the transfer function when leaving a sampled one instead of rescaling it, re-applying `invert` so an inverted volume stays inverted across a switch. A colormap is untouched when the function does not change.
- Apply the VOI LUT function before the range in `setProperties` / `resetToDefaultProperties`, so a combined `{ VOILUTFunction, voiRange }` call does not apply the range through the previous curve and then read it back out.
- Map `VOILUTFunction` onto the planar data presentation and pass `voiLUTFunction` to the volume slice transfer function, so the generic viewport applies it too. This also affects the stack sigmoid example in compatibility mode (mean grey difference between the two functions measured 0 before, 19.6 after), and it is what makes `props.voiLUTFunction` reachable at all from `setProperties` on those viewports.
- `getVOIModifiedEventDetail` read the emitted range off the transfer function's **mapping range**. That is the VOI only for a linear function - a sampled sigmoid bakes its curve into the nodes, so its mapping range is the whole node domain, `[c - 1.733w, c + 1.560w]`, about **3.3x** the window width and off center. Once the two fixes above make the sigmoid actually apply on the first click, that bogus range reaches every `VOI_MODIFIED` consumer: `voiSyncCallback` copies `detail.range` verbatim into `setProperties({ voiRange })` on the target, and `ViewportColorbar` reads the same field, so a synced or colorbar'd viewport washed out at `WW 3293 / WC -387` for a `-800..200` window. It now takes `voiRange` off the `getProperties(volumeId)` call the method already makes, which decodes the sigmoid back to the real window. Linear viewports are byte identical - vtk's `getRange()` *is* `getMappingRange()`, which is what `getProperties` returns on the non sigmoid branch. Pre existing on `main` (the same range was emitted on the next window level drag), but this PR is what makes it reachable from a single click.
- `getVoiFromSigmoidRGBTransferFunction` had two defects. It returned a **reversed** range for an inverted curve, since the derived window width comes out negative, which flipped a linear function rebuilt from it; the range is now returned in order. And it converted window width/center to a range as `c ± w/2` - LINEAR_EXACT semantics - while every other range/window conversion in the codebase uses the C.11.2.1.2.1 note 4 convention that `createSigmoidRGBTransferFunction` reads the range through. The mismatch widened the window by one on every round trip (`-800..200` came back as `-800..201`, drifting to about `WW 1050` after 50 LINEAR/SIGMOID toggles without a re-supplied `voiRange`). It now converts through `toLowHighRange`, the exact inverse of that `toWindowLevel` call, so a range survives any number of round trips unchanged. Two rounding calls went with it: the window width was rounded before being used to derive the center, and the center is a half integer whenever the range bounds sum to an even number, so `Math.round(-299.5)` shifted the range half a unit in each direction.

`toWindowLevel` is deliberately **not** touched. Its `+1` / `+0.5` terms cancel against `toLowHighRange`, so the sigmoid curve built from DICOM tags was already spec exact (PS3.3 C.11.2.1.3.1, `y = 1 / (1 + exp(-4(x - c)/w))`, with `logit` its exact analytical inverse and no LINEAR `-0.5`/`(w-1)` offsets applied). Adding a sigmoid branch to only one side of that pair would inject a one unit drift per drag into `WindowLevelTool` and `Colorbar`, which call both.

Example:

- Both buttons apply the **same** window, `WW 1000 / WC -300`, so only the shape of the curve changes. The window is placed to clip both ends of this CT: the linear LUT maps lung/air below -800 to solid black and bone above 200 to solid white, where the sigmoid rolls off instead and keeps a gradient. Measured on the canvas: **86% of pixels change, 61% of them by 10 or more grey levels, mean 11.6/255, peak 31/255** - 31 is the maximum possible at identical WW/WC, since the curves differ by 0.119 at the window edges.
- A toolbar label shows the active function and window (the issue also noted there was no feedback that anything happened), and an info section says where to look. With the exact read back it now reads `WW 1000 / WC -300` under **both** functions, rather than `WW 1001 / WC -299` under the sigmoid - which is the point of the example, the same window with only the curve changing.

| Linear | Sigmoid |
| --- | --- |
| lung and bone flat | vertebral bodies and lung keep gradation |

### Testing

1. Visit the `volumeVoiSigmoid` example.
2. Click **Set Sigmoid VOI**: the blown out vertebral bodies gain internal texture and the lung/air lifts off pure black. The label reads `SIGMOID - WW 1000 / WC -300`.
3. Click **Set Linear VOI**: the render returns exactly to the starting image, and the label reads `LINEAR` on the same window.
4. Left click drag still window levels, and both functions follow the new window.

Automated:

- `tests/volumeVoiSigmoid.spec.ts` - new, snapshot free (canvas pixel comparison): the sigmoid render must differ from the linear one, and switching back must restore it. Passes in legacy and compatibility modes, and **fails with `Received: 0` when the core changes are reverted**, so it guards the reported bug.
- `packages/core/test/utilities/getVoiFromSigmoidRGBTransferFunction.jest.js` - new, 4 cases: the ordered range for an inverted curve, exact recovery of the range a sigmoid was built from, a range whose bounds sum to an even number (half integer center, which the old rounding shifted), and 50 consecutive round trips with no drift.
- Full jest suite passes (118 suites, 1850 tests, 4 skipped). `volumeBasic` snapshot unchanged.

### Notes

- Complementary to #2856 and conflict free (no shared files). That PR fixes the stack/CPU/metadata side and keeps the same `forceRecreateLUTFunction` mechanism on the stack path that this ports to the volume path; it does not touch volume viewport switching. One follow up once it lands: use its shared `getValidVOILUTFunction` in `setVOILUTFunction` instead of the inline enum check. (The other follow up previously listed here, an exactly invertible `getVoiFromSigmoidRGBTransferFunction`, is done in this PR - the 1 HU rounding in the label is gone.)
- Still open, and not addressed here: a SIGMOID tagged series renders linear on a volume viewport, because `setDefaultVolumeVOI` writes only the range to the actor and never sets `viewportProperties.VOILUTFunction` - it takes an actor, not a viewport, so it cannot.
- A colormap is still dropped when switching to sigmoid. Pre existing, and the `TODO` for PET is carried forward rather than widened.
- Also still open: `toLowHighRange` has no sigmoid branch of its own - it routes `SAMPLED_SIGMOID` through the linear formula, with the asymptotic 1%/99% version sitting commented out. Consistent as it stands, and changing it would change range semantics app wide, so it is left alone.
- Known sampling deviation, unchanged by this PR: the standard's sigmoid is asymptotic over all x, while the implementation samples 1024 nodes uniformly in y over `[1/1026, 1024/1026]`, so vtk clamps outside the node domain at y = 0.00097 and y = 0.99805 rather than 0 and 1 - about 0.2% of full output range at the bright end. Uniform in y is the right way round, since it puts the dense samples where the slope is steep. It is also the asymmetry of that interval that makes the node domain off center, and hence why the mapping range could not be used as the VOI above.

#### PR

#### Code

#### Tested Environment

- [x] "OS: Windows 11"
- [x] "Node version: 24.2.0"
- [x] "Browser: Chromium (Playwright)"

## Summary by CodeRabbit

- **New Features**
  - Added support for switching volume visualization between linear and sampled-sigmoid VOI modes.
  - Updated the example interface with the active VOI mode and guidance for comparing rendering results.
  - Window-level changes now update the displayed VOI state.

- **Bug Fixes**
  - Preserved VOI ranges and inversion when switching transfer functions.
  - Ensured inverted sigmoid VOI ranges remain correctly ordered.
  - Improved compatibility with legacy VOI LUT settings.

### DIFF
--- a/packages/core/examples/volumeVoiSigmoid/index.ts
+++ b/packages/core/examples/volumeVoiSigmoid/index.ts
@@ -10,6 +10,8 @@ import {
   createImageIdsAndCacheMetaData,
   setTitleAndDescription,
   addButtonToToolbar,
+  addLabelToToolbar,
+  createInfoSection,
 } from '../../../../utils/demo/helpers';
 
 import * as cornerstoneTools from '@cornerstonejs/tools';
@@ -35,44 +37,67 @@ const { ViewportType } = Enums;
 // ======== Set up page ======== //
 setTitleAndDescription(
   'Sigmoid VOI Volume',
-  'This example shows how to set a Sigmoid VOI on a Volume viewport.'
+  'This example shows how to set a Sigmoid VOI on a Volume viewport, and how it ' +
+    'differs from the linear VOI at the same window level.'
 );
+
+// Both buttons apply this same window (WW 1000 / WC -300) so that only the shape
+// of the VOI LUT curve changes between them, which is the point of the example.
+// The window is deliberately placed to clip both ends of this CT: the linear LUT
+// maps everything below -800 to solid black and everything above 200 to solid
+// white, while the sigmoid LUT rolls off instead of clipping and keeps mapping
+// values out to about WC +/- 1.73 * WW. The lung and the bone are therefore
+// where the two functions differ - flat with the linear LUT, and still showing
+// gradation with the sigmoid one, up to about 30 of 255 grey levels of it right
+// at the window edges. In the middle of the window the two curves nearly
+// coincide, which is why an unconstrained window looks the same either way.
+const voiRange = { lower: -800, upper: 200 };
+
+let voiLabel: HTMLLabelElement;
+
+function getViewport() {
+  // Get the rendering engine
+  const renderingEngine = getRenderingEngine(renderingEngineId);
+
+  // Get the volume viewport
+  return renderingEngine.getViewport(viewportId) as Types.IVolumeViewport;
+}
+
+function setVOILUTFunction(VOILUTFunction: Enums.VOILUTFunctionType) {
+  // The same voiRange is set with both functions, so the only thing that
+  // changes is how the values inside (and outside) that range are mapped.
+  const viewport = getViewport();
+
+  viewport.setProperties({ VOILUTFunction, voiRange });
+  viewport.render();
+  updateVoiLabel();
+}
+
+// Shows which function is active and on which window, since the window level
+// tool can change the window afterwards.
+function updateVoiLabel() {
+  const { voiRange: range, VOILUTFunction } = getViewport().getProperties();
+  const windowWidth = Math.round(range.upper - range.lower);
+  const windowCenter = Math.round((range.lower + range.upper) / 2);
+
+  voiLabel.innerText = `${VOILUTFunction} - WW ${windowWidth} / WC ${windowCenter}`;
+}
 
 addButtonToToolbar({
   title: 'Set Linear VOI',
-  onClick: () => {
-    // Get the rendering engine
-    const renderingEngine = getRenderingEngine(renderingEngineId);
-
-    // Get the volume viewport
-    const viewport = renderingEngine.getViewport(
-      viewportId
-    ) as Types.IVolumeViewport;
-
-    // Set a range to highlight bones
-    viewport.setProperties({ VOILUTFunction: Enums.VOILUTFunctionType.LINEAR });
-
-    viewport.render();
-  },
+  onClick: () => setVOILUTFunction(Enums.VOILUTFunctionType.LINEAR),
 });
 
 addButtonToToolbar({
   title: 'Set Sigmoid VOI',
-  onClick: () => {
-    // Get the rendering engine
-    const renderingEngine = getRenderingEngine(renderingEngineId);
+  onClick: () => setVOILUTFunction(Enums.VOILUTFunctionType.SAMPLED_SIGMOID),
+});
 
-    // Get the volume viewport
-    const viewport = renderingEngine.getViewport(
-      viewportId
-    ) as Types.IVolumeViewport;
-
-    // Set a range to highlight bones
-    viewport.setProperties({
-      VOILUTFunction: Enums.VOILUTFunctionType.SAMPLED_SIGMOID,
-    });
-
-    viewport.render();
+voiLabel = addLabelToToolbar({
+  id: 'voiState',
+  title: 'VOI:',
+  style: {
+    paddingLeft: '10px',
   },
 });
 
@@ -83,6 +108,20 @@ element.style.width = '512px';
 element.style.height = '512px';
 
 content.appendChild(element);
+
+createInfoSection(content)
+  .addInstruction(
+    'Both buttons apply the same window (WW 1000 / WC -300), so only the VOI LUT function changes'
+  )
+  .addInstruction(
+    'Linear clips at the window: lung and air below -800 go solid black, bone above 200 goes solid white'
+  )
+  .addInstruction(
+    'Sigmoid rolls off instead of clipping, so the lung and the bone keep some gradation - that is where to look for the difference'
+  )
+  .addInstruction(
+    'Left click drag window levels the volume, and both functions follow the new window'
+  );
 // ============================= //
 
 /**
@@ -156,10 +195,14 @@ async function run() {
   volume.load();
 
   // Set the volume on the viewport
-  viewport.setVolumes([{ volumeId }]);
+  await viewport.setVolumes([{ volumeId }]);
 
-  // Render the image
-  viewport.render();
+  // Keeps the label honest while the window level tool is used.
+  element.addEventListener(Enums.Events.VOI_MODIFIED, updateVoiLabel);
+
+  // Start off in the linear state that the buttons switch between, so that the
+  // first click on either button is a pure change of the VOI LUT function.
+  setVOILUTFunction(Enums.VOILUTFunctionType.LINEAR);
 }
 
 run();
--- a/packages/core/src/RenderingEngine/BaseVolumeViewport.ts
+++ b/packages/core/src/RenderingEngine/BaseVolumeViewport.ts
@@ -462,22 +462,18 @@ abstract class BaseVolumeViewport extends Viewport {
       throw new Error(`No actor found for the given volumeId: ${volumeId}`);
     }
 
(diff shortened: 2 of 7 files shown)


wayfarer3130:
From @sedghi Fix before merge — finding 1 (BaseVolumeViewport.ts:471). getVOIModifiedEventDetail builds range from transferFunction.getMappingRange(). For a linear TF that's the real VOI (setRange sets it), but for the sampled sigmoid it's the full node domain — 3.3× too wide. voiSyncCallback (packages/tools/src/synchronizers/callbacks/voiSyncCallback.ts:39) copies it verbatim into setProperties({voiRange}) on the target, and ViewportColorbar consumes the same field. So clicking "Sigmoid" on a VOI-synced or colorbar'd viewport now washes out the target at WW 3296. Reading this.viewportProperties.voiRange (or getProperties(volumeId).voiRange, which decodes sigmoid correctly) instead of the mapping range fixes it. Caveat that keeps it off the blocker list: the sigmoid path on main already emits the same garbage range on the next window-level drag — the PR moves it to button-press time rather than creating it.

wayfarer3130:
Closing: every fix here has been folded into #2856, which is the PR that makes a SIGMOID tagged series apply a sigmoid on a volume viewport in the first place. The description above records what moved across and what was dropped as duplicated.

One loose end for #2856: it does not list `Fixes #2458` yet, so that issue will not close automatically when it lands. Worth adding, since the reworked `volumeVoiSigmoid` example and `tests/volumeVoiSigmoid.spec.ts` that demonstrate and guard the reported bug went over with it.

PR #2880, opened 2026-08-25
fix(tools): guard surface polySeg conversion against concurrent re-renders

TFRadicalImaging:
## Context

On a large segmentation (big volume, several segments) shown in a 3D viewport, the **"Converting Labelmap to Surface"** progress toast never completes — the surface conversion churns and the workers never reach `progress:100`.

## Root cause

`displayTools/Surface/surfaceDisplay.ts` `render()` starts a fresh polySeg surface conversion on **every** re-render while a previous conversion for the same viewport is still running. Each concurrent conversion allocates the full labelmap scalar array (via `computeSurfaceData` → `getCompleteScalarDataArray`), so several segments in flight at once exhaust memory; the conversion never settles and the toast hangs.

The sibling `displayTools/Contour/contourDisplay.ts` already guards against exactly this with a per-viewport in-progress flag (`polySegConversionInProgressForViewportId`). The Surface path was missing that guard.

## Fix

Mirror the contour guard in `surfaceDisplay.ts`:

- Track a per-`viewport.id` in-progress flag and skip re-entry while a conversion for that viewport is running.
- Set the flag before `computeAndAddRepresentation` and clear it after it settles.
- Replace the throw-on-empty-result with a `console.warn` (as `contourDisplay` does), so a failed conversion cannot break the render loop.

The change is a faithful, line-for-line mirror of the existing contour implementation.

## Testing

Adds `surfaceDisplay.spec.ts` (tools jest project) covering:

1. Under two concurrent `render()` calls for the same viewport, the conversion is launched **once** (without the guard it is launched twice — the concurrent churn behind the bug).
2. The guard is **per viewport**: two different viewports each convert.
3. The guard is **released** after the conversion settles, so a later render can retry.

All three pass; removing the guard condition makes test (1) fail with two launches, confirming the test exercises the fixed behavior.

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented duplicate surface conversions during simultaneous renders, including across different viewports.
  - Ensured surface conversions remain independent for different segmentations.
  - Allowed subsequent render attempts after a conversion completes or fails.
  - Improved handling of conversion failures so unavailable surface data is skipped gracefully with a warning.

- **Tests**
  - Added coverage for concurrent rendering, viewport behavior, segmentation isolation, retry handling, and conversion failures.

### DIFF
--- a/packages/tools/src/tools/displayTools/Surface/surfaceDisplay.ts
+++ b/packages/tools/src/tools/displayTools/Surface/surfaceDisplay.ts
@@ -11,6 +11,16 @@ import { getPolySeg } from '../../../config';
 import { computeAndAddRepresentation } from '../../../utilities/segmentation/computeAndAddRepresentation';
 import { internalGetHiddenSegmentIndices } from '../../../stateManagement/segmentation/helpers/internalGetHiddenSegmentIndices';
 
+// Guards against launching a second polySeg surface conversion for a
+// segmentation while one is already running. Every re-render (and every viewport
+// showing the segmentation) would otherwise kick off another concurrent
+// conversion; each allocates the full labelmap scalar array, so a large volume
+// with several segments exhausts memory, the workers never reach progress:100,
+// and the "Converting Labelmap to Surface" toast hangs forever. Keyed by
+// segmentationId (not viewport) so several 3D viewports of the same segmentation
+// share a single conversion rather than each starting their own.
+const polySegConversionInProgressForSegmentation = new Set<string>();
+
 /**
  * It removes a segmentation representation from the tool group's viewports and
  * from the segmentation state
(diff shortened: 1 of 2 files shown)


TFRadicalImaging:
@wayfarer3130 this is the Surface-side mirror of the existing `contourDisplay` polySeg conversion guard — the fix for the stuck "Converting Labelmap to Surface" toast on large segmentations. Assigning it to you as discussed so you can get it merged.

packages/tools/src/tools/displayTools/Surface/surfaceDisplay.ts
```diff
@@ -67,23 +74,30 @@ async function render(
     getPolySeg()?.canComputeRequestedRepresentation(
       segmentationId,
       Representations.Surface
-    )
+    ) &&
+    !polySegConversionInProgressForViewportId.get(viewport.id)
```
wayfarer3130: Why is it viewport specific?  If there were two 3d displays, would it double up on the conversion?  It feels like it should be specific to the segmentationId

packages/tools/src/tools/displayTools/Surface/surfaceDisplay.ts
```diff
@@ -67,23 +74,30 @@ async function render(
...
+      );
+    } catch (error) {
+      console.warn(
+        'Unable to compute surface data for segmentationId',
+        segmentationId,
+        error
       );
     }
+
+    polySegConversionInProgressForViewportId.set(viewport.id, false);
```
wayfarer3130: Drop this in a finally block - I know you currenlty catch everything, but having it in a finally is more descriptive.

packages/tools/src/tools/displayTools/Surface/surfaceDisplay.ts
TFRadicalImaging: Changed to key the guard by segmentationId instead of viewport.id, so several 3D viewports of the same segmentation share a single conversion — the 3D four-up layout that triggered this no longer starts one conversion per viewport. Pushed. Note contourDisplay still keys by viewport.id and has the same latent double-up; happy to align it in a follow-up if you want.

packages/tools/src/tools/displayTools/Surface/surfaceDisplay.ts
TFRadicalImaging: Moved the release into a finally — it now deletes the segmentationId from the in-progress set there, so it runs on both success and failure. Pushed.

TFRadicalImaging:
@wayfarer3130 ping on this one — both of your comments were addressed on Aug 25 (the conversion guard is keyed by `segmentationId`, and the lock is released in a `finally`). It is still sitting at `REVIEW_REQUIRED`, so it needs your approval to move. Could you take another look when you have a moment?

wayfarer3130 (approved):

PR #2881, opened 2026-08-25
fix(tools): cancel a pending SegmentSelectTool hover activation on mouseleave

TFRadicalImaging:
## Context

With the "hover on segment border to activate" tool (`SegmentSelectTool`) Active, the active segment could change on its own: a user clicks segment N in the segmentation panel, N becomes active, and ~100 ms later the active segment reverts to a different one (whichever the pointer was last over in the viewport). It looks random, but it is deterministic once the tool is Active.

## Root cause

`SegmentSelectTool.mouseMoveCallback` debounces hover activation: every mouse move over the viewport arms a `hoverTimeout` (100 ms default) timer that, on fire, sets the active segment to whatever is under the pointer. The timer is only cleared by the next mouse move or by disabling the tool.

When the user moves off the viewport to click a segment in the panel, no further mouse move happens on the viewport element, so the last-armed timer survives. It fires ~100 ms later, re-derives the segment under the (now stale) last in-viewport pointer position, and calls `setActiveSegmentIndex` directly — overriding the selection the user just made in the panel.

## Fix

Cancel the pending hover activation when the pointer leaves the viewport element (`mouseleave`), and remove the listener when the tool is disabled. Hovering and staying still activates the hovered segment after `hoverTimeout` as before — only the stale-after-leave case is cancelled.

## Testing

Adds `SegmentSelectTool.spec.ts` (tools jest project) covering:

1. Hovering and staying activates the hovered segment after `hoverTimeout` (feature preserved).
2. Leaving the viewport (`mouseleave`) cancels the pending activation, so it cannot override a later selection (the fix).
3. No timer is armed while the tool is not Active.

All pass; removing the `mouseleave` cancellation makes test (2) fail (the stale timer fires once), confirming the test exercises the fixed behavior. Also verified end-to-end in a running viewer: with the tool Active, selecting one segment then moving off the viewport no longer reverts the active segment, while hovering-and-staying still activates.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved segment selection behavior when hovering over elements.
  - Pending hover selections are now canceled when the pointer leaves the viewport or the tool is disabled.
  - Prevented unnecessary hover activation scheduling when passive mode is enabled.

- **Tests**
  - Added coverage for delayed activation, cancellation, cleanup, and passive-mode behavior.


TFRadicalImaging:
@wayfarer3130 this is the CS3D fix for the "segment index sometimes randomly changes" report — the SegmentSelectTool hover timer surviving the pointer leaving the viewport and reverting the active segment ~100ms after a panel selection. Root cause and the fix are confirmed live (A/B) in a running viewer. Sending it upstream so it can go into the single combined build, alongside #2880.

wayfarer3130 (approved):

PR #2883, opened 2026-08-27
fix(examples): repoint dead ohif-assets URLs and fix the Windows example build

wayfarer3130:
### Context

Two Playwright tests have been failing on `main`, in both the **legacy** and **compatibility** jobs, on the initial attempt and every retry:

- `tests/contourRendering.spec.ts:13` — Contour Rendering › should add a contour as a segmentation to a volume viewport
- `tests/contourRenderingTiled.spec.ts:14` — same test, tiled variant

```
TimeoutError: locator.waitFor: Timeout 30000ms exceeded.
Call log:
  - waiting for locator('canvas.cornerstone-canvas').first() to be visible
   at utils/checkForCanvasSnapshot.ts:126
```

This is **not** a snapshot/pixel-diff failure — the comparison never runs. The canvas never becomes visible because the example page isn't rendering at all.

Not caused by any open PR. The same two tests fail the same way on unrelated PR runs days earlier: [32519385664](https://github.com/cornerstonejs/cornerstone3D/actions/runs/32519385664) (Aug 21), [32586300257](https://github.com/cornerstonejs/cornerstone3D/actions/runs/32586300257) (Aug 22), [32863905323](https://github.com/cornerstonejs/cornerstone3D/actions/runs/32863905323) (Aug 25). Playwright (next) passes in all of them, but that is not evidence the test works — per `scripts/run-playright.sh:314-322`, `--next` runs only `tests/genericViewport/**`, so the contour tests never run in that mode.

#### Root cause

**The `ohif-assets` S3 bucket has been deleted.** Every URL in `utils/assets/assetsURL.json` returns:

```
HTTP 404 — <Code>NoSuchBucket</Code><BucketName>ohif-assets</BucketName>
```

The chain to the timeout:

1. `createAndCacheGeometriesFromContours` does `fetch(assetsURL[name]).then(res => res.json())` with no status check — the 404 body is XML, so `res.json()` throws `SyntaxError`.
2. That rejects at `packages/tools/examples/contourRendering/index.ts:131`, **before** `renderingEngine.setViewports()` on line 165.
3. No viewport → no `canvas.cornerstone-canvas` → the `waitFor({ state: 'visible' })` times out.

No 5.0 API break, and nothing wrong in library code. It also explains why only these two tests were affected: `contourRendering` is the only asset-dependent example with a live test. `tests/surfaceRendering.spec.ts` has its entire body commented out ("this is too much for the CI"), and `contourRenderingConfiguration` / `contourRenderingMultiple` have no specs — so the other broken examples fail silently.

While fixing this I also found the local example build is broken on Windows in a way that makes it silently produce nothing. Details in the second section below.

#### 1. Repoint demo assets at the live bucket (`68328ad0f`)

The assets were migrated to `ohif-assets-new.s3.us-east-1.amazonaws.com` under identical paths. Verified all eight keys return `200` with `Access-Control-Allow-Origin: *`, and that the served `Contour.json` is **byte-identical** to the unused copy checked in at `packages/tools/examples/contourRendering/Contour.json`.

- `utils/assets/assetsURL.json` — all 8 entries (contour, surface and labelmap demos).
- `packages/nifti-volume-loader/examples/{niftiBasic,niftiWithTools}/index.ts` — same dead bucket, hard-coded. Untested by Playwright but broken for real users.
- `packages/docs/docs/tutorials/basic-video.md` and `packages/docs/docs/migration-guides/2x/7-nifti-volume-loader.md` — copy-pasteable snippets that 404 as written.

`grep 'ohif-assets\.s3'` across the repo now returns nothing outside `node_modules`.

#### 2. Fix the example build (`be7e276ee`)

Three separate cwd/platform bugs that combined to make a local build silently produce nothing:

**The rspack spawn was a silent no-op on Windows.** `build-all-examples-cli.js` spawned `node_modules/.bin/rspack`, which on Windows is a `.cmd` — and since Node 18.20, `spawnSync` refuses to execute `.cmd` without `shell: true`, so it failed with `ENOENT`. The result was never checked, so `./scripts/run-playright.sh --force` printed the example list, exited 0, and rebuilt nothing. It now resolves the CLI's own JS entry from `@rspack/cli/package.json`'s `bin` field and runs it with `process.execPath` — no shim, no shell, identical on every platform — and propagates both `result.error` and a non-zero `result.status`.

**`resolve.modules` pointed outside the repo.** The generated config emitted `path.resolve('../node_modules/')` as *literal text*, so it was evaluated when rspack ran rather than when the config was written. From the repo root that resolves to a directory outside the checkout, and since `resolve.modules` replaces the default lookup, every bare import failed — 403 `Can't resolve` errors, mostly `@kitware/vtk.js/*`. The same bug in the two CopyPlugin `from` paths is why `ort/` and `dicom-microscopy-viewer/` were never copied into `.static-examples`. All are now anchored to the `root` argument already being passed in, along with the `@cornerstonejs/*` aliases, which had the same latent cwd dependency.

**`npm run all-examples` could not work on a fresh clone.** Only the `--build` branch wrote the generated config; the `serve` branch read whatever was on disk, and that file is gitignored. Config generation now happens for both branches.

Before vs after, run from the repo root on Windows:

| | before | after |
|---|---|---|
| `--build` | 403 module-resolution errors, **exit 0** | `Rspack compiled successfully`, **exit 0** |
| `--build` with a broken import | **exit 0** (silent) | `rspack build exited with code 1`, **exit 1** |
| `serve` | `ENOENT`, nothing starts | serves on the configured port |
| `ort/`, `dicom-microscopy-viewer/` | not copied | present in `.static-examples` |

Note: the pre-commit hook's prettier reformatted `template-multiexample-config.js` beyond the logical change — the file had been formatted under an older prettier, so that diff is larger than it looks. The build and both test modes were re-run after the reformat.

### Testing

Both failing tests now pass in **legacy** and **compatibility** modes, against the **existing committed snapshots** — no regenerated screenshots, no raised timeouts, no `test.skip`/`test.fixme`:

```bash
# build once
./scripts/run-playright.sh --project=chromium --force

# legacy
PLAYWRIGHT_SKIP_REBUILD=true PLAYWRIGHT_USE_BUNDLED_CHROMIUM=true \
  npx playwright test tests/contourRendering.spec.ts tests/contourRenderingTiled.spec.ts --project=chromium
#   2 passed (12.4s)

# compatibility
PLAYWRIGHT_SKIP_REBUILD=true PLAYWRIGHT_USE_BUNDLED_CHROMIUM=true PLAYWRIGHT_FORCE_COMPAT=true \
  npx playwright test tests/contourRendering.spec.ts tests/contourRenderingTiled.spec.ts --project=chromium
#   2 passed (12.6s)
```

To check the build fixes:

```bash
# should print "Rspack compiled successfully" and exit 0
node utils/ExampleRunner/build-all-examples-cli.js --build --fromRoot --packages core,tools,dicomImageLoader
echo $?

# add `import 'this-module-does-not-exist';` to any example — should now exit 1 instead of 0

# dev server should start and serve pages
npm run all-examples
```

To check the asset move by hand, open `contourRendering` in the dev server: the contour segmentation renders instead of the page staying blank.

Out of scope, noted for the record: the legacy job also reports three flaky-but-recovered tests (`splineContourSegmentationTools.spec.ts:66`, `stackAnnotationTiled.spec.ts:15`, `stackManipulationTools.spec.ts:34`). Nothing here touches them and they do not share the asset path.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] "OS: Windows 11 (10.0.26200)"
- [x] "Node version: v24.2.0"
- [x] "Browser: Chromium 147.0.7727.15 (Playwright 1.59.1 bundled)"

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated sample imaging and video assets to use the current asset hosting location.
  * Corrected NIFTI examples to load the intended sample dataset.

* **Documentation**
  * Refreshed migration guides and tutorials with updated sample asset links.

* **Developer Experience**
  * Improved example building and serving across platforms.
  * Ensured generated example configurations resolve paths consistently from the repository root.
  * Refreshed all configured sample asset links without changing their file paths.


utils/ExampleRunner/template-multiexample-config.js
```diff
@@ -1,180 +1,186 @@
...
+      // vtk.js 36 pulls in xmlbuilder2@4 -> @oozcitak/url which requires
+      // node's built-in 'url' module, unavailable in the browser bundle.
+      url: false,
+    },
+  },
+  devServer: {
+    hot: true,
+    open: false,
+    port: ${process.env.CS3D_PORT || 3000},
+    historyApiFallback: true,
```
jbocce, line 160: Is it worth adding `static: { directory: require('path').resolve(dir) },` so that `pnpm run all-examples` can be run locally.

jbocce (approved):
Looks fine. See my one comment.

utils/ExampleRunner/template-multiexample-config.js
wayfarer3130: Good catch — it's needed, and your `path.resolve` is load-bearing rather than cosmetic. Added in f228d0ca4.

I checked what the dev server actually serves before and after. Everything rspack emits was already fine; the gap is the index:

| | before | after |
|---|---|---|
| `/` | 404 `Cannot GET /` | 200, the examples index |
| `/index.html` | 404 | 200 |
| `/stackBasic.html` | 200 | 200 |
| `/ort/…`, `/dicom-microscopy-viewer/…`, `/serve.json` | 200 | 200 |

`.static-examples/index.html` is written straight to disk by `build-all-examples-cli.js` rather than emitted by rspack, so dev middleware never knows about it — you land on "Cannot GET /" and have to guess each example's URL by hand.

On `path.resolve`: I first tried plain `directory: dir` and it threw `Using a URL as static.directory is not supported`. `Server.isAbsoluteURL` exempts Windows paths with `/^[a-zA-Z]:\/` — backslashes only — so the posix-slashed `Z:/src/...` form this config uses everywhere falls through to the scheme regex and gets read as a `Z:` URL. `path.resolve` normalises it back to native separators, and is a no-op on POSIX. I left a comment on the line so nobody "simplifies" it away later.

Re-ran the build and both contour tests after the change: `Rspack compiled successfully`, legacy 2 passed, compat 2 passed.

PR #2884, opened 2026-08-27
fix(StackViewport): cancel the pending scroll timer when the viewport is disabled

TFRadicalImaging:
### Context

Navigating quickly between stacks (e.g. rapidly stepping through studies with the arrow keys) intermittently throws an uncaught error and trips the app's error boundary.

The cause is a pending debounced frame change that outlives its viewport. `StackViewport.scroll()` does not always change the frame synchronously: when the target image is not yet cached it defers the change behind a 40 ms `debouncedTimeout` that calls `setImageIdIndex()`. `setImageIdIndex()` opens with `_throwIfDestroyed()`, so if the viewport is torn down within that 40 ms window the timer fires against a destroyed viewport and throws. Every stack viewport arms such a timer on mount, because restoring the position presentation calls `scroll(0)`.

`initializeElementDisabledHandler()` was meant to clear that timer on teardown (all teardown paths — `disableElement`, the rendering engine reset, and `destroy()` — emit `ELEMENT_DISABLED`), but the handler never worked:

```js
private initializeElementDisabledHandler() {
  eventTarget.addEventListener(
    Events.ELEMENT_DISABLED,
    function elementDisabledHandler() {
      clearTimeout(this.debouncedTimeout); // `this` is the event target, not the viewport
      eventTarget.removeEventListener(Events.ELEMENT_DISABLED, elementDisabledHandler);
    }
  );
}
```

It is broken three ways:

1. It is a plain function, so `this` is the `eventTarget`, not the viewport — `clearTimeout(this.debouncedTimeout)` clears `undefined`.
2. It listens on the global `eventTarget` and removes itself on the **first** `ELEMENT_DISABLED` from **any** viewport, so it is usually gone before its own viewport is disabled.
3. It never checks that the event is for its own element.

As a result no teardown path actually cancels the timer.

### Changes & Results

Bind the handler as an arrow function so `this` is the viewport, and gate it on `evt.detail.element === this.element` so each viewport clears only its own pending timer before removing its listener. `ELEMENT_DISABLED` is the single choke-point every teardown path emits, so this covers `disableElement`, the rendering engine reset, and `destroy()`.

Before: a burst of fast navigation throws `_throwIfDestroyed` from the debounced `setImageIdIndex` and the error boundary appears.
After: the pending timer is cancelled on teardown, no error is thrown.

### Testing

1. Open a stacked study and step forward through studies as fast as possible (repeated "next study") so viewports are torn down within ~40 ms of mounting.
2. Before the fix, this throws "Something went wrong" (the `_throwIfDestroyed` error from the debounced `setImageIdIndex`).
3. After the fix, navigation stays clean.

Also verified directly: arm `viewport.debouncedTimeout` with a `setImageIdIndex` call, trigger a real teardown via `renderingEngine.disableElement(viewportId)` (which emits `ELEMENT_DISABLED`), and confirm the timer is cancelled — while an equivalent unguarded timer still fires (positive control).

#### PR

#### Code

- [ ] I have run the `yarn build:update-api` to update the API documentation. (No public API changes.)

- [ ] The documentation page has been updated as necessary for any public API additions or removals. (No public API changes.)

#### Tested Environment

- [x] "OS: macOS
- [x] "Node version: v22
- [x] "Browser: Chrome

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed viewport event handling so disabling one viewport no longer affects other viewports.
  * Improved cleanup of pending actions when a viewport is disabled, preventing delayed or unintended behavior.


TFRadicalImaging:
@wayfarer3130 could you take a look when you get a chance? This is the upstream root-cause fix for the fast-navigation "Something went wrong" — the broken `ELEMENT_DISABLED` handler in `StackViewport` that let the debounced `setImageIdIndex` timer fire against a torn-down viewport. It supersedes the viewport-service-side workaround we carried downstream, since it cancels the timer at the single teardown choke-point instead.

wayfarer3130 (approved):

PR #2885, opened 2026-08-27
feat(core): let a deployment store labelmaps as RLE instead of a frame per slice

wayfarer3130:
### Context

A large multi-segment SEG — a whole-body AI segmentation, say — costs one byte
per pixel per slice in the current labelmap representation, whether or not a
slice contains anything. A labelmap frame is mostly background: a few contiguous
runs of segment values per row. RLE holds the same content in a fraction of the
memory, and lets a per-row question ("is segment N on this slice?") be answered
from the runs instead of by scanning a whole frame.

Cornerstone already has the RLE voxel manager, and
`createAndCacheDerivedImage`'s `voxelRepresentation` option already selects it.
What was missing was a way to reach it: the labelmaps in question are created
deep inside the SEG adapter, where no per-call option can be threaded through
from the host. This adds the configuration that covers those, plus the work
needed to make an RLE labelmap behave like the frame buffer it stands in for.

### Changes & Results

**The configuration.** New `segmentation.labelmapVoxelRepresentation` in the
cornerstone configuration, read by
`createAndCacheDerivedLabelmapImage(s)` when the caller does not pass
`voxelRepresentation` itself:

```js
csInit({
  segmentation: { labelmapVoxelRepresentation: 'RLE' },
});
```

- `'Volume'` is the default and is exactly the existing behaviour.
- The bare string is accepted alongside the enum member, so the value can come
  straight from a host's JSON/deployment configuration.
- An unrecognized value (`'rle'`, the likely typo) warns and falls back to
  `'Volume'`, rather than being a flag that looks set and silently does nothing.
- A per-call `voxelRepresentation` still wins.

It is deliberately opt-in, because it changes what a host reads back out of a
labelmap: an RLE frame's `getScalarData()` is a fresh expansion rather than the
live buffer, so in-place writes to it are discarded.

**Ancillary work**, so that an RLE labelmap stands in for a typed array
correctly:

- `createRLE{Volume,Image}VoxelManager` take `pixelDataConstructor` and
  `defaultValue`. An expanded frame then comes back as the same array type as
  the buffer it replaces, and an unwritten voxel reads as that array's zero
  rather than `undefined`. Choosing a type is also what installs
  `_getConstructor`; without it `getConstructor()` had no scalar data to read
  the type from and fell back to `Float32Array`, which the stack labelmap actor
  would then build its texture as. A map whose creator chose nothing — the
  editing history managers, `addInstanceToImage` — keeps the expansion type and
  the constructor fallback it had before the option existed, so no existing RLE
  user is moved onto a type it never asked for.
- `VoxelManager.getScalarDataLength` and `bytePerVoxel` fall back to deriving
  from the dimensions and the array constructor when there is no backing array.
  Without this, anything sizing a buffer from a voxel manager —
  `getCompleteScalarDataArray`, and so segmentation statistics — threw for an
  RLE labelmap.
- New `VoxelManager.setFromScalarData`, for a caller replacing a manager's whole
  contents. `getScalarData().set(...)` writes into a throwaway expansion on a
  manager that is not array backed; unlike `setScalarData` this leaves the RLE
  map as the source of truth for later reads and edits.
  `convertContourToStackLabelmap` now uses it.
- New `VoxelManager.getLiveScalarData`, which answers whether there is a real
  backing store to read and write, as opposed to an expansion whose reads are a
  snapshot and whose writes are discarded.
- The SEG adapter uses that to keep touching the frame buffer directly — those
  are its hot loops, one access per non-zero voxel — and to fall back to a call
  per voxel only for a representation that has no buffer. The planar overlap
  test in particular read `getPixelData()`, which on an RLE labelmap would have
  missed every collision. The group fill is now also bounded by the frame as
  well as the group, so an over-long group cannot write past the labelmap.
- An RLE labelmap image allocates a 1-element stand-in buffer — allocating a
  full frame would be the exact per-slice cost the encoding exists to avoid,
  and the pixel/bit-depth metadata is still derived from its type — and is
  charged `DEFAULT_RLE_SIZE` in the cache, matching what `addInstanceToImage`
  already does for an RLE image.
- `rleForEach` no longer resets `map.defaultValue` to `undefined`. Nothing in it
  reads through `get`, so the reset only undid a default the map's creator had
  chosen.

**Backward compatibility.** With the flag unset, `Volume` is passed explicitly
where `undefined` used to be, and every consumer branches on
`=== VoxelManagerEnum.RLE`, so the representation and every allocation are
unchanged. The adapter still reads and writes the frame buffer directly, and an
RLE manager created without a chosen pixel type expands to `Uint8ClampedArray`
and reports the same constructor as before, so no existing RLE user changes
either. Existing hosts need no changes.

### Related work

The motivating case is segment navigation in a downstream viewer. Clicking a
segment label to jump to that segment has to either wait on the async statistics
worker for a centroid — which has not settled yet on a large SEG, so the jump
silently does nothing — or find the segment's slices from the labelmap itself.
The labelmap search is the reliable option, since the pixel data is there the
moment the SEG loads.

That search does not require this PR: it works against a full frame buffer by
scanning it. But scanning every pixel of every frame is what it costs today,
whereas an RLE frame answers "is segment N on this slice?" from a handful of runs
per row. On a large multi-segment SEG that is roughly an order of magnitude
faster, which is the difference between navigation that feels instant and
navigation the reader waits on. So the two are related but independent: this PR
is what makes that navigation fast at scale, not what makes it work.

### Known limitations (RLE mode only)

- **Cache accounting under-reports.** `sizeInBytes` is the fixed
  `DEFAULT_RLE_SIZE` while the runs grow as voxels are written, so the cache
  limit is not a true bound in RLE mode. This matches the pre-existing
  `addInstanceToImage` behaviour rather than introducing it.
- **The render path expands per update.** `updateVTKImageDataWithCornerstoneImage`
  and `vtkStreamingOpenGLTexture` call `voxelManager.getScalarData()`, which
  re-expands the frame each time. The trade is resident memory for transient
  allocation: only the slices being rendered expand, instead of every slice
  staying resident. `getScalarData(true)` exists to cache an expansion but has
  no caller yet — wiring it into the labelmap sync is the obvious follow-up if
  brush-edit churn on the active slice measures badly.

### Testing

Unit tests, all green (`npx jest packages/core packages/adapters packages/tools`
— 100 suites, 1398 passing):

- `packages/core/test/labelmapVoxelRepresentation.jest.js` (new) — the config
  resolution: unset, absent section, enum member, bare string, and the
  warn-and-fall-back on an unrecognized value.
- `packages/core/test/utilities/VoxelManager.jest.js` — an RLE manager built
  with a pixel type and default value, and one built without, which is pinned to
  the expansion type and constructor fallback it had before; `getLiveScalarData`
  across an array-backed manager, an RLE manager, and an RLE manager holding a
  cached expansion; scalar-data length and voxel size derived with no backing
  array; and `getCompleteScalarDataArray` over RLE-backed slices, which is the
  segmentation-statistics path.
- `packages/adapters/test/seg*RoundTrip.jest.js` — the SEG round trips, with
  `getAtIndex` added to the voxel-manager mocks for the read-through path.

Manually, with a multi-segment SEG:

1. Load the SEG with no configuration change — unchanged behaviour, and a good
   before-baseline for memory.
2. Set `segmentation: { labelmapVoxelRepresentation: 'RLE' }` in the
   cornerstone init and load the same SEG. It should render identically; segment
   statistics, brush edits, undo/redo, and contour-to-labelmap conversion should
   all still work, at markedly lower resident memory.
3. Set the value to `'rle'` and confirm the console warning and the fall back to
   the full-frame representation.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] "OS: Windows 11"
- [x] "Node version: v24.2.0"
- [ ] "Browser: not yet exercised in a browser — verification so far is the unit
  suites above plus `tsc --noEmit` on core and tools. The manual steps in
  Testing are what still needs a browser run."

## Summary by CodeRabbit

* **New Features**
  * Added configurable Volume or RLE storage for labelmaps.
  * Added support for preserving voxel data types, including 16-bit labelmaps.
  * Improved voxel data access and transfer across segmentation workflows.

* **Bug Fixes**
  * Improved overlap handling and labelmap generation for RLE data.
  * Preserved supplied voxel data when creating RLE labelmaps.
  * Prevented invalid or empty segmentation data from affecting statistics.
  * Added safe fallbacks for unsupported configuration values.
  * Ensured unwritten RLE voxels use the configured default value.


packages/core/src/loaders/imageLoader.ts
```diff
@@ -535,10 +541,19 @@ export function createAndCacheLocalImage(
...
+      VoxelManager.createRLEImageVoxelManager<number>({
+        dimensions,
+        id,
+        // The RLE map is standing in for `scalarDataToUse`, so expanded frames
+        // keep that array's type and an unwritten voxel reads as its zero.
+        pixelDataConstructor: scalarDataToUse.constructor as new (
+          length: number
+        ) => PixelDataTypedArray,
+        defaultValue: 0,
+      })) ||
```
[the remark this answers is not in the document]
wayfarer3130: Fixed in 61ce57177 — this was a real silent-data-loss path.

It isn't reachable through `createAndCacheDerivedImage`, which always hands over a freshly-allocated buffer (deliberately 1 element for RLE). But `createAndCacheLocalImage` is exported and `voxelRepresentation` is part of `LocalImageOptions`, so a direct caller passing populated `scalarData` did lose it.

Took the populate option rather than the reject one — the premise of the feature is that RLE is a storage decision, so it shouldn't change what the image holds. `VoxelManager.setFromScalarData` already does exactly this (clear + `fillFrom`), so:

```js
if (isRle && scalarDataToUse.length >= length) {
  voxelManager.setFromScalarData(scalarDataToUse);
}
```

The length guard matters: without it the derived-image path would scan a full frame on behalf of a 1-element stand-in that holds no voxels, which is the exact per-slice cost the encoding exists to avoid. `>= width * height` is the same test `addInstanceToImage` uses to tell a real frame from dummy data.

Three tests in `labelmapVoxelRepresentation.jest.js`: supplied voxels survive, RLE and Volume agree on identical input, and the stand-in buffer is still not treated as data. I confirmed the first two fail against the previous commit; the third passes either way and is there to keep the guard from regressing.

packages/core/src/utilities/VoxelManager.ts
```diff
@@ -475,10 +478,73 @@ export default class VoxelManager<T> {
...
+    if (map instanceof RLEVoxelMap) {
+      const [width, height, depth] = this.dimensions;
+      // fillFrom appends runs to the rows it finds, so start from empty to get
+      // replace-the-whole-thing semantics.
+      map.clear();
+      map.fillFrom(
+        (i, j, k) => {
+          const value = scalarData[i + j * width + k * this.frameSize];
+          // fillFrom skips undefined, which is what keeps the map sparse.
+          return (value === 0 ? undefined : value) as T;
```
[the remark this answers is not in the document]
wayfarer3130: Fixed in 61ce57177, and the inconsistency was one method wider than the comment noted.

`defaultValue` was honoured by `RLEVoxelMap.get` but by nothing that expands or replaces a frame — three places, not two:

- `updateScalarData` did `scalarData.fill(0)`, so `getScalarData()` disagreed with `getAtIndex()` on every unwritten voxel.
- `getPixelData` had the same bug (`pixelData.fill(0)` for a reused buffer, and a fresh allocation arrives zeroed). This one is a live read path — `voxelManager.getPixelData` is bound to it — so fixing only the two flagged lines would have left it inconsistent with them. Fixed it too.
- `setFromScalarData` treated literal zero as the sparse value, so on a map defaulting to 5 the supplied 0s were dropped and read back as 5.

All three now use `defaultValue ?? 0`. Worth noting `defaultValue` is declared without an initializer (`public defaultValue: T`) despite the docstring saying "Set to 0 by default", so it really is `undefined` on maps that never set it — `?? 0` is what keeps those on the zero behaviour they've always had. No caller in the repo sets a non-zero default today, so nothing changes for existing code; this makes the option self-consistent for when one does.

Four tests in `VoxelManager.jest.js` under `a non-zero defaultValue`: the expansion value, `getPixelData` into both a fresh and a reused array, zeros surviving a replace, and a zero-default map unchanged. The first three fail against the previous commit with exactly the reported symptoms (`Expected: 5, Received: 0`, and `Expected: 0, Received: 5`); the fourth passes either way as the backward-compatibility guard.

wayfarer3130:
@jbocce  - this change just allows using the RLE labelmap representation as an opt-in option for cs3d rendering.  I've tested it with OHIF and although it needs a couple of small OHIF tweaks to avoid two places where direct buffer access is used, it does speed up large labelmaps.  I don't think there is anything int he PR that is particularly concerning - we have been using RLE maps for a while now for various things, just not defaulting to them for new labelmaps.

packages/core/src/utilities/VoxelManager.ts
```diff
@@ -492,6 +563,11 @@ export default class VoxelManager<T> {
       return this._getScalarDataLength();
     }
 
+    const [width, height, depth] = this.dimensions ?? [];
+    if (width && height) {
+      return width * height * (depth || 1) * this.numberOfComponents;
```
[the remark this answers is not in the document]
wayfarer3130: Fixed in 5d537f784. The finding is correct, and both of the failure modes occur.

I reproduced each one. An empty untyped RLE map throws `Cannot read properties of undefined (reading 'BYTES_PER_ELEMENT')`. A map that sets `defaultValue: 0` and holds a voxel gives `NaN`. The second mode does not need an empty map: `_get(0)` on an RLE map returns a number, and a number has no `BYTES_PER_ELEMENT`, so the `_get(0)` fallback cannot give a width for any RLE map. Both modes are new in this PR, because `getScalarDataLength()` derives a length now instead of throwing first.

I read the width from the type that the map expands to, and not from a constant:

```js
if (this.map instanceof RLEVoxelMap) {
  const ExpansionConstructor = this.map.pixelDataConstructor
    ? this.map.pixelDataConstructor
    : Uint8ClampedArray;
  return ExpansionConstructor.BYTES_PER_ELEMENT;
}
```

For an untyped map this gives the same 1 byte as the proposed fix, because `RLEVoxelMap.getScalarData` falls back to `Uint8ClampedArray`. The chain also stays correct if a caller sets `map.pixelDataConstructor` after construction, which leaves `_getConstructor` unwired. `bytePerVoxel` then always agrees with the array that `getScalarData()` returns.

One test in `VoxelManager.jest.js`, under `scalar data length without a backing array`: it covers the empty map and the map with a zero default, and it checks `bytePerVoxel` and `sizeInBytes` for each.

packages/core/src/init.ts
```diff
@@ -54,6 +59,14 @@ const defaultConfig: Cornerstone3DConfig = {
     },
   },
 
+  segmentation: {
```
sedghi, line 62: technically segmentation exist on cstools not cs core, i think we should move it there?

packages/core/src/utilities/VoxelManager.ts
```diff
@@ -478,10 +481,78 @@ export default class VoxelManager<T> {
...
+   *
+   * A manager that keeps its voxels in something else (an RLE map) can still
+   * produce an array from `getScalarData()`, but that is a fresh expansion:
+   * reads of it are a snapshot and writes into it are discarded. A caller with
+   * a whole frame to read or fill wants to know which one it has, so that it
+   * can touch the array directly where that is correct and only fall back to a
+   * call per voxel (`getAtIndex` / `setAtIndex`, or `setFromScalarData`) where
+   * it is not.
+   */
+  public getLiveScalarData(): PixelDataTypedArray | undefined {
```
sedghi: Hmmm maybe getWritableScalarData?

packages/adapters/src/adapters/Cornerstone3D/Segmentation/labelmapImagesFromBuffer.ts
```diff
@@ -1131,8 +1131,18 @@ export function insertPixelDataPlanar({
...
+        // The live frame buffer, where there is one. The overlap test below runs
+        // once per non-zero voxel of the SEG, so it reads that array directly
+        // rather than paying for a call through the voxel manager.
+        //
+        // An RLE labelmap has no live buffer: `getPixelData()` hands back a
+        // fresh expansion that the writes below never show up in, so reading it
+        // would make this test miss every collision. That one is read through
+        // the voxel manager instead.
+        const labelmap2DView = imageVoxelManager
+          ? imageVoxelManager.getLiveScalarData()
```
sedghi: this might be under feature flag off and not on by default right? so might return undefined

sedghi (approved):
See my comments

packages/adapters/src/adapters/Cornerstone3D/Segmentation/labelmapImagesFromBuffer.ts
wayfarer3130: Correct, and that is the design. `Volume` is the default, so this returns the frame buffer today. With `RLE` it returns `undefined`, and every read and every write here handles that.

The method is now `getWritableScalarData` (see the rename below), which says more clearly why it can be `undefined`: an RLE map has no array that a caller can write into. `getPixelData()` gives an array, but the array is a fresh expansion, so a write into the expansion goes nowhere, and a read of the expansion does not show the `setAtIndex` writes in this same loop. That is why the code does not simply fall back to `getPixelData()`.

I checked each of the four places:

- Line 1156, the overlap test: `labelmap2DView ? labelmap2DView[x] : imageVoxelManager.getAtIndex(x)`. The RLE case reads through the voxel manager.
- Line 1182, the write: the `else` branch of `if (imageVoxelManager)`. Only a labelmap that has no voxel manager writes to the array, and that labelmap gets the array from `getPixelData()`.
- Line 1311, `insertOverlappingPixelDataPlanar`: this one does not even ask. It is `imageVoxelManager ? undefined : labelmapImage.getPixelData()`, and the write at line 1328 has the same guard as line 1182.
- Line 1461, `labelmapImagesFromSegBuffer`: `if (frame) { ... } else { ... }`, and the `else` branch writes a voxel at a time.

I have tests for both paths. `packages/adapters/test/` has four SEG round-trip suites, and `packages/core/test/labelmapVoxelRepresentation.jest.js` asserts that an RLE labelmap returns `undefined` here.

One change that does affect a caller outside this repository: code that passes its own mock image to these adapter functions must now give the mock a `getWritableScalarData` method. The four adapter test suites show the shape. A real cornerstone image is not affected, because its `voxelManager` is always a `VoxelManager`.

packages/core/src/utilities/VoxelManager.ts
wayfarer3130: Agreed, that name is better. Renamed in 868850eb5.

The property that each caller tests is whether a write into the array reaches the voxel manager, and `getWritableScalarData` names that property. "Live" also suggested that a read is current, which is true but is not the reason a caller asks.

The rename covers the method, the two adapter call sites, and the test suites in `packages/core/test/` and `packages/adapters/test/`. The docstring now also says that `undefined` is the normal answer for an RLE labelmap, and not an error.

packages/core/src/init.ts
wayfarer3130: I looked at this, and I keep the key in core. Here is the reason.

The two settings are two different kinds of setting. `SegmentationConfig.overwriteMode` in `packages/tools/src/config.ts` is display and editing configuration for the tools: the setting describes what an edit does to the other segments under the brush, and cstools owns the editing. `labelmapVoxelRepresentation` is core configuration: the setting describes how core allocates the voxels of an image that core itself creates. The word `segmentation` in the key names the kind of image, and the word does not move the ownership.

Core already owns the labelmap creation. `createAndCacheDerivedLabelmapImage`, `createAndCacheDerivedLabelmapImages` and `VoxelManagerEnum` are all in core, and `packages/core/src/loaders/imageLoader.ts` is the file that reads the configuration. The key sits next to the function that reads the key. Core cannot import cstools, so a key in cstools is a key that core cannot read.

I also checked what a move needs. Every caller of `createAndCacheDerivedLabelmapImage(s)` must then read the value from cstools and pass the value explicitly. There are four such call sites, two in `packages/adapters` and two in `packages/polymorphic-segmentation`. Both packages have a peer dependency on cstools, so a move is possible. But a host that calls the core function directly then gets `Volume` always, and the host cannot change that from the configuration. One deployment-level setting must apply everywhere, so I do not want that gap.

I continue with the key in core. Tell me if you disagree, and I will make the change; the change is mechanical.

PR #2890, opened 2026-09-01
feat(dicomImageLoader): decode truncated HTJ2K byte ranges at full resolution

wayfarer3130:
### Context

HTJ2K progressive loading had to work around a decoder limitation: OpenJPH
threw on a truncated codestream, so a partial byte range could not be decoded
at full resolution. The workaround was to decode a partial range at a reduced
resolution and let the worker scale the result back up, plus fetch a fairly
large initial buffer (128k) and carry a fallback ladder of successively
coarser `decodeLevel`s in case a decode still failed.

`@cornerstonejs/codec-openjph` 2.4.10 rebases our OpenJPH fork onto upstream
0.30.1 and decodes a truncated codestream instead of throwing. That removes the
reason for both the reduced-resolution decode and the fallback ladder: a full
resolution decode of 32k of codestream loses considerably less than decoding a
quarter-size image and scaling it up afterwards.

### Changes & Results

**Codecs** — bumped to the current releases: `codec-openjph` 2.4.10,
`codec-charls` 1.2.6, `codec-openjpeg` 1.3.3, `codec-libjpeg-turbo-8bit` 1.2.5.
The pinned `minimumReleaseAgeExclude` entries in `pnpm-workspace.yaml` move with
them.

**Full resolution decode of a partial stream** — an explicit `decodeLevel: 0`
on a range or streaming retrieve now means "decode at full resolution from
whatever bytes have arrived", rather than choosing a level from how much of the
frame is present:

- `getImageQualityStatus` reports such an image as `LOSSY` rather than
  `SUBRESOLUTION`. It is full size and only the codestream is incomplete, so it
  is a different quality from a sub-resolution decode, and ordering the two
  correctly is what lets a later stage replace an earlier one. It stays
  `FULL_RESOLUTION` only when the whole frame fit in the first chunk.
- `loadImage` skips `decodeLevelFromComplete` entirely when level 0 is
  requested. Its "no point decoding again" guard now applies only to
  sub-resolution decodes, because at full resolution every additional chunk
  refines the same level and the level alone cannot say whether the work is
  worthwhile.
- The default range `chunkSize` drops from 65536 to 32768, which is enough of
  an HTJ2K codestream to put a recognisable full resolution image up.

**Examples** — the sub-resolution fallback stages (`singleFastFailure`,
`singleMiddleFailure`) existed only because partial decode used to throw and
are removed, taking `stackProgressive` and `htj2kStackBasic` from five stages
to three. The HTJ2K byte range configurations now use `decodeLevel: 0` with a
32k first chunk.

`stackProgressive` was additionally still pointing at rendition names from the
retired `mkdicomweb create --alternate` CLI. It now uses the names
`createdicomweb alternates` actually writes — `htj2k/` and `htj2kLossy/`
instead of primary `frames/` and `/lossy/` — matching what `volumeProgressive`
already does. The `HTJ2K Thumbnail` button is dropped: there is no
`htj2kThumbnail` rendition in the new CLI to retrieve, and a 32k byte range
decoded at full resolution is a better fast preview than a quarter-size
thumbnail anyway.

**Not changed:** sub-resolution decode plus scale-up remains the right route
for renditions that are genuinely small, so the JLS thumbnail configurations
and the `scaleImageFrame`/`bilinear`/`replicate` path in the decode worker are
untouched.

### Testing

The rendition directories the examples retrieve are generated with static
DICOMweb:

```bash
cd /src/static-dicomweb
createdicomweb create "/dicom/DE Images for Rad"
createdicomweb alternates 1.3.6.1.4.1.9590.100.1.2.19841440611855834937505752510708699165 \
  --jls --jls-thumbnail --htj2k --htj2k-lossy

createdicomweb create /src/viewer-testdata/dcm/Juno
createdicomweb alternates 1.3.6.1.4.1.25403.345050719074.3824.20170125113417.1 \
  --series-uid 1.3.6.1.4.1.25403.345050719074.3824.20170125113545.4 \
  --jls --jls-thumbnail --htj2k --htj2k-lossy
```

Serve with `dicomwebserver`, then open the `stackProgressive` example with
`?useLocal=true` and exercise `HTJ2K`, `HTJ2K Bytes` and `HTJ2K Lossy`. The
first rendering should arrive at full frame size rather than a scaled-up
quarter-size image, and no decode failures should appear in the console.
`volumeProgressive` covers the same configurations for a volume viewport.

Automated: `packages/dicomImageLoader` unit tests pass (398), and `tsc --noEmit`
is clean for both `core` and `dicomImageLoader`. Truncated decode was verified
directly against the published 2.4.10 wasm, which decodes a 512x512 frame from
both 10k and 32k truncated slices without throwing.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] "OS: Windows 11"
- [x] "Node version: 22"
- [x] "Browser: Chrome"

## Summary by CodeRabbit

- **New Features**
  - Improved progressive HTJ2K loading with full-resolution decoding from partial image data.
  - Added configurable initial and subsequent range sizes, plus timing controls between progressive decodes.
  - Updated image quality reporting to distinguish full-resolution lossy results from sub-resolution results.

- **Bug Fixes**
  - Corrected decode-level handling for explicit full-resolution requests.
  - Prevented unnecessary repeated decodes during progressive loading.

- **Documentation**
  - Updated progressive-loading guidance, HTJ2K examples, and migration instructions for new retrieval settings.


packages/dicomImageLoader/src/imageLoader/wadors/loadImage.ts
```diff
@@ -185,15 +185,22 @@ function loadImageFromNetwork(
...
-            : decodeLevelFromComplete(
-                percentComplete,
-                options.retrieveOptions?.decodeLevel
-              ));
-        if (!done && lastDecodeLevel <= decodeLevel) {
+            : decodeLevelFromComplete(percentComplete, retrieveDecodeLevel));
+        // At full resolution each additional chunk refines the same level, so
+        // the level alone cannot say whether decoding again is worthwhile -
+        // only a sub-resolution decode can be skipped for not improving.
+        if (!done && decodeLevel > 0 && lastDecodeLevel <= decodeLevel) {
```
jbocce: Adding `decodeLevel > 0` here means the "no point decoding again" shortcut can never fire at level 0. That's the right instinct — at full resolution every extra chunk really does improve the image — but nothing else limits how often it happens either. The only pacing left is the 128KB streaming threshold in `streamRequest.ts:37`.

On the test data that's harmless: a 185KB frame arrives in about two chunks, so two decodes. A 10MB frame arrives in about eighty, so eighty full-resolution decodes for one frame, each allocating a fresh buffer and pushing a render. Before this change that frame decoded about three times.

The brake probably wants to be how much new data has arrived, not the decode level — something like "only decode again once the bytes have grown by half." That keeps the refinement on big frames without doing it eighty times.

Separately, the rest of the change here is a no-op worth noting: `decodeLevelFromComplete(percent, 0)` already returned 0 from every branch, so the new short-circuit doesn't change any result. It just makes the intent readable, which is an improvement on its own.

🤖 Reviewed with [Claude Code](https://claude.com/claude-code)

packages/core/examples/stackProgressive/index.ts
```diff
@@ -173,88 +184,77 @@ const jlsMixedOptions = {
...
+const htj2kFrames = { framesPath: '/htj2k/' };
+
 const htj2kProgressiveOptions = {
   retrieveOptions: {
+    default: htj2kFrames,
     single: {
+      ...htj2kFrames,
       streaming: true,
-      decodeLevel: 1,
+      decodeLevel: 0,
```
jbocce, line 199: The multi-stage configs look after themselves — `ProgressiveRetrieveImages` queues the next stage from a `.finally()`, so it runs whether the previous stage succeeded or threw. That's the real safety net, which makes dropping the `singleFastFailure` / `singleMiddleFailure` stages cost much less than it first appears.

This config is the exception: it's a single stage, so there's no `next` to fall through to. If a truncated decode ever throws, the frame just fails. And it's the config that used to be safe by construction, since `decodeLevel: 1` only ever needed the front of the codestream.

Not saying don't do it — but this is the one place where the PR's central assumption has no backstop. The exact version pin (`"2.4.10"`, not `^`) keeps it honest in this repo; a downstream consumer who dedupes openjph to an older version would break in a way that's hard to trace back. A peer-dependency floor, or a release-note line saying the codec bump isn't optional, would cover it.

For transparency: I took the truncated-decode verification on the description's word rather than re-running it against the wasm.

🤖 Reviewed with [Claude Code](https://claude.com/claude-code)

packages/dicomImageLoader/src/imageLoader/internal/rangeRequest.ts
```diff
@@ -11,6 +11,15 @@ import type { CornerstoneWadoRsLoaderOptions } from '../wadors/loadImage';
...
 type RangeRetrieveOptions = Types.RangeRetrieveOptions;
 
+/**
+ * Bytes fetched by a range stage that does not configure its own chunkSize.
+ *
+ * 32k is enough of an HTJ2K codestream to decode a recognisable full
+ * resolution image, and OpenJPH decodes the truncated remainder rather than
+ * throwing, so there is no reason to buy a larger initial buffer.
+ */
+const DEFAULT_CHUNK_SIZE = 32768;
```
jbocce: 32KB is clearly right for HTJ2K, and the comment explains why well. The thing is this constant applies to *any* range retrieve that doesn't set its own `chunkSize`, including transfer syntaxes where a third of the previous buffer may not be enough for the level being asked for.

Nothing in this repo is affected — the configs that matter all set `chunkSize` explicitly now. Downstream consumers are where it would land, and it would show up as a silent quality change rather than an error. Since this ships as a `feat`, a release-note line should cover it.

Also worth calling out that the docs and code disagreed before this (the type doc said 64,000 while the code used 65536). Good to see them reconciled.

🤖 Reviewed with [Claude Code](https://claude.com/claude-code)

jbocce (changes requested):
See the comments. Also check coderabbit's comments as well. Thanks.

packages/dicomImageLoader/src/imageLoader/wadors/loadImage.ts
wayfarer3130: Fixed in 6ffd196 — you were right that nothing was pacing it.

The brake is now how much new codestream arrived rather than the decode level, as you suggested: a repeat decode at level 0 needs the retrieved bytes to have grown by half (`REDECODE_GROWTH_FACTOR = 1.5`). Your 10MB/80-chunk case comes down to ~10 decodes instead of 80, and the 185KB frame still decodes twice. Sub-resolution is untouched — those still only redecode when the level itself improves, so that path behaves exactly as it did before this PR.

I pulled the decision out into `shouldDecodeAgain` and unit tested it (`shouldDecodeAgain.spec.ts`), including a replay that counts decodes for an 8MB frame over 128k reads, so the bound is pinned rather than reasoned about.

And yes — `decodeLevelFromComplete(percent, 0)` returning 0 from every branch means the short-circuit was a no-op. Kept it for the reason you gave.

packages/core/examples/stackProgressive/index.ts
wayfarer3130: Good catch that this is the one config with no backstop — single stage, so no `next` for the `.finally()` to queue.

I went with the release-note option rather than a peer-dependency floor. `codec-openjph` is a regular pinned dependency of `dicom-image-loader`, not something the consumer supplies, so declaring it as a peer would misdescribe the relationship; the exposure you're pointing at is a consumer *deduping* it to an older copy, which a peer range wouldn't prevent either. So 6ffd196 adds a migration-notes entry saying plainly that the codec floor is not optional, spelling out that it degrades to a slower load in a multi-stage config but fails the frame in a single-stage one, and a tip in the progressive-loading docs naming 2.4.10 as the minimum.

If you'd still rather have a hard floor enforced at install time, say so and I'll add it — happy to be overruled.

On the verification: fair to flag. I did run it — the published 2.4.10 wasm decodes CT1.j2c at full 512x512 from 10k and 32k truncated slices without throwing, and openjphjs' own `truncated.test.js` passes 9/9 against that build. What I have *not* done is load the examples in a browser against regenerated renditions.

packages/dicomImageLoader/src/imageLoader/internal/rangeRequest.ts
wayfarer3130: Agreed — the constant is transfer-syntax blind, and a silent quality change downstream is the bad failure mode.

6ffd196 adds a migration-notes entry that leads with exactly that: the default applies to **any** range retrieve that doesn't set its own `chunkSize`, not just HTJ2K ones, that a third less data may not be enough for the level being asked for on other syntaxes, and that it surfaces as a quality change rather than an error. Migration guidance is to set `chunkSize` explicitly on configurations tuned around the old 64kb.

Also fixed the docs while in there — `retrieve-Configuration.md` still said 64kb in two places, so the docs/code disagreement you noticed had a third instance I'd missed.

packages/dicomImageLoader/examples/htj2kStackBasic/index.ts
```diff
@@ -133,54 +133,42 @@ const htj2kProgressiveOptions = {
   retrieveOptions: {
```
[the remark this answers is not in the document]
wayfarer3130: Valid — verified and fixed in 6ffd196.

`htj2kStackBasic` uses the same study as `stackProgressive` (`...9590.100.1.2.1984...`), which this PR repointed to `/htj2k/` precisely because its primary frames aren't HTJ2K, and `htj2kVolumeBasic` uses the Juno CT whose primary frames are JPEG-LS lossless. Both were reading primary `frames/`, so neither exercised the truncated HTJ2K path.

All three now route through a shared `htj2kFrames` const, matching how `volumeProgressive` and `stackProgressive` do it. I also added it to `configHtj2k` in `htj2kVolumeBasic`, which had the same problem but wasn't in your list, and replaced both files' doc comments — they still referenced the retired `mkdicomweb --alternate` CLI and an `htj2kThumbnail` rendition that `createdicomweb alternates` doesn't produce.

packages/dicomImageLoader/src/imageLoader/wadors/loadImage.ts
wayfarer3130: Correcting my earlier reply — the growth-factor brake was the wrong mechanism, and 6a8ac17 replaces it.

Two things govern this, and a single ratio conflated them:

```
initialChunkSize  32k    byte range for the first decode
chunkSize         128k   each range after the first, and the streaming
                         accumulation threshold
msBetweenDecode   500    minimum gap between decodes of one partial image
```

The first range stays small because 32k of HTJ2K is enough for a usable full resolution decode — the point there is time to first image. Later ranges are larger because the image is already up and the point is refinement, where 32k steps would only mean more requests for the same result. Range boundaries follow: the end of range `n` is now `initialChunkSize + n * chunkSize`.

The part my growth factor got wrong is that chunk size alone doesn't bound decode cost. 128k arrives in a few milliseconds against a local server, so your large frame would still decode dozens of times regardless of any size-based rule — work that costs far more than the receive it's keeping up with, and that no display can show. A clock is the right brake for that, so `msBetweenDecode` is what limits it now. It's timed from the *end* of the previous decode, so a slow decode doesn't immediately qualify the chunk behind it. A completed image is always decoded, so only intermediate versions are ever delayed. Sub-resolution decoding is exempt and stays bound by the level improving, as before.

I also folded streamRequest's untyped `minChunkSize` into `chunkSize` (old name kept as an alias) so the streaming and range paths are configured by one knob rather than two.

`shouldDecodeAgain` and its tests are reworked accordingly — the replay now models chunk arrival rate and decode duration, and pins both ends: a fast 8MB transfer stays at a handful of decodes, a slow one still refines ~every 500ms rather than being starved.

One caveat worth flagging: `chunkSize` changes meaning. It used to size the first range. The example configs that set it to 32k are updated — left alone they'd have shrunk every subsequent range to 32k — and the migration notes call out the rename.

packages/dicomImageLoader/src/imageLoader/internal/rangeRequest.ts
wayfarer3130: Updated the chunk size to have an initial chunk size and a general one, as well as a ms between retrieves so make it work for larger images without decoding too often.    Note that other than htj2k, the current codebase doesn't allow byte range retrieves or streaming decode - it needs an extra setting to enable it for a transfer syntax since more TSUIDs will just throw.

wayfarer3130: On where the HTJ2K gate actually lives — it is not a setting, which is worth stating plainly.

It is a hard-coded module constant, `streamableTransferSyntaxes` in [`packages/dicomImageLoader/src/imageLoader/wadors/loadImage.ts`](https://github.com/cornerstonejs/cornerstone3D/blob/fix/progressive-render/packages/dicomImageLoader/src/imageLoader/wadors/loadImage.ts#L26-L34):

```ts
const streamableTransferSyntaxes = new Set<string>([
  '3.2.840.10008.1.2.4.96',   // HTJ2K, private/pre-ratification UID
  '1.2.840.10008.1.2.4.202',  // HTJ2K RPCL - reliably streamable
  '1.2.840.10008.1.2.4.203',  // HTJ2K lossy - attempted, may not be
]);
```

and the single place it is consulted:

```ts
if (!extractDone && !streamableTransferSyntaxes.has(transferSyntax)) {
  continue;  // partial buffer, non-streamable syntax -> wait for more
}
```

It is not exported, and there is no option or registration API, so enabling another syntax means editing that array. That is deliberate — a codec that cannot read a truncated codestream throws rather than degrading, so the list is really "codecs known to survive truncation", which today is openjph only.

**This changes your point, and mine.** You wrote that the default "applies to *any* range retrieve, including transfer syntaxes where a third of the previous buffer may not be enough for the level being asked for", and I agreed. Having checked the gate, that is not what happens: a non-HTJ2K partial buffer is never decoded at all, and `decodeLevel` is only honoured by the HTJ2K decoder (the JPEG2000 one has it commented out). Those syntaxes fetch in ranges but decode exactly once, when the frame is complete. So a smaller `initialChunkSize` costs them a round trip, not quality.

1827f3b corrects the migration note, which had repeated the quality-change framing, and documents the gate in retrieve-Configuration.md — the three UIDs, that `decodeLevel` is HTJ2K-only, and that the list is a constant rather than a setting. It was undocumented before, which is probably why we both reasoned about it the wrong way round.

packages/dicomImageLoader/src/imageLoader/internal/streamRequest.ts
```diff
@@ -33,8 +39,13 @@ export default function streamRequest(
     streamingData = {} as StreamingData,
   } = options;
 
-  // @ts-expect-error
-  const minChunkSize = retrieveOptions.minChunkSize || 128 * 1024;
+  // How much new data to accumulate before decoding the partial codestream
+  // again. `minChunkSize` is the former name for this and is still honoured.
+  const minChunkSize =
+    (retrieveOptions.chunkSize as number) ||
```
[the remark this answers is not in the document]
wayfarer3130: Confirmed and fixed in b4a8bf67. This one was mine and it was a real defect.

`chunkSize` is declared `number | ((metadata) => number)`. A function is truthy, so it passed the `||` chain and landed in `lastSize + minChunkSize` — NaN comparison, threshold never satisfied, so every received chunk decoded. Exactly backwards from what setting `chunkSize` is for.

`rangeRequest` already had the metadata-aware reader for this, so rather than duplicate it I extracted it as `getRetrieveValue` and pointed both paths at it — two readers of the same option is how they drifted apart in the first place. It also handles the deprecated `minChunkSize` alias, which isn't declared on the option types at all.

Added `getRetrieveValue.spec.ts` (5 tests), including one that pins the resolved value is a `number` and not the function.

packages/core/examples/stackProgressive/index.ts
wayfarer3130: Following up on this thread — CodeRabbit pushed back on how I'd described the fallback in the migration note, and it was right in a way that strengthens your original point.

I'd written that a multi-stage config "degrades to a slower load". That's only true when a **later stage selects the same image**. `ProgressiveRetrieveImages.createStageRequests` chains `next` per image ID:

```js
const existingRequest = imageRequests.get(imageId);
if (existingRequest) { existingRequest.next = request; }
else { interleaved.push(request); }
```

So checking each built-in:

- `sequentialRetrieveStages` — both stages select every image, chained.
- `interleavedRetrieveStages` — offsets 0 and 2 appear in only one `multipleFinal` stage, **but** the final `errorRetrieve` stage has no `decimate`/`offset`, so it covers every image. Chained.
- `singleRetrieveStages` — one stage, and its `errorRetrieve` is *commented out* in `singleRetrieve.ts`. No `next`.

That last one is the default when a config sets no `stages`, which is exactly `htj2kProgressiveOptions` — so your "this config is the exception" was precisely right, and the commented-out `errorRetrieve` is the reason. A hand-written multi-stage config with disjoint `decimate`/`offset` selections has the same gap for any image only one stage covers.

b4a8bf67 rewrites the note to say all of that, and to record that a lost frame surfaces as a permanent `errorCallback` on the listener rather than an uncaught error — so it is reported, not silent.

The codec floor guidance stands; it's just now accurate about how far the damage spreads.

packages/dicomImageLoader/src/imageLoader/wadors/loadImage.ts
wayfarer3130: One more on this thread, since it affects the mechanism you were reviewing rather than just the code around it.

The `chunkSize` half of the brake was silently broken for one input shape. `chunkSize` is declared `number | ((metadata) => number)`, and the streaming path read it without resolving the callback form. A function is truthy, so it passed the `||` chain and landed in `lastSize + minChunkSize` — NaN comparison, threshold never satisfied, every received chunk decoded. So for anyone using the metadata-callback form, the accumulation half of what I described to you was doing nothing at all; only `msBetweenDecode` was actually holding it back.

Fixed in b4a8bf67 by extracting `rangeRequest`'s existing metadata-aware reader as `getRetrieveValue` and pointing both paths at it, with tests pinning that the resolved value is a number. CodeRabbit caught it — thread is [here](https://github.com/cornerstonejs/cornerstone3D/pull/2890#discussion_r3929574525).

Net position on your original point: the pacing is now `initialChunkSize` 32k / `chunkSize` 128k / `msBetweenDecode` 500, both halves actually functioning, and `shouldDecodeAgain` is unit tested at both ends — a fast 8MB transfer stays at a handful of decodes, a slow one still refines roughly every 500ms.

wayfarer3130:
@jbocce ready for another look. Summary of what changed since your review, and where I ended up disagreeing with my own earlier replies.

**Your three comments**

1. **Repeat decodes at level 0 unbounded** ([thread](https://github.com/cornerstonejs/cornerstone3D/pull/2890#discussion_r3925274814)) — you were right, and my first fix was wrong. I initially added a growth-factor brake; that conflated two separate things. It's now three explicit options: `initialChunkSize` 32k for the first range, `chunkSize` 128k for each range after, `msBetweenDecode` 500 between decodes of one partial image. Your 10MB/80-chunk case comes down to a handful. Timed from the *end* of the previous decode so a slow decode doesn't immediately qualify the next chunk; sub-resolution is untouched and still bound by the level improving. `shouldDecodeAgain` is extracted and unit tested at both ends. A later CodeRabbit catch found the `chunkSize` half was broken for callback-valued input — fixed too.

2. **Single-stage config has no backstop** ([thread](https://github.com/cornerstonejs/cornerstone3D/pull/2890#discussion_r3925274826)) — went with the release-note option rather than a peer-dependency floor, since `codec-openjph` is a pinned direct dependency and the risk you named is a consumer *deduping* it, which a peer range wouldn't prevent either. Checking the code also proved your point sharper than I'd put it: `next` is chained per image ID, and `singleRetrieveStages` has its `errorRetrieve` stage **commented out**, so that config genuinely has no retry. The migration note now spells out which built-ins do and don't, and that a lost frame surfaces as a permanent `errorCallback`.

3. **32k default is transfer-syntax blind** ([thread](https://github.com/cornerstonejs/cornerstone3D/pull/2890#discussion_r3925274837)) — I agreed with you here and **we were both wrong**. Partial decoding is gated on `streamableTransferSyntaxes` in `loadImage.ts`, a hard-coded set of three HTJ2K UIDs. A non-HTJ2K partial buffer is never decoded, and `decodeLevel` is only honoured by the HTJ2K decoder. So those syntaxes decode once when complete — a smaller first range costs a round trip, not quality. The gate was undocumented, which is likely why we both reasoned about it backwards; it's now documented in `retrieve-Configuration.md` and the migration note is corrected.

**CodeRabbit's**: `framesPath` on the two htj2k examples (they were reading primary `frames/`, which is JPEG-LS for both studies, so they exercised no HTJ2K path at all) and the `chunkSize` callback bug. Both fixed, both threads resolved.

**Two things I have not done**, so the checkboxes don't oversell it:
- No browser pass over `stackProgressive` against regenerated renditions. Playwright green is the repo's own suite. The Tested Environment box says unit-tests-only.
- The truncated-decode verification is real — published 2.4.10 wasm decodes CT1.j2c at full 512x512 from 10k and 32k slices, and openjphjs' `truncated.test.js` passes 9/9 — but that's codec-level, not viewer-level.

414 tests pass in `dicomImageLoader`, `tsc` clean on core and dicomImageLoader, 17/17 checks green. I left your three threads open rather than resolving them myself.

wayfarer3130:
@jbocce three more items are fixed in eaccf56f1, on top of a merge of current `main`. The peer dependency floor is the one open question, and that decision is yours.

**1. A streaming retrieve now shows the first image at 32k.**

Your comment on the byte range default caused me to examine the streaming path again. That path did not read `initialChunkSize` at all — it waited for the full 128k `chunkSize` before the first decode. So the 32k time to first image reached the byte range path only.

`htj2kProgressiveOptions` is a streaming configuration, which means the one config you identified as having no backstop received no benefit from the change.

`initialChunkSize` moves from `RangeRetrieveOptions` to `BaseRetrieveOptions`, `streamRequest` reads it, and the threshold is now `nextDecodeAt`, a pure function with 6 unit tests.

**2. `msBetweenDecode` has no effect on the byte range path.**

I documented the option under Byte Range Options, and that position is wrong. `rangeRequest` sends no `done` flag, `done` defaults to `true` in `loadImage.ts`, and `shouldDecodeAgain` returns `true` immediately when `done` is true. `ProgressiveIterator.as` also yields a range result one time. A range stage decodes one time, when its range arrives, so there is no second decode for the clock to hold back.

The behaviour is correct. The documentation described a control that the path does not have. The option is now in the Streaming Options section, and both the concept document and the migration notes state that the option applies to a streaming retrieve only.

**3. The first stage of a frame fixes both sizes for that frame.**

`rangeRequest` reads `initialChunkSize` and `chunkSize` from `streamingData`, and only the first stage writes those two values. A `chunkSize` on a later stage is ignored.

The behaviour is correct, because the ranges of one frame must agree on the position of each boundary. But a reader gets it wrong easily: the new text says `chunkSize` is the size of each range after the first, so the later stage is where an author writes it, and there the option does nothing. A configuration with `initialChunkSize: 16384` on stage 1 and `chunkSize: 262144` on stage 2 fetches to `16384 + 5 * 128kb`, not to `16384 + 5 * 256kb`.

The concept document, the migration notes, the option comments and the example comment all state this now, with an example of the correct form.

**Also**, the three default values move into one `retrieveDefaults` module. `rangeRequest`, `streamRequest` and `loadImage` each held a copy, which is how the streaming path kept its own 128k threshold and missed `initialChunkSize`.

**Open: the peer dependency floor.** My position is in [the thread above](https://github.com/cornerstonejs/cornerstone3D/pull/2890#discussion_r3928206489): a peer range does not stop a consumer that dedupes `codec-openjph` to an older copy, which is the exposure you named. The migration notes and the progressive loading document both state the 2.4.10 floor. If you prefer a hard floor at install time, tell me and I will add it.

**Tests.** 420 jest tests pass, and 6 of the tests are new. `tsc --noEmit` is clean for `core` and for `dicom-image-loader`.

**One item that is not mine.** A clean `pnpm --filter @cornerstonejs/core run build:esm` fails on `packages/core/src/utilities/getTargetVolumeAndSpacingInNormalDir.ts:175`:

```
error TS2551: Property 'getLevel' does not exist on type 'Logger'. Did you mean 'setLevel'?
```

The line is `if ((sliceStepLog.getLevel?.() ?? 5) <= 1) {`, and `getLevel` does not exist in `@cornerstonejs/metadata` at all. The file is identical on `origin/main` and this PR does not touch it. The optional chain makes the line safe at run time, and `tsc` emits the declarations anyway, so the error blocks neither this PR nor a build that ignores the exit code. It needs a separate fix.

jbocce (approved):
approving on manual verification of the examples; a truncated-HTJ2K regression test would be worth adding separately"

Issue #2889, opened 2026-09-01
[Feature Request] Define voxel-relative annotation statistics and implement a shared oblique-capable voxel iterator


wayfarer3130:
## What feature or change would you like to see made?

Cornerstone3D computes ROI statistics (min/max/mean/stdDev/count) for `CircleROI`,
`EllipticalROI`, `RectangleROI`, `PlanarFreehandROI` and the StartEndThreshold tools, but it
has **never defined which voxels those statistics are computed over**. Each tool improvised
its own iteration, the improvisations have diverged, and none of them is correct in an
oblique/MPR view.

This issue defines the behaviour normatively, and proposes a single shared voxel iterator
that delivers it.

### Why the current approach cannot be repaired

The current oblique path (and [#2744](https://github.com/cornerstonejs/cornerstone3D/pull/2744),
adapted from [#2619](https://github.com/cornerstonejs/cornerstone3D/pull/2619)) rasterises
the ROI **in canvas space** and maps each canvas pixel back to a voxel. That assumption is
unsound for two independent reasons:

1. **Depth collapse.** One canvas pixel can correspond to several voxels stacked along the
   view plane normal. Enlarging the canvas never fixes this. A circle drawn at 2 mm
   thickness over two 1 mm frames in acquisition orientation has two voxels behind every
   displayed pixel, and canvas sampling can only ever see one of them.
2. **Conversion and rounding drift.** Assumptions in the world↔index conversion, plus
   rounding to integer indices, cause some voxels to be missed entirely.

Either failure means `max`, `mean` and every other statistic can silently omit voxels.
Canvas-space sampling also makes results depend on zoom and canvas resolution, which is not
acceptable for a measurement.

`#2744` fixes a real, user-visible bug (freehand ROI on oblique MPR — see
[OHIF#5889](https://github.com/OHIF/Viewers/issues/5889)) and should stay open; the plan
below adapts it onto the new iterator rather than discarding it.

---

### Terms

| Symbol | Meaning |
| --- | --- |
| `n` | View plane normal. **`annotation.metadata.viewPlaneNormal` is authoritative.** The `planeRestriction` in-plane vectors are for orientation compatibility only and must never be used to derive `n` — they are chosen heuristically and can be as little as ~18° apart, making their cross product badly conditioned. |
| `P₀` | Annotation plane anchor: `annotation.metadata.planeRestriction.point`. Populated at creation from the first annotation point. |
[...]

PR #2893, opened 2026-09-02
feat(core): define voxel-relative annotation statistics and add a shared oblique-capable voxel iterator

### LINKED ISSUE #5889: [Bug] Freehand ROI not renders in Volume Viewport for Oblique data

wayfarer3130:
This PR implements the core half of #2889. The PR defines which voxels an area annotation covers, and it adds a shared iterator that evaluates the definition exactly at any orientation.

This PR adds **core primitives only**. No annotation tool uses the iterator yet. Read [What this does not do](#what-this-does-not-do) before you review the PR for a change of behaviour. The PR touches no file in `packages/tools`, and it changes no reported measurement.

## Why

Cornerstone3D calculates ROI statistics for the annotation tools, but it never defined *which voxels* those statistics cover. Each tool wrote its own traversal, and no traversal is correct in an oblique view. The tools of today have three failure modes:

- **Lattice coverage under rotation.** A nearest-neighbour sample on a rotated grid cannot cover an integer lattice at any sample pitch. At 45° in-plane oblique, samples along `(0.707, 0.707)` in IJK round to `(0,0), (1,1), (2,2)…` and never visit `(1,0)` or `(0,1)`. The sub-pixel phase of the camera decides which voxels the tool skips, so a half-pixel pan changes the reported maximum.
- **Dependence on the display.** Where a tool derives the voxel set from the canvas geometry, that set is a function of the zoom, the pan, the canvas size and `devicePixelRatio`. The same annotation on the same data reports a different mean, because the display state differed when the tool calculated the statistics.
- **No extent through the plane.** A traversal of one plane returns a sheet one voxel thick. Draw a freehand annotation on an NM series with 1 mm slices, and fuse that series with a CT series at 0.5 mm in the same orientation. The correct CT maximum must then examine **two** CT voxels for each in-plane location. That extent is a necessary dimension, not a tuning parameter.

## The rules

Issue #2889 states both rules normatively.

**Rule M (voxel membership).** A voxel belongs to an area annotation when the voxel obeys two conditions. The voxel centre lies within `(T + T_v) / 2` of the annotation plane along the normal. The projection of that centre along the normal onto the plane falls inside the 2D shape. `T` is the thickness of the annotation, and `T_v` is the voxel thickness along the normal.

The viewport slab thickness `t` does not appear in Rule M. That absence is the purpose of the rule: the statistics cannot change because a user zoomed the viewport, resized the canvas, or increased the slab.

**Rule D (display).** A viewport shows a plane when the distance from the plane point to the focal point along the normal is within `(t + T) / 2`. The effects across modalities are intended. An annotation on one thick NM slice can correctly appear on two thin CT slices, and an annotation that spans two CT slices can correctly appear on one NM slice.

**The source of `T`.** A new annotation takes `T` once, at creation, from the slab thickness of the viewport that the user drew in. `Viewport.getReferenceThickness` supplies it, and `BaseVolumeViewport` overrides that method. After creation, `T` belongs to the annotation. When a reference records no thickness, `T` defaults to one voxel along the normal. A stack viewport uses that default, and so does every annotation that precedes this field, so this PR does not change an existing annotation.

This behaviour does not contradict the independence of Rule M from `t`. The code reads the slab once, when it creates the reference. The code never reads the slab when it calculates the statistics again.

## How the iterator stays exact and cheap

The depth half of Rule M is exactly linear in the integer voxel indices: `depth(p) = p · g + c₀`, with `g = Mᵀn`. The voxels that qualify along one axis therefore form an interval in closed form. The iterator emits exact integer runs instead of a test for each voxel. It puts the depth in the outer loop and the in-plane runs in the inner loop, so the runs of a shape intersect directly with the depth interval.

The cost is proportional to the voxels that the iterator emits, plus the rows that it touches. The cost is not proportional to the volume of a bounding box.

The same reduction makes the shapes exact. For a fixed outer index and row index, a voxel centre moves along a straight line in the column index, and so does its projection onto the annotation plane. Every shape test therefore becomes a one-dimensional intersection in the column index: a quadratic for the ellipse and the ellipsoid, intersected half-spaces for the rectangle and the box, and sorted crossings for the contour. The code tests no voxel on its own, and a non-convex contour gives the exact multiple runs that the iterator already supports.

## Commits

**`f9e97fd` — the iterator and the plane thickness.** `iterateVoxelsInSlab` and `collectVoxelsInSlab`, the index-space run arithmetic (`indexSpaceSlab`), the Rule M predicates (`slabMembership`), `getVoxelThicknessAlongNormal`, and `PlaneRestriction.thickness`. This commit also puts the depth half of Rule D into `isPlaneDepthViewable`.

**`a304366` — the clipping shapes.** The plane-anchored shapes that the annotation tools need, each one an exact run provider: the in-plane ellipse, the out-of-plane ellipsoid, the in-plane rectangle, the out-of-plane box, and the contour prism. The ellipse and the rectangle take identical options — a major-axis orientation vector and the extents — and an optional third extent changes the flat shape into a solid. The two are therefore interchangeable.

Each shape gives `containsPoint` as the definition and `getRuns` as the optimisation, which is the same relation that the brute-force reference has to the iterator. The tests require that the two select identical voxels, and that both agree with the reference.

**`2da0f33` — internal holes in the contour shapes.** `createContourShape` accepted one ring, so a user could not express an annulus. The obvious alternative failed without a message: a flat array of two rings adds an edge from the end of the first ring to the start of the second. The probe measured 222 voxels where the annulus holds 264. The code gave no error, and it measured a different shape.

`polyline` now accepts one ring or an array of rings. The even-odd rule over every edge of every ring gives the interior, and the parity accumulates across the rings instead of per ring, so a hole ring turns its interior back to outside. That one rule gives holes, nesting to any depth (a ring inside a hole is solid again), and disjoint rings as separate regions. The rule does not depend on the winding direction.

**The seven review-fix commits.** See [Review fixes](#review-fixes) below.

## Boundary handling

One rule holds for all shapes: **a voxel centre on a shape outline is inside the shape.** Three independent cases made that rule necessary, and a reviewer must know all three:

- A circle of radius 5 on an integer grid puts voxel centres exactly on its outline, at `(5, 0)` and at every Pythagorean point such as `(3, 4)`. `containsPoint` adds the squares and can give a little more than 1, while `getRuns` solves for the roots and gives exactly 5. Both therefore compare against a boundary that a relative epsilon widens.
- The even-odd rule gives the interior of a contour, but `containsPoint` casts a ray along one plane axis while `getRuns` intersects a line along the direction that the column axis projects to. The even-odd rule is independent of direction away from the boundary, but the two tie rules degenerate at different geometry. A rectangular contour drawn on voxel boundaries kept a row at one end and lost it at the other.
- The crossing test cannot see an outline edge that runs *along* a run line, because both end points lie on the same side of a line that holds them. Such an edge now supplies its extent directly, and so does any vertex that touches the line. The code merges every contribution, so it emits no voxel twice.

## Two incidental defects fixed

This code depends on both fixes, so they are here and not in a separate PR.

- `Viewport.getViewReference` passed `target.planeRestriction` to `updatePlaneRestriction`, which expects the whole reference and does `reference.planeRestriction ||= …`. A `PlaneRestriction` satisfies the structure of the all-optional `ViewReference`, so the code passed the type check, built a nested `planeRestriction.planeRestriction`, and changed *that* object. The code discarded the in-plane vectors that it derives from the points, and it gave no message. Legacy viewports had this defect.
- `updatePlaneRestriction` tested collinearity on one side only. An anti-parallel candidate has a dot product of about `-length`, which passes an unsigned comparison although the candidate is as collinear as a parallel one. A symmetric polyline that returns along its own path therefore gave a second "in-plane vector" that fixed no orientation.

## Verification

A brute-force reference implementation of Rule M visits every voxel and applies the predicates literally. The fast iterator must return an identical set. The tests cover oblique angles from 1° to 89°, anisotropic spacing, rotated direction matrices, thick slabs, clipped results and empty results, and both choices of the column axis. Every test also asserts that the result holds no duplicate.

I ran these checks locally on this branch:

| Check | Result |
| --- | --- |
| `jest packages/core/test/voxelSlab* planeRestrictionThickness` | 5 suites, **122 passed** |
| `jest packages/core` (full suite, for the shared-code changes) | 39 suites, **684 passed**, 2 skipped |
| `tsc --noEmit -p packages/core/tsconfig.json` | clean |
| `oxlint` on changed files | 0 errors, 2 warnings. Both warnings are in `BaseVolumeViewport.ts`, both precede this PR, and this PR does not touch those lines. |

I merged `main` into this branch, and the merge gave no conflict.

## Review fixes

The review found five items. Each item has its own commit:

- **The caller bounds now clamp to the volume.** `iterateVoxelsInSlab` defaulted `bounds` to the volume extent, but it never intersected a supplied `bounds` with that extent. A caller who supplied a box that reaches outside the volume received an index outside the volume. Each axis now intersects the volume extent, so the bounds only narrow. Two tests cover the clamp and the empty result.
- **A planar contour shape now reports no required thickness.** `createContourShape` returned `voxelThickness` when the caller gave no `depth`. Under the documented idiom `shape.getRequiredThickness() || annotationThickness`, that value replaced the recorded thickness of the annotation with one voxel. The shape now returns 0, as the `VoxelSlabShape` contract requires of a planar shape. This change needed a second one, because #2744 passes the value straight through. See [What #2744 found in this PR](#what-2744-found-in-this-pr).
- **The doc comment for the default column axis is correct.** `buildIndexSpaceSlab` uses the lower-numbered of the two non-outer axes, and the comment said "the higher-numbered one".
- **An empty voxel selection has a test.** The test named "a shape that selects nothing" centred the disc on a voxel centre and asserted a length of 1. That test keeps its assertion under a correct name, and a new test beside it covers the empty result.
- **The rectangle boundary tolerance has a floor.** `createRectangleShape` scaled its slack by the half extent alone. `createContourShape` already floors its own slack at one unit, and the rectangle now uses the same rule. This change removes an inconsistency between two shapes in one module. The 1e9 origin of the review comment is not a real patient coordinate, so the stated failure is not reachable.

A sixth commit shortens the comments that the five fixes added. A seventh commit, `3d1ff8c`, repairs the defect that the second fix introduced in #2744.
## The second step: the oblique planar freehand demo

This PR ships the primitives, and **PR #2744 is the real demonstration.** That PR already targets this branch, so a reviewer can check out #2744 and see the oblique defect disappear in a running viewer.

PR #2744 fixes the freehand ROI tool in a volume viewport that shows oblique data. @arul-trenser wrote the original fix, which used a rasterizer in canvas space. The sampling now uses `iterateVoxelsInSlab` and `createContourShape` instead, so #2744 is the first consumer of this PR. The PR fixes OHIF/Viewers#5889.

### The demo

Follow the steps of #2744:

1. Open the study `https://viewer-dev.ohif.org/viewer?StudyInstanceUIDs=1.3.6.1.4.1.14519.5.2.1.9823.1001.134394060407147891170882809392`
2. Change to the 2D MPR layout.
3. Draw a freehand ROI.

PR #2744 holds a video of the behaviour before the fix and a video after the fix.

Two further checks show the properties that Rule M adds, and neither works before the conversion:

- **Independence from the display.** Write down the mean and the maximum. Pan the viewport by about half a pixel, zoom by a small amount, and force a new calculation. The values must stay identical. Before the conversion they change, because the sub-pixel phase of the camera selects a different voxel set.
- **The extent through the plane.** Load an NM series with 1 mm slices and a CT series at 0.5 mm in the same orientation. Draw the annotation on the NM series. The CT statistics must cover two CT voxels for each in-plane location, not one.

### The code a tool writes

`sampleVoxelsInContour` in #2744 is the whole consumer side:

```ts
const shape = createContourShape({
  volume,       // { dimensions, direction, spacing, origin } from imageData
  planePoint,   // points[0] - the contour's own points define the plane depth
  normal,       // annotation.metadata.viewPlaneNormal, normalized
  polyline: points,
  depth: annotation.metadata.planeRestriction?.thickness,   // T
});

for (const { ijk, center } of iterateVoxelsInSlab({
  volume,
  planePoint,
  normal,
  annotationThickness: shape.getRequiredThickness(),
  bounds: this.getContourIndexBounds(points, imageData),
  getShapeRuns: shape.getRuns,
})) {
  // accumulate
}
```

Nothing there reads the viewport. **A tool with a different shape swaps `createContourShape` for `createEllipseShape` or `createRectangleShape` and changes nothing else.** That is the adoption path for `CircleROITool`, `EllipticalROITool`, `RectangleROITool` and the StartEndThreshold tools.

Two details matter for a consumer:

- `ijk` and `center` are **reused between iterations**. Copy either one before you retain it.
- When `planeRestriction.thickness` is absent, `T` defaults to one voxel along the normal. Stack annotations and every annotation drawn before that field existed therefore behave as before.

Swap `getShapeRuns: shape.getRuns` for `isInShape: shape.containsPoint` and the voxel set must stay identical, only slower. That swap is the cheapest way to debug a shape.

### What #2744 found in this PR

The review of this PR asked `createContourShape` to return 0 from `getRequiredThickness` when the caller gives no depth. #2744 passes `getRequiredThickness()` straight to the iterator, and it does not use the `|| annotationThickness` idiom. `resolveAnnotationThickness` accepted that 0 as a literal `T`, which gave a half width of `T_v / 2` in place of `T_v`.

A plane on the voxel centres still selected one layer, so every test passed. A plane between voxel centres selected nothing. A probe over an 8×8×8 volume returned 128 voxels for an unrecorded `T`, and 0 voxels for `T = 0`. An oblique annotation rarely lands on the voxel centres, so the pair of changes emptied the exact measurement that #2744 demonstrates.

`resolveAnnotationThickness` now takes 0 or less as unrecorded, and it applies the one-voxel default. Rule D is unaffected, because `isPlaneDepthViewable` tests `Number.isFinite` on its own and never calls the resolver. Commit `3d1ff8c` carries the fix and two tests.

## What this does not do

I state the limits so the review scope is clear:

- **No annotation tool is converted.** `CircleROITool`, `EllipticalROITool`, `RectangleROITool`, `PlanarFreehandROITool` and the StartEndThreshold tools all keep their existing traversals. This PR therefore does **not yet fix** the oblique statistics defect. The PR is the foundation that the follow-ups use.
- **No reported measurement changes.** The PR touches nothing in `packages/tools`. When the tools do convert, the statistics *will* change for existing annotations. That change is the intent, and the migration needs its own discussion.
- **The documentation page now exists.** `voxelSlab/index.ts` cited `docs/docs/concepts/annotations/voxel-statistics.md`, which was neither the repository convention nor a real file. The page is now `concepts/cornerstone-tools/annotation/voxel-statistics.md`, beside `measurementTargets` in the Annotations section, and the citation names the real path.
- **The relation to #2744.** The original #2744 fixes the freehand oblique case with a rasterizer in canvas space, which has all three failure modes above. This PR does not supersede #2744. Instead, **#2744 now targets this branch and uses these utilities**, so #2744 is the demonstration. Its `viewUp` normalization in `getSubPixelSpacingAndXYDirections` and its `computeEffectiveVoxelSpacing` are correct, and both stay. See [The second step](#the-second-step-the-oblique-planar-freehand-demo).

## Suggested review order

1. `PlaneRestriction.thickness` in `types/IViewport.ts` — the doc comment is the specification for `T`.
2. `voxelSlab/slabMembership.ts` — Rule M as predicates.
3. `voxelSlab/indexSpaceSlab.ts` — the linear-depth argument that makes the runs exact.
4. `voxelSlab/iterateVoxelsInSlab.ts` — the traversal.
5. `voxelSlab/shapes/` — each `containsPoint` beside its `getRuns`.
6. `test/utils/voxelSlabReference.js` — the brute-force oracle that every test compares against.

### PR

### Code

  etc.)

### Public Documentation Updates

  additions or removals.

The PR adds `utilities.voxelSlab` as public API, and the new page
`concepts/cornerstone-tools/annotation/voxel-statistics.md` documents it. The page holds Rule M, Rule D, the source of `T`, the way a tool uses the iterator, the reason the runs are exact, the boundary and epsilon rules, and the L1-versus-L2 distinction for `T_v`. The doc comments that carried that reference text now keep one fact each and point at the page.

### Tested Environment

- [x] "OS: Windows 11"
- [x] "Node version: 24.20.0"
- [ ] "Browser: not tested"

The browser box stays open on purpose. The jest suite gives the verification above, and jest runs under Node. The karma browser suite (`pnpm run test:ci`) did not run on this branch.

Two changes do reach a display path, so a browser check has value:

- `isPlaneDepthViewable` now decides plane visibility from the slab depth. `packages/core/test/planeRestrictionThickness.jest.js` covers the new behaviour.
- `Viewport.getViewReference` no longer builds a nested `planeRestriction.planeRestriction`, so a legacy viewport keeps its in-plane vectors.

No annotation tool uses the voxel iterator yet, so no measurement changes.

Ref: #2889

## Summary by CodeRabbit

* **New Features**
  * Added support for thickness-aware annotation visibility across viewport slabs.
  * Added voxel-slab tools for selecting and iterating volume voxels within annotation depth.
  * Added support for rectangle, ellipse, circle, and multi-ring contour shapes, including holes and oblique orientations.
  * Improved plane references by preserving orientation details and recording slab thickness.

* **Bug Fixes**
  * Improved handling of parallel and anti-parallel annotation vectors to prevent incorrect orientations.
  * Improved boundary and floating-point stability for slab visibility and voxel selection.


packages/docs/docs/concepts/cornerstone-tools/annotation/voxel-statistics.md
wayfarer3130, line 1: @sedghi - please review this file as it provides the rules and reasoning for the rest of the PR, along with enough detail to actually determine whether or not the code is correct - we previously never had a good way to tell if some of the code was actually correct and now we do.

packages/core/src/RenderingEngine/BaseVolumeViewport.ts
```diff
@@ -2417,6 +2417,26 @@ abstract class BaseVolumeViewport extends Viewport {
...
+    const slabThickness = this.getSlabThickness();
+
+    if (
+      !Number.isFinite(slabThickness) ||
+      slabThickness <= RENDERING_DEFAULTS.MINIMUM_SLAB_THICKNESS
+    ) {
+      return undefined;
+    }
+
+    return slabThickness * 2;
```
sedghi, line 2439: why not getReferenceSlabThickness? i wonder if we use thickness and slabthickness interchangebly

packages/core/src/utilities/voxelSlab/shapes/createContourShape.ts
```diff
@@ -0,0 +1,315 @@
...
+   *
+   * Unlike the ellipsoid and box, the shape does *not* apply this itself: the
+   * slab enforces it, via `getRequiredThickness()`. Omit to let the slab
+   * decide, in which case `getRequiredThickness()` returns 0.
+   */
+  depth?: number;
+}
+
+/** A point in the plane's 2D basis, relative to the plane anchor. */
+type PlanePoint = [number, number];
```
sedghi: we already have point2D or point2 if i remember

packages/core/src/utilities/voxelSlab/shapes/createContourShape.ts
```diff
@@ -0,0 +1,315 @@
...
+import { projectPointOntoPlane } from '../slabMembership';
+import type { PlaneBasis, VoxelSlabShape } from './shapeGeometry';
+import {
+  SHAPE_BOUNDARY_EPSILON,
+  createColumnLineResolver,
+  createPlaneBasis,
+  toIntegerRun,
+} from './shapeGeometry';
+
+export interface ContourShapeOptions {
```
sedghi: we already have contour segmentation i would avoid calling it contour too, maybe PolylineShapeOptions?

packages/core/src/utilities/voxelSlab/shapes/createPolylineShape.ts
```diff
@@ -0,0 +1,315 @@
...
+import {
+  SHAPE_BOUNDARY_EPSILON,
+  createColumnLineResolver,
+  createPlaneBasis,
+  toIntegerRun,
+} from './shapeGeometry';
+
+export interface ContourShapeOptions {
+  /** Geometry of the volume being measured. */
+  volume: VolumeGeometry;
```
sedghi, line 19: I'm not sure what is geometry of the volume being measured. is it volume? as in Ivolume? since we already have Ivolume and Volume

packages/core/src/utilities/voxelSlab/shapes/createContourShape.ts
```diff
@@ -0,0 +1,315 @@
...
+  /** Geometry of the volume being measured. */
+  volume: VolumeGeometry;
+  /** The annotation plane anchor. Defines the plane's depth. */
+  planePoint: Point3;
+  /** The annotation view plane normal. Unit length. */
+  normal: Point3;
```
sedghi: viewPlaneNormal? since it is viewPlaneNormal as comment says

packages/core/src/utilities/voxelSlab/shapes/createPolylineShape.ts
```diff
@@ -0,0 +1,315 @@
...
+  /** The annotation plane anchor. Defines the plane's depth. */
+  planePoint: Point3;
```
sedghi: is this a must? since i see required. isn't all the points of a polyline in the plane? can't we just choose polyline[0]?

packages/core/src/utilities/voxelSlab/shapes/createPolylineShape.ts
```diff
@@ -0,0 +1,315 @@
...
+   */
+  polyline: Point3[] | Point3[][];
+  /**
+   * The contour's extent along the normal, in mm.
+   *
+   * Unlike the ellipsoid and box, the shape does *not* apply this itself: the
+   * slab enforces it, via `getRequiredThickness()`. Omit to let the slab
+   * decide, in which case `getRequiredThickness()` returns 0.
+   */
+  depth?: number;
```
sedghi: is this required? since you have already thickenss

packages/core/src/utilities/voxelSlab/shapes/createPolylineShape.ts
sedghi: How about simple single Image as IImage

packages/core/src/utilities/voxelSlab/shapes/createEllipseShape.ts
```diff
@@ -0,0 +1,202 @@
...
+
+/**
+ * The squared-radius threshold, expanded so a voxel centre lying exactly on the
+ * outline is inside it. See SHAPE_BOUNDARY_EPSILON.
+ */
+const UNIT_BOUNDARY = 1 + SHAPE_BOUNDARY_EPSILON;
+
+export interface EllipseShapeOptions {
+  /** Geometry of the volume being measured. */
+  volume: VolumeGeometry;
```
sedghi, line 27: same comment how about image

packages/core/src/utilities/voxelSlab/shapes/createEllipseShape.ts
```diff
@@ -0,0 +1,202 @@
...
+  /** The annotation plane anchor. Defines the plane's depth. */
+  planePoint: Point3;
+  /** The annotation view plane normal. Unit length. */
```
sedghi, line 30: same comments

packages/core/src/utilities/voxelSlab/shapes/createEllipseShape.ts
```diff
@@ -0,0 +1,202 @@
...
+  normal: Point3;
+  /**
+   * The ellipse centre in world coordinates. For the planar form it is
+   * projected onto the annotation plane, so a centre carrying a little depth
+   * error is harmless.
+   */
+  center: Point3;
```
sedghi: centerWorld? since we use that elsewhere

packages/core/src/utilities/voxelSlab/index.ts
```diff
@@ -0,0 +1,46 @@
...
+ * Rule M (voxel membership) and Rule D (display) are normative. See
+ * `docs/docs/concepts/cornerstone-tools/annotation/voxel-statistics.md` and
+ * https://github.com/cornerstonejs/cornerstone3D/issues/2889.
+ */
+export { default as getVoxelThicknessAlongNormal } from './getVoxelThicknessAlongNormal';
+
+export {
+  SLAB_RELATIVE_EPSILON,
+  getSlabEpsilon,
+  resolveAnnotationThickness,
```
sedghi: i'm not sure what is "annotation" thickenss, it sounds like a thickness really, and not annotation specific

packages/core/src/utilities/voxelSlab/iterateVoxelsInSlab.ts
```diff
@@ -0,0 +1,225 @@
...
+
+/**
+ * One voxel visited by {@link iterateVoxelsInSlab}.
+ *
+ * The arrays are **reused between iterations** so that iterating a large ROI
+ * does not allocate three arrays per voxel. Copy anything you intend to retain.
+ * `iterateVoxelsInSlab` is a generator, so a consumer that reads the fields and
+ * moves on - which is what a statistics accumulator does - never notices.
+ */
+export interface VoxelSlabVisit {
```
sedghi: I think this one and all others can follow the naming of the poinInShapeCallback existing namings

voxelSlab -> voxelsInShape
iterateVoxelsInSlab -> iterateVoxelsInShape
collectVoxelsInSlab -> collectVoxelsInShape
VoxelSlabIterationOptions -> VoxelsInShapeOptions
VoxelSlabVisit -> VoxelInShape

sedghi (approved):
see suggestions, logic seems fine

packages/core/src/utilities/voxelSlab/shapes/createPolylineShape.ts
wayfarer3130: `IImage` cannot carry this data. The iterator walks a 3D index space, so it needs `direction`, `spacing`, `origin` and `dimensions`. An `IImage` is one 2D frame and has no direction matrix and no dimension triple.

`VolumeGeometry` is `IImageVolume | { direction, spacing, origin, dimensions }`. The structural alternative lets a test, or a caller that holds no cached volume, use the code without the construction of a full volume.

I agree that the comment did not say any of this. The comment now reads:

```
/**
 * The volume being measured. Pass an `IImageVolume`, or any object that has
 * `direction`, `spacing`, `origin` and `dimensions`. The structural form
 * lets a caller that holds no cached volume, such as a test, use this code.
 */
volume: VolumeGeometry;
```

packages/core/src/utilities/voxelSlab/iterateVoxelsInSlab.ts
wayfarer3130: I agree about the iterator, and I want to keep the module name. My decision is a split:

Renamed, to follow `pointInShapeCallback`:

- `iterateVoxelsInSlab` becomes `iterateVoxelsInShape`
- `collectVoxelsInSlab` becomes `collectVoxelsInShape`
- `VoxelSlabIterationOptions` becomes `VoxelsInShapeOptions`
- `VoxelSlabVisit` becomes `VoxelInShape`

Kept:

- the `voxelSlab` directory
- the `IndexSpaceSlab` type

The reason for the split: the directory holds three things, and only one of the three is a walk over a shape. The directory also holds Rule D (`isPlaneDepthViewable`), the index space slab, and the membership mathematics. The word "slab" is the term that the specification uses for Rule M, and `voxelsInShape` names one of the three jobs and cuts the link to the specification.

Tell me if you prefer all five renames, and I will make them. If you want the directory renamed to a name that covers all three jobs, `voxelStatistics` describes the directory better than either candidate.

I applied your other naming comments in 2780f97.

packages/core/src/utilities/voxelSlab/shapes/createPolylineShape.ts
wayfarer3130: Agreed, and I removed the `depth` option in 5547ff4.

The option never affected the in-plane test. The option only fed `getRequiredThickness()`, which the caller then combined with the recorded thickness, so the option repeated a job that `referencePlaneThickness` already does. `getRequiredThickness()` now always returns 0 for a polyline, and a caller that wants a prism passes the prism depth to the iterator as the thickness.

The ellipse and the rectangle keep their depth, because each of those applies its own depth to the in-plane test.

I also made `planePoint` optional on the same shape, per your other comment: the default is the first point of the first ring.

packages/core/src/utilities/voxelSlab/shapes/createPolylineShape.ts
wayfarer3130: You are right, and I made `planePoint` optional for `createPolylineShape` in 5547ff4. The default is the first point of the first ring.

Two notes are now in the comment:

- Pass the annotation`s own anchor when you have one. A drawn vertex carries rounding error that the anchor does not.
- Give the shape and the iterator the SAME anchor. If the shape picks its own default and the caller passes a different anchor to the iterator, the two describe different slabs.

`planePoint` stays required on `createEllipseShape` and `createRectangleShape`, because neither shape has a point list to fall back on.

packages/core/test/voxelSlabShapes.jest.js
```diff
@@ -0,0 +1,971 @@
...
+
+      const defaulted = collectVoxelsInSlab({
+        volume: cube,
+        planePoint: anchor,
+        normal: AXIAL,
+        getShapeRuns: shape.getRuns,
+      });
+
+      expect(passedThrough.length).toBeGreaterThan(0);
+      expect(passedThrough).toEqual(defaulted);
```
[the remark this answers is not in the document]
wayfarer3130: Applied in 5547ff4, but not as suggested. The suggested assertion is wrong for this test.

The test loops over two anchors. The second anchor, `[8, 8, 8.5]`, lies midway between two voxel centres, and Rule M selects BOTH layers there: the half width is `(T + T_v) / 2 = 1`, and layers 8 and 9 sit at a depth of -0.5 and +0.5. An assertion of exactly one distinct depth index therefore fails on the second anchor.

The finding itself is correct: equality between the two paths does hold if a regression widens the slab in both paths. The test now names the layers per anchor:

```js
[
  { anchor: [8, 8, 8], layers: [8] },
  { anchor: [8, 8, 8.5], layers: [8, 9] },
].forEach(({ anchor, layers }) => {
  ...
  expect(
    [...new Set(passedThrough.map(([, , k]) => k))].sort((a, b) => a - b)
  ).toEqual(layers);
});
```

That assertion catches the regression, and it also pins the midway-between-centres rule that the specification states.

packages/core/src/utilities/voxelSlab/iterateVoxelsInSlab.ts
wayfarer3130: Applied in a1e844a, with the split as described.

Renamed:

- `iterateVoxelsInSlab` becomes `iterateVoxelsInShape`
- `collectVoxelsInSlab` becomes `collectVoxelsInShape`
- `VoxelSlabIterationOptions` becomes `VoxelsInShapeOptions`
- `VoxelSlabVisit` becomes `VoxelInShape`

Kept: the `voxelSlab` directory and the `IndexSpaceSlab` type.

Tell me if you want the directory renamed too, and I will do it.

PR #2898, opened 2026-09-04
feat(dicomImageLoader): decode JPEG XL, Encapsulated Uncompressed and Deflated Image Frame Compression

wayfarer3130:
### Context

Three transfer syntaxes were asked for and all three are implemented here:
JPEG XL, Encapsulated Uncompressed, and Deflated Image Frame Compression.

**This branch lands after #2890 and has it merged in**, so the diff against
`main` shows that PR's progressive-loading changes too until it merges. Review
against #2890's head, or wait for it to land.

The details below come from the standard rather than from existing practice,
because two of the three syntaxes are recent and the repository's existing JPEG
XL constant turned out to be wrong.

### Changes & Results

**Encapsulated Uncompressed Explicit VR Little Endian — `1.2.840.10008.1.2.1.98`**

Nothing is compressed. The syntax exists so uncompressed pixel data can use the
encapsulated format, one frame per fragment, which makes a single frame
addressable without reading the whole Pixel Data element (PS3.5 A.4.11).
Decoding is therefore trimming the fragment's padding and reading the rest as
Explicit VR Little Endian.

**Deflated Image Frame Compression — `1.2.840.10008.1.2.8.1`**

Each frame is separately compressed with **raw** DEFLATE per RFC 1951 — no zlib
header, no Adler-32 trailer — and encapsulated as one fragment per frame
(PS3.5 A.4.13). Raw is load bearing, so this is `pako.inflateRaw`, and a test
pins that a zlib-wrapped stream is rejected rather than silently accepted.

Worth distinguishing from **Deflated Explicit VR Little Endian**
(`1.2.840.10008.1.2.1.99`), which already worked: that deflates the *whole data
set*, so dicomParser inflates it at parse time and a frame reaches the decoder
already uncompressed. The new syntax deflates *per frame*, so inflating happens
in the decoder. Both are now commented to say which is which.

**Padding.** Both syntaxes pad: encapsulated fragments to an even length, and
Deflated Image Frame Compression appends a NULL when the deflated stream itself
is odd. Both decoders trim to the frame's native pixel length via a shared
`nativeFrameBytes` helper, which also handles bit-packed (`BitsAllocated` 1)
frames. A frame *shorter* than its pixel data throws rather than rendering
partially.

**Latent bug found on the way.** `framesAreFragmented` compared
`NumberOfFrames` against the fragment count, but a single-frame image carries no
`NumberOfFrames` element — so `undefined !== 1` read as "fragmented" and fell
back to `createJPEGBasicOffsetTable`, a scan for JPEG SOI markers. That scan
finds nothing in a syntax that is not JPEG, which would have broken both new
syntaxes (and JPEG XL later) for the most common case of all: a single-frame
image. It now defaults to 1. Existing JPEG behaviour is unchanged — a genuinely
multi-fragment frame still takes the marker-scan path.

**JPEG XL UIDs — corrected, not implemented**

`image/jxl` mapped to `1.2.840.10008.1.2.4.140`, which is not a JPEG XL UID at
all. Supplement 232 assigns:

| UID                       | Name                       |
| ------------------------- | -------------------------- |
| `1.2.840.10008.1.2.4.110` | JPEG XL Lossless           |
| `1.2.840.10008.1.2.4.111` | JPEG XL JPEG Recompression |
| `1.2.840.10008.1.2.4.112` | JPEG XL                    |

`image/jxl` now resolves to `.110`, which PS3.18 Table 8.7.3-5 gives as the
default when a response carries no `transfer-syntax` parameter.
`application/x-deflate` was added from the same table. All three UIDs are in
`constants/transferSyntaxes`.

### Testing

`packages/dicomImageLoader/src/__tests__/encapsulatedAndDeflatedFrames.spec.ts`
adds 13 tests: native frame sizing including bit-packed and multi-sample,
padding trimmed / truncation thrown, both decoders over signed and unsigned
data, a deflate frame that is a view into a larger buffer (which is how frames
actually arrive), and the raw-vs-zlib distinction.

Full suite: 428 tests pass in `dicomImageLoader`, `tsc --noEmit` clean.

I have **not** exercised any of the three syntaxes against real DICOM instances
in a browser - I had no conformant sample files for them. The unit tests
construct frames directly, and the JPEG XL codec is not exercised at all. Sample
data through the wadouri and wadors paths is the check this most needs before
merging; for JPEG XL it would be the only end-to-end evidence there is.

### JPEG XL

All three syntaxes decode through `@cornerstonejs/codec-libjxl` 1.1.0, published
2026-09-04. One decoder covers the set.

Two details from the codec that are easy to get wrong:

- **Signedness comes from `PixelRepresentation`, not the codestream.** JPEG XL
  has no signed sample type, so the decoder always reports `isSigned: false`.
  This is the arrangement JPEG-LS already uses, so the shared `getPixelData`
  already had the `signedOverride` argument for it. A test pins a 16 bit frame
  reading as `Int16Array` on the override alone - trusting the frame info there
  would render signed CT as large positive values.
- **JPEG XL is deliberately *not* in `streamableTransferSyntaxes`.** The codec
  calls `JxlDecoderCloseInput()` up front and throws on
  `JXL_DEC_NEED_MORE_INPUT`, so a partial buffer must wait for the whole frame
  rather than be handed to a decoder that will reject it. The format does
  support progressive decoding; this build does not use libjxl's
  `SetProgressiveDetail`/`FlushImage` path. Both the gate documentation and the
  migration note record this.

**The decoder has no unit test**, and I would rather say why than leave it
looking covered. No WASM decoder in this package has one: jest's
`^@cornerstonejs/(.*)$` mapping rewrites codec packages to `packages/<name>/src`
and takes precedence over a virtual mock, so faking the codec means changing the
jest config. That did not seem proportionate for one decoder. The substance that
*is* testable without the codec - the signedness rule - is covered in
`getPixelData.spec.ts`.

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] "OS: Windows 11"
- [x] "Node version: 22"
- [x] "Browser: not exercised in a browser — unit tests only, see Testing"

## Summary by CodeRabbit

* **New Features**
  * Added JPEG XL decoding support.
  * Added decoding for encapsulated uncompressed and per-frame deflated DICOM images.
  * Added WADO-RS mappings for JPEG XL and deflated image content types.
  * Added configurable progressive-loading chunk sizes and decode throttling.

* **Bug Fixes**
  * Improved handling of single-frame images, fragment padding, and truncated frames.
  * Corrected JPEG XL transfer-syntax mapping and partial-decoding quality reporting.

* **Documentation**
  * Added guidance for progressive loading, JPEG XL, and partial decoding.


wayfarer3130:
@jbocce - I re-enabled the cornerstone tests for the various transfer syntaxes - you can use this PR to test the codec changes if you want by linking in those.  I'm just going to add colour samples for testing.

wayfarer3130:
Note this requires https://github.com/cornerstonejs/codecs/pull/94 to release first because one of the tests is failing without that PR.

wayfarer3130:
## The pending `1.2.840.10008.1.2.4.70` case: root-caused and fixed upstream

The note on that pending case guessed it was "a stream some decoders tolerate and this one does not". It is narrower and more definite than that — a real decoder bug, now fixed.

`jpeg-lossless-decoder-js` 2.1.2 drops the final sample of any frame whose last Huffman code ends exactly on a byte boundary. Its three end-of-scan guards test `index < markerIndex` (9); once the `0xFF` introducing EOI has been shifted into `temp`, only `index - 8` bits are real data, so consuming the last of them leaves `index === 8` — a legal decode that those guards reject, abandoning the scan one sample early. T.81 B.1.1.2 pads only an *incomplete* final byte, so a scan that tiles its last byte exactly is perfectly legal, and DCMTK writes one whenever a frame ends in a run of a single value. That is why the DCMTK-encoded fixture here failed while dcm4che's SV1 frame of the same shape and depth in viewer-testdata did not: not the encoder being sloppy, just whether the last byte happens to come out full.

Fixed in [cornerstonejs/JPEGLosslessDecoderJS@03bb80c](https://github.com/cornerstonejs/JPEGLosslessDecoderJS/commit/03bb80c073e34369893e468b615dac9dcb0dcee9), which replaces all three guard sites with a named `readPastEntropyData` putting the boundary at `index < 8` and drops the `isLastPixel()` special case that had been papering over the same off-by-one at one of them.

### Verified against this branch's suite

I enabled the pending case and ran `decoders_test.ts` in headless Chrome twice, changing only which build webpack resolves for `jpeg-lossless-decoder-js`:

| decoder build | result |
| --- | --- |
| published 2.1.2 (as depended on today) | **16 passed, 1 failed** — `JPEGProcess14SV1TransferSyntax: pixel 262143 is -1024, expected -3024` |
| the fixed build (as vendored by cornerstonejs/codecs#94) | **17 passed, 0 failed**, karma exit 0 |

Pixel 262143 is the last of 512x512 and the fixture is RescaleIntercept -1024 / slope 1, so those are stored samples 0 and -2000 — exactly the "one sample wrong, the last one" the pending note recorded. Every other case in both the grayscale and colour suites stayed green with the fixed build, including grayscale and colour `.57`, which go through the same decoder.

### What this PR should do about it

**Nothing yet, and the test change is not committed here.** `dicomImageLoader` depends on `jpeg-lossless-decoder-js@2.1.2` directly, so there is no path by which the fix reaches this branch: enabling the case now would just turn a documented pending into a red CI. The options, in the order I would rank them:

1. **Publish the fork** (as `jpeg-lossless-decoder-js` or under a scoped name) and bump the dependency here. Cleanest, and the only one that does not leave a build artifact in a repo. A git-URL dependency is not a real option — this is a runtime dependency of a published package.
2. Route `.57`/`.70` through `@cornerstonejs/dicom-codec`, which already carries the fixed build after cornerstonejs/codecs#94. Removes a direct dependency, but it is a decoder-wiring change well outside this PR.
3. Vendor the built bundle here too, as codecs did. Fastest, least appealing to carry in two repos.

Until one of those lands, the pending entry is worth updating so it records the actual cause rather than the encoder-tolerance guess — I can push that alone if you want it in this PR.

wayfarer3130:
Correction to my comment above: the test change **is** now pushed (a41a7b29b) rather than held back.

So `1.2.840.10008.1.2.4.70` is an active case in `decoders_test.ts` instead of a pending one, and **it will be red in CI** until `dicomImageLoader` actually gets a fixed `jpeg-lossless-decoder-js` — the dependency is still `2.1.2` and this commit does not touch it. That is deliberate: a failing case that names its cause is harder to lose track of than a pending note. The three ways to make it green are unchanged and listed above; publishing the fork is still the cleanest.

If you would rather this branch stayed green while the dependency question is settled, reverting that one commit puts the case back to pending.

wayfarer3130:
@jbocce - I'm good with your changes.  I merged the htj2k progressive work in already so if you approve then we can merge.

jbocce (approved):

Issue #2894, opened 2026-09-02
[Feature Request] vtk-wasm render backend, delivered as a separable rendering module


wayfarer3130:
## What feature or change would you like to see made?

A **vtk-wasm render backend** for Cornerstone3D — covering planar stacks, volume slice (MPR) and 3D volume/geometry rendering — delivered as a **separately packaged rendering module**, so that a build can include the vtk.js renderer, the vtk-wasm renderer, or both.

This replaces #2796, which added an experimental **vtk.js WebGPU** backend for planar and MPR GenericViewports.

### Why not #2796

#2796 is being rejected on direction, not on quality: vtk.js's WebGPU work is not going to be completed, and the supported route to WebGPU is `vtk-wasm`, with vtk.js expected to be deprecated in its favour. Extending the vtk.js WebGPU view API therefore builds on a path that will not be maintained.

> This is the rationale from the #2796 review, and it is the premise of this request. I could not find a public upstream deprecation announcement for vtk.js, so anyone acting on this should confirm the current Kitware position before committing the work — that confirmation is the first item under "Verify before starting" below.

What #2796 established is still worth keeping, and this request assumes it:

- the **render-backend registry** (`registerRenderBackend()`, already on `main` in `packages/core/src/RenderingEngine/helpers/renderBackendRegistry.ts`) as the extension seam;
- **resolving render-mode kind and surface through the registry** rather than through hardcoded `ActorRenderMode` literals, so an extension backend participates in view resolution, view references, scrolling and labelmap styling;
- the **examples with live backend switching** and a `?renderBackend=` parameter, and the debug panel reporting effective backend / mounted render mode;
- the **performance fixes** in that branch (per-stroke brush binding resolution, incremental flat-index labelmap slice extraction, slice buffer reuse) — those are independent of the backend and should land on their own regardless of what happens here.

The part that gets thrown away and rewritten is the backend implementation itself: `WebGPUImageMapperRenderPath.ts`, `WebGPUVolumeSliceRenderPath.ts`, `webgpuViewportRenderWindow.ts`.

---

## Part 1 — Split rendering into a separate module

**Goal:** `@cornerstonejs/core` stops depending on a specific renderer, so vtk.js and vtk-wasm can be included independently or together in a single build.

[...]

PR #2904, opened 2026-09-08
feat(webgpu): move the WebGPU backend out of core into its own package

wayfarer3130:
This pull request is on top of #2796. The base branch is `cornerstoneWebGPUCornerstone`, thus this diff shows only the two new commits. Both branches are up to date with `main` at 5.8.9.

This is a proposal. It is not a correction. It takes the WebGPU work as it is, and it shows that the work can be outside of core and lose nothing. Then a renderer that no person wants to maintain in the tree does not have to be in the tree.

## The problem

The developers had to write the backend inside the tree. A planar render path needs nine modules, which do these functions:

- they construct the actor entry
- they create the vtkImageData
- they apply the camera
- they resolve the projection
- they apply the presentation
- they give the shapes for the mounted rendering

Core did not export these modules. The developers could get them only if they imported deeply into the internal files of core. This is a fact about the packages. It is not a fact about WebGPU, because the same problem also applies to a vtk-wasm backend.

## The two commits

**1. `feat(core)`: the SDK to write a render backend, and the capabilities of a render mode**

This commit adds `@cornerstonejs/core/renderBackend`. This is one experimental entry point, and it has documentation. It holds only the items that a person needs to write a backend. It is separate from the viewport API that applications use.

The commit also adds two capabilities of a render mode to the registry. These capabilities replace a check of the surface, because that check cannot tell what the backends of an extension do.

- `supportsOverlayActors` tells you if a mount can hold overlay actors. `addImages()` needs this.
- `usesVtkActors` tells you if the mode mounts vtk actors, or if the mode draws through a `CanvasActor`.

`PlanarViewport.addImages()` used the type of the output surface to find if the viewport supports overlays. If a backend copies its pixels into the CPU canvas, and also mounts actors, the method removes the overlays and gives no message. This backend does exactly this. The method now asks the registry.

Both capabilities have the default value `true`. Thus **this commit does not change the behavior of the backend in the tree**. The tree continues to build and to run with WebGPU in its current location.

**2. `feat(webgpu)`: the backend as `@cornerstonejs/rendering-webgpu-vtkjs`**

The backend is now its own workspace package. It has a `@deprecated` mark from the start. The install does not include it, and no bundle includes it, unless an application asks for it.

The source files of the render path are the same, except for their import lines. This change moves the code into a new package. It is not a rewrite. Three items move out of core:

- the `'webgpu'` entries in the registry and in the constants. The new package now augments the module, and the registry file tells extensions to declare their backend IDs in this way.
- the four public exports.
- one direct use of `renderingEngineCache`. The code now calls the public function `getRenderingEngine()`.

Both modes now declare their capabilities. Thus the code no longer needs the special conditions for `renderMode !== CPU_VOLUME`.

## The proof that the change works

- The package does its type check against the **built `.d.ts` files** of core, and not against the sources of core. It imports only `@cornerstonejs/core` and these subpaths of core: `renderBackend`, `enums`, `types`, `utilities` and `loaders`. It has no deep imports. This shows that the change is possible, and the build makes sure of it.
- `git grep -i webgpu packages/core/src` finds only comments and examples in the documentation.
- The three `genericWebGPU*` examples stay in their current location. Thus **their URLs and the Playwright spec do not change**. These examples import the four entry points from the package, and not from core. An application does the same. All three examples bundle correctly, and the backend goes into **its own chunk**.
- The test `webgpuViewportRenderWindow` from Sedghi moved together with the code, and it continues to pass. It runs in a new jest project for the package.

## Notes for the review

- I had to divide the example alias for `@cornerstonejs/core` into an exact `$` alias and a directory alias. Webpack matched the prefix of the bare key, and then it changed the subpath imports to `src/index/<sub>`. This change also corrects the `@cornerstonejs/core/<subpath>` imports in all of the other examples.
- `utils/fixJSDOMJest.js` is an ESM file, but it called `require('util')`. The call ran only if the environment did not have `TextEncoder`, and no existing suite had this condition. Thus the `ReferenceError` stayed hidden until this package added a jest project. The file now uses a static import.
- **Registration for the release.** `lerna.json` and `pnpm-workspace.yaml` both include the package. The workspace file controls the install and the build. `lerna.json` is the file that `publish-version.mjs` (`lerna version`), `publish-package.mjs` (`lerna publish from-package`) and `validate:packaging` read. If only the workspace file includes a package, the tools build and test that package, but they never version it or publish it. `validate:packaging` now covers the package, thus it covers 11 packages and not 10. On my machine, the package fails there with the same `attw` message as all ten existing packages. The package has no problem of its own with publint or with the import in node.
- **The first publish is work for the release. It does not stop the merge.** The name is free, because `npm view` gives a 404. No person must do work before this PR merges. `packages/rendering-webgpu-vtkjs/PUBLISHING.md` records what an owner of `@cornerstonejs` must check on the npm side before the first release that includes the package. The most important item is the `NPM_TOKEN` in CI, which can possibly create a new package in the scope. The publish command uses `--no-verify-access`, thus it does not check this first. The file also records the optional `npm deprecate` command. This command is the only method that shows the deprecation on npm.
- The examples import the package statically, and not dynamically, because they drive it from synchronous callbacks. The README recommends the dynamic pattern to applications, and it tells why the examples are different.

## Not in this change

This change does not include the path for 3D volumes. It also does not include the literals of the render mode on the side of the tools. These literals stay in the code until their own change.

## Failures that were already present

Three tests in `packages/tools` fail on this branch. The same three tests also fail on `cornerstoneWebGPUCornerstone` without these commits: two tests in `labelmapImageMapperSupport.spec.ts` and one test in `labelmapRenderPlan.spec.ts`. Their mock of `@cornerstonejs/core` does not define `isVolumeRenderMode`, and `labelmapImageMapperSupport.ts:86` calls that function. This PR does not touch these tests, but I record them here because they are red.

Refs #2894, #2796


wayfarer3130:
@sedghi - I moved the WebGPU code into a separate module, and I marked that module as deprecated. Thus we can run the code and test the code, but core does not include the code. We need exactly this structure for vtk-wasm. Thus I think it is good to merge this PR now.

We can also add the demos to the public URL. But we must not show WebGPU as a supported feature. Then only a small number of persons will tell us that it does not work.

sedghi (approved):
Thanks

PR #2907, opened 2026-09-09
fix(adapters): apply the predecessor to the dataset, and number an unnumbered one

TFRadicalImaging:
OHIF_REF: refs/pull/6261/head

### Downstream test

The downstream job builds OHIF against the head of this branch. The line above
points the job at OHIF pull request 6261, because that pull request holds the
test update that the job needs.

cornerstone3D 2744 gave the area annotations one rule for the voxels that an
annotation covers. The rule changed the mean and the standard deviation that
`tests/Rectangle.spec.ts` reads, and the test still expects the old two values.
OHIF master pins `@cornerstonejs` to 5.8.2, which is older than 2744, so the
tests of OHIF master pass today. The test fails only in this downstream job.
This pull request does not cause the failure.

### Context

Two defects on the path that makes a saved segmentation a revision of the one it supersedes. Both sit in `packages/adapters`, they sit close together, and a consumer has had to work around the first of them.

**1. The predecessor merge lands on the wrapper, not on the dataset.**

`generateSegmentation` returns a wrapper whose `dataset` property holds what a caller stores, and both export branches merged the predecessor onto that wrapper:

```js
Object.assign(segmentationResult, predecessor);
```

The same function writes `segmentationResult.dataset.SOPClassUID` a few lines above each of those two places, which shows the shape. So the chosen `SeriesInstanceUID`, the incremented `InstanceNumber` and the `PredecessorDocumentsSequence` never reached the stored object: a save on top of an existing series wrote a **new** series instead, however the caller answered its save dialog, and the series dcmjs made up for the derivation survived untouched.

**2. `InstanceNumber` can become `NaN`.**

The `PREDECESSOR_SEQUENCE` module of `referencedMetadataProvider` incremented without a guard:

```js
result.InstanceNumber = 1 + Number(generalImage.instanceNumber);
```

A predecessor that arrived with no instance number gives `NaN`, and the caller then stores `NaN` in a Type 2 element. An artifact written back by another system is exactly that case, and it has to stay indistinguishable from one this library produced.

### Changes & Results

- `packages/adapters/src/adapters/Cornerstone3D/Segmentation/generateSegmentation.ts` — both branches merge onto `segmentationResult.dataset`. The RTSS and SR adapters already merge onto the dataset and need no change.
- `packages/adapters/src/utilities/referencedMetadataProvider.ts` — an unnumbered predecessor offers nothing to increment, so the value falls back to the `1` that `NEW_INSTANCE_DATA` gives a first instance in this same provider. The order of the revisions comes from the chain of `PredecessorDocumentsSequence` and not from this element, so the fallback loses nothing.

Before: a second save of the same data opened another series, and a save on top of an unnumbered instance stored `InstanceNumber` as `NaN`.
After: the second save extends the series it was told to extend, and an unnumbered predecessor yields a revision numbered `1`.

Only the returned wrapper stops carrying these values. Within this repository nothing reads them off the wrapper, and a consumer reads `.dataset`.

### Testing

`packages/adapters/test/predecessorMergeTarget.jest.js` (new, 5 cases) drives the real export through the existing `segRoundTrip` harness, for the bitmap branch and the labelmap branch alike: the predecessor's series, series number, instance number and document sequence are on `dataset`; the wrapper no longer carries them; and a save with no predecessor still writes a new series.

`packages/adapters/test/predecessorInstanceNumber.jest.js` (new, 5 cases): a numeric or string instance number is incremented, an absent or non-numeric one gives `1`, and the link back to the predecessor survives either way.

Whole repository: **1873 pass, 4 skipped, 121 suites across 5 projects**. The `adapters` project alone is 20 suites / 111 tests, up from 18 / 101.

Reverting each change fails only its own tests — 4 of the 5 merge-target cases (the no-predecessor control keeps passing), and 2 of the 5 instance-number cases.

#### PR

#### Code

#### Tested Environment

- [x] OS: macOS 15
- [x] Node version: 24.3.0
- [x] Browser: Chrome

@wayfarer3130

## Summary by CodeRabbit

- **New Features**
  - Segmentation data now records whether labels were generated and can retain the source image relationship.
  - Exported segmentations preserve predecessor series information in the dataset for bitmap and labelmap workflows.
  - Added a utility to safely retain only defined metadata attributes.

- **Bug Fixes**
  - Improved predecessor metadata handling when identifiers or study details are unavailable.
  - Prevented invalid or missing instance numbers from producing unusable metadata.
  - Improved numeric conversion fallback behavior for empty, invalid, or non-finite values.


wayfarer3130:
@TFRadicalImaging — I reviewed this PR for the merge of cornerstone3D into the fork of the University of Calgary, and I pushed one commit, `2816ec1bb`. The commit fixes two defects that sit next to the instance number of your second commit. Both defects come from one cause, and the fix is in the provider.

Both of your commits are correct. I checked the two premises: `fillLabelmapSegmentation` and `fillBitmapSegmentation` each return a wrapper whose `.dataset` is the object that the viewer stores, and the SR path and `createInstance` already assign to the dataset. The SEG path was the one path that did not.

## Defect 1 — an attribute with no value cleared a real value

Every consumer merges the result of `PREDECESSOR_SEQUENCE` onto a dataset with `Object.assign`, and `Object.assign` copies a key whose value is `undefined` over a real value. `SERIES_DATA` comes from `getNormalized`, which uses `for (const key in data)` and drops no key (`packages/metadata/src/metaData.ts:335`). A provider answers a module as a whole: OHIF returns the full object literal of `generalSeriesModule`, so an attribute that the instance does not carry is present as a key with the value `undefined`.

A predecessor with no Series Number therefore cleared the Series Number that the derivation gave the revision. dcmjs drops the element when dcmjs denaturalizes the dataset, so the revision lost a Type 1 attribute because of a gap in the instance that the revision supersedes. `Modality`, `SeriesInstanceUID` and `StudyInstanceUID` have the same exposure.

**Your first commit made the SEG path reach this defect.** Before that commit the keys with no value landed on the wrapper, and the viewer discarded the wrapper. The SR path and `createInstance` already assigned to the dataset, so both already reached the defect.

The class of predecessor that reaches this defect is the same class that your second commit exists for: an artifact that another system wrote back, with sparse metadata.

## Defect 2 — a predecessor that no provider holds threw

`generalImage.instanceNumber` on an `undefined` module raises `TypeError: Cannot read properties of undefined`, and `study.studyInstanceUID` raises the same. `getMetaData` does not catch an exception from a provider, so the exception took the whole save with it. A stale predecessor imageId reaches this case, as does an instance that a provider never ingested.

Your guard reads the *value* of `instanceNumber`, and the guard does not read `generalImage` itself. Line 106 already tolerates an absent `SERIES_DATA`, so a partly known predecessor got two more lines and then threw.

## What `2816ec1bb` does

- `definedAttributesOf` drops a key with no value before the result leaves the provider. **The fix is in the provider, and not in the four consumers**, so one place holds the rule and no consumer must know the rule. The four consumers are `generateSegmentation.ts` twice, `MeasurementReport.ts:759` and `createInstance.ts:56`. `assignDefined` in `createInstance.ts` stays as it is.
- The provider answers `undefined` when no provider names the SOP Instance UID of the predecessor. `undefined` is the answer of a provider for "I have nothing", `Object.assign` takes `undefined` as a no-op, and `applyPredecessorLink` in OHIF already tests the result. The two alternatives are both worse: an exception loses the save of the user, and a `PredecessorDocumentsSequence` that the provider builds from nothing carries a Type 1 `ReferencedSOPInstanceUID` that names no instance. An absent study module no longer throws either.

Six new tests, and **each of the six fails without the commit**. I verified that: I put the change of the source aside and ran the file, and the result was 6 failed and 6 passed.

- an instance that no provider holds
- an instance with no SOP Instance UID
- a merge of the `undefined` answer onto a dataset
- an absent study module
- an attribute with no value, which the result now omits
- the value that the derivation gave the revision, which the merge now keeps

Results: the two predecessor files give 17 of 17 tests, and the whole `adapters` suite gives **118 of 118** tests, across 20 suites. `tsc --noEmit` reports no new error. The `TS6059` messages about `rootDir` are already in the baseline of the package.

## One item that I did not change

`result.InstanceNumber` falls back to `1` for an unnumbered predecessor. A consumer that orders a series by instance number can then read the revision and the instance that the revision supersedes as one rank, and `2` avoids that tie. Your comment states the reason for `1`, and the order of the revisions comes from `PredecessorDocumentsSequence`, so the value is a choice and not a defect. I left the choice to you. @wayfarer3130 has `updateNewInstanceMetadata` in the fork, which computes the number again, so the fork does not depend on the value.

## One item for the fork, and not for this PR

`applyPredecessorLink` in OHIF does the same bare `Object.assign(dataset, predecessor)`. The helper is therefore safe once this PR lands, because the provider no longer gives a key with no value. The helper must stay until the fork takes the release of cornerstone3D that carries this PR, and the doc comment of the helper already says so.

TFRadicalImaging:
`InstanceNumber` stays at `1`.

`2` states a fact the save does not have. An instance that carries no number is not thereby the first instance of its series — it can be the fifth one an external system wrote back — so `2` asserts that the predecessor was `1`, and wherever it was not, the revision sorts below the instances it supersedes. `1` asserts only what `NEW_INSTANCE_DATA` already asserts for an instance that supersedes nothing.

The tie needs a consumer that reads an absent number as `1`, and the consumers on this path do not. The fork resolves which instance a display set shows from the chain (`resolveCurrentRevision`, through `createAddInstances`) and not from the number, and `applyPredecessorLink` already falls back to this same `1`, with a test on it. A reader that must know the order of the revisions has `PredecessorDocumentsSequence`, which is the element that carries it.

Verified at `2816ec1bb`: `adapters` gives 118 of 118 tests across 20 suites, and the two predecessor files 17 of 17.

wayfarer3130 (approved):

PR #2913, opened 2026-09-11
fix(publish): publish from GitHub Actions with trusted publishing, and drop lerna

wayfarer3130:
## The problem

npm holds no release after 5.8.2, which it received on 2026-08-17. The versions
5.8.3 to 5.10.2 are all absent from the registry.

The CircleCI job `NPM_PUBLISH` bumps the version and pushes the tag. Then
`lerna publish` fails with `E404 Not found` for each package, and it publishes
nothing. The tags and the GitHub Releases therefore name versions that npm does
not hold.

npm answers `E404` when a token does not permit a write, because a `403` tells
the caller that the package exists. Both `@cornerstonejs/utils` and
`@cornerstonejs/codemods` are public on the registry, so the packages exist. The
`NPM_TOKEN` in CircleCI no longer permits a publish.

The release machinery did not change: `git log v5.8.2..main` reports no change
to `.circleci/config.yml`, to `publish-version.mjs`, or to
`publish-package.mjs`.

## Why the failures were not visible

Three signals stayed green:

- `publish-version.mjs` runs before the publish step, and it succeeds. The
  version commits and the `v*` tags therefore continue to appear.
- The version commit starts `TEST_AND_RELEASE` again. The anti-loop guard calls
  `circleci-agent step halt`, and a halted job reports success. A red release
  run is therefore followed by a green run for the same version.
- The workflow `release.yml` needs only the tag, so each GitHub Release appears.

## What this pull request changes

- It adds `.github/workflows/publish.yml`. The workflow publishes the packages
  when a change lands on `main`.
- npm trusted publishing (OIDC) authenticates the publish, so the repository
  holds no npm token, and no one must rotate a token again.
- `publish-package.mjs` runs `npm publish` for each package that the registry
  does not hold. Lerna cannot use trusted publishing: lerna calls
  `libnpmpublish` directly, and it never runs the npm CLI, which is the program
  that exchanges the OIDC token.
- It removes lerna, `lerna.json`, and the 8 overrides that only lerna needed.
- It moves the creation of the GitHub Release into the publish workflow, and it
  removes `.github/workflows/release.yml`.
- It removes the job `NPM_PUBLISH` and the workflow `TEST_AND_RELEASE` from
  `.circleci/config.yml`. The workflow `PULL_REQUEST` stays, and it keeps the
  audit of the lockfile.

Lerna had little left to do, because this repository already did the difficult
part by hand. `publish-version.mjs` already rewrote the dependency ranges
between the packages. The write of the version, the commit, and the tag are now
direct. The option `--conventional-commits` generated a `CHANGELOG.md` file for
each package, and the GitHub Release carries generated notes instead.

## The order of the release

git receives the version commit and the tag **before** npm receives the
packages. git can reject a push, and npm cannot take a version back, so the step
that fails cheaply runs first.

A change that lands on `main` during a run therefore costs nothing: git rejects
the push, npm holds nothing, and the run stops. The run for the next merge
releases the same changes.

The state that a failed publish leaves is a tag that npm does not hold yet, and
that state repairs itself. The run reads one of three modes:

| Mode | When | What it does |
| --- | --- | --- |
| `skip` | `main` moved past this run | Nothing. No install, and no build. |
| `recover` | The tip is a version commit, or npm lacks the version that `main` carries | Publishes what npm lacks of that version. Takes no new version. |
| `release` | Otherwise | The normal path. |

The `recover` mode holds the count of unpublished tags at one. Without it, a
broken credential gives every merge a new version, and each one leaves a tag
that npm never receives.

## What someone must do before the merge

**1. Trust this workflow on each of the 11 packages.** `npm trust` does this
from the command line, so no one must use the web pages. Log in first as a user
with owner rights, then run:

```bash
for pkg in adapters ai codemods core dicom-image-loader \
           labelmap-interpolation metadata nifti-volume-loader \
           polymorphic-segmentation tools utils; do
  npm trust github "@cornerstonejs/$pkg" \
    --file publish.yml \
    --repository cornerstonejs/cornerstone3D \
    --allow-publish -y
done
```

Read the result back with `npm trust list @cornerstonejs/core`. Add `--dry-run`
to a command to see the effect before you make the change.

The name of the workflow file is a part of the trust configuration. A rename of
`publish.yml` stops the publish until someone runs `npm trust` again.

**2. Let the workflow push the version commit to `main`.** The branch `main`
carries a classic protection rule that asks for 1 approving review, and that
rule holds no bypass list. `github-actions[bot]` therefore cannot push to `main`
today. This command adds the bypass:

```bash
gh api -X PATCH \
  repos/cornerstonejs/cornerstone3D/branches/main/protection/required_pull_request_reviews \
  --input - <<'JSON'
{"bypass_pull_request_allowances":{"apps":["github-actions"],"users":[],"teams":[]}}
JSON
```

The ruleset `main and 1.x` needs no change. That ruleset holds a `deletion` rule
and a `non_fast_forward` rule only, and the push of this workflow is neither of
those two.

## What happens on the first run after the merge

`main` carries 5.10.2, and npm holds 5.8.2. The first run therefore reads the
mode `recover`: it publishes 5.10.2, and it takes no new version. That fills the
version that the OHIF viewer names in its dependencies today.

The tag `v5.10.2` exists already, and it names the commit that CircleCI made, so
the packages of that version carry the changes of this pull request as well. The
release for that tag exists, and the workflow leaves it as it is.

## What this pull request does not do

It does not publish the versions 5.8.3 to 5.10.1. `npm publish` sends only the
version that the manifest holds, so those versions stay absent from npm. To add
them, check out each tag and publish it.

## Tests

The version step ran from end to end on this branch:

- `node ./version.mjs` computed 5.10.3.
- `node ./publish-version.mjs` set all 11 packages to 5.10.3, rewrote the
  dependency ranges, updated the lockfile, generated the version files,
  committed, and tagged.
- `pnpm install --frozen-lockfile` then passed on the release commit, which is
  the check that CI makes.
- `node ./publish-package.mjs` ran against 5.8.2, which npm holds. All 11
  packages took the path that skips, and `npm publish` never ran.
- `node ./scripts/release-mode.mjs` answered `recover` for 5.10.2, which npm
  lacks, and `release` for 5.8.2, which npm holds.
- `npm view @cornerstonejs/utils@5.8.2 version` answers `5.8.2`. The same
  command answers `E404` for a version that npm lacks, and for a name that npm
  lacks.
- `npm publish --dry-run --access public` packs `@cornerstonejs/utils@5.10.3`.

The removal of the overrides changes the lockfile by 8 lines and nothing else:
no package changes its version, no package appears, and no package goes away.

Three defects appeared during those tests, and this pull request holds the
corrections:

- `getPublishablePackages` called `getAllPackages` again, so it answered a
  second set of objects. The new version reached those objects only, and the
  manifests kept the old version.
- `git tag` asks for a message on a machine that signs its tags. The command now
  sets `tag.gpgsign=false`, which matches every release tag of this repository.
- `scripts/validate-esm-packaging.mjs` read the list of packages from
  `lerna.json`. It reads the directories under `packages/` now.

I did not run `publish-package.mjs` against a version that npm lacks, because a
local npm session with owner rights would publish it for real.


package.json
```diff
@@ -138,7 +138,6 @@
     "karma-junit-reporter": "2.0.1",
     "karma-spec-reporter": "0.0.36",
     "karma-webpack": "5.0.1",
-    "lerna": "9.0.4",
```
jbocce, line 141: Could we please go through and remove any and all overrides/resolutions that were done because of having lerna? There are a few I'm sure.

sedghi (approved):
Thanks

PR #2916, opened 2026-09-14
chore(ci): harden the workflows that run on the self-hosted runner

jbocce:
Gate, pin and code-owner the two workflows that run fork PR code on the shared self-hosted runner; fix a shell-injection path and an unpinned dependency install in one of them.

## Context

`playwright.yml` and `ohif-downstream.yml` run on the self-hosted "nashua" runner, a fixed machine shared with codecs and OHIF. Both trigger on `pull_request`, so on a fork PR the code they run is the fork's. This brings them level with the protections codecs already has.

Along the way, the OHIF downstream job turned out to still install OHIF with bun, three months after OHIF moved to pnpm — so it was ignoring OHIF's lockfile on every run. That is fixed here too.

## Changes

- **Gate.** A metadata-only `gate` job on `ubuntu-latest` now fronts each nashua job. It never checks out PR code. A fork PR that touches a CI-defining path — workflows, the scripts they exec on the box, or the root manifests — is not dispatched to nashua; it runs on `main` once merged. Same-repo PRs always proceed. Same-repo is `head.repo.full_name == github.repository`; `author_association` is not a substitute (see the comment in the workflow).
- **CODEOWNERS.** Routes every workflow file, the scripts the nashua workflows exec, and the two scripts `publish.yml` runs with write/publish rights to @jbocce, @sedghi and @wayfarer3130. **Require review from Code Owners** is enabled on `main`, so it binds when this PR is merged.
- **SHA-pinned actions.** All nine `uses:` in the two workflows, at the versions already in use, tag in a trailing comment.
- **Shell injection.** `ohif-downstream.yml` interpolated `${{ env.OHIF_REF }}` — a value parsed from the PR body — into `run:` steps, where a crafted value would execute on the runner. Now read as `$OHIF_REF`, and `ohif-ref-resolve.sh` validates it before writing to `GITHUB_ENV`.
- **Triggers.** Both `pull_request` triggers bounded to `[main, beta]`. `master` doesn't exist here; `ohif-downstream.yml` had no filter.
- **OHIF install (bun → pnpm).** OHIF uses pnpm since OHIF/Viewers#6031, but `ohif-downstream.yml` still installed OHIF with `bun install`. Bun does not read `pnpm-lock.yaml` or `pnpm-workspace.yaml`. Every run therefore resolved the dependency tree again from the registry, and every run ignored the security `overrides` of OHIF. The job now runs `pnpm install --frozen-lockfile`, and the job no longer uses `oven-sh/setup-bun`. A frozen lockfile installs the exact versions that OHIF ships, and pnpm applies the `overrides` of OHIF as well. A frozen lockfile does not apply the `minimumReleaseAge` quarantine of OHIF: pnpm applies that setting when pnpm resolves a version range, and a frozen lockfile resolves no range. The lockfile is the stronger guarantee, because the lockfile pins one exact version for each package.

## Changes after the code review

A code review of this branch found four defects. The last three commits correct the four
defects.

- **The gate lists missed the manifest of a workspace project.** A root `pnpm install` runs
  the lifecycle scripts of every workspace project, and not only of the root project.
  `packages/docs/package.json` already declares `"preinstall": "node preinstall.js"`. The
  gate lists matched only the root `package.json` and the root `preinstall.js`. A fork PR
  that edited `packages/docs/preinstall.js`, or that added a `postinstall` to any
  `packages/*/package.json`, matched no pattern and ran on nashua. The `package.json`,
  `preinstall.js` and `.npmrc` patterns now carry a `*/` alternative, which matches at every
  depth. A `case` pattern is not a path glob, and `*` matches `/` as well, so one alternative
  covers a workspace project that a fork adds. The `pnpmfile` patterns stay at the root,
  because pnpm loads only a root `pnpmfile`.
- **CODEOWNERS had the same gap.** The `/package.json`, `/preinstall.js` and `/.npmrc` rules
  lose the leading slash, so the three rules now match at every depth.
- **`ohif-ref-resolve.sh` fell back to the default ref when `gh api` failed.** The script set
  `set -e` without `pipefail`, and the script piped `gh api` into `sed`. A pipeline reports
  only the exit status of the last command, so a failed API call gave an empty ref and a
  silent fall back to `master`. The log then read "PR body, defaulted" — the same text that a
  PR body with no `OHIF_REF` line produces. The script now sets `-eo pipefail`, reads the body
  into `PR_BODY` first, and stops with an `::error::` when `gh api` fails.
- **Two comments stated something that `ohif-downstream.yml` does not hold.** One comment said
  that the workflow never interpolates `${{ env.OHIF_REF }}`, but the `Checkout OHIF` step does
  exactly that. That use is safe, because an action input is not a shell, so the comment now
  scopes the claim to the `run:` steps and names the one exception. The other comment said that
  `--frozen-lockfile` honours the `minimumReleaseAge` quarantine, which is incorrect for the
  reason above.

## What changes for contributors

- A fork PR that touches a CI-defining path does not run Playwright or OHIF downstream until a maintainer merges the PR. The CI-defining paths include `pnpm-lock.yaml`. The CI-defining paths also include a `package.json`, a `preinstall.js` or an `.npmrc` in any workspace project, and not only at the root. This rule is deliberate, because `pnpm install` runs the install scripts of the packages on the box.
- A PR that changes a `package.json`, a `preinstall.js` or an `.npmrc` in any workspace project now needs an approval from a code owner. A version bump and a dependency change both match this rule. The cost is deliberate: the same manifest that carries the version carries the lifecycle scripts as well.
- PRs against branches other than `main`/`beta` no longer trigger OHIF downstream.
- OHIF downstream now installs from OHIF's lockfile instead of a fresh bun resolve, so it tests the dependency tree OHIF actually ships — security overrides included. If a run fails after this merges where it passed before, check for a real incompatibility the bun install had been hiding before assuming the workflow is broken.

## Results

- Both workflows parse; `gate → nashua job` wiring confirmed.
- Gate path matching dry-run on 14 real paths: 9 CI-defining flagged, 5 ordinary not.
- I ran the `case` block of each workflow against the paths that the code review named. The gate lists flag `packages/docs/preinstall.js`, `packages/docs/package.json`, `addOns/local/<name>/package.json` and `packages/core/.npmrc`. The gate lists still ignore `packages/core/src/RenderingEngine.ts` and `README.md`.
- A local `pnpm install` in this repository prints `packages/docs preinstall$ node preinstall.js`. That output confirms that pnpm runs the lifecycle script of a workspace project.
- I ran `ohif-ref-resolve.sh` against a stub `gh`. A body with an `OHIF_REF` line resolves to that ref. A body with no line defaults to `master`. A failed API call exits 1. The two invalid refs exit 1.
- `ohif-ref-resolve.sh`: `bash -n` clean; grammar accepts `master`, `release/3.12`, `1.0.0+build`, rejects `$(id)`, backticks, `../`, leading `-`, whitespace, `;`.
- Every action resolves to a 40-char SHA taken from its tag via the GitHub API.
- Every CODEOWNERS path exists; all three owners hold admin, so the rule binds.
- **This PR's own run** is the same-repo path through the gate and the first run of the pnpm install. A green "Install OHIF dependencies" step confirms OHIF's workspace file isolates the install; if red, that commit drops on its own. Fork paths need a fork PR after merge. actionlint not run.

## Testing

**What this PR's own CI run covers**

- The gate's same-repo path: `proceed=true`, both nashua jobs dispatched.
- The first pnpm install of OHIF. A green "Install OHIF dependencies" step confirms OHIF's `pnpm-workspace.yaml` isolates the install from this repo's workspace; a red one means it did not, and that commit can be dropped on its own.
- The default `OHIF_REF` path — no line in the PR body, so `master`.

**What it does not cover**

- The gate's fork paths: deferral when a fork PR touches CI files, and a plain fork run. Both need a fork PR after merge.

**`OHIF_REF`, exercised on this PR** by editing the description and re-running the downstream job — the resolve script reads the body live, so no new commit is needed:

1. No line → `OHIF ref (PR body, defaulted): master`. Run: https://github.com/cornerstonejs/cornerstone3D/actions/runs/34883666926/job/104108899638
2. `OHIF_REF: test/OHIF_REF-in-CS3D` (a branch cut from OHIF master) → `OHIF ref (PR body): test/OHIF_REF-in-CS3D`, same green result. Covers the parse path plus `/`, uppercase and `_` in a ref. Run https://github.com/cornerstonejs/cornerstone3D/actions/runs/34883666926/job/104120459957
3. `OHIF_REF: $(id)` → resolve step fails with `Rejected the OHIF ref…`, no later step runs, the value is not echoed. Run https://github.com/cornerstonejs/cornerstone3D/actions/runs/34883666926/job/104117716177
4. Restored to step 2's value.

#### PR

#### Code

  etc.)

  additions or removals.

## CI and Automation
- Updated pull request checks for supported primary branches.
- Added safeguards to prevent untrusted fork changes from running on shared runners when CI-related files are modified.
- Standardized automated build and test workflows around the project’s package manager.
- Pinned workflow actions for more consistent and secure execution.
- Added validation to reject invalid build references with clear errors.
- Improved review routing for critical automation and release-related changes.

## Documentation
- Corrected Playwright script references and snapshot-update guidance across documentation and examples.


.github/workflows/playwright.yml
```diff
@@ -11,7 +15,81 @@ permissions:
...
+
+          # CI-defining paths: the workflows themselves, the scripts they exec
+          # on the nashua box, and the root manifests that steer what gets
+          # installed and run there (a modified lockfile can pull a package
+          # whose install scripts execute on the box).
+          ci_touched=false
+          while IFS= read -r f; do
+            [ -n "$f" ] || continue
+            case "$f" in
+              .github/*|scripts/ci/*|scripts/run-playright.sh|scripts/link-ohif-cornerstone-node-modules.mjs|package.json|pnpm-lock.yaml|pnpm-workspace.yaml)
```
wayfarer3130: The path list does not contain `.pnpmfile.cjs`, `pnpmfile.cjs` or `.npmrc`.

The repository does not track any of these three files today, and that absence is the problem. A fork PR can *add* a root `.pnpmfile.cjs`. The gate does not flag that PR, because the new path matches no pattern in the list. The step `pnpm install --frozen-lockfile` then executes the top-level code of that file on the nashua box.

Please add the three names to this list, and to the same list in `ohif-downstream.yml` (line 89).

.github/CODEOWNERS
```diff
@@ -0,0 +1,44 @@
...
+
+# ── Self-hosted runner ────────────────────────────────────────────────────────
+# Scripts that the workflows exec on the shared self-hosted "nashua" box. Keep
+# this group in step with the protected-path list in the `gate` job of
+# .github/workflows/playwright.yml and ohif-downstream.yml — that list is what
+# keeps fork PRs that touch these paths off the box until merged.
+
+/scripts/ci/                                     @jbocce @sedghi @wayfarer3130
+/scripts/run-playright.sh                        @jbocce @sedghi @wayfarer3130
+/scripts/link-ohif-cornerstone-node-modules.mjs  @jbocce @sedghi @wayfarer3130
```
wayfarer3130, line 34: The comment above tells the reader to keep this group in step with the gate lists. The gate lists also contain `package.json`, `pnpm-lock.yaml` and `pnpm-workspace.yaml`. This group does not contain those three files.

The result: a same-repo change to `pnpm-lock.yaml` runs on the nashua box, and no code owner reviews the change first.

One correction to the PR description. I read the branch protection of `main`: `require_code_owner_reviews` is already `true`. The description says that this file stays inert until an admin enables **Require review from Code Owners**. An admin enabled that setting already. This file therefore binds on the day you merge the PR, and the gap above is live on that same day.

.github/workflows/playwright.yml
```diff
@@ -11,7 +15,81 @@ permissions:
   contents: read
 
 jobs:
+  # Metadata-only gate on a hosted VM: it never checks out or executes PR code.
```
wayfarer3130, line 18: Please record a dependency in this comment.

The `gate` job is a part of the workflow that the `gate` job protects. A `pull_request` event runs the workflow definition from the merge ref of the PR. A fork PR can therefore delete the `gate` job, delete `needs:` and delete `if:` in that same PR, and reach the nashua box.

One repository setting stops that attack today:

    actions/permissions/fork-pr-contributor-approval
      -> {"approval_policy": "all_external_contributors"}

This is the strictest of the three available policies: a maintainer must approve every fork PR run, and not only the run of a first-time contributor.

The `gate` job is therefore a second layer behind a human decision. The `gate` job still has value, but the value is specific. The `gate` job protects against an approval that a maintainer gives without full attention, for example on a dependency bump that looks safe. The `gate` job does not protect against a deliberate attack, because a deliberate attacker deletes the `gate` job.

This comment reads as though the `gate` job is sufficient alone. If someone later sets the policy to `first_time_contributors`, the `gate` job no longer binds, no job fails, and no person sees a notice. Two sentences here about the setting prevent that silent change.

wayfarer3130 (commented):
Three items from a review of this PR. Each one is small, and each one leaves a live gap today.

.github/workflows/playwright.yml
jbocce: Fixed in `fc55abf`, both gates.

This also turned up `preinstall.js`, which was missing and matters more: `package.json` declares it as the `preinstall` script, so it already runs on the box, and it already spawns child processes — a fork could edit it rather than add anything. CodeRabbit then caught `.pnpmfile.mjs`, which pnpm prefers over `.cjs` (`e0b8fc2`).

.github/CODEOWNERS
jbocce: Fixed in `fc55abf` — the three root manifests are in, and the comment now says what each mechanism does: the gate keeps a fork PR off the box until merge, these rules put an owner in front of that merge.

You're right about the branch protection, and the PR description is wrong. `require_code_owner_reviews` is already on, so this binds on merge. I'll fix the description.

.github/workflows/playwright.yml
jbocce: Added in `fc55abf`, both gates — the setting, that relaxing it fails silently, and that the gate catches an inattentive approval rather than a deliberate attacker.

I added one more line: other layers exist outside this repository, so this is not the last line of defence.

.github/CODEOWNERS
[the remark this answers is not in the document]
jbocce: Root manifests added in `fc55abf`.

On the path: the file really is `run-playright.sh`, missing the `w`. Rather than match the typo, `70feeba` renames it and fixes all 28 references — these rules match on the exact path, so the correct spelling matched nothing, silently.

.github/workflows/ohif-downstream.yml
```diff
@@ -23,20 +30,99 @@ concurrency:
...
+
+          # CI-defining paths: the workflows themselves, the scripts they exec
+          # on the nashua box, and the root manifests that steer what gets
+          # installed and run there (a modified lockfile can pull a package
+          # whose install scripts execute on the box).
+          ci_touched=false
+          while IFS= read -r f; do
+            [ -n "$f" ] || continue
+            case "$f" in
+              .github/*|scripts/ci/*|scripts/run-playright.sh|scripts/link-ohif-cornerstone-node-modules.mjs|package.json|pnpm-lock.yaml|pnpm-workspace.yaml)
```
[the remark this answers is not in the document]
jbocce: Fixed in `e0b8fc2` — all four spellings, both gates and CODEOWNERS.

Good catch on `.mjs`. The list committed an hour earlier had only the `.cjs` fallback.

scripts/ci/ohif-ref-resolve.sh
```diff
@@ -14,23 +15,46 @@ DEFAULT_REF="${DEFAULT_REF:-master}"
...
+#   - a leading '-' or '.': makes the value look like a flag to a later command
+#   - '..': a ref path escape
+#   - shell metacharacters generally: the workflow reads this as $OHIF_REF
+#     rather than interpolating it, so this is the second line of defence
+#     rather than the only one
+#
+# Fails loudly instead of silently falling back to the default, so a rejected
+# ref is diagnosable from the job log. The rejected value is deliberately NOT
+# echoed — no need to put attacker-supplied text into the log to explain this.
+if [[ ! "$REF" =~ ^[A-Za-z0-9][A-Za-z0-9._/+-]*$ ]] || [[ "$REF" == *..* ]]; then
```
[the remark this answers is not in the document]
jbocce: Fixed in `79f7657`. All four examples confirmed: they pass the character grammar and git rejects them.

Used `git check-ref-format "refs/heads/$REF"` rather than `--branch`, which also expands `@{-1}` and so consults repository state. The grammar bars `@` anyway, but this form needs no repository.

wayfarer3130 (approved):

PR #2919, opened 2026-09-16
fix(Segmentation): store a labelmap SEG drawn on a multiframe source

TFRadicalImaging:
### Context

Storing a **LABELMAP** SEG whose referenced **source** series is a multiframe (enhanced) instance throws and the SEG is never written:

```
Cannot set properties of undefined (setting 'PixelValueTransformationSequence')
```

The error surfaces from dcmjs' `SEGImageNormalizer.normalizeMultiframe`, which requires a `SharedFunctionalGroupsSequence` the dataset never received. A SEG drawn on a stack of single-frame source images saves fine; only a multiframe (enhanced CT/MR/US/PET) source hits this.

**Root cause (not a symptom guard).** `_createMultiframeSegmentationFromReferencedImages` assembles a *virtual* multiframe reference object frame-by-frame — one virtual dataset per referenced image — by spreading the source instance's `IMAGE_DATA` module:

```js
{ ...studyData, ...seriesData, ...imageData, PixelData, BitsAllocated: 16, ... }
```

For a multiframe source, `IMAGE_DATA` carries the source instance's `NumberOfFrames` (e.g. 16) but **none of its functional-group sequences**. That leaked `NumberOfFrames > 1` makes the existing `isSingleNonMultiFrame` guard (`datasets.length === 1 && !(NumberOfFrames > 1)`) skip the single-image duplicate-frame workaround. dcmjs' `SEGImageNormalizer` then takes its "already a multiframe" branch for the single virtual dataset and runs `normalizeMultiframe` on it — which dereferences the absent `SharedFunctionalGroupsSequence` and throws. The virtual dataset is explicitly *not* a real multiframe (see the comment at the normalizer call): it stands for the single frame its image references.

### Changes & Results

Force `NumberOfFrames: 1` on each per-frame virtual dataset so its frame count reflects the single frame it represents rather than inheriting the source instance's multiframe framing. A multiframe source is then normalized on the exact same virtual-multiframe assembly path a single-frame source stack already uses: the N single-frame virtual datasets are normalized into one virtual multiframe reference object, and the existing per-frame functional-group rebuild applies unchanged.

- **Before:** labelmap SEG on a multiframe source → throws, SEG not written.
- **After:** labelmap SEG on a multiframe source → produces a valid SEG dataset; single-frame-source behaviour is byte-identical (that path already had `NumberOfFrames` absent/1, and the derivation path overwrites `NumberOfFrames` with the frame count regardless).

This is a cause-level fix at the point the multiframe-ness leaks in, not a guard around the throw. It lands the downstream workaround we currently carry in our fork (RadicalImaging/UCalgary#497), which will be reduced to a forward-port once this merges.

### Testing

New jest spec `packages/adapters/test/segMultiframeSource.jest.js` drives the public `generateSegmentation`:

- A metadata provider whose `IMAGE_DATA` reports `NumberOfFrames > 1` and an enhanced (multiframe) SOP Class UID with **no** `SharedFunctionalGroupsSequence`, with a labelmap segment on a single frame — reproduces the throw on `main`, and now asserts a valid LABELMAP SEG (`SOPClassUID`, `SegmentationType`, `NumberOfFrames`, per-frame source reference to the enhanced instance).
- A single-frame-source control asserting unchanged output.

Verified the spec **fails on `main`** with the exact `PixelValueTransformationSequence` TypeError and **passes with the fix**. Full adapters suite green (21 suites, 128 tests).

```
yarn jest --selectProjects adapters
```

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] "OS: macOS"
- [x] "Node version: v24"
- [x] "Browser: Chrome"

/cc @wayfarer3130 for review — thanks!

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed segmentation export for labelmaps referencing multiframe source images.
  * Ensured generated single-frame segmentation datasets correctly reference their source image.
  * Preserved existing segmentation behavior for single-frame source stacks.


packages/adapters/src/adapters/Cornerstone3D/Segmentation/generateSegmentation.ts
```diff
@@ -563,6 +563,16 @@ function _createMultiframeSegmentationFromReferencedImages(
...
+      // Each virtual dataset stands for the single frame this image references,
+      // not the whole source instance. When the source is a multiframe
+      // (enhanced) instance, IMAGE_DATA carries its NumberOfFrames (e.g. 16) but
+      // none of the functional-group sequences. Left in place, that leaked frame
+      // count makes SEGImageNormalizer treat the single virtual dataset as an
+      // already-multiframe one and run normalizeMultiframe on it (which needs the
+      // absent SharedFunctionalGroupsSequence) → it throws. Forcing 1 frame lets
+      // the normal virtual-multiframe assembly path — the same one a single-frame
+      // source stack uses — handle a multiframe source identically.
+      NumberOfFrames: 1,
```
wayfarer3130, line 575: Would it work to use:
SharedFunctionalGroupsSequence: { Value: [] },
PerFrameFunctionalGroupsSequence: [],

TFRadicalImaging: I tried that. Keeping the source's `NumberOfFrames` (> 1) and adding empty `SharedFunctionalGroupsSequence: []` / `PerFrameFunctionalGroupsSequence: []` doesn't fix it: with the frame count still > 1 the single virtual dataset stays on `SEGImageNormalizer.normalizeMultiframe`, whose per-frame loop then iterates `PerFrameFunctionalGroupsSequence` and dereferences per-frame entries the empty array doesn't provide, so it throws again inside `normalizeMultiframe` (dcmjs `normalizers.js`, the per-frame `forEach`) rather than at the earlier `SharedFunctionalGroupsSequence` guard. The regression spec fails on that variant.

The reason `NumberOfFrames: 1` is the fix rather than a workaround: each virtual dataset here stands for exactly one referenced frame, so 1 is its true frame count. Setting it routes the dataset down the same virtual per-frame assembly path a single-frame source stack already uses (the well-exercised one), instead of sending a one-frame dataset through the multiframe normalizer at all. Empty functional-group sequences would only be needed to survive a path this dataset shouldn't be on.

PR #2920, opened 2026-09-17
fix(SliceIntersectionTool): keep slab handles reachable when hovering slab boundary lines

julianlinaress:
Slab thickness handles only showed while hovering the center line, so once the slab was widened the handles disappeared before the cursor could reach them. Hovering a dashed slab boundary line now also activates its plane group (lower priority than any center line or handle). Mouse down hit testing is unchanged, so slab lines do not capture clicks.

### Context

The slab thickness handles of the slice intersection lines only appeared while
hovering the center line (the one that drives the slice index). Once the slab
was widened, the handles moved away from that center line and became
unreachable: hovering the center line shows the handles, but moving the cursor
towards the outer handle leaves the center line's proximity halfway there, the
line stops being active, and the handles disappear before the cursor gets to
them.

The wider the slab, the harder it gets, and past a certain thickness the slab
can no longer be adjusted with the mouse at all.

Reproducible on the basic setup of the tool's own example:
https://www.cornerstonejs.org/live-examples/sliceintersections

Cause: hover hit testing (`mouseMoveCallback` -> `_findLineNear`) only
considered the center line and the handle circles. The dashed slab boundary
lines are computed for every line on every render, but were never part of the
hit test, so there was nothing keeping the line active in the gap between the
center line and the handle sitting on the boundary line.

### Changes & Results

`packages/tools/src/tools/SliceIntersectionTool.ts`

- `_findLineNear` takes an options object (`includeHandles`, `includeSlabLines`)
  instead of a single positional boolean.
- Hover now also matches the dashed slab boundary lines (`includeSlabLines`), so
  hovering either boundary line activates its plane group and shows its handles.
  Since the handles sit on those boundary lines, the cursor can now travel along
  the line straight onto the handle without the line losing its active state.
- Slab boundary lines are only checked after every line and every handle of all
  groups, so a boundary line crossing another group's center line never steals
  that line's hover.
- Only hover changed: `isPointNearTool` and `toolSelectedCallback` are
  untouched, so slab boundary lines do not capture mouse down and the active
  mouse tool (window level, etc.) keeps working when clicking on them. Dragging
  is still limited to the center line and the handles, exactly as before.

`packages/tools/src/tools/SliceIntersectionTool.spec.ts`

- New test: hovering either slab boundary line resolves to that plane group,
  and a boundary line crossing another group's line does not steal its hover.

Before — the handles vanish on the way to them, so the slab cannot be adjusted
once it is wide:

https://github.com/user-attachments/assets/c46e7451-4036-45ed-8c60-853bff0b8e88

After — hovering a boundary line keeps the handles visible all the way to the
handle:

https://github.com/user-attachments/assets/28f3a329-7a52-47cd-a4dd-e1f8877c4323

### Testing

1. Run the slice intersection example: `pnpm run example sliceIntersections`
   (or open https://www.cornerstonejs.org/live-examples/sliceintersections for
   the "before" behavior).
2. Hover one of the intersection lines: the rotation handles and the two slab
   thickness handles (hollow circles) appear.
3. Drag a slab handle away from the line to widen the slab. Two dashed boundary
   lines appear at the slab edges.
4. Move the cursor away and then hover one of the dashed boundary lines: the
   line's handles show up, and the cursor can reach the handle sitting on that
   boundary line and drag it. On `main`, the handles only appear over the center
   line and disappear halfway to the handle.
5. Check that clicking on a dashed boundary line does not select or drag the
   line, and still goes to the active mouse tool.
6. With two crossing intersection lines, hover the point where one line's slab
   boundary crosses the other line: the other line (the real line) is the one
   that gets activated.

Unit tests: `npx jest packages/tools/src/tools/SliceIntersectionTool.spec.ts`

#### PR

#### Code

  etc.)

  additions or removals.

#### Tested Environment

- [x] OS: Ubuntu 22.04.5 LTS (WSL2, Windows 11)
- [x] Node version: 24.15.0
- [x] Browser: Chrome 152.0.7977.83

## Summary by CodeRabbit

- **New Features**
  - Improved slice tool hover behavior: moving over slab boundary lines keeps the related line and handles visible.
  - Added slab-thickness controls and an option to show or hide slab handles in the slice intersections example.
  - Slab handles now align accurately with oblique slab boundaries.
  - Hovering near a slab boundary selects the closest boundary line.

- **Bug Fixes**
  - Preserved existing proximity behavior for rotation and slab handles.
  - Slab boundaries no longer trigger hover when slab handles are disabled.

- **Documentation**
  - Added user-facing specifications for crosshairs and slice intersection controls.


LucasLeone:
That happens to my project, nice fix!

wayfarer3130:
I pushed one commit to this branch: `f10904a5`. The commit fixes three problems in the new hover region, and it adds a test for each problem.

Thank you for the fix. The hover region is the correct approach. The three problems are below.

## 1. The slab handles leave the boundary line on an oblique plane

`_computeSlabHandles` made each handle from a world point. The code offset the anchor point along the leader normal, and then the code projected the result to the canvas. That projection lands on the drawn boundary line only when the leader normal lies in the target plane.

Write `s` for the sine of the angle between the two planes:

- The drawn boundary line sits at `(slabThickness / 2) / s` from the center line.
- The old projection sits at `(slabThickness / 2) * s`.

The two values are equal only when `s` is 1, which is the perpendicular case. The tests of the first version use the perpendicular case only.

A user gets the oblique case easily, because `_startManipulation` sets the rotation axis to the normal of the **target** viewport. A rotation in one viewport tilts the leader normal out of the plane of the **third** viewport.

Each handle now takes the closest point of its own drawn segment. Two results follow:

- The handle sits on the line that the user sees.
- The handle follows the cursor during a drag, because `_applySlabThickness` sets the slab from the distance between the cursor and the leader plane. The old handle lagged the cursor by the factor `s * s`.

## 2. The hover ignored `slabThicknessControls`

`_appendSlabGeometry` tests `showSlabThickness` only, but `_appendHandles` also tests `slabThicknessControls`. With `slabThicknessControls: false`, the tool still draws the dashed lines, and no handle sits on those lines. The old code gave the hover to a distant plane group. `_findLineNear` now tests `slabThicknessControls` before the code uses the slab boundary lines.

## 3. `lines.find` selected the first line, and not the closest line

Two slab boundary lines can cross each other. Near the point where the two lines cross, both lines are within the 6 px proximity. The old code gave the hover to the first group of `_getPlaneGroups()`. The new code keeps the closest line, and a tie keeps the first group.

## Tests

`createFakePlanarViewport` takes an optional `axes` parameter, which gives the spec an oblique plane. Three new tests cover the three fixes. Each new test fails against the previous version of `SliceIntersectionTool.ts`, and each new test passes with the new version. The tools package passes: 59 suites and 834 tests.

## Example

The `sliceIntersections` example gets two controls:

- A slider sets the slab thickness of all three viewports. The slider writes the display-set presentation, which is the path that the slab handles also use.
- A toggle turns `slabThicknessControls` on and off. The toggle is necessary for the demonstration of problem 2, because you cannot widen the slab with the handles when the handles do not exist.

Steps for problem 1:

1. Move the slider to a wide slab.
2. Rotate the yellow sagittal line in the axial viewport by about 30 degrees.
3. Look at the yellow line in the coronal viewport. The hollow handles stay on the dashed boundary lines.

Steps for problem 2:

1. Move the slider to a wide slab.
2. Set the toggle to `OFF`.
3. Hover a dashed line. The tool does not select the line.

wayfarer3130:
I pushed a second commit: `ab976f66`. The commit adds a specification of the crosshairs control at `packages/docs/docs/specs/tools/control/Crosshairs.md`.

The specification describes what the user expects, and not how the tool works:

- The intersection lines, and what a drag of a line does.
- The control points that appear when the user hovers a line, and the rule that the control points stay visible until the cursor reaches them.
- The dashed slab boundary lines, and the rule that the hollow control points sit on those lines.
- The behaviour on an oblique plane, where the control points must stay on the dashed lines.

The user requirements and the implementation requirements stay in separate sections, because the two sections have different change rules. The Scope section states those rules.

The Past defects section records the four defects of this pull request, and it names the requirement of each defect. `packages/docs/docs/specs/` is a new directory, so `sidebars.js` gets a Specifications category for it.

wayfarer3130 (approved):

wayfarer3130 (approved):

